PR #190, opened 2015-03-10
Bug Fixes

goehle:
This fixes two bugs
-  This is the corresponding PG pull request which removes error messages from Applet.pm that were preventing problems from running if the applet was part of a hint.  Test by going to a problem like: `Library/CSUOhio/calculus/trigonometric_substitution/trigSub12.pg` and trying the problem as a student.  You should get lots of error messages and none after the pull. 

-  This fixes bug 3337 where PopUps that use itemized lists and are included in a unionTable fail hardcopy generation.  The issue was the itemize environment inside a halign environment. The itemize environment was replaced with a custom compatible solution.  Test by trying to generate a hardcopy for a problem like `Library/LoyolaChicago/Precalc/Chap5Sec1/graph-shifting-01.pg`.  Before the patch the hardcopy should fail and after the patch it should succeeed.

### DIFF
--- a/lib/Applet.pm
+++ b/lib/Applet.pm
@@ -693,7 +693,9 @@ use constant DEFAULT_HEADER_TEXT =><<'END_HEADER_SCRIPT';
 	 	  if (obj ) {   //RECENT FIX to ==
 	 		  return( obj );
 	 	  } else {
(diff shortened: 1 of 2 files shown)


mgage:
I'll try to add a new method to alert a user if the applet doesn't load properly.

PR #191, opened 2015-03-10
Bug Fixes (Develop)

goehle:
This is the develop version of #190.


mgage:
I'll try to add a new method to alert a user if the applet doesn't load properly.

PR #192, opened 2015-03-31
Develop cmplx bug

mgage:
Fix stupid bug that forced complex number answer evaluators to use absolute error, not matter the setting of the relTol flag.


goehle:
This pull is from a different branch with different commit numbers.  Its better if you just make a new pull request which pulls the mgage:master_cmplx_bug branch into develop.  Having different commits for the same change has the potential to cause conflicts down the road.

PR #193, opened 2015-03-31
Fix bug that forces the use of absolute error in evaluating complex numb...

mgage:
...er answers regardless of the answer evaluator setting.

Also remove PGcomplexmacros2Exp.pl,  an unused file that duplicates PGcomplexmacros2.pl


goehle:
I tested and this seems to work with the right parameters included.  I wasn't able to reproduce the original problem, but the commented out parameter is clearly an issue.

PR #195, opened 2015-04-15
UTF8 Encoding

goehle:
Don't encode utf characters in encode_pg_and_html because it breaks rendering.

Test this bug by entering something like ñeumé into an answer blank and check.  The preloaded answer should look the same as you entered it.  Pre patch behavior is to mangle the accented characters.


mgage:
Works well with all of the "option" keys on my mac.  Automatic capitalization still causes trouble with the TeX preview -- we'll deal with that later.

mgage:
You need to be cautious when testing since some of the preview images are cached and may not be fixed immediately with an upgrade. I restarted the server just in case.

PR #197, opened 2015-04-17
modifications for AskSage

mgage:
Add https  address.  

add key for asking for tos (Terms of Service).  Each author of a pg problem has to set this to true.


goehle:
Tested with a problem and it works.  If an existing problem does not explicitly accept the ToS then the user will be asked to do so.

PR #198, opened 2015-04-17
Develop grep fix

mgage:
1. Replaced \1 by $1 in grep strings. This quiets the warnings that
newer versions of perl emit when they see the old style grep strings.

2. Cleaned up the unit test code for the "Units.pm" module.  Make it use the Units.pm file that is actually being used by WW instead of using a copy.


goehle:
When I test this by submitting an essay answer with some math (e.g. `x+y`) I get the error 
```
Use of uninitialized value $1 in regexp compilation at line 2037 of [PG]/macros/PGbasicmacros.pl
```
I'm not sure `$1` works in matches.

goehle:
Actually its something more subtle than that, since I only get a problem when I use \`x+y\` type equations and not `\(x+y\)` type ones

goehle:
Ignore the comment I deleted.  LaTeX style math strings are not interpolated by `EV3P_parser`. 

As far as I can tell you can't use $1 in a match since it gets bound to character groups from previous matches.  In this case you use \1 to match backreferences.  

http://stackoverflow.com/questions/1068840/what-is-the-difference-between-1-and-1-in-a-perl-regex

macros/PGbasicmacros.pl
```diff
@@ -2032,7 +2032,7 @@ sub EV3P_parser {
   my $start = ''; my %end = ('\('=>'\)','\['=>'\]');
   my @parts = split(/(``.*?``\*?|`.+?`\*?|(?:\\[()\[\]]))/s,$string);
   foreach my $part (@parts) {
-    if ($part =~ m/^(``?)(.*)\1(\*?)$/s) {
+    if ($part =~ m/^(``?)(.*)$1(\*?)$/s) {
```
dpvc: Geoff is correct, this should not have been changed.  The `\1` is used here to make sure the end quotes are the same as the beginning quotes.  `$1` doesn't work in this situation.

mgage:
I'm also puzzled as to how the new merges of development into the develop_grep_fix branch seem to have appeared.  I think I'll close this pull request  also and resubmit it.

PR #200, opened 2015-04-19
Develop fix pretty print

mgage:
Fix the way pretty_print is called to conform with the new, refactored, version of pretty_print in PGUtils.pm

### DIFF
--- a/lib/AnswerHash.pm
+++ b/lib/AnswerHash.pm
@@ -594,7 +594,7 @@ sub print_result_if_debug {
 	if (defined($self->{debug}) and $self->{debug}>0) {
 	    	$rh_ans->{rh_options} = \%options;  #include the options in the debug information
 	    	my $name = (defined($rh_ans->{_filter_name})) ? $rh_ans->{_filter_name}: 'unnamed';
-	    	warn "\n $count. Result from queue $queue:  name: \"$name\"n", $rh_ans->pretty_print();
+	    	warn "\n $count. Result from queue $queue:  name: \"$name\"n", pretty_print($rh_ans,'html');
 	    	++$count; 	
 	 }
 	$rh_ans->{_filter_name} = undef;
--- a/lib/AnswerHash.pm
+++ b/lib/AnswerHash.pm
@@ -332,16 +332,16 @@ sub error_message {
 
 # error print out method
 
-=head4 pretty_print
-
-
-	Useage:     $rh_ans -> pretty_print();
-	
-	
-	Returns a string containing a representation of the AnswerHash as an HTML table.
-
-=cut
-
+# =head4 pretty_print
+# 
+# 
+# 	Useage:     $rh_ans -> pretty_print();
+# 	
+# 	
+# 	Returns a string containing a representation of the AnswerHash as an HTML table.
+# 
+# =cut
+# 
 # sub pretty_print {
 #     my $r_input = shift;
 #     my $level = shift;
@@ -503,7 +503,7 @@ sub get_student_answer {
 	my %answer_options = @_;
 	my $display_input  = $input;
 	$display_input =~ s/\0/\\0/g;  # make null spacings visible
-	warn "Raw student answer is |$display_input|" if $self->{debug};
+	eval (q!main::DEBUG_MESSAGE(  "Raw student answer is |$display_input|")!) if $self->{debug};
 	$input = '' unless defined($input); 
 	if (ref($input) =~/AnswerHash/) {
 		# in this case nothing needs to be done, since the student's answer is already in an answerhash.
@@ -544,7 +544,7 @@ sub evaluate {
 	$rh_ans->{error_flag}=undef;  #reset the error flags in case 
 	$rh_ans->{done}=undef;        #the answer evaluator is called twice
 	
-    warn "<H3> Answer evaluator information: </H3>\n" if defined($self->{debug}) and $self->{debug}>0;
+    eval (q!main::DEBUG_MESSAGE( "<H3> Answer evaluator information: </H3>")!) if defined($self->{debug}) and $self->{debug}>0;
     $self->print_result_if_debug('pre_filter',$rh_ans);
     
 	my @prefilters	= @{$self -> {pre_filters}};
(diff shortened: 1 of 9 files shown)


goehle:
Could you post some testing instructions for this?

mgage:
To test:

Create a new homework set. "check answer debug info"

From the library choose:
Subject = WeBWorK
Chapter=Demos
Section=Answers

view the results and import all of the problems into your new homework set.

2.  Near the bottom of problem2 make this change:

$new_eval->{debug}=1;
# this prints out a trace of the answer evaluation process
ANS($new_eval);

Submit the problem with various answers.

3. In problem 3 you can add debug=>1 to the arguments for fun_cmp
e.g.
ANS(fun_cmp($ans,limits=>[-1.5,1.5],vars=>'t', debug=>1));

I only tested these first two problems but you can inspect the answer evaluator printouts for other problems as well by adding debug=>1

goehle:
Which one is problem2?  I've added this: 
```
$new_eval->{debug}=1;
# this prints out a trace of the answer evaluation process
ANS($new_eval);
```
to several problems and cant make it work right.  Is there supposed to be an associated ans_rule?

lib/AnswerHash.pm
```diff
@@ -115,7 +115,7 @@ The answer hash class is guaranteed to contain the following instance variables:
 =cut
 
 BEGIN {
-	be_strict(); # an alias for use strict.  This means that all global variable must contain main:: as a prefix.
+#	main::be_strict(); # an alias for use strict.  This means that all global variable must contain main:: as a prefix.
```
goehle: I'm just trying to figure out this change.  If I uncomment this I get warnings about not being able to find `main::be_strict()`.  If I change it back to the way it was I dont get any warnings, but I dont know whats different.

mgage:
Problem 2 is Library/Rochester/setSampleAnswers/sample_compound_ans_eval.pg

Problem 3 is Library/Rochester/setSampleAnswers/sample_fun_ans.pg

I'll have to get back to you about the main::be_strict() issue.  be_strict() arose because you 
can't (or couldn't) use the pragma   "use strict;"  inside the Safe compartment.  I'm not sure 
about the current status of that issue.  Not using it at all simply means that you don't get the extra warnings that Perl will give you if you do things like used uninitialized variables, etc. etc.

goehle:
You have to add the "strict" flag to the safe container. But it breaks so much in PG I wasn't able to get anywhere with adding it.

goehle:
-  I tested the debug out put for various problems.  It seemed to work fine. In particular it is now formatted as html and uses the new pretty print setup.  
- I tested the new level setting and it works fine
- I tested the new text output and it seems fine.  However, I am slightly worried that I was able to do 
```
\{$PG->pretty_print(Context(),'html',)\}
```
But if I change it to 'text' output my server bogs down and times out.  Is there a reason text output should be more inefficient?

lib/PGUtil.pm
```diff
@@ -163,4 +164,51 @@ sub pretty_print_tex {
...
+		$out .= " type = UNDEFINED; ";
+	}
+	return $out." " unless defined($r_input);
+	
+	if ( ref($r_input) =~/HASH/ or "$r_input" =~/HASH/ ) {
+	    $out .= "{\n";
+	    $indent++;
+ 		foreach my $key (sort keys %{$r_input})  {
+ 			$out .= "  "x$indent."$key => " . pretty_print_text( $r_input->{$key}, $indent ) . "\n";
+ 		}
```
goehle: I think the two `$indent`s here and line 198 need to be `$level`s.  I'm getting infinite loops with nested hashes otherwise.

mgage: You're right.  I could have sworn I did a global replace on this subroutine when I transferred it from renderProblem.pl -- but I guess not.  Thanks for catching it. 

This will be useful for printing things to log files but I think the text version of pretty_print is going to be most useful on the WW2 side of the applications.  There are several versions of it in the code already and in my opinion they should be consolidated.

mgage:
use strict; has never worked within our safe compartment because "use" (and require) are commands
that are forbidden within the PG language. (Allowing them opens everything up.)

At one point we were able to spoof the effect of use strict by wrapping it within a subroutine be_strict().  I think that ceased to work with newer versions of Perl.  The commented out main::be_strict was an attempt to keep it working. 

The purpose of "use strict" is to warn about undefined variables and similar bad coding practices.  I don't know what the current best replacement would be with recent versions of perl and the Safe compartment.  Figuring out what the modern replacement for be_strict should be would be a worthwhile project, but not a small one I'm afraid. Not to be addressed in this pull request.

goehle:
Its not `use strict`; its a parameter passed when the safe environment is created:  http://perldoc.perl.org/Safe.html#reval-%28STRING%2C-STRICT%29

mgage:
OK.  I remember that as well but I don't recall exactly what went  wrong when we tried to use it.  Part of the problem is that you definitely don't want to have the PG problems themselves to be compiled under the strict protocal.  The thinking was that subroutines and certainly the underlying core code for WeBWorK and PG should be compiled following strict programming protocols (for reliability) but that the problems themselves would be allowed a lot of leeway so that the instructors wouldn't have to follow all those rules such as declaring variables.  I don't know whether this was a good idea or not in the long run, but it was probably the right thing to do when we (and everyone else using the internet at the time) were just getting started.

goehle:
As I recall, it turns strict on for everything in the safe compartment,
including problems. I think thats why we didn't pursue it.

mgage:
I've added some .pg problems that test non-Null and pretty_print to the pg/t directory.  

Next step (some day) is to hook those up to an automatic testing scheme such as TestHarness.  Meantime you can copy them into a homework set for testing.

t/settest_notnull_and_prettyprint/settest_notnull_and_prettyprint.def
```diff
@@ -0,0 +1,17 @@
...
+openDate          = 09/03/2014 at 10:34am EDT
+dueDate           = 09/10/2014 at 10:34am EDT
+answerDate        = 09/10/2014 at 10:34am EDT
+paperHeaderFile   = defaultHeader
+screenHeaderFile  = defaultHeader
+description       = Test the notnull() function.  This returns 0 for an undefined variable, an empty string, a blank string, an empty array and an empty hash.
+<n>
+<n>--------
+<n>Test the $PG-> pretty_print(object, html/text/TeX, level) function for pretty printing objects.  When called directly as a subroutine (without $PG->) it uses level=5 and the current displayMode to print the object.
+problemList       =
```
goehle: These <n> are causing an issue for me on import?  Is that something specific to my setup?

mgage:
They were \n that I had put into the description field to trigger a line break -- they were ignored in popup description.  I also tried <br/> and that didn't seem to work either.  Is there a way to create a line break in these pop up descriptions?  If not that's ok -- since encouraging the descriptions to be one paragraph about one idea is probably the right thing to do.

goehle:
Not right now.  I suppose I could do search replaces to make them into <br>'s but I didn't because I didn't want to make them too long.

PR #202, opened 2015-04-24
change regexes to silence perl warnings

bldewolf:
Perl whines that \1 could be better written as $1 in these regexes, so this
changes them.


goehle:
This has already been fixed in #198.  Thanks for putting in the time to make a pull request, though. 

Cheers.
Geoff.

bldewolf:
Oh woops, I checked master instead of develop to see if it was fixed yet.  Well, consider this pull request another vote for merging it to master. :)

PR #207, opened 2015-05-15
Fix undef msgs

mgage:
This pull request makes minor changes in constructions which sometimes generate "undefined" errors.

It also makes one more replacement of \1 by $1 in a grep substitution 

Because it uses the // operator these files will not be compatible with perl 5.8 and earlier.  Do we need to continue to support these earlier installations?

### DIFF
--- a/lib/Chromatic.pm
+++ b/lib/Chromatic.pm
@@ -5,9 +5,16 @@ BEGIN {
 package Chromatic;
 
 our $webwork_directory = $WeBWorK::Constants::WEBWORK_DIRECTORY; #'/opt/webwork/webwork2';
-our $seed_ce = new WeBWorK::CourseEnvironment({ webwork_dir => $webwork_directory });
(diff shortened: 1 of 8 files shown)


goehle:
For what its worth, this is branched from develop_minorfixes and not develop.

mgage:
OK. I never got around to pushing that branch and this is more of the same kind of minor warning fixes. 
I've deleted that minorfixes branch and removed the kink from the graph. :-)

mgage:
1.To check:  warning methods about encode_base64 being redefined should no longer appear
2.Warnings on isHash($obj) should not appear even if $obj is undefined or an empty hash.
3.Check that complex variables objects produce correct perl code $cmplobj->perl
4. Warning messages about AskSage being redefined (these appear in the apache error log) should disappear -- even when using a problem that uses AskSage
5. Misc warning messages that appear sporadically in the log files -- these will be hard to check -- mostly make sure that no harm has been done.

PR #210, opened 2015-05-26
Javascript LiveGraph Replacement

goehle:
This is the pg pull request for the LiveGraph replacement.  It will need to be tested and pulled at the same time as its counterpart in webwork2 (number 587)


dpvc:
Looks like this is to master rather than develop (which is probably why there are so many extraneous commits).

goehle:
Whoops.  Thanks, nice catch.

PR #211, opened 2015-05-26
Javascript LiveGraph Replacement

goehle:
This is the pg pull request for the LiveGraph replacement. It will need to be tested and pulled at the same time as its counterpart in webwork2 (number 587)


dpvc:
That looks better.  I'm glad you did this, as I was thinking I would need to do it.  

I hope to be able to do testing next week, but probably won't get to it before then.

mgage:
The y axis seems to be labeled as the v axis

goehle:
I double checked and its a "y" but it definitely comes across as a "v".  I've changed the font style and size to make it a little clearer.

mgage:
That's good -- it doesn't come across as a y in every example, but it does in most of the early ones.  I'll look at your changes in a minute but my feeling was that a slightly lighter weight font would help (the axes seem to be in bold).  The other thing I wished for was a completion of the viewing box -- just a dotted line outline, not numbered axes.  I think that would help keep me oriented as I look at the figure.  I've answered the first 10 questions and am working my way through the remaining 40 or so.  I'll probably have some suggestions for the questions themselves since I'm going through them carefully.

goehle:
Issues with the questions themselves should be pg bugs and not part of this pull, of course.  What do you mean a completion of the viewing box?

mgage:
They will be.  I mean outlining the other 9 sides of the cube parallel to the axes. I think Mathematica usually does this to help give a better perspective of the surface inside.  This should probably be an option since it is not desirable in every case.

goehle:
Ah.  I added it.  Its turned on by default since its what mathematica does.

mgage:
I've used it on my laptop and on my iPad. If anything it works better on the ipad since you can do a flat rotation of the image with two fingers. (rotates in the plane of the screen). I haven't figured out the key combination to do that on the laptop even after looking at the instructions.

There may eventually need to be some tweaks needed to the theme. On the laptop I found reducing the type size to 0.1 from 0.15 and changing the coloring of the tick figures to red made it easier to solve the problems. I solved 15 or so of them in detail.

I think this should be pulled -- minor changes in default settings for colors and size can come later.

Issue #209, opened 2015-05-25
AskSage is partially broken


mgage:
Hi, 

AskSage seems to have stopped working — at least when it is using the fancier WEBWORK user variable return method.

I think the code needs to be revisited — in particular I suspect there are changes on the Sage end that will make the 
communication over JSON simpler.  I’m hoping someone will see some suggestions.  

All of the relevant WeBWorK code is at https://github.com/openwebwork/pg/blob/master/lib/WeBWorK/PG/IO.pm
the function query_sage_server https://github.com/openwebwork/pg/blob/master/lib/WeBWorK/PG/IO.pm#L223
and most importantly the function AskSage https://github.com/openwebwork/pg/blob/master/lib/WeBWorK/PG/IO.pm#L259

The formatting being done in lines 273–296 return output that looks like this:

IO::askSage: We have some kind of value |{"user_variables": {"WEBWORK": {"status": "ok", "data": {"text/plain": "{'diff': [4*u*v^2*cos(2*u^2*v^2) + 2*v^2*cos(2*u*v^2) 4*u^2*v*cos(2*u^2*v^2) + 4*u*v*cos(2*u*v^2)]\n [ 4/u 4/v]\n [ 14*u*v^2 14*u^2*v],\n 'function': (sin(2*u^2*v^2) + sin(2*u*v^2), log(9*u^2*v^2) + log(6*u^2*v^2), 7*u^2*v^2)}"}, "metadata": {}}}, "success": true, "execute_reply": {"status": "ok", "execution_count": 1, "payload": [], "user_expressions": {}, "user_variables": {"WEBWORK": {"status": "ok", "data": {"text/plain": "{'diff': [4*u*v^2*cos(2*u^2*v^2) + 2*v^2*cos(2*u*v^2) 4*u^2*v*cos(2*u^2*v^2) + 4*u*v*cos(2*u*v^2)]\n [ 4/u 4/v]\n [ 14*u*v^2 14*u^2*v],\n 'function': (sin(2*u^2*v^2) + sin(2*u*v^2), log(9*u^2*v^2) + log(6*u^2*v^2), 7*u^2*v^2)}"}, "metadata": {}}}}}| returned from sage at [PG]/lib/WeBWorK/PG/IO.pm line 306
This is relatively complicated and the only part we are using is:
my $sage_WEBWORK_data = $decoded->{execute_reply}{user_variables}{WEBWORK}{data}{'application/json'};
[...]

PR #212, opened 2015-05-27
Update AskSage to correspond to current configuration of SageMathCell…

mgage:
Update AskSage to correspond to current configuration of SageMathCell server


goehle:
For posterity.  This pull is a fix for issue #209

PR #214, opened 2015-06-08
Add the macros for communication with a remote R server

cubranic:
Add the macros for communication with a remote R server

### DIFF
--- a/lib/Rserve.pm
+++ b/lib/Rserve.pm
@@ -1,0 +1,67 @@
+package Rserve;
+
+use strict;
+use warnings;
+
+my $rserve_loaded = eval {
+    require Statistics::R::IO::Rserve;
+    1
+};
+
+sub access {
+    die 'Statistics::R::IO::Rserve could not be loaded. Have you installed the module?'
+        unless $rserve_loaded;
+    
+    Statistics::R::IO::Rserve->new(@_)
+};
+
+
+## Evaluates an R expression guarding it inside an R `try` function
+##
+## Returns the result as a REXP if no exceptions were raised, or
+## `die`s with the text of the exception message.
+sub try_eval {
+    my ($rserve, $query) = @_;
+
+    my $result = $rserve->eval("try({ $query }, silent=TRUE)");
+    die $result->to_pl->[0] if _inherits($result, 'try-error');
+    # die $result->to_pl->[0] if $result->inherits('try-error');
+
+    $result
+}
+
+
+## Returns a REXP's Perl representation, dereferencing it if it's an
+## array reference
+##
+## `REXP::to_pl` returns a string scalar for Symbol, undef for Null,
+## and an array reference to contents for all vector types. This
+## function is a utility wrapper to make it easy to assign a Vector's
+## representation to an array variable, while still working sensibly
+## for non-arrays.
+sub unref_rexp {
+    my $rexp = shift;
+    
+    my $value = $rexp->to_pl;
+    if (ref($value) eq ref([])) {
+        @{$value}
+    } else {
+        $value
+    }
+}
+
+
+## Reimplements method C<inherits> of class L<Statistics::R::REXP>
+## until I figure out why calling it directly doesn't work in the safe
+## compartment
+sub _inherits {
+    my ($rexp, $class) = @_;
+
+    my $attributes = $rexp->attributes;
+    return unless $attributes && $attributes->{'class'};
+    
+    grep {/^$class$/} @{$attributes->{'class'}->to_pl}
+}
+
+
+1;
--- a/macros/RserveClient.pl
+++ b/macros/RserveClient.pl
@@ -71,11 +71,13 @@ Other than backward compatibility, the only reason for using these
 functions is to start a new clean session within a single problem,
 which shouldn't be a common occurrence.
 
-=item rserve_start_plot [IMG_TYPE]
+=item rserve_start_plot [IMG_TYPE, [WIDTH, HEIGHT]]
 
 Opens a new R graphics device to capture subsequent graphics output in
 a temporary file on the R server. IMG_TYPE can be 'png', 'jpg', or
-'pdf', with 'png' as the default. Returns the name of the remote file.
+'pdf', with 'png' as the default. If left unspecified, WIDTH and
+HEIGHT, will use the R graphics device's default size. Returns the
+name of the remote file.
 
 
 =item rserve_finish_plot REMOTE_NAME
@@ -160,12 +162,15 @@ sub rserve_start_plot {
     _rserve_warn_no_config && return unless $Rserve->{host};
     
     my $device = shift // 'png';
+    my $width = shift // '';
+    my $height = shift // '';
 
     die "Unsupported image type $device" unless $device =~ /^(png|pdf|jpg)$/;
     my $remote_image = (rserve_eval("tempfile(fileext='.$device')"))[0];
     
     $device =~ s/jpg/jpeg/;
-    rserve_eval("$device('$remote_image')");
+    
+    rserve_eval("$device('$remote_image', width = ${width}, height = ${height})");
 
     $remote_image
 }
--- a/macros/RserveClient.pl
+++ b/macros/RserveClient.pl
@@ -1,0 +1,199 @@
+=head1 NAME
+
+RserveClient.pl - Macros for evaluating R code on an Rserve server
+
+=head1 SYNPOSIS
+
+=head1 SYNOPSIS
+
+    loadMacros('RserveClient.pl');
+    
+    rserve_start();
+    my @rnorm = rserve_eval("rnorm(15, mean=$m, sd=$sd)");
+    rserve_eval(data(stackloss));
+    my @coeff = rserve_eval('lm(stack.loss ~ stack.x, stackloss)$coeff');
+    rserve_finish();
+
+
+=head1 DESCRIPTION
+
+The macros in this file provide access to facilities of L<R
+statistical computing environment|http://www.r-project.org>,
+optionally located on another server, by using the
+L<Rserve|http://www.rforge.net/Rserve/> protocol. 
+
+B<IMPORTANT:> Before you can use these macros, you will need to
+configure the location of your Rserve host by adding it to
+C<$pg{specialPGEnvironmentVars}{Rserve}{host}>, for instance by
+appending the following line to F<webwork2/conf/localOverrides.conf>:
+
+    $pg{specialPGEnvironmentVars}{Rserve} = {host => "localhost"};
+
+Without this configuration in place, Rserve macros will only print out
+a warning about missing configuration and return C<undef>.
+
+=head1 MACROS
+
+The macros in this file set up a connection to the R server and
+pass a string parameter to R for evaluation.  The resulting
+vector is returned as a perl array object.
+
+=over 4
+
+=item rserve_eval REXPR
+
+Evaluates an R expression, given as text string in REXPR, on the
+L<Rserve|http://www.rforge.net/Rserve/> server and returns its result
+as a Perl representation of the L<Statistics::R::REXP> object.
+Multiple calls within the same problem share the R session and the
+object workspace.
+
+=item rserve_query
+
+Evaluates an R expression, given as text string in REXPR, in a
+single-use session on the L<Rserve|http://www.rforge.net/Rserve/>
+server and returns its result as a Perl representation of the
+L<Statistics::R::REXP> object.
+
+This function is different from C<rserve_eval> in that each call is
+completely self-enclosed and its R session is discarded after it
+returns.
+
+=item rserve_start, rserve_finish
+
+Start up and close the current connection to the Rserve server. In
+normal use, these functions are completely optional because the first
+call to C<rserve_eval> will call start the session if one is not
+already open. Similarly, the current session will be closed in its
+destructor when the current question goes out of scope.
+
+Other than backward compatibility, the only reason for using these
+functions is to start a new clean session within a single problem,
+which shouldn't be a common occurrence.
+
+=item rserve_start_plot [IMG_TYPE]
+
+Opens a new R graphics device to capture subsequent graphics output in
+a temporary file on the R server. IMG_TYPE can be 'png', 'jpg', or
+'pdf', with 'png' as the default. Returns the name of the remote file.
+
+
+=item rserve_finish_plot REMOTE_NAME
+
+Closes the R graphics capture to file REMOTE_NAME, transfers the file
+to WebWork's temporary file area, and returns the name of the local
+file that can then be used by the image macro.
+
+=item rserve_get_file REMOTE_NAME, [LOCAL_NAME]
+
+Transfer the file REMOTE_NAME from the R server to WebWork's temporary
+file area, and returns the name of the local file that can then be
+used by the C<htmlLink> macro. If LOCAL_NAME is not specified, the
+filename portion of the REMOTE_NAME is used.
+
+=back
+
+
+=head1 DEPENDENCIES
+
+Requires perl 5.010 or newer and CPAN module Statistics::R::IO, which
+has to be loaded in WebWork's Safe compartment by adding it to
+${pg}{modules}.
+
+
+=cut
+
+
+my $rserve;                     # Statistics::R::IO::Rserve instance
+
+
+sub _rserve_warn_no_config {
+    my @trace = split /\n/, Value::traceback();
+    my ($function, $line, $file) = $trace[0] =~ /^\s*in ([^ ]+) at line (\d+) of (.*)/;
+    
+    $PG->warning_message('Calling ' . $function .
+                         ' is disabled unless Rserve host is configured in $pg{specialPGEnvironmentVars}{Rserve}{host}')
+}
+
+
+sub rserve_start {
+    _rserve_warn_no_config && return unless $Rserve->{host};
+    
+    $rserve = Rserve::access(server => $Rserve->{host}, _usesocket => 1);
+
+    # Keep R's RNG reproducible for this problem
+    $rserve->eval("set.seed($problemSeed)")
+}
+
+
+sub rserve_finish {
+    $rserve->close() if $rserve;
+    undef $rserve
+}
+
+
+sub rserve_eval {
+    _rserve_warn_no_config && return unless $Rserve->{host};
+    
+    my $query = shift;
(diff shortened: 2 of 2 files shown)


jwj61:
Since this request is to pg, I am leaving it to Mike/Davide/Geoff to decide about (should the new file go here, or in the OPL macros?).

dlglin:
This actually raises the larger question of where contributed macros should live, as some are currently in the pg directory, and some are in the OPL.

On Jun 16, 2015, at 2:00 PM, John Jones <notifications@github.com<mailto:notifications@github.com>> wrote:


Since this request is to pg, I am leaving it to Mike/Davide/Geoff to decide about (should the new file go here, or in the OPL macros?).

mgage:
Danny,

It probably depends.  In this case -- where the macros will be widely used (eventually) I think importing these macros into pg is appropriate.    The macros in the OPL are also searched if the file they are in is placed in the macro search path in defaults.conf (or the override configurations).  That's probably the best place for macros associated with one or a small group of problems from a specific school.

We have in the past moved macros from an OPL location into the pg/macros file as they became more widely used.  We don't have a specific policy for doing this yet.

goehle:
A couple of questions. (Sorry for getting to this so late, I was away for a couple weeks.)

One for @cubranic Do you have a pg file I can use to test this macro?  

One for @mgage.  This macro file requires additional configuration to use.  Specifically you have to specify `$pg{specialPGEnvironmentVars}{Rserve} = {host => "localhost"};` in a conf file.  Does this mean that this macro should go in the OPL or are we ok with having macros in PG that dont work "out of the box".

cubranic:
@goehle Here is a PG file from our problem library: https://github.com/ubc/webwork-open-problem-library/blob/master/OpenProblemLibrary/UBC/STAT/STAT300/hw01/stat300_hw01_q02.pg

Also, the reason the additional configuration has to be done that way is so the macro can have a soft-fail with a reasonable message if it's not present. This was requested by @jwj61 and the way I implemented it was proposed on the mailing list a while back (probably around a year).

goehle:
That makes sense.  Since there needs to be an additional configuration line in localOverrides.conf could you make a pull request to webwork2 that has a commentented out configuration line in localOverrides.conf.dist along with a couple of lines of documentation explaining what it is and why you would set it.

goehle:
I'm having trouble getting this to run.  I've installed Statistics::R::IO::Rserve, and I installed an rserver and have it running.  When I try to run the problem I get
```
Can't locate object method "read" via package "IO::File" at /usr/local/share/perl/5.14.2/Statistics/R/IO/Rserve.pm line 38
```
Is there other configuration that I"m missing? 


Some other things you will need to add to the webwork2 pull. 
-  You should add Statistics::R::IO::Rserve to check modules (or at least write in localOverrides that it needs to be installed in order for the problems to work)
-  You should add Statistics::R::IO::Rserve to the `${pg}{modules}` list in `defaults.config` so that the package can be used in pg.

cubranic:
@goehle Sorry, I think I left you out of an email in which I described the pull request (it was part of a separate discussion on OPL). There are indeed several modules that need to be in `${pg}{modules}`, we have the list in our [installation guide](http://wiki.ubc.ca/Documentation:WeBWork/The_WeBWorKiR_Project:_Integrating_WeBWorK_with_R/Installation_guide).

dlglin:
My preference for add-ons like this would be to have a separate config file which includes all of the static configuration needed, and then a small section in localOverrides.conf referring to it, and including the server-specific settings.  Something like the following:

# To enable the Rserve integration, uncomment the following lines, and set the location of your R server:
# $pg{specialPGEnvironmentVars}{Rserve} = {host => "localhost”};
# include(“conf/Rserve.conf”);


This way, the user is only bothered with the things that they need to set.

Danny

On Jun 24, 2015, at 12:42 PM, Davor Cubranic <notifications@github.com<mailto:notifications@github.com>> wrote:


@goehle<https://github.com/goehle> Sorry, I think I left you out of an email in which I described the pull request (it was part of a separate discussion on OPL). There are indeed several modules that need to be in ${pg}{modules}, we have the list in our installation guide<http://wiki.ubc.ca/Documentation:WeBWork/The_WeBWorKiR_Project:_Integrating_WeBWorK_with_R/Installation_guide>.

goehle:
I'm not sure there is any static information other than the server address.  If its more than 5 or 10 lines of configuration then another file is definitely worth it.  Right now its just the one (and some modules that need to be added to the existing pg module list.)  

However ... the list you have on the installation guide includes IO::File.  This is not safe for us to include in the OK list for PG because pg problems could use it to open files on the server.  I'm not sure what to do at this point. It can be put in the OPL macros folder without any real issues, but I'm getting a little wary about adding it to pg.

mgage:
What exactly is used in IO::File?  Sometimes we can construct our own module that can be included and which allows only restricted
access to the IO::File (which does not need to be included in Safe).  This works well on older .pm modules that tend to do all of their setup at compile time.  If IO::File does a lot of runtime evaluation then it will be a pain in the neck.

Take care,
Mike

cubranic:
Maybe you can clarify for me an issue that led me to including `IO::File` in that list. It's not used in the macro file itself, only in the Statistics::R::IO distribution. Yet, without `IO::File` in `${pg}{modules}`, I was getting errors from WWSafe. Is there no way around it?

goehle:
Like mike said, sometimes we can create "dummy" packages which will fool Statistics::R::IO into thinking it has access to IO::File, and as long as you don't try to use anything in Statistics::R::IO that actually uses IO::File it can work.  However, I think part of the problem is that this is all buried under several levels of abstraction which makes it tough to tell what is exactly being used from the packages.

cubranic:
RserveClient.pl creates an instance of [Statistics::R::IO::Rserve](https://github.com/cubranic/Statistics-R-IO/blob/master/lib/Statistics/R/IO/Rserve.pm), which in turn loads:

```perl
use Moose;

use Statistics::R::IO::REXPFactory;
use Statistics::R::IO::QapEncoding;

use Socket;
use IO::Socket::INET ();
use Scalar::Util qw(blessed looks_like_number);
use Carp;

use namespace::clean;
```

So it doesn't really need it directly, but I guess via `IO::Socket::INET`.

mgage:
The usual problem is that Statistics::R::IO is trying to compile some part of itself at runtime.

That was what was going on with maketext() in Locale — the localization module. That was difficult to 
get to run inside of Safe.

Take care,

Mike

goehle:
Sockets in linux can be created and controled by opening files.  Its possible that is how it is being used.  If so then its likely integral to the network functionality of the whole thing.  Is there a client only version of the package that connects via http or something safer?

mgage:
The fact that it has use Moose;
could cause a problem.  I believe Moose does a lot of run time setup. 
Is there any clue where the problem is occuring?  I would expect some error 
such as “eval” caught at line …..  when operating inside Safe.

Take care,

Mike

cubranic:
@mgage This is the error:
```
 Can't locate object method "read" via package "IO::File" at /usr/local/share/perl/5.14.2/Statistics/R/IO/Rserve.pm line 38
   Died within Statistics::R::IO::Rserve::fh called at line 306 of /usr/local/share/perl/5.14.2/Statistics/R/IO/Rserve.pm
   from within Statistics::R::IO::Rserve::_send_command called at line 217 of /usr/local/share/perl/5.14.2/Statistics/R/IO/Rserve.pm
   from within Statistics::R::IO::Rserve::eval called at line 127 of [PG]/macros/RserveClient.pl
   from within main::rserve_start called at line 142 of [PG]/macros/RserveClient.pl
   from within main::rserve_eval called at line 46 of (eval 2393)
```

The line apparently causing it is a [call](https://github.com/cubranic/Statistics-R-IO/blob/master/lib/Statistics/R/IO/Rserve.pm#L306) to the `print` method on the socket handle:
```perl
$self->fh->print(pack('V4', $command, length($parameters), 0, 0) .
                     $parameters);
```

cubranic:
@goehle You're probably thinking of domain sockets (PF_UNIX) -- I don't think any file is created for IP sockets (PF_INET), which I use here. 

What I don't see is how is connecting via HTTP any safer. It would still ultimately have to use a socket, would it not?

goehle:
I was thinking along the lines of how geogebra or sage connects.  (Although
that is all client side so I suppose this is pretty different.)  In any
case it looks like the system is actually trying to use IO::File::read,
which can't be in the safe compartment.

dpvc:
What you probably want to do is make your own `.pm` file in `pg/lib` that provides an interface to the `Statistics::R::IO::Rserve` package and that can be included into the Safe compartment.  Your `.pm` file can use IO::File, since it is not itself in the safe compartment, but it would provide a package that *can* be imported into the safe compartment.  That package could be called from PG, and in term call the `Statistics::R::IO::Rserve` package.  Since your package is compiled outside the safe compartment, that would allow access to R through your interface.  It might be enough simply to provide a method that calls `Statistics::R::IO::Rserve->new()` and the rest could be in your `.pl` file as is.  If the `new` call is done in a `.pm` file rather than a `.pl` file, I think you will be OK.

cubranic:
@goehle If you're thinking of AskSage macro, that actually communicates by running an external `curl` process. It's "safe" because the function is defined in PG.pm, but I can't believe that's the best way to do it.

cubranic:
@dpvc Thanks! Would this file, let's call it lib/Rserve.pm, also be part of this pull request?

goehle:
I think its the same basic principle.  AskSage is defined outside of the safe compartment in Translator.pm (if its the one I'm thinking of) and the function itself is exported to the safe compartment.  You would add a `lib/Rserve.pm` file to this pull request as Davide suggested. Then you wouldn't need to add those modules to the safe list.  (Although you should still add them to check_modules.pm)

dpvc:
Yes.  You would also add `lib/Rserve.pm` to the `$pg{modules} list (in `webwork2/config/defaults.config`) so that your module is loaded, as I assume you did for `Statistics::R::IO::Rserve` for your current macros.  Of course, you would remove any such modules now, and only load `lib/Rserver.pm` instead.  That library would include `use Statistics::R` or whatever is needed.

It has been a while since I did this sort of thing, but that is my recollection of how it works.  Your `lib/Rserve.pm` could potentially be as simple as

```
use Statistics::R::IO::Rserve;

package Rserve;
sub access {Statistics::R::IO::Rserve->new(@_)};

1;
```

and then your `.pl` file would use `Rserve::access(...)` in place of `Statistics::R::IO::Rserve->new(...)`.

cubranic:
That's great, thanks @dpvc and @goehle, I'll update this request as you suggest.

cubranic:
I have updated the code (after rebasing it onto current `develop`) to refactor the interface to "Statistics::R::IO" to `lib/Rserve.pm`, which is now the only module required inside `${pg}{modules}`. (Actually, the required addition is: `[qw(Rserve Moose IO::File IO::Handle IO::Seekable)]`; without the additional modules, the code was not finding required methods when opening the connection to Rserve.) I tested it, and it works fine.

I have a couple of tidying-up questions at this point:

1. should I submit a pull request to openwebwork/webwork2 with the required entry for `${pg}{modules}` added to `defaults.conf`?
2. if the user doesn't have Statistics::R::IO installed, there will be a warning about `Rserve.pm` on every question render, even if it doesn't use the `RserveClient` macros. Should I replace `use ...` with a `require` in a `BEGIN` block and check this status in each call into it?
3. can I squash the various fixup commits in my branch?

goehle:
Unfortunately still requires us to add IO::File to the safe compartment, which we can't do.   However after many hours of poking I've found some magic which may fix things.  Or at the very least I was able to run the test problem posted above.  You should run tests yourself to make sure I'm not missing anything.  The main downside is that it requires a change to the Statistics::R::IO::Rserve package.  First, add `[Rserve IO::Handle Moose]` to the module list.  Then, make a local copy of Statistics::R::IO::Rserve and add the line 
```
           bless $fh, 'IO::Handle';
```
to line 30.  That will change the `$fh` generated by `socket` to be explicitly an instance of IO::Handle.  
In particular, the socket handle is being generated by `socket` and isn't actually a subclass of IO::Handle.  There is some magic typecasting going on when you call `$fh->read` which is turning $fh into an IO::File object.  However, since this typecasting happens at runtime, this occurs in the safe compartment which doesn't have access to IO::File. 

As for your questions.  
1.  Yes, you will need to add these modules to defaults.conf.  Once we get this pull request worked out that one will be quick. 
2.  In order for this fix to work WeBWorK  will have to have a special copy of Statistics::R::IO::Rserve, so there will always be a copy of that module available.  
3.  A good commit structure is nice, but not required.  :)

cubranic:
Thanks for looking into this in such detail Geoff. I’m happy to go with your suggestion, but can you clarify something about the safe compartment for me? When I have an entry such as `[Rserve Moose IO::Handle]` in `${pg}{modules}`, where the entry is a multi-package list, I thought only the very first one is put in the safe compartment and the rest are OK only during the initial load? So there is no danger of user code (i.e., a problem file) being able to directly call either Moose or IO::Handle, e.g., to instantiate a Moose object. (Or open files, as might be the case with IO::File.)

Re your answer #2: I’m the maintainer of Statistics::R::IO, so there is no problem of modifying the official release on CPAN, especially since your change is so minor. In addition, since S::R::IO depends on Moose, unless you want to add Moose as WebWork dependency, we should still probably replace `use` with `require` for a more fault-tolerant loading of Rserve.pm

goehle:
That's what I thought too, but I was wrong.  Turns out the intended purpose
is if you have multiple packages in one file you would do [mainfile
subpackage1 subpackage2].  The first file is evaled and imported and the
rest are just imported.  I think its mostly just luck that they all end up
being available.  In any case they are all shared with the safe compartment
https://github.com/openwebwork/pg/blob/master/lib/WeBWorK/PG/Translator.pm#L298

I was under the impression that use was just BEGIN { require thing; import
thing; }  What makes require more fault tolerant?

Also, since you have used Moose, you may be able to answer this.  Is it
dangerous to include it in the safe compartment.  Its big and complicated
and does dynamic method creation, but unless it allows users to eval code
or access files its probably ok.

Geoff.

cubranic:
> I was under the impression that use was just BEGIN { require thing; import
> thing; }  What makes require more fault tolerant?

I can include it in an `eval {}` block and set a status variable that's checked in `Rserve::access` to make 
sure the package is imported before calling it.

goehle:
Ah, sure.  Thats a good idea.

cubranic:
> Also, since you have used Moose, you may be able to answer this.  Is it
> dangerous to include it in the safe compartment.  Its big and complicated
> and does dynamic method creation, but unless it allows users to eval code
> or access files its probably ok.

I'm just a Moose user, so I can't really answer that. FWIW, it's the first OO package mentioned by `perlootut` and is probably the de-facto/future standard Perl's OO framework. If that's really a no-go, I can look into replacing it with one of the others mentioned in `perlootut`, although (as you might imagine), I'd love not to have to do that at this point.

goehle:
You aren't really using a lot of the fancier Moose functionality and you
dont have any inheritance or subclassing going on.  This afternoon I was
able to get a working version of Statistics::R::IO::Rserve running without
Moose at all.  (Basically you just have to manually define the constructor
and all of the accessors.)  If you were to do anything, I would remove
Moose altogether.  Doing so would make the package a lot lighter and may
improve load times (since this stuff is getting compiled every time a
student views a problem or hits submit).  However, its not really that big
a deal and Moose is probably fine.

mgage:
Hi Goeff,

Did you use Class::Accessor ?  I used that in attempts table to streamline
the constructor/accessor/setter code. It does the job without all the infrastructure 
required by Moose.

I even added the “antlers” option as an option since
that impliments moose-like syntax, but I ended up not using it and I believe we’re removing it 
because it caused some trouble with installation on Danny’s system.

Take care,

mike

cubranic:
Hm, [`Class::Accessor`](https://metacpan.org/pod/Class::Accessor) doesn't have destructors, which I use for closing the connection to Rserve. (Which may be done by Perl too during GC, but I'm not sure about it.) Following up on [`perlootut`](https://metacpan.org/pod/distribution/perl/pod/perlootut.pod), I looked at `Class::Tiny` and saw that by using [`Class::Tiny::Antlers`](https://metacpan.org/pod/Class::Tiny::Antlers), I'd get a near-Moose syntax so hopefully significantly reducing my work to switch to it. And all this with no non-core dependencies!

I hope there isn't a catch in the fine print that I missed. I'll have to give this a try and will report back with my findings.

cubranic:
I created a very simple class hierarchy using `Class::Tiny::Antlers` that captures pretty much everything I used in Moose and tested out using this from a Webwork macro similar to RserveClient.pl. It worked once I added `Class::Tiny` to the safe compartment (the line in defaults.conf is `[qw(Rserve Class::Tiny)]`. So if this would be an acceptable solution to you, I'll proceed to convert Statistics::R::IO to use this OO framework and then update this PR.

@mgage, when you used `Class::Accessor`, did you need to add it to the safe compartment? I was mildly disappointed to see that Class::Tiny was still required.

goehle:
That solution works for me.  Thanks for going the extra mile and slimming
the code down a bit.

P.G.  Mike used Class::Accessor on the WeBWorK side, so Class::Accessor is
not in the safe cmpt.


Geoff.

cubranic:
Just a quick update: I converted Statistics::R::IO to use `Class::Accessor` and it looks good so far. All of my tests pass and the test WebWork problem still works correctly. I still need to release it on CPAN and then just need to clean up the code in this pull request before it's ready.

cubranic:
I released a new version of Statistics::R::IO that uses Class::Tiny and it is available through CPAN (v1.0001). So this should now work.

The line in wwk2/defaults.conf for adding the required modules to the safe compartment now looks like this:
```perl
[qw(Rserve Class::Tiny IO::Handle)],
```
If you can test it and give me OK from your end, I'll rebase onto current `develop` and squish into a single commit, and then submit a separate pull request on `webwork2/develop` for the configuration change.

goehle:
I tested this and it worked with a sample problem.  I also tested it on a server that didn't have Rserve or R::IO::Statistics and it didn't complain when running regular problems.  Looks good to me.  Thanks for going the extra mile.

cubranic:
OK, I cleaned up the commits and opened the PR for the config file: openwebwork/webwork2#700.

goehle:
This works for me on a couple of problems and doesn't cause issues when its not configured on non rserve problems.

PR #215, opened 2015-06-13
Add more moodle prefixes

mgage:
Add more moodle prefixes


goehle:
I tested this and it works with a vanilla installation and the prefix is added if I set the appropriate flag.

PR #216, opened 2015-06-25
Add zero height paragraph break to html at start of problems

jwj61:
This will help knowls open in better places.  I tried different variations (with and without tagging menus, with and without filepaths printed) and saw now difference (linux/chrome) except for where knowls open.


goehle:
Was this supposed to change the behavior in this bug:  http://bugs.webwork.maa.org/show_bug.cgi?id=3384

In particular if I use the problem PG code in that bug the knowls still opens below the paragraph, which seemed to be the original complaint.

jwj61:
It was motivated by that bug report.

I think the consensus was that with the code in that bug report, the knowl should open at the bottom because it is one big paragraph.  If you insert a $PAR somewhere in the middle of the problem (or at the blank line right after the knowlLink), then without this change the knowl opens at the end, but with this change it should open at the paragraph break.

goehle:
Ok, in that case this works as advertised.

PR #221, opened 2015-07-02
Create contextTrigDegrees.pl

paultpearson:
A new macro that defines a context TrigDegrees where trig functions are defined in degrees.  Using this context, trig functions will be evaluated in Perl mode as well as within MathObjects in degrees.  Derivative methods are defined on these trig functions as well.  This version of contextTrigDegrees.pl does not have the capability for using these trig functions in vector or matrices.  Most of the heavy lifting was done by Davide Cervone (thanks, Davide!).

### DIFF
--- a/macros/contextTrigDegrees.pl
+++ b/macros/contextTrigDegrees.pl
@@ -1,0 +1,177 @@
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright � 2000-2007 The WeBWorK Project, http://openwebwork.sf.net/
+# $CVSHeader$
+# 
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+# 
+# This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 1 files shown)


goehle:
After testing this for a bit it looks good.  The functions all behave appropriately and it works very well with math objects.  (I was impressed that it even changed the automatic derivative of sin/cos/etc... to be correct.)

One thing for posterity.  Through no fault of the authors, if this module is used in parserCustomization this will lead to some broken problems.  In particular problems with various components specified "manually" will likely be wrong.  For example: 
1.  Derivatives of trig functions where the solution has been manually specified (e.g. Utah/Calculus_I/set4_The_Derivative/1210s4p3.pg).
2.  Problems with graphs of trig functions (e.g. ASU-topics/setTrigGraphs/p5.pg).  If you add contextTrigDegrees to this problem (manually) then it changes how the graph and the answer behave, but breaks the problem overall.

dpvc:
@Geoff, this context was not really intended to be used as a course-wide preference; I think Paul just intended it to be used in specific problems where degrees were desired by the problem author.  I suspect that changing to log base 10 in the preferences would have a similar (though not quite so widespread) effect.

goehle:
Makes sense.  I just wanted to put it out there since was something that
sprang to mind, for me at least.  For single problem use it works just
fine.

PR #224, opened 2015-07-07
Adjust assignments so that they are not treated as lists

dpvc:
This fixes a problem where an assignment can be incorrectly treated as a list during answer checking.  You can test the patch using

```
loadMacros("parserAssignment.pl");

Context("Numeric")->variables->add(y=>'Real');
parser::Assignment->Allow;

$f = List("3x+1");

TEXT(ans_rule(20));
ANS($f->cmp);
```
and entering the answer `y=1` in the answer blank.

Before the patch, you will get no warning message about the incorrect type of answer, and after, you will get a message indicating that the answer isn't a formula returning a number.

### DIFF
--- a/macros/parserAssignment.pl
+++ b/macros/parserAssignment.pl
@@ -260,6 +260,20 @@ sub new {
   }
 }
 
+#
+#  Mark assignments so that they are not treated as lists by classMatch()
+#
+sub make {
+  my $self = shift;
+  $self = $self->SUPER::make(@_);
+  $self->{isList} = 0;
+  return $self;
+}
+sub class {"Assignment"}
+
+#
+#  Produce proper output
+#
 sub string {
   my $self = shift; my ($x,$v) = $self->value;
   $x->string . ' = ' . $v->string;
@@ -296,6 +310,7 @@ sub typeRef {
   my $self = shift;
   Value::Type('Assignment',2,$self->{data}[1]->typeRef,list=>1);
 }
+
 
 ######################################################################
 


goehle:
I'm getting that error message before the patch as well.  Is there something else to the example problem?

dpvc:
Sorry, the answer should have been `y=1` not `y=x` (it is only in the constant case that the problem occurs).

goehle:
OK.  Works now.  I just wanted to be sure that the problem didn't also depend on some other component that needed some thing or another.

dpvc:
Here is another example that shows what is going on:

```
loadMacros("parserAssignment.pl");

Context("Numeric")->variables->add(y=>'Real');
parser::Assignment->Allow;

$f = List("x");

TEXT(ans_rule(20));
ANS($f->cmp(checker=>sub {
  my ($correct,$student) = @_;
  WARN_MESSAGE($correct,$student);
  return 0;
}));
```

Without the patch, `y=1` as an answer should give you four warning message blocks showing that the `x` is checked against the `y` and the `1` separately (incorrect), while with the patch, you should see two checks, both with `x` against `y=1` (correct).

goehle:
Does this mean there is a context out there where = is a list separator?
Or is that some behind the scenes stuff?

dpvc:
It is behind-the-scenes stuff.  The object that implements the equality in this case is a subclass of List (it is a pair that has the variable together with the equation it is equal to).  But the context does not have `=` as a list separator in the sense that `,` is a list separator.

PR #226, opened 2015-07-08
Create unique UUID for graphs created on the fly

mgage:
This abstracts one subroutine in PGalias.pm so that PGresources can be created independently of being placed in the PGalias resource lists.

This is used in three lines in PGgraphmacros.pl to take an onTheFlyImage with a label like image1
and create a unique_id which contains no obvious information.  These methods from PGresources
are already used to create unique ids for .html auxiliary files and for static .png and .gif files so this is seems a good way to resolve the image conflict problem.  

It also gives the names for on the fly graphic files the same anonymity that other auxiliary files have had.  You cannot deduce the student login, set or problem number from these names although they are involved in creating it.

(PGalias could be extended to handle other files in the same way -- .js files, .css files, etc. )

### DIFF
--- a/lib/PGalias.pm
+++ b/lib/PGalias.pm
@@ -158,7 +158,19 @@ sub check_parameters {
 	warn "htmlURL is not defined." unless $self->{htmlURL};
 	warn "tempURL is not defined." unless $self->{tempURL};
 }
-
+sub make_resource_object {
+	my $self = shift;
+	my $aux_file_id =shift;
+	my $ext = shift;
+	my $resource = PGresource->new(
+		$self,                    #parent alias of resource
+		$aux_file_id,             # resource file name
+		$ext,                     # resource type
+		WARNING_messages => $self->{WARNING_messages},  #connect warning message channels
+		DEBUG_messages   => $self->{DEBUG_messages},
+	);	
+	return $resource;
+}
 sub make_alias {
    	my $self = shift;   	
    	my $aux_file_id = shift;
@@ -226,13 +238,12 @@ sub make_alias {
 	###################################################################
 	unless ( defined $self->get_resource($aux_file_id.".$ext") ) {
     	$self->add_resource($aux_file_id.".$ext", 
-    	                    PGresource->new(
-    	                            $self,                    #parent alias of resource
-    	                            $aux_file_id,             # resource file name
-    	                            $ext,                     # resource type
-    	                            WARNING_messages => $self->{WARNING_messages},  #connect warning message channels
-                                    DEBUG_messages   => $self->{DEBUG_messages},
-    	));
+    						$self->make_resource_object(
+    							$aux_file_id,  # resource file name
+    							$ext           # resource type
+    						)
+    	                   
+    	);
 
     } else {
     	#$self->debug_message( "found existing resource_object $aux_file_id");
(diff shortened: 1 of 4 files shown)


goehle:
There are a couple of other places I found where the image name is set which will also need this technique. 
-  `graph_cmp` in AnswerChecker.pm (although this may be a debugging tool)
-  `init_graph_no_labels` in PGgraphmacros.pl

Also, you might consider altering `insertGraph` so that it checks the set name instead of the imageName for Undefined_Set.  Right now it will never find undefined set which means that check is useless.  I would recommend keeping it though, because the hashing process takes the pg path into account, but not the pg code itself.  I can see issues with collisions using the xmlrpc service with pg code provided directly and no path, or set, or problem number, or any other useful identifying information.  

Other than that, this seems like a fine solution.  It saves reproducing the hashing process.

dpvc:
The `graph_cmp` function is part of the Formula debugging process when `ANS($f->cmp(diagnostics=>1))` is set.  It is not something that will be in problems generally, and will only be used when viewing an individual problem, so I don't think there should be collisions.

mgage:
Good suggestions.  I'll do that.

Mike

Sent from my iPad

On Jul 8, 2015, at 8:54 AM, Geoff Goehle <notifications@github.com<mailto:notifications@github.com>> wrote:


There are a couple of other places I found where the image name is set which will also need this technique.

  *   graph_cmp in AnswerChecker.pm (although this may be a debugging tool)
  *   init_graph_no_labels in PGgraphmacros.pl

Also, you might consider altering insertGraph so that it checks the set name instead of the imageName for Undefined_Set. Right now it will never find undefined set which means that check is useless. I would recommend keeping it though, because the hashing process takes the pg path into account, but not the pg code itself. I can see issues with collisions using the xmlrpc service with pg code provided directly and no path, or set, or problem number, or any other useful identifying information.

Other than that, this seems like a fine solution. It saves reproducing the hashing process.

-

goehle:
Some problems with multiple graphs are not getting unique ids.  See for instance 
Library/FortLewis/Algebra/5-4-Equations-for-lines/MCH1-5-4-11-Equations-for-lines.pg

goehle:
The issue is that you are setting path to __ in PGgraphmacros.  However the path is what is used to create the unique ID and needs to be unique to the resource (locally within the problem).  So changing the path to be equal to the aux_file_id (i.e. `image$imageNum`) or using the aux_file_id in the unique_id would fix it.

mgage:
I'll check this when I get back to the house.  It's supposed to use image$imageNum in the unique_id but not in the uniq_id stub (where path is a required segment).   For static .png the path is set to the location on the hard disk.  Since this is "onthefly" there is not yet any hard disk location, hence __.   If necessary we can use path to further differentiate (enter some hash of the code for example) but it shouldn't be needed.   Thanks.

mgage:
The PGresource code wasn't including the resource id in the UUID !!! (sigh).  The image number was updating properly but it wasn't being added to the string from which the UUID was created.

goehle:
That seems to have fixed it.  Now I get different UUID's when using multiple images in one problem.  I checked and all the places I was getting collisions before have been fixed.  This also fixes issues of graphs getting switched around when problems are reordered.

PR #228, opened 2015-07-08
Fix a number of issues with PGML 

dpvc:
This pull request fixes some bugs in PGML and adds some missing features that were identified during the recent PREP2015 problem-authoring course.  Among these are:

* Fix problems with variable substitution for array elements and hash references.
* Add support for performing LaTeX math substitution on the results command and variable substitutions.
* Add support for automatic reductions in math specified using MathObject algebra notation
* Add code blocks via triple back-ticks, with optional class name that can be used for syntax highlighting (though that is not included in PGML itself).

To test these features, use

```
loadMacros("PGML.pl");

@a = (" OK "); $i = 0;
$a = {x => " OK "};

BEGIN_PGML
x[$a[$i]]x

x[$a->{x}]x
END_PGML
```

to test that array and hash references can be substituted (the unpatched version should give two lines with two adjacent x's, while the patched version should have " OK " between the two x's).

---

```
loadMacros("PGML.pl");

$a = "math: \(x+1\over x-1\)";

$b = PGML::Format('[$a]***');
TEXT ($b =~ m/~~~~~~(/ ? "YES" : "NO"," ");
$b = PGML::Format('[@ $a @]***');
TEXT ($b =~ m/~~~~~~(/ ? "YES" : "NO");
```

This tests for processing of math mode (using the new triple star) after variable and command substitution.  Without the patch you should see "YES YES" indicating that the `\(` is still present in the output,while with the patch you should get "NO NO".

---

```
loadMacros("PGML.pl");

BEGIN_PGML
[:1+(2-3)x:]

[:1+(2-3)x:]**

[:1+(2-3)x:]***
END_PGML
```

This tests the new double and triple star for algebra-format math.  In the unpatched version, you will see three copies of the equation `1+(2-3)x`, the second and third with stars after them.  In the patched version, you should see `1+(2-3)x`, then `1+(-1)x`, and finally `1-x`.

---

    loadMacros("PGML.pl");
    
    BEGIN_PGML
    ``` html
    <p>
      <b>bold</b> and <i>italics</i>
    </p>
    ```
    END_PGML

Here, the output should be verbatim (you will see the HTML tags as they appear here), and the result should be in a `<code>` tag that has `class="html"`.

### DIFF
--- a/macros/PGML.pl
+++ b/macros/PGML.pl
@@ -39,6 +39,7 @@ my $heading = '#+';
 my $rule = '(?:---+|===+)';
 my $list = '(?:^|(?<=[\t ]))(?:[-+o*]|(?:\d|[ivx]+|[IVX]+|[a-zA-Z])[.)]) +';
 my $align = '>> *| *<<';
+my $code = '```';
 my $pre = ':   ';
 my $emphasis = '\*+|_+';
 my $chars = '\\\\.|[{}[\]\'"]';
@@ -48,7 +49,7 @@ my $close = '(?:[!>%@$]|::?|``?| ?\|+)\]';
 my $noop = '\[\]';
 
 my $splitPattern =
-  qr/($indent|$open|$ansrule|$close|$linebreak|$lineend|$heading|$rule|$list|$align|$pre|$emphasis|$noop|$chars)/m;
+  qr/($indent|$open|$ansrule|$close|$linebreak|$lineend|$heading|$rule|$list|$align|$code|$pre|$emphasis|$noop|$chars)/m;
 
 my %BlockDefs;
 
@@ -171,6 +172,7 @@ sub All {
     /-|=/         && do {return $self->Rule($token)};
     /<</          && do {return $self->Center($token)};
     />>/          && do {return $self->Align($token)};
+    /```/         && do {return $self->Code($token)};
     /:   /        && do {return $self->Preformatted($token)};
     $self->Text($token);
   }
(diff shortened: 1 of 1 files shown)


goehle:
All of these changes check out.  I have a couple of questions if you will indulge me.  
-  Is there a cheat sheet with all of the symbols and brackets and what they do?  
-  What context is the math mode *** using to reduce.  In particular having a similar math objects function and using `$f->reduce()` only gets me to the ** level of reduction.  I'm having trouble following the difference between double stars and triple stars in `sub Math`.

dpvc:
> Is there a cheat sheet with all of the symbols and brackets and what they do?

The wiki page for [PGML](http://webwork.maa.org/wiki/Category:PGML) does document them all, but they aren't all in one place.  The [PGML lab](https://courses1.webwork.maa.org/webwork2/cervone_course/PGML/1?login_practice_user=true) had reference menus that list all the notations, but it hasn't been updated to include the new ones that are in this pull request.  It is also a little out of date in terms of the layout of the lab (due to changes in the theme from when it was originally written).  I suppose an additional page in the wiki that lists everything would be a good idea.

dpvc:
> What context is the math mode `***` using to reduce.

The `[: ... :]` algebra notation is in the `Typeset` context from the `contextTypeset.pl` macro file.  With one star, it uses the current context instead.  With two stars it is the `Typeset` context with `reduce()`, and with three stars, it is the current context with `reduce()`.  The number of stars is basically a 2-bit number where the low-order bit determines the context (0 = `Typeset`, 1 = current) and the high-order bit determines whether `reduce()` is used (0 = not reduced, 1 = reduced).

PR #230, opened 2015-07-22
Popup menus

dpvc:
This pull request updates the MathObject popup menu object to allow randomization of the choices (similar to the new RadioButton mechanism), and also allows arbitrary strings for the menu choices.  Since menus can't include math or images or other markup, the text of the choices is protected so that it appears correctly in the menus as well as in the results table.  You can also use numeric values to select the correct answer, just as with the RadioButtons object.

To check, use the problem

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
  "parserPopUp.pl",
);

TEXT(beginproblem());

$item1 = '"> <b> \ &%{}_^$#~';
$popup = PopUp(["?",$item1,"Item 2","Item 3"],$item1);
BEGIN_TEXT
\{$popup->menu\}
END_TEXT
ANS($popup->cmp);

ENDDOCUMENT();
```

Without the patch, the first menu item will not show the `">`, the `<b>` or the `</b>`, and the rest of the menu item will be repeated.  Selecting the item will not result in a correct answer, and no answer will show in the results table.  With the patch, the menu item will appear correctly in the menu, as well as in the entered, preview, and correct answer columns of the results table.  It will also show correctly in hardcopy.

You can use

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
  "parserPopUp.pl",
);

TEXT(beginproblem());

$popup = PopUp(["?",["Item 1","Item 2","Item 3"]],1);
BEGIN_TEXT
\{$popup->menu\}
END_TEXT
ANS($popup->cmp);

ENDDOCUMENT();
```

to check that randomization works, and that the numeric selection of correct answers works (the correct answer should be "Item 1" no matter where it falls in the randomization).

### DIFF
--- a/macros/parserPopUp.pl
+++ b/macros/parserPopUp.pl
@@ -165,7 +165,7 @@ sub MENU {
   my $answer_value = (defined($main::inputs_ref->{$name}) ? $main::inputs_ref->{$name} : '');
   my $label = main::generate_aria_label($name);
   if ($main::displayMode =~ m/^HTML/) {
-    $menu = qq!<select class="pg-select" name="$name" id="$name" aria-label="$label" size="1">\n!;
+    $menu = qq!<select class="pg-select" name="$name" id="$name" aria-label="$label" size="1" style="max-width:100%">\n!;
     foreach my $item (@list) {
       my $selected = ($item eq $answer_value) ? " selected" : "";
       my $option = $self->quoteHTML($item,true);
--- a/macros/parserPopUp.pl
+++ b/macros/parserPopUp.pl
@@ -20,8 +20,9 @@ parserPopUp.pl - Pop-up menus compatible with Value objects.
 
 =head1 DESCRIPTION
 
-This file implements a pop-up menu object that is compatible
-with Value objects, and in particular, with the MultiAnswer object.
+This file implements a pop-up menu object that is compatible with
+MathObjects, and in particular, with the MultiAnswer object, and with
+PGML.
 
 To create a PopUp object, use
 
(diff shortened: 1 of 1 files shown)


macros/parserPopUp.pl
```diff
@@ -90,51 +121,73 @@ sub new {
...
+  my @list = @{$self->{choices}}; my $menu = "";
   $name = main::NEW_ANS_NAME() unless $name;
   my $answer_value = (defined($main::inputs_ref->{$name}) ? $main::inputs_ref->{$name} : '');
   my $label = main::generate_aria_label($name);
   if ($main::displayMode =~ m/^HTML/) {
-    $menu = qq!<SELECT class="pg-select" NAME="$name" id="$name" aria-label="$label" SIZE=1>\n!;
-    foreach my $option (@$list) {
-      my $selected = ($option eq $answer_value) ? " SELECTED" : "";
-      $menu .= qq!<OPTION$selected VALUE="$option">$option</OPTION>\n!;
+    $menu = qq!<select class="pg-select" name="$name" id="$name" aria-label="$label" size="1" style="max-width:100%">\n!;
```
goehle: If this is to address a math4 issue should it be changed in math4 instead?

dpvc: Probably, but I don't know why the `max-width` is set to 95% in math4.css (I can't think of any reason for that), and so didn't want to make a change whose effect I didn't understand.  Do you think it turns to specify 100% explicitly here?

goehle: Max width was 95% in math4 because without a max width input boxes and their like were spilling over the problem well on small screens.  Changing it to 95% made it so that even if the screen is small there is always a little boundary between the input elements and the problem well.   If there is a better way to acomplish that, I'm happy to change it.  I don't think it hurts anything to specify 100% here, other than the general principle that explicitly setting css makes it harder for things like themes to modify the look of page elements.

dpvc: OK, the `max-width` of 95% was causing the longest menu item to be truncated when it was selected.  That was because there was nothing else in the test problem except the menu, and so the container for the menu was just big enough for the menu, and 95% of that was smaller than the menu.  In a real problem, there is likely to be text that makes the problem wider, so I guess it is not a problem, though I still think it is a work-around that can have unexpected consequences.  For example, if you use popups inside a table (popups in their own cells), then they will by 95% of their usual size even on large screens.  I'm wondering if a media-specific CSS setting might be a better solution for small screen issues than making global settings like this.

I will remove the commit that made this change.  I suspect that there will be people who make wide menus by putting in long choice texts, and reducing the size on small screens may be important.

goehle: Making the max-width media specific is a good idea.  I'll put that in my tweaks pull request.

goehle:
Works as described.  I was able to make radio buttons, use the randomization, use the alternate way of providing the index.  It also deals with special characters correctly and all of the old radio button problems I tried still worked.

Issue #184, opened 2014-12-19
Gateways and Labelled Answers


goehle:
I'm getting the following error when viewing the problem pasted below on gateways, but not on standard homeworks.  

```
There is no answer evaluator for the question labeled first_answer at /opt/webwork/pg/lib/WeBWorK/PG/Translator.pm line 1218
Error in Translator.pm::process_answers: Answer first_answer:<br/>
Unrecognized evaluator type || at /opt/webwork/pg/lib/WeBWorK/PG/Translator.pm line 1242
Error in Translator.pm::process_answers: Answer first_answer:<br/>
```

from the following problem

```
DOCUMENT();
loadMacros(
        "PGbasicmacros.pl",
        "PGchoicemacros.pl",
        "PGanswermacros.pl"
);
TEXT(beginproblem(), $BR,$BBOLD, "Conditional questions example", $EBOLD, $BR,$BR);
$showPartialCorrectAnswers = 1;

$a1 = random(3,25,1);
$b1 = random(2,27,1);
$x1 = random(-11,11,1);
$a2 = $a1+5;

BEGIN_TEXT
If \( f(x) = $a1 x + $b1  \), find \( f'( $x1 ) \).
$BR $BR \{NAMED_ANS_RULE('first_answer',10) \}
$BR
END_TEXT



$ans_eval1 = num_cmp($a1);       
NAMED_ANS(first_answer => $ans_eval1);                                  

# Using named answers allows for more control.  Any unique label can be
# used for an answer.    
# (see http://webwork.math.rochester.edu/docs/docs/pglanguage/pgreference/managinganswers.html
# for more details on answer evaluator formats and on naming answers
# so that you can  refer to them later.  Look also at the pod documentation in
# PG.pl and PGbasicmacros.pl  which you can also reach at
# http://webwork.math.rochester.edu/docs/techdescription/pglanguage/index.html)

# Check to see that the first answer was answered correctly.  If it was then we
# will ask further questions.
$first_Answer = $inputs_ref->{first_answer};  # We need to know what the answer
                                                                                          # was named.
$rh_ans_hash = $ans_eval1->evaluate($first_Answer);

# warn pretty_print($rh_ans_hash);  # this is useful technique for finding errors.
                                    # When uncommented it prints out the contents of 
                                    # the ans_hash for debugging

# The output of each answer evaluator consists of a single %ans_hash with (at
# least) these entries:
#       $ans_hash{score}        -- a number between 0 and 1
[...]

PR #232, opened 2015-08-03
Make named answer rules work properly with MathObject answers in PGML

dpvc:
This fixes a problem where named answer rules didn't work with answers in PGML that come from MathObjects.  To test, use

```
loadMacros("PGML.pl");

BEGIN_PGML
[______]{1}{name=>"myAnswer"}
END_PGML
```

Before the patch, this should produce an error about the answer with label "myAnswer" not having an associated answer checker.  With the patch, the problem should produce no error and will properly grade student answers.

You can also try

```
loadMacros("PGML.pl");

BEGIN_PGML
[______]{name=>"myAnswer"}
END_PGML
NAMED_ANS(myAnswer => Real(1)->cmp);
```

### DIFF
--- a/macros/PGML.pl
+++ b/macros/PGML.pl
@@ -985,7 +985,13 @@ sub Answer {
       }
       $rule = $ans->$method(@options);
       $rule = PGML::LaTeX($rule) if $item->{hasStar};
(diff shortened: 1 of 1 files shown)


goehle:
This fix works as advertised.  

Note:  Named answers don't work in gateway mode, but they don't work for either PG or PGML as noted in https://github.com/openwebwork/pg/issues/184

PR #237, opened 2015-09-03
loadFiles subdirectory bug

goehle:
This was an extra line left over from some testing which prevented `loadModules` from using paths.  Test by trying to load setOrientation problem 14.  Before the pull you should get a missing macro warning and after you shouldn't.

### DIFF
--- a/lib/PGloadfiles.pm
+++ b/lib/PGloadfiles.pm
@@ -141,7 +141,6 @@ sub loadMacros {
    
     while (@files) {
         $fileName = shift @files;
-        $fileName = WeBWorK::PG::IO::fileFromPath($fileName);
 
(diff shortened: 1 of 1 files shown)


mgage:
These are ok.  There needs to be some changes to localOverrides.conf which I'll submit as part of the ltiups3 pull request. 

 $pg{specialPGEnvironmentVars}{CAPA_Tools}             = "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_Tools/",
 $pg{specialPGEnvironmentVars}{CAPA_MCTools}           = "$courseDirs{templates}/Contrib/CAPA/macros/CAPA_MCTools/",

(before the paths were written in full and read_whole_problem_file won't handle those. )
This also requires a link Contrib in the templates directory of each course.

PR #240, opened 2015-09-23
Patch PGML

dpvc:
This patch fixes two issues with PGML and empty strings in `[@ ... @]` substitutions, and also has PGML use `WARN_MESSAGE` rather than `warn` for its error reporting.

The first error was with empty string literals in `[@ ... @]`, as in `[@ "" . "X" @]`, which would produce no output, and would generate

```
==================
Errors parsing PGML:
Error evaluating command: Can't find string terminator '"' anywhere before EOF.

==================
```

in the Apache error log file.  A test case is the problem:

```
DOCUMENT();
loadMacros("PGstandard.pl","PGML");
BEGIN_PGML
No [@ "" . "X" @] output
END_PGML
ENDDOCUMENT();
```

Before the patch, this will generate `No  output` and the warning message in the log file.  With the patch, it should generate `No X output` and no error message in the log.

The second issue is that paragraph breaks are not being combined when an empty `[@ ... @]**` substitution appears between them.  So a test case would be

```
DOCUMENT();
loadMacros("PGstandard.pl","PGML");
BEGIN_PGML
x

[@ @]**

x
END_PGML
ENDDOCUMENT();
```

WIthout the patch, this should produce the error

    Can't locate object method "pushItem" via package "PGML::Item"

in the web page.  After the patch, the problem should display as

    x
    
    x

with no error messages.

To test the use of `WARN_MESSAGE` use

```
DOCUMENT();
loadMacros("PGstandard.pl","PGML.pl");
BEGIN_PGML
[@ xyz() @]
END_PGML
ENDDOCUMENT();
```
With the patch, you should see

```
==================
Errors parsing PGML:
Error evaluating command: Undefined subroutine &main::xyz called. 
==================
```
in the pink PG warning messages area.

### DIFF
--- a/macros/PGML.pl
+++ b/macros/PGML.pl
@@ -504,7 +504,9 @@ sub terminateGetString {
 
 sub terminateBalance {
   my $self = shift; my $token = shift;
-  my $block = $self->{block}; my $stackString = $self->stackString;
+  my $block = $self->{block};
+  if (!$block->{stack} || scalar(@{$block->{stack}}) == 0) {$self->Text("",1)}
+  my $stackString = $self->stackString;
   $self->{block} = $block->{prev}; $self->{block}->popItem;
   if ($block->{token} eq '"' || $block->{token} eq "'") {
     $self->Item("quote",$block->{token});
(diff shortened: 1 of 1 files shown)


dpvc:
PS, this addresses bugzilla issues [3409](http://bugs.webwork.maa.org/show_bug.cgi?id=3409) and [3460]( http://bugs.webwork.maa.org/show_bug.cgi?id=3460).

PR #243, opened 2015-09-29
Fix Parser::UOP so that it will add parentheses for implied multiplication when needed

dpvc:
This resolves [bugzilla 3463](http://bugs.webwork.maa.org/show_bug.cgi?id=3463).

To test, use

```
$F = Formula("3*(-x)");
BEGIN_TEXT
\(\{$F->TeX\}\)
END_TEXT
```

Without the patch, the output will be "3-x", and with the patch it should be "3(-x)".  (Of course, `$F = Formula("3*(-x)")->reduce` produces a better result.)


goehle:
The test case didn't work for me, I had to use 
```
$F = Compute("3*(-x)");
Context()->normalStrings;
BEGIN_TEXT
\(\{$F->TeX\}\)
\{$F->ans_rule()\}
END_TEXT
ANS($F->cmp);
```
and actually look at the "correct answer".  However the patch fixes the issue.

PR #244, opened 2015-09-30
Add problem seed to PGalias.

goehle:
Add problem seed to string used to generate aliases.  Causes images to be regenerated if seed changes, but also causes some duplication when linking to files that are not seed dependent.

To test:  
-  Look at a problem with an image (say Demo set #7).  Change the seed using Problem Set Detail.  Before the patch the image won't change.  After the patch the image will have a new file name and will be updated. 

This addresses webwork2 issue 644


mgage:
Checked that the file name changes when the seed is changed in the editor and then viewed.  Only the first section of the file name changes since it is the uniq stub which includes the problem seed. 

Also used the check suggested above on problem 7 in Demo set.

PR #248, opened 2015-10-30
Develop include pg updates

mgage:
These are minor changes that insure that values that are evaluated in Translator are always defined.  The errors are normally trapped in the standard usage of WeBWorK but they produce runtime warnings printed to STDERR when the problems are run using the standaloneProblemRender.

This should be tested to see that no harm has been done -- it has already been tested to see that the error messages cease and that no additional warnings occur when run on all problems in the OPL.  Visual testing of the code fixes might also be helpful. 

I'm afraid these are one of those "what could possibly go wrong" fixes, but I honestly can't see anything going wrong.  :-)


mgage:
Replaced by fix_undefined_PG_values

PR #249, opened 2015-11-01
Fix undefined pg values

mgage:
Minor cleanups.  Defines some MathObject variables to prevent spurious warnings. 

Some adjustments to the includePGfiles fix.

(The perl "undefined" warnings were already trapped by HTML output but appear when you run problems using the standalongPGrenderer.)


goehle:
I didn't test each change, but I did test a bunch of pg problems.  They all worked correctly and these changes are all small.

PR #250, opened 2015-11-10
Patch BOP to add needed parentheses

dpvc:
This fixes the [parenthesis problem](http://webwork.maa.org/moodle/mod/forum/discuss.php?d=3793) reported by Gavin.  To test, use

```
Context("Numeric");
$f = Compute("x - (x-3)");
BEGIN_TEXT
\(\{$t->TeX\}\) = \{$f->ans_rule(20)\}
END_TEXT
ANS($f->cmp);
```
and check the correct answer to see if it has parentheses or not.  With the patch, the parentheses should be there; without the patch, the parens will be missing.


mgage:
Works.  Thanks Davide.

PR #252, opened 2015-11-18
add maketext calls

heiderich:


mgage:
I've also added this commit by hand to the pull request "develop_translator_upd" so that it can be added to develop without merging master into develop.

dpvc:
@mgage, FYI, the usual way to handle the `master`/`develop` branch issue is to make a second pull request (or do the merge by hand) from the same original branch into `develop`.  That is, hot fixes branch from `master` and are pulled into both `master` and `develop`.  Technically, it is a pull from master that way (two steps back), but it at least keeps the history clear about what is happening, since it is listed as from the patch branch.

goehle:
Whoops.  Looks like I wasn't paying enough attention where I was pulling things or I would have redirected this to develop.  Adding the same changes in by hand *might* cause some conflicts some day in the future, but its my own fault!

PR #255, opened 2015-11-22
Develop translator upd

mgage:
The main change is to how answer evaluators and answer blanks are matched up.  

Now the list of answer evaluators guides the process answer subroutine and each each evaluator
knows (via the PGanswergroup ) which answer blanks to evaluate.  Before the match up was done by dead reckoning and some trickery was used to allow multiple answer evaluators and matrix evaluators. Some of that trickery is still being used -- later updates may refactor some of those answer evaluator routines to be more straightforward using the facilities of the PGanswergroup and PGresponsegroup objects. 

There are some minor (but important) fixes for MathObject parentheses from Davide.

There is support for debug information which can be obtained for those with admin privileges using checkBoxes defined for Problem.pm. These changes are in the develop_wwclients7 pull request for webwork2. 

There is also support for a showResources checkbox which exposes the data collected by PGalias.

You should check that any resources that used alias has been listed in the resources hash and that the entries are appropriate.  The resources hash should contain information of the type of resource  it's url
and/or its directory path.  Whether any conversion is needed (e.g. .gif to.png ) and whether a link is needed from it's current location (usually in the same directory as the pg file) to a public directory -- usually myClass/html .  There should be enough information so that if a library pg file with resources is moved to a local file then we can either transfer the resources as well or make links. Similarly if we export a question (or export a set containing many questions) there is enough information so that the action can also export the accompanying resource files, wrap them together.  The idea is to get seamless transport of a set in one instructors course to the course of another instructor (not necessarily at the same site.)

Try to alias some things that don't exist and make sure that the error warnings are reasonable.  Alias actually checks to see if the url works -- I believe I have fixed am old  bug so that the call times out even if the url site doesn't respond.


VERSION
```diff
@@ -1,4 +1,4 @@
-$PG_VERSION  ='2.11';
+$PG_VERSION  ='develop';
 $PG_COPYRIGHT_YEARS = '1996-2016';
```
goehle: I dont know why this is here since VERSION in the develop branch already says develop.  Its not causing conflicts though so it may not matter.  Merging develop into develop_translator_upd may fix it.

lib/PGalias.pm
```diff
@@ -205,41 +206,36 @@ sub make_alias {
...
+	
+	#$self->debug_message("This auxiliary file id is $aux_file_id" );
+	if ($aux_file_id =~ m/\.([^\.]+)$/ ) {
 		$ext = $1;
 	} else {
 		$self->warning_message( "The file name $aux_file_id did not have an extension.<BR> " .
 		     "Every file name used as an argument to alias must have an extension.<BR> " .
-		     "The permissable extensions are .jpg, .pdf, .gif, .png, and .html .<BR>");
+		     "The permissable extensions are .jpg, .pdf, .gif, .png, .mpg, .mp4, .ogg, .webm and .html .<BR>");
 		$ext  = undef;
```
goehle: My changes from pull #264 seem to have snuck in here some how.  If this gets pulled before #264 then #264 will be moot.

mgage: Is that ok with you?  It seemed useful to just combine the two pulls.  As it turns out I didn't do any more refactoring (no time) but eventually I want to look at this file again and see if the html and hardcopy paths can be refactored.  They still share a lot of code.

goehle: Its fine with me as long as my stuff gets tested.  Also, after thinking
about it adding a `video` command might not be a bad idea.  (See my other
pull)

lib/WeBWorK/PG/Translator.pm
```diff
@@ -1198,50 +1202,172 @@ sub process_answers{
...
+        	my @tmp = ();
+        	foreach my $key (sort keys %$new_temp_ans) {
+        		push @tmp, $key if $new_temp_ans->{$key} eq 'CHECKED';
+        	}
+        	$new_temp_ans = join("\0",@tmp);
+        	$PG->debug_message("Hash answer associated with $ans_name is $new_temp_ans. ref=". ref($new_temp_ans)."<br/>".pretty_print($new_temp_ans) ) if $local_debug;
+        }
+        #FIXME -- hack to allow answers such as <4,6,7>  -- < and > have been escaped.
+        unless( $skip_evaluation) { 
+			$new_temp_ans =~ s/\&lt\;/</g; # <
```
goehle: I don't think this is the right way to solve your encoding problem.  I think the answer is to move the `encode_pg_and_html` line here https://github.com/mgage/pg/blob/develop_translator_upd/macros/PGbasicmacros.pl#L364 to after the `RECORD_ANS_NAME` call.  I think this maybe should always have been the case but the PGanswergroup set up by `RECORD_ANS_NAME` wasn't used before. The answer value should be encoded before it is parsed by ev3 or printed to the screen, but after its been saved for processing. 

If this is the right thing to do then you will have to move encode_pg_and_html after `INSERT_RESPONSE` as well at
- https://github.com/mgage/pg/blob/develop_translator_upd/macros/PGbasicmacros.pl#L364
- https://github.com/mgage/pg/blob/develop_translator_upd/macros/PGbasicmacros.pl#L364

goehle:
Could you post a problem to test the sage stuff you added?

lib/PGUtil.pm
```diff
@@ -137,7 +137,8 @@ sub pretty_print_tex {
 		$out =~ s/_/\\\_/g;   # protect tex
 		$out =~ s/&/\\\&/g;
 		$out =~ s/\$/\\\$/g;
-	} elsif ("$r_input" =~/hash/i) {  # this will pick up objects whose '$self' is hash and so works better than ref($r_iput).
+		#FIXME -- how should mathobjects be handled??
```
goehle: Your set `test_notnull_and_prettyprint` fails with this change.  

```

Context()->texStrings;
BEGIN_TEXT
$BR
Enter pi \{ans_rule(20)\}
$BR $BR
Test ${DOLLAR}PG->pretty_print (${DOLLAR}pi->cmp, "TeX", level=3);
$BR$BR
\{$PG->pretty_print($pi->cmp, "TeX", 3)\}
$BR $BR
Test ${DOLLAR}PG->pretty_print (${DOLLAR}pi->cmp, "TeX", level=4);
$BR $BR
\{$PG->pretty_print($pi->cmp, "TeX", 4)\}
$BR $BR
Test pretty_print(${DOLLAR}pi->cmp) -- this should give current display
\{pretty_print($pi->cmp)\}
END_TEXT
```

goehle:
I'm not getting images in hardcopies after  your changes to alias.  Try `setDemo/limits.pg` (problem 2 of the demo set) to see an example.  Things with `insertGraph` or directly included `png` files are broken but stuff that directly includes a `gif` are ok.

lib/WeBWorK/PG/Translator.pm
```diff
@@ -1198,50 +1202,172 @@ sub process_answers{
...
-			$rh_ans_evaluation_result = $self->{safe} ->reval('$rf_fun->evaluate($temp_ans, ans_label => \''.$ans_name.'\')');
+		my ($rh_ans_evaluation_result, $new_rh_ans_evaluation_result);
+		
+		if (ref($new_rf_fun) eq 'CODE' ) {
+#			$rh_ans_evaluation_result = $self->{safe} ->reval( '&{ $rf_fun }($temp_ans, ans_label => \''.$ans_name.'\')' ) ;
+#			warn "Error in Translator.pm::process_answers: Answer $ans_name: |$temp_ans|\n $@\n" if $@;
+			$PG->warning_message( "Can no longer used CODE objects directly as answer evaluators.  Use AnswerEvaluator");
+		} elsif (not $skip_evaluation )   {
+# 			$rh_ans_evaluation_result = $self->{safe} ->reval('$rf_fun->evaluate($temp_ans, ans_label => \''.$ans_name.'\')');
+# 			$@ = $errorTable{$@} if $@ && defined($errorTable{$@});  #Are we redefining error messages here?
```
dpvc: The original commit that added the errorTable says this:

		  #  Get full traceback, but save it in local variable so that
		  #  we can add it later.  This is because some evaluators use
		  #  eval to trap errors and then report them in the message
		  #  column of the results table, and we don't want to include
		  #  the traceback there.

The "local variable" is the `errorTable`.  So this line is adding the complete traceback for the error (and that traceback has the file names inserted rather than `(eval nnn)`, and has paths replaced by `[TMPL]`, `[PG]`, etc.).

dpvc:
>  there is enough information so that the action can also export the accompanying resource files, wrap them together

Is there support for adding files that are not actually used in the problem?  For example, if a problem selects an image randomly from among several choices, ALL the choices need to be included if the file is copied, not just the one used.  The problem should register all the images so that they can be copied as well, even though only one is actually used in any particular incarnation of the problem.

This does require adding commands to those problems, but it would be good if those commands were actually available to be called.

lib/WeBWorK/PG/Translator.pm
dpvc: I agree with Goeff.  The encoding is being done too soon.  I don't think the saved data should be HTML-ecoded, it should be the original (raw) student input.  The encoding should occur only when the data is being added to the HTML output, not before.  In fact, I would not even remove the multiple spaces as is being done jut before the line Goeff links to.  (I can see situations where multiple spaces are important, e.g., essay answers where indenting is needed, or perhaps a new matrix format that allows you to enter the matrix in one large array and you want to line up the numbers in columns).

If you are finding that you have to decode the data like you are doing here, then that means it was encoded too soon.  You should leather encode it at the point where you are inserting it into HTML (and need the encoding), or have two versions (the original and the HTML encoded) so that the proper one is available when needed (like the text and LaTeX fields in the AnswerHash).  If you have to decode, then something is wrong.  Decoding by hand will lead to maintenance issues in the future.  If the encoding changes, you will have to go back and fix all the places where decoding was done.  If you must encode early, at least make a corresponding decode routine and put it with the encoding one, so that changes in one can be reflected in the other.

mgage: Thanks Goeff and Davide.  I'll change the order of encoding -- I may not get to this until tomorrow 1/16.

lib/WeBWorK/PG/Translator.pm
mgage: I gather that this comment about errorTable is in response to the comment/question I inserted in the code and that you are not suggesting any changes here.  I will restore the original comment since that might help in the future.

mgage:
I’m aware of the problem but I haven’t yet included support for this use case.  (I also haven’t started actually using the data
to collected in PGresponse to actually export questions bundled with all of their companion files.)

I think however that the existing support in PGresponse is about all that is needed.  The responsibility for gathering the other
unused images rests with the problem.  This can be accomplished by calling alias() on all of the images whether they are used or 
not.  The output (a url for the image) would be discarded but the side effect (registering the image in a PGresponse object) would 
still take place.  

At the moment I don’t see why a separate command other than alias() would be needed to register the unused images.

lib/WeBWorK/PG/Translator.pm
dpvc: Correct.  Sorry for not being clear.

dpvc:
OK, so `alias()` itself is the needed command.  I hadn't thought of that.  For static images, does this also cause the web-accessible version of the image to be created, or is that done by `image()` or some other call?

goehle:
I think `image()` calls alias for you: https://github.com/openwebwork/pg/blob/master/macros/PGbasicmacros.pl#L2680

mgage:
It also causes the accessible image and the link to be created (if needed).  In addition it checks that the file
and it’s link are actually accessible.  I would regard this as a feature since it insures that running the problem once
actually checks all variations of the problem.  You wouldn’t be blind sided by a seed which you hadn’t explicitly tested.

The code for this is actually called make_alias() in   the PGalias.pm  module.  alias() is simply a cover for that.

The code is still somewhat in flux.  It is about one half to three quarters refactored.  There is still substantial 
overlap between the handling of auxiliary files in html mode and in tex mode.  In addition we have recently
added a half dozen file types to keep track of.  They are certainly substantially similar to html files and are being
treated exactly that way at the moment.  As we examine use cases we may find that the handling of some files
might need additional tweaking.  


There is one more loose end — that I’m in the middle of figuring out.   I’d like alias to pass back the url (or file path in the tex case)
but also pass back the PGresponse object so that more precise error messages can be generated (e.g. by image()).  Often
the second value returned (the PGresponse object) would be ignored — especially by earlier macros.  My first attempt to do
this breaks if the auxiliary file can’t be found.  I discovered this just before JMM and haven’t yet had time to resolve that issue.

dpvc:
As a testing feature, I understand the desire to check that all possible files are available, but it seems wasteful to create copies of all files for each student using the problem.  If a problem picks an image from among six possible ones, then that would mean that problem will take up six times as much space as needed.  I guess disk space isn't as tight as it once was, but it still rubs me the wrong way.

It also seems that this is conflating problem testing with problem usage unnecessarily.  The problem doesn't have to go through the sanity check that call files are available *every time a student submits an answer*, does it?

mgage:
It's a little wasteful, but I'd need to see some timing data to be convinced that it is really a problem.  In general images are not created, just links.  And if the link already exists nothing is done, (not sure whether the url checking is even done in that case).  By the way the on-the-fly graphics is not handled through alias() (as I recall) which meant that some fixes for naming files in alias() were later discovered to be required in WWPlot as well.  That's a place where some refactoring might or might not be desirable. 

You certainly want the testing to take place when the author is working and when the instructor is assigning the problem or when the instructor is exporting the problem.  One might turn the testing off when the targeted user is a student, implicitly assuming that  nothing could go wrong at that point. I'd be willing to add that feature once everything else is tested and stabilized.  I have my doubts that it would actually speed anything up significantly.

goehle:
Part of why images go through alias now, and why they are checked every time a problem is rendered, is that it finally puts the image cache'ing problem to bed.  One of the things I've been keeping an eye on is that the same url is used for images as long as you are just submitting answers.

dpvc:
> In general images are not created, just links.

OK, if it is just links, I am less concerned.

mgage:
This next change moves calls to encode_pg_and_html() until after calls to RECORD_ANS_NAME. This means that the hacks in Translator which undid the html entity codes for <, >, prime ', $, _ and * are no longer needed.

Check that entries which contain <  > (e.g. vectors) behave properly, also answers that contain '  and _ and * behave properly.  For example:
Library/AlfredUniv/diffeq/ReductionOfOrder/LRC.pg
Library/CollegeOfIdaho/setStatistics_Ch18CompareTwoMeans/18Stats_06_CompareTwoMeans.pg
Library/CollegeOfIdaho/setAlgebra_02_01_IntroFunctions/21IntAlg_14_function.pg

goehle:
Moving problems to this pull since its likely where they are located. 

-  [ ]  There is an issue with `pretty_print_tex`.  This is likely a problem with the PG branch `develop_translator_upd`.  You can debug it with the following problem 

  ```
Context()->texStrings;
BEGIN_TEXT
$BR
Enter pi \{ans_rule(20)\}
$BR $BR
Test ${DOLLAR}PG->pretty_print (${DOLLAR}pi->cmp, "TeX", level=3);
$BR$BR
\{$PG->pretty_print($pi->cmp, "TeX", 3)\}
$BR $BR
Test ${DOLLAR}PG->pretty_print (${DOLLAR}pi->cmp, "TeX", level=4);
$BR $BR
\{$PG->pretty_print($pi->cmp, "TeX", 4)\}
$BR $BR
Test pretty_print(${DOLLAR}pi->cmp) -- this should give current display
\{pretty_print($pi->cmp)\}
END_TEXT
 ```
- [ ]  This is likely a problem with the PG branch `develop_translator_upd`.  I want to test the sage additions you made.  Could you post a pg file which uses the functions defined at https://github.com/openwebwork/pg/pull/255/files#diff-ec688a0529a3964cd5ab6e67145adba3R82 and at https://github.com/openwebwork/pg/pull/255/files#diff-ec688a0529a3964cd5ab6e67145adba3R102
- [ ]  This is likely a problem with the PG branch `develop_translator_upd`.  There are some reversions with using the xmlrpc pathway to render problems.   The easiest way to test these is to use the problems below in a set and try to render them using the Problem Set Detail 2 page.   In particular 
  -  Using the following problem I get 
    ```
faultstring: Can't locate object method "_string" via package "InequalitySetBuilder::Union" at (eval 1206) line 365. 
    ```
    The problem is: 
    ```
DOCUMENT();      
loadMacros(
   "PGstandard.pl",     # Standard macros for PG language
   "MathObjects.pl",
   "contextAlternateIntervals.pl",
   "contextInequalitySetBuilder.pl"
);
Context("InequalitySetBuilder");
context::AlternateIntervals->Enable;
Context()->flags->set(enterIntervals=>"either", displayIntervals=>"alternate");
 $S1 = Compute("{ x : 1 < x <= 4 }");
 $S2 = SetBuilder("]1,4[");     # force interval to be set in set-builder notation
 $S3 = Interval("[1,4[");
 $S4 = Compute("{ x : x < -2 or x > 2 }");  # forms the Union (-inf,-2) U (2,inf)
 $S5 = Compute("{ x : x > 2 and x <= 4 }"); # forms the Interval (2,4]
 $S6 = Compute("{ x : x = 1 }");            # forms the Set {1}
 $S7 = Compute("{ x : x != 1 }");           # forms the Union (-inf,1) U (1,inf)
Context()->texStrings;
BEGIN_TEXT
\($S1\) \{$S1->ans_rule\} $BR
\($S2\) \{$S2->ans_rule\} $BR
\($S3\) \{$S3->ans_rule\} $BR
\($S4\) \{$S4->ans_rule\} $BR
\($S5\) \{$S5->ans_rule\} $BR
\($S6\) \{$S6->ans_rule\} $BR
\($S7\) \{$S7->ans_rule\} $BR
END_TEXT
Context()->normalStrings;
ANS($S1->cmp());
ANS($S2->cmp());
ANS($S3->cmp());
ANS($S4->cmp());
ANS($S5->cmp());
ANS($S6->cmp());
ANS($S7->cmp());
ENDDOCUMENT();        
    ```
  -  Similarly I get 
    ```
faultstring: Can't locate object method "string" via package "context::AlternateDecimal::Real" at /opt/webwork/pg/lib/Parser/Number.pm line 66
    ```
    with the problem 
    ```
DOCUMENT();      
loadMacros(
   "PGstandard.pl",     # Standard macros for PG language
   "MathObjects.pl",
   "contextAlternateDecimal.pl"
);

Context("Matrix");
context::AlternateDecimal->Default(",",",");

TEXT(beginproblem());
$pi = Compute("[[3,14; 2,18]; [3,22; 3,14*x^(2,12)]]");
Context()->texStrings;
BEGIN_TEXT
\($pi\) and \{$pi->ans_rule\} $BR
END_TEXT
Context()->normalStrings;
ANS($pi->cmp);
ENDDOCUMENT();       
    ```
-  Images dont work in hardcopies.    This is likely a problem with the PG branch `develop_translator_upd`.  If you take the following pg file and add it to a set, the pdf hardcopy is generated correctly on develop but with `develop_translator_upd` it fails.  
  ```
Library/NAU/setCalcI/PaperFold/Paper_Fold_1.pg
```

**Comment by mgage** [comment]
Additional items transferred from wwclientsup7:

Fix pretty_print_tex -- test against problem 6 of   test_notnull_and_prettyprint set in gage_test
 Fix images and alias:  see Library/NAU/setCalcI/PaperFold/Paper_Fold_1.pg  example above.
 embedCSV and embedPDF dont work in latex. -- postpone briefly -- these were added recently and probably need further work somewhere in add Resource.

**Comment by mgage** [comment]
Code for testing sage Calculator pad.  I'll see if I can find an actual .pg file somewhere.



As a wandering vagabond I just found out how easy it is to also embed Sage in the course info files. See the image below for a simple example.

The txt file is also pasted here. Interesting how transparently this works. What a great system.

JT

<script src="https://sagecell.sagemath.org/static/jquery.min.js"></script>
    <script src="https://sagecell.sagemath.org/static/embedded_sagecell.js"></script>
    <script>$(function () {
    // Make the div with id 'mycell' a Sage cell
    sagecell.makeSagecell({inputLocation:  '#mycell',
                           template:       sagecell.templates.minimal,
                           evalButtonText: 'Activate'});
    // Make *any* div with class 'compute' a Sage cell
    sagecell.makeSagecell({inputLocation: 'div.compute',
                           autoeval: 1,
                           evalButtonText: 'Evaluate'});
    });
    </script>
  </head>

<center><font size=+2><b>Introduction To Mathematical Statistics</b></font></center>
<p align=center>

<p>

blah, blah, blah

<p>
You can type your own Sage computation below and click “Evaluate”.
    <div class="compute"><script type="text/x-sage">
data = [1, 3, 4, 1, 7, 4, 2, 3, 2, 4, 2, 5, 4, 1, 3, 3, 2]
n = len(data); print "Number of data values =",n

s = sum(data); print "          Sum of data = ",s

s2 = sum((x^2 for x in data)); print "       Sum of squares = ",s2

s3 = sum((x^3 for x in data)); print "         Sum of cubes = ",s3

s4 = sum((x^4 for x in data)); print "        Sum of forths = ",s4;print

mu = mean(data); print "             The mean =",mu

var = variance(data); print "  The sample variance = ",var
</script></div>

**Comment by mgage** [comment]
This last commit above fixes error in problem 6 of the set test_nonnull_prettyprint

**Comment by mgage** [comment]
The problem setDemo/images.pg    now prints the image correctly in TeX mode.

**Comment by mgage** [comment]
For testing the sageCalculatorPad.  This is still very experimental:

DOCUMENT();      

loadMacros(
   "PGstandard.pl",     # Standard macros for PG language
   "MathObjects.pl",
   "sage.pl",
   #"source.pl",        # allows code to be displayed on certain sites.
   #"PGcourse.pl",      # Customization file for the course
);

# Print problem number and point value (weight) for the problem
TEXT(beginproblem());

# Show which answers are correct and which ones are incorrect

sageCalculatorHeader();  # set up javaScript needed for the sageCalculatorPad

$pi = Real("pi");
##############################################################
#
#  Setup
#
#
Context("Numeric");


##############################################################
#
#  Text
#
#

Context()->texStrings;
TEXT(
   sageCalculatorPad( "Use this calculator pad to make calculations",  
q!                                                                          
data = [1, 3, 4, 1, 7, 4, 2, 3, 2, 4, 2, 5, 4, 1, 3, 3, 2]
n = len(data); print "Number of data values =",n
s = sum(data); print "          Sum of data = ",s
s2 = sum((x^2 for x in data)); print "       Sum of squares = ",s2
s3 = sum((x^3 for x in data)); print "         Sum of cubes = ",s3
s4 = sum((x^4 for x in data)); print "        Sum of forths = ",s4;print
mu = mean(data); print "             The mean =",mu
var = variance(data); print "  The sample variance = ",var
!
   )
);



BEGIN_TEXT





Enter a value for \(\pi\)

\{$pi->ans_rule\}
END_TEXT
Context()->normalStrings;

##############################################################
#
#  Answers
#
#

ANS($pi->with(tolerance=>.0001)->cmp);
# relative tolerance --3.1412 is incorrect but 3.1413 is correct
# default tolerance is .01 or one percent.


ENDDOCUMENT();

**Comment by goehle** [comment]
The error when you are missing a file from alias is a little wierd: 
```
The macro alias cannot find the HTML file: |normaddddn.png| anywhere|
```
It refers to the file as an HTML file even though its png and there is a | after anywhere.

**Comment by goehle** [comment]
The Show Aux Resources Hash doesn't properly show if conversion is needed.  If you use `image` with a gif it shows
```
convert	=>	 HASH(0x7f12c3f16c38)
from_path	=>	 
from_type	=>	 
needed	=>	 0
to_path	=>	 
to_type	=>	 
```

File: lib/PGUtil.pm [hunk 6]
```diff
@@ -92,7 +92,17 @@ sub pretty_print_html {    # provides html output -- NOT a method
     $level--;
     return "PGalias has too much info. Try \$PG->{PG_alias}->{resource_list}" if ref($r_input) eq 'PGalias';  # PGalias just has too much information
     return 'too deep' unless $level > 0;  # only print four levels of hashes (safety feature)
-    my $out = '';
+	my $out = '';
+	    # protect against modules defined in Safe which can't find their stringify procedure.
+		$r_input = $r_input//'';
+		eval { "$r_input" };
```
goehle: Line 98 here is giving me the compilation error: Useless use of string in void context at /opt/webwork/pg/lib/PGUtil.pm line 98.

PR #257, opened 2015-11-26
new macro parserWordBlank.pl

paultpearson:
New macro parserWordBlank.pl.  Provides free response, fill in the blank questions with interactive help.  As a student types their answer into the answer blank, jQuery's autocomplete feature generates a drop-down list of allowable answers that match what has already been typed.  When the student presses the "Check Answers" or "Submit Answers" button, jQuery generates a warning message if the student answer is not one of the allowable answers.  Choices in the drop-down list and the correct answer are specified by the problem author.  WordBlank objects are compatible with Value objects, and in particular, can be used with MultiAnswer objects.


paultpearson:
I'm going to rename this parserWordCompletion.pl, so I'm closing this pull request and will open another.

PR #259, opened 2015-12-05
Update for MultiAnswer compatibility

paultpearson:
Apparently, the way that answer blanks were named by parserMultiAnswer.pl changed sometime in the last few years.  This update makes parserWordCompletion.pl compatible with the new answer blank names produced by parserMultiAnswer.pl.


mgage:
Works for me.  It's a nice answer type.

PR #260, opened 2015-12-12
Fix incorrect requirement for extraParens in stringification of UOP.

dpvc:
This fixes the [parenthesis problem](http://bugs.webwork.maa.org/show_bug.cgi?id=3547) reported by Mike. To test, use

```
Context("Numeric");
$f = Compute("(-1)^x");
BEGIN_TEXT
\(\{$f->TeX\}\) = \{$f->ans_rule(20)\}
END_TEXT
ANS($f->cmp);
```

and check the correct answer to see if it has parentheses or not. With the patch, the parentheses should be there; without the patch, the parens will be missing.


goehle:
Works as advertised.  Thanks for the quick turnaround.

PR #261, opened 2015-12-14
Fix problem with correct_ans being incorrectly set

dpvc:
This patch uses of the `noInherit` list to remove unwanted fields from propagating to copies of existing objects.  This affects the MathObject's Value class' `make()` method (and the overridden ones in the Complex and Interval objects).  This means that it is fairly far reaching, so we should keep an eye out for unexpected consequences, though I don't think there should be any.

This resolves [bug 3550](http://bugs.webwork.maa.org/show_bug.cgi?id=3550), and the unreported but similarly caused problems in Complex and Interval objects.

To test, use

```
$f = Compute("sin(x)");
$a = Compute("pi");
$b = $f->eval(x=>$a);

BEGIN_TEXT
\(\sin(\pi)\) = \{$b->ans_rule(5)\}
END_TEXT

ANS($b->cmp());
```

Without the patch, the correct answer will show up as pi rather than the correct value of 0.  With the patch, the correct value should be shown.

To test the Complex object, use

```
Context("Complex");
$f = Compute("sin(z)");
$a = Compute("pi+i");
$b = $f->eval(z=>$a);

BEGIN_TEXT
\(\sin(\pi+i)\) = \{$b->ans_rule(5)\}
END_TEXT

ANS($b->cmp());
```

and for the corresponding Interval situation, use

```
Context("Interval");
$I = Compute("[1,5]")->new(2,3);

BEGIN_TEXT
\((2,3)\) = \{$I->ans_rule(10)\}
END_TEXT

ANS($I->cmp());
```

Here, make sure the correct answer shows as `(2,3)` rather than `[1,5]`.

### DIFF
--- a/lib/Value.pm
+++ b/lib/Value.pm
@@ -568,7 +568,7 @@ sub formula {
 sub make {
   my $self = shift; my $class = ref($self) || $self;
   my $context = (Value::isContext($_[0]) ? shift : $self->context);
-  bless {$self->hash, data => [@_], context => $context}, $class;
+  bless {$self->hashNoInherit, data => [@_], context => $context}, $class;
 }
 
 #
@@ -600,6 +600,13 @@ sub hash {
   my $self = shift;
   return %$self if isHash($self);
   return ();
+}
+
+sub hashNoInherit {
+  my $self = shift;
+  my %hash = $self->hash;
+  foreach my $id ($self->noinherit) {delete $hash{$id}}
+  return %hash;
 }
 
 #
--- a/lib/Value/Complex.pm
+++ b/lib/Value/Complex.pm
@@ -39,7 +39,7 @@ sub make {
   my $self = shift; my $class = ref($self) || $self;
   my $context = (Value::isContext($_[0]) ? shift : $self->context);
   while (scalar(@_) < 2) {push(@_,0)}
-  my $c = bless {$self->hash, data => [@_[0,1]], context => $context}, $class;
+  my $c = bless {$self->hashNoInherit, data => [@_[0,1]], context => $context}, $class;
   foreach my $x (@{$c->{data}}) {$x = $context->Package("Real")->make($context,$x) unless Value::isValue($x)}
   return $c;
 }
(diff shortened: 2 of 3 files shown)


goehle:
-  I'm having trouble getting your third test case working.  I get `2,3` (no parens) as the correct answer before and after the patch. 

-  What sort of stuff should I be looking out for with regard to "unexpected concequences".  It looks like "correct_ans", "correct_ans_latex_string", "original_formula", "equation" won't be inherited anymore.  Is the issue is that some things may have been making use of these inherited values?

goehle:
The test cases all check out, although I'm still curious about the answer to the second question.  I'm merging this in any case.

dpvc:
> What sort of stuff should I be looking out for with regard to "unexpected concequences". It looks like "correct_ans", "correct_ans_latex_string", "original_formula", "equation" won't be inherited anymore. Is the issue is that some things may have been making use of these inherited values?

Yes, that is my concern.  The `make()` method that is changed here is used by every MathObject class (unless it is overridden, as in the Complex and Interval objects), and so changing it means that everything is affected.  That is always a source of concern, though I can't think of anything specific that will go wrong.

The `correct_ans` and `correct_ans_latex_string` fields are the ones that needed to be removed to get the answers to show up correctly, and these are not used anywhere internally for MathObjects, so that is certainly safe.  The `original_formula` should be removed, so that is also OK.  The `equation` field is one that gets used, so there is some chance that that is a problem, but it appears to be added in again later in the process, so that it points to the correct location, so that should not be a problem either.

Note, however, that the `noInherit` list can be modified by the various classes, so there may be other losses as well.  For example, the Formula class adds a bunch of Formula-specific fields to the list; but these all look good to me, so I think it is all OK.  It's jus that MathObjects is a *big* library that has lots of complicated interactions, so something unexpected could occur.  I ran the test suite that I have for the core MathObject classes, and everything was fine, but it is not as comprehensive as I would like (it hasn't been updated in years).

In any case, I think everything should be OK but just wanted you to have it in mind in case something strange shows up in the future.

PR #262, opened 2015-12-17
Fixbasiscmp

jwj61:
This fixes bug 1134 using Davide's suggestion there.  A problem set with

Library/WHFreeman/Holt_linear_algebra/Chaps_1-4/4.3.5.pg

generating hardcopy showing correct answers can be used to test.


dpvc:
Works for me.

PR #263, opened 2015-12-22
Release 2.11

mgage:
Fixed overzealous definition of undefined quantity. Commit comment includes Davide's explanation of the problem.


goehle:
This fixes issues I was having with various random `correct_ans` fields ending up blank when they didn't used to.

PR #264, opened 2015-12-23
Video Extensions For Alias

goehle:
PGalias will handle video files without any chage with a small change to add video extensions. This is useful if you want to serve video in pg directly with the <video> tag.

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2765,16 +2765,17 @@ sub video {
 	        $video =~ /.*\.(\w*)/;
 	        my $type = $1;
 		my $out;
-
+	        my $htmlmessage = maketext("Your browser does not support the video tag.");
+	  
 		if ($displayMode eq 'TeX') {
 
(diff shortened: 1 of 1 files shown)


goehle:
Would it be worth adding a `video` command like the `image` command?  Right now you have to set things up manually with something like 
```
$videoaddress = alias('VideoFile.mp4');

TEXT(MODES(TeX=>'The video must be viewed online.', HTML=><<"EOH"));
<video width="400"  controls>
  <source src="$videoaddress" type="video/mp4">
</video>
EOH
```

mgage:
go ahead and add a video embedding command (like image).  You can put it in PGbasicmacros.pl since it is short enough.

goehle:
I added a video tag.  You can test this by 
1) Put a video file in a location where `alias` will find it (e.g. in the same folder as your .pg file).
2) Use the `video` command to print the video.  Options are `height` and `width`.  E.G:
```
\{video('filename.ogg',height=>300,width=>350)\}
```

mgage:
I checked this with the file moodle_gradebook.mp4.    Both the hand embedding using alias and the video() macro made the proper links.  They also recorded things correctly in a PGresource object.

This is ready to go.

PR #265, opened 2016-01-03
Develop beginproblem bugfix

mgage:
Prevents beginproblem() from printing points and file path when question is viewed independently
such as after clicking on the "eye" in the library browser. 

To test:
View files in library and check that the path is not displayed within the problem even after randomizing
the problem. 

File paths are still shown when the problem is rendered within homework sets detail 2 which is a bit redundant but not too obnoxious.


mgage:
I'll update this pull later after the wwclients series has been pulled.

Issue #266, opened 2016-01-10
Differentiation for parserRoot.pl


Alex-Jordan:
This is a feature request. Can we make it so that the D method applies to a Formula that has the root(,) function in it (once enabled by parserRoot,pl)?

dpvc:
It should be able to be added.  I'll look into it.

dpvc:
I've made a pull request that implements the differentiation (and better handling of parentheses for these roots in TeX output).

PR #267, opened 2016-01-11
Add differentiation to parserRoot, and improve parentheses in TeX output

dpvc:
This resolves issue #266 by adding the `D()` method needed for differentiation.  To test it, use

```
loadMacros("parserRoot.pl");
parser::Root->Enable;
$f = Compute("root(4,x)")->D;
TEXT($f);
```

Without the patch, the code will throw an error (`Differentiation of 'root' not implemented`), and with the patch, you should get the result `0.25*[root(4,x)]^(-3)`.

This patch also improves the TeX output so that parentheses are used when appropriate.  To test this, use

```
loadMacros("parserRoot.pl");
parser::Root->Enable;
$f = Compute("root(3,x)^2");
TEXT($f->TeX);
```

With the patch, you should get `\left(\sqrt[3]{x}\right)^{2}`, but without it, you would get `\sqrt[3]{x}^{2}`.

### DIFF
--- a/macros/parserRoot.pl
+++ b/macros/parserRoot.pl
@@ -164,12 +164,44 @@ sub root {
 }
 
 #
+#  Implement differentiation: (u^(1/n))' -> (1/n)(u^(1/n))^(1-n) * u'
+#  (We use (u^(1/n))^(1-n) rather than u^(1/n-1) so that we have the
+#  same domain as u^(1/n) does originally).
+#
+sub D {
+  my $self = shift; my $x = shift;
+  my $equation = $self->{equation};
+  my $BOP = $self->Item("BOP");
+  my $NUM = $self->Item("Number");
+  my $n = $self->{params}[0];
+  return
+    $BOP->new($equation,'*',
+      $BOP->new($equation,'*',
+        $BOP->new($equation,'/',$NUM->new($equation,1),$n->copy($equation)),
+        $BOP->new($equation,'^',
+          $self->copy($equation),
+	  $BOP->new($equation,'-',$NUM->new($equation,1),$n->copy($equation))
+        )
+      ),
+      $self->{params}[1]->D($x)
+    )->reduce;
+}
+
+#
 #  Output TeX using \sqrt[n]{x}
 #
 sub TeX {
-  my $self = shift;
+  my ($self,$precedence,$showparens,$position,$outerRight,$power) = @_;
+  $showparens = '' unless defined $showparens;
+  my $fn = $self->{equation}{context}{operators}{'fn'};
+  my $fn_precedence = $fn->{parenPrecedence} || $fn->{precedence};
   my ($n,$x) = @{$self->{params}};
-  return '\sqrt['.$n->TeX."]{".$x->TeX."}";
+  my $TeX = '\sqrt['.$n->TeX."]{".$x->TeX."}";
+  $TeX = '\left('.$TeX.'\right)'
+    if $showparens eq 'all' or $showparens eq 'extra' or
+       (defined($precedence) and $precedence > $fn_precedence) or
+       (defined($precedence) and $precedence == $fn_precedence and $showparens eq 'same');
+  return $TeX;
 }
 
 ########################################################################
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
In `Formula("root(n,x)")`, is it a requirement that `n` be a constant?  Or that it be an integer? Or can `n` be anything, including a formula in `x`?

And then if so, it appears based on the comment in the new code that the derivative is only taken with respect to the second variable. Is it doable to make `root(u, v)` differentiate to 

``1/u * root(u,v)^(1-u) * v' + root(u,v) * ln(v) * -1/u^2 * u'``?  

I'm hoping I differentiated that right.  It might be a good precedent for functions in `numeric2` to try to use the multivariable chain rule this way. It would allow for differentiating things like `root(x, 2)` and `root(x,x)`, not just `root(2, x)`.

There is a puzzle here with how to handle differentiating `root(u, v)` when `v` is taking a negative value. If `u` is varying while `v` is negative, then it makes sense for the derivative to be undefined, which would be recognized by the `ln(v)`. But if you use the formula above, and `u` is constant, then it shouldn't let the `ln(v)` throw an error when `v` is negative. You almost want something like

``1/u * root(u,v)^(1-u) * v' + {if (v> 0 or u'!=0) then (root(u,v) * ln(v) * -1/u^2 * u') else 0}``

I don't need to use `root` this way right now, but it occurred to me that it might be useful some day.

Separate observation: if you open this up again, it might be nice to have a flag that controls whether `root(2,x)` is typeset as `\sqrt[2]{x}` or just `\sqrt{x}`.

dpvc:
> In `Formula("root(n,x)")`, is it a requirement that `n` be a constant? 

No, but it does have to be an integer, so it can be a formula as long as that formula only produces integers.

The purpose of `root(n,x)` was to was to extend the domain of `x^(1/n)` to include negative `x` when `n` is an odd integer.  As such, it requires `n` to be an integer.  So in terms of derivatives, it doesn't make sense to consider `n` to be a formula in `x`.  The implementation of `root(n,x)` will allow you to do things like `root(x,x)`, but when it is used, it will throw an error if it is evaluated at any non-integer value of `x` (or at `x = 0`).  So it doesn't make sense to ask about the limit needed for the derivative.

As you know, in general, `a^b` is only defined for `a > 0`, which is why `root(n,x)` and `x^(1/n)` are not the same thing, and `root(n,x)` is not a synonym for `x^(1/n)`.  If you are looking to use derivatives of things like `root(x,x)` then I would recommend `x^(1/x)` instead, since this allows non-integer values.

Alex-Jordan:
That all sounds fair. I was also seeing this as a convenience for display purposes. So if I want to display a function to students as `\sqrt[x]{2}` (defined only on positive `x`), and if I want to do anything with that that involves evaluation of that formula, I'll need to separately manage behind the scenes computation (`2^(1/x)`) from presentation (`\sqrt[x]{2}`).

Would you mind adding a line to the documentation about the restrictions on `n`? I could see a future problem author restricting domain to positive `x` and trying to use `Formula("root(x,2)")->D('x')` and then not understanding why the result is `0`.

dpvc:
I should probably add a check in the differentiation routine that `n` is constant and throw an error if the variable of differentiation appears in it.

goehle:
My tests are all good.  You can even use this to take a complicated derivative of x^2: `root(.25,x)`.  Do you want me to wait for the check that `n` is constant or should I merge?

dpvc:
I'm still thinking about Alex's comments, so wait for a bit.  I'll probably make another commit with at least something that gives a warning if `n` is not constant.

dpvc:
OK, I've patched the file so that the root can be a non-integer (where defined), and the derivative can be taken when the root is non-constant.  So I think that covers the situations that you need.

goehle:
Works for me.  It even marks you as incorrect if you answer 1/3 x^(-2/3) to the derivative of root(3,x) and says you have the wrong domain.  I'm looking forward to explaining the difference between principle roots and real roots to someone who uses this :)  (I had to look it up myself.  For completeness: http://blog.wolframalpha.com/2013/04/26/get-real-with-wolframalpha-computing-roots/)

PR #269, opened 2016-02-03
Solution Hardcopy Bug

goehle:
There is a bug in develop where hardcopies with solutions don't compile because of a \bfSolution command.  This fixes that and adds maketext to Hint for good measure. 
-  To Test:  Generate a hardcopy with a problem that has a solution and a hint.  Before the patch it will fail and after the patch both the solution and hint will work.

I also found another really subtle bug where Hardcopy was failing to find macro files under very specific circumstances.  Basically if it failed to find one particular file then it was replacing parts of the paths with abbreviations and it would then fail to find more files as a result.


mgage:
This fix works.  I can now print solutions in hardcopy.

PR #270, opened 2016-02-04
Hardcopy Themes

goehle:
This is the PG pull corresponding to: https://github.com/openwebwork/webwork2/pull/682

It should be tested along side that pull request.


mgage:
As long as we are modifying these files I suggest that we remove all of the copyright material in all the .tex files.  This will be replaced with the text of the OPL_LICENSE https://github.com/openwebwork/webwork-open-problem-library/blob/master/OPL_LICENSE
placed only in the hardcopyPreamble.tex file.   This will dramatically reduce the clutter in the TeX output of homework sets.  The OPL_LICENSE is more appropriate than the WeBWorK GPL license since it is really the content of the homework sets that we are protecting.

Does anyone else have thoughts on this?

mgage:
It looks very nice. I'd be willing to keep one copy of the OPL_LIcense embedded in the tex preamble file, but I'm not sure it's really that useful. I'm satisfied with the way things are.

I'm actually going to find this ability quite useful. It's almost at the point where you could easily print an exam from a webwork homework setby printing the TeX output and then doctoring slightly. The only iffy part would be downloading the graphs and other auxiliary files properly. At the moment that would have to be done by hand.

I'll pull this.

PR #271, opened 2016-02-09
MathObjects and Pi

goehle:
Change pi and Infinity to be defined by non MathObjects expressions when MathObjects is not loaded.

To test use the problem:  
```
DOCUMENT();      
loadMacros("PG.pl",
           "PGbasicmacros.pl",
);
TEXT(beginproblem());

$a = pi;

BEGIN_TEXT
$a
\{ref($a)\}
END_TEXT
ENDDOCUMENT();
```
Before the patch you should get
```
3.14159265 Value::Real 
```
and after you should get 
```
3.14159265358979 
```
If you add `MathObjects.pl` to the macro list after the patch it should go back to the first answer.

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -551,12 +551,20 @@ sub k () {
 }
 
 # ^function pi
+# ^uses $_parser_loaded
 # ^uses &Value::Package
-sub pi () {Value->Package("Formula")->new('pi')->eval}
+sub pi () {
+  if (!eval(q!$main::_parser_loaded!)) {return 4*atan2(1,1)}
+  Value->Package("Formula")->new('pi')->eval;
+}
 
 # ^function Infinity
+# ^uses $_parser_loaded
 # ^uses &Value::Package
-sub Infinity () {Value->Package("Infinity")->new()}
+sub Infinity () {
+  if (!eval(q!$main::_parser_loaded!)) {return 'infinity'}
+  Value->Package("Infinity")->new();
+}
 
 
 # ^function abs
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -5,10 +5,17 @@
 # initialize PGcore and PGrandom
 
 
-	$main::VERSION ="WW2";
+$main::VERSION ="WW2";
 
 sub _PG_init{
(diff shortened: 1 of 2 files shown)


macros/PG.pl
```diff
@@ -551,12 +551,20 @@ ()
...
+  if (!eval(q!$main::_parser_loaded!)) {return 4*atan2(1,1)}
+  Value->Package("Formula")->new('pi')->eval;
+}
 
 # ^function Infinity
+# ^uses $_parser_loaded
 # ^uses &Value::Package
-sub Infinity () {Value->Package("Infinity")->new()}
+sub Infinity () {
+  if (!eval(q!$main::_parser_loaded!)) {return 'infinity'}
```
dpvc: I would make the string be `Infinity` rather than `infinity`, since that is what a bare `Infinity` would mean without MathObjects.

goehle:
Added code from Davde which will e nsure that all problems get their own copy of the context hash, even if they dont explicitly load mathobjects.  

To test use this problem: 
```
DOCUMENT();      
loadMacros("PG.pl",
           "PGbasicmacros.pl",
);
TEXT(beginproblem());
BEGIN_TEXT
\{Parser::Context->current()\}
END_TEXT
ENDDOCUMENT();
```
Before the patch if you reload the page over and over you should always get the same address for the hash.  After the patch refreshing the page will give different hash addresses.

macros/PG.pl
```diff
@@ -5,10 +5,17 @@
 # initialize PGcore and PGrandom
 
 
-	$main::VERSION ="WW2";
+$main::VERSION ="WW2";
 
 sub _PG_init{
-	$main::VERSION ="WW2.9+";
+  $main::VERSION ="WW2.9+";
```
dpvc: Should this really be 2.11 at this point?

goehle:
I added the code to prevent Context from initializing twice. 

I'm not sure why that VERSION code is there.  I'm sure someone added it to check if WeBWorK was new enough, but without knowing what they were doing I dont know if changing the version will break anything.

dpvc:
OK, sounds good.  Merging...

PR #276, opened 2016-06-07
Fix dumb typo in init subroutine of PG.pl

mgage:
This is such a dumb mistake I wanted to fix it now.  I don't know how it has lasted this long.

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -13,7 +13,7 @@ sub _PG_init{
   #  Set up MathObject context for use in problems
   #  that don't load MathObjects.pl
   #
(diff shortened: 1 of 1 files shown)


mgage:
Too test:

1. Download github.com/mgage/standaloneProblemRenderer  
to a convenient directory (I use  /opt/webwork/standaloneProblemRender).

2. Type `wwperl standalonePGproblemRenderer.pl  input.pg` and get response
`Reference found where even-sized list expected at /Volumes/WW_test/opt/webwork/pg/macros/PG.pl line 16.`

3. After update there is no error message.

PR #278, opened 2016-06-21
UTF8 Support

goehle:
This is the pg pull associated to openwebwork/webwork2#712


mgage:
I believe that all of these changes have been incorporated into PR #319

mgage:
included in #319 and merged into develop_international

PR #279, opened 2016-08-04
Error Numbering Bug

goehle:
This changes the white space character in the pg code preprocessing regexs to the `horizontal` whitespace character.  This has the affect of not removing blank lines from preprocessed code and means that the error numbers reported by PG will match the lines in the PG problem.  

To test: 
-  Check that problems still work.  Mike/Davide may have an opinion if the full `\s` is necessary for preprocessing.  
-  Try the following problem.  Before the patch the error will be reported on line 5.  After it will be reported with the correct line 7
  ```
DOCUMENT();

BEGIN_TEXT
spoon
END_TEXT

,difl.poeldjva;

ENDDOCUMENT(); 
  ```


mgage:
Looks good.  Nice catch on this long standing annoying but not fatal error.

Issue #280, opened 2016-08-29
Two solutions using Context("Inequalities")


d-torrance:
Hello!

Consider the following problem:

    DOCUMENT();

    loadMacros(
      "PGstandard.pl",
      "MathObjects.pl",
      "contextInequalities.pl"
    );

    Context("Inequalities");

    $ans = Compute("x=1 or x=2");

    TEXT(beginproblem());
    BEGIN_TEXT
    \{ ans_rule(20) \}
    END_TEXT

    ANS($ans->cmp());

    ENDDOCUMENT();

The solution "{1,2}" is marked correct, but "x=1 or x=2" (which we used to define the answer in the first place) is marked incorrect.

This is an issue for many of the problems in [1].

[1] https://github.com/openwebwork/webwork-open-problem-library/tree/master/OpenProblemLibrary/CollegeOfIdaho/setAlgebra_04_03_AbsoluteValue

dpvc:
The `x=1 or x=2` expression is supposed to produce a Set object equivalent to `{1,2}`, but it turns out that it is producing a Union object (with one element, the set `{1,2}`).  Due to some technical issues in how Union's are checked, this causes the correct answer to be marked incorrect.  It is easily fixed in the `lib/Value/Set.pm` file, but until that is is an official release, one way to handle it in the problem itself is to add

    $ans = $ans->{data}[0] if $ans->type eq "Union";  # hack to fix problem with inequality above

directly after the line that computes `$ans`.  That will fix it for now, and when the library is fixed, the `if` statement will be false, and it will stop doing anything.  So the problem should continue to work without modification (though it would be good to remove the hack eventually).

dpvc:
The `issue280` branch of my fork of PG includes the fix.

dpvc:
The branch has been merged into `develop`.

PR #281, opened 2016-08-29
Don't force the result of adding two sets to be a Union.  #280

dpvc:
Don't force the result of adding two sets to be a Union (in particular, let it be a Set).  Resolves issue #280.  That issue includes a test case.  Alternatively, use

```
TEXT((Set(1) + Set(2))->type);
```

Before the patch it should produce `Union`, but after the patch it should produce `Set`.


goehle:
The fix works for me.  Merging!

dpvc:
Thanks!

PR #283, opened 2016-08-31
Fix issue with radio buttons and numeric correct answers.

dpvc:
Fix issue with radio buttons not giving an error if the index is out of range, and add noindex option to prevent numeric value from being used as an index.

http://bugs.webwork.maa.org/show_bug.cgi?id=3718

includes one example of this.  You can also test it by using

```
loadMacros("parserRadioButtons.pl");
TEXT(RadioButtons([1,2,3],2)->{data}[0],",");
TEXT(RadioButtons([1,2,3],2,noindex=>1)->{data}[0]);
```

This should give you `B2,B1` as the result after the patch (before it should be `B2,B2`).

Also

```
loadMacros("parserRadioButtons.pl");
TEXT(RadioButtons([1,2,3],5));
```

shovel produce an error about the correct answer index being outside the range of choices.  (Before the patch, it produces a warning about `$value` not being defined).

### DIFF
--- a/macros/parserRadioButtons.pl
+++ b/macros/parserRadioButtons.pl
@@ -32,7 +32,9 @@ To create a RadioButtons object, use
 where "choices" are the strings for the items in the radio buttons,
 "correct" is the choice that is the correct answer for the group (or
 its index, with 0 being the first one), and options are chosen from
-among those listed below.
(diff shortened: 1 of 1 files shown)


goehle:
Fixes work as advertised.

PR #284, opened 2016-09-02
PGmatrixmacros and "cellspacing"

goehle:
PGmatrixmacros uses the "cellspacing" attribute to put space between the table cells it uses to build matrices.  This attribute is not supported in HTML5 and this patch replaces that attribute with inline css.  

To test:  Look at the problem `Library/WHFreeman/Holt_linear_algebra/Chaps_1-4/4.3.12.pg` and check that the spacing of the matrix is OK.

### DIFF
--- a/macros/PGmatrixmacros.pl
+++ b/macros/PGmatrixmacros.pl
@@ -289,7 +289,7 @@ sub dm_begin_matrix {
                       or $main::displayMode eq 'HTML_LaTeXMathML'
                       or $main::displayMode eq 'HTML'
                       or $main::displayMode eq 'HTML_img') {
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
The horizontal spacing looks OK. The vertical spacing might be too much imho. You _could_ use something like `border-spacing:10px 0px` to put zero additional vertical spacing. However I think the real excess vertical spacing is coming from something else: vertical margins or line height or something.

I can live with a little excess vertical spacing in matrices.

pstaabp:
This looks better, but I think Alex's css looks better.  I can resubmit with his suggestions.

mgage:
Hi Peter,

That's fine with me.  Go ahead and branch it.
-- Mike

pstaabp:
This is being closed in favor of PR #303.

PR #287, opened 2016-11-24
bugfixes for niceTables.pl

Alex-Jordan:
This patch addresses the two bugs that are described:
http://webwork.maa.org/moodle/mod/forum/discuss.php?d=3998

One issue: variable $footerline was referenced that is not mentioned anywhere else. I think it was an accidental remnant from a feature I tried to add one time but abandoned. Patch simply removes the reference.

Next issue: At line 367, inserted a check that a certain value is defined before using it in a comparison. 

The above two errors address the report at the forum post. Additionally, I discovered my own bug. It was possible for the LaTeX table output to create a cell with content like:
``
&\bfseries \multicolumn{...} &
``

And that construct is not allowed. You can't have something like `\bfseries` outside a `\multicolumn`. So the remaining edits to this file ensure that `\multicolumn` is outermost, if it arises.

Tested all this with the forum post's example problem. Also with all of the problems that the MBX ORCCA OER is using (around 500 at this point) many of which use `niceTables.pl` and all compiled correctly after these edits. If this leads to any issues reported in the forum or bugZilla I'll act quickly to address them.

I'm making this as a PR to `master`. Please advise if I should make a PR to `develop` instead or in addition.

### DIFF
--- a/macros/niceTables.pl
+++ b/macros/niceTables.pl
@@ -142,7 +142,7 @@ sub _niceTables_init {}; # don't reload this file
  #        texpre => tex code,           # For more fussy cell-by-cell alteration of the tex version of the table
  #                                      # TeX code here will precede the cell entry...
  #        texpost => tex code           # and code here will follow the cell entry
- #                                      # The ordering will be: tex texpre data texpost
+ #                                      # The ordering will be: texpre tex data texpost
  #        texencase => array ref        # This is just a shortcut for entering [texpre,texpost] at once.
  #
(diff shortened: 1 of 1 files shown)


mgage:
This looks good.  The vast majority of problems using niceTables.pl (or as previously named pccTables.pl)  have all been authored at PCC so those have all been checked.  This should not affect things even in the master branch so I'll patch this to both master and devel

mgage:
These work fine.  Adding the comments can only make the TeX output clearer.

PR #288, opened 2016-11-24
Adjustments to PGbasicmacros.pl for compatibility with MathBook XML

Alex-Jordan:
Two things are done here. 

One puts harmless LaTeX comments into hardcopy output for a problem that designate the beginning and ending of any solution, any hint, and the problem preamble that the beginproblem macro creates. These comments are going to be used by MathBook XML to parse a problem into chunks: preamble, statement(s), hint(s), solution(s). 

The preamble is largely ignored, although PGML preamble content is put into the document preamble of MBX latex output.

Statement(s), hint(s), solution(s) are each handled separately by MBX in different ways depending on processing flags.

The other thing that is done here is that `$envir->{setNumber}) ne 'not defined'` is added to the conditions for the `beginproblem` macro to refrain from printing a header. This is the condition in play when a problem is accessed as outlined at http://michaelgage.blogspot.com/2015/06/whether-writing-full-text-book-or-just.html.


mgage:
These work fine.  Adding the comments can only make the TeX output clearer.

PR #291, opened 2016-12-11
Develop update1

mgage:
Match hotfix updates 287 and 288 made to master branch.


mgage:
These are minor updates and won't cause troubles.  Since the updates were aggregated and then updated from mgage the history might look a bit more complicated than it actually is. These changes were made from release 2.12 and are independent of other changes.

PR #293, opened 2017-01-12
Extend capabilities

paultpearson:
Add a third (optional) argument to DifferenceQuotient that is a numerical value $zp to be substituted into the variable $dx named in the second argument.  This third argument is the "division by zero point" (denoted $zp in the code), and when $zp is substituted in for $dx in the student answer and results in a division by zero error for an otherwise correct answer, then students are told to simplify their answer further.

### DIFF
--- a/macros/parserDifferenceQuotient.pl
+++ b/macros/parserDifferenceQuotient.pl
@@ -1,6 +1,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright © 2000-2007 The WeBWorK Project, http://openwebwork.sf.net/
+# Copyright Â© 2000-2007 The WeBWorK Project, http://openwebwork.sf.net/
 # $CVSHeader$
 # 
 # This program is free software; you can redistribute it and/or modify it under
@@ -15,11 +15,8 @@
 ################################################################################
 
 =head1 NAME
-
 parserDifferenceQuotient.pl - An answer checker for difference quotients.
-
 =head1 DESCRIPTION
-
 This is a Parser class that implements an answer checker for
 difference quotients as a subclass of the Formula class.  The
 standard ->cmp routine will work for this.  The difference quotient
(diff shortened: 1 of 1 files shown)


dpvc:
I'm a little confused by the changes.  Can you give a use case for changing `$zp`?  In what setting would `$dx` not be going to 0?

paultpearson:
# Suppose you want students to simplify (x^2 - 4) / (x-2) to (x+2)
$df = DifferenceQuotient("x+2","x",2);
ANS($df->cmp);

paultpearson:
Another use case would be difference quotients of the form $\lim_{x \to c} \frac{f(x) - f(c)}{x-c}$.

dpvc:
Got it, thanks.  Hadn't considered that setup.

macros/parserDifferenceQuotient.pl
```diff
@@ -15,28 +15,27 @@
 ################################################################################
 
 =head1 NAME
-
 parserDifferenceQuotient.pl - An answer checker for difference quotients.
-
 =head1 DESCRIPTION
-
```
dpvc: Why are these blank lines being removed?  (Here and below.)  I particularly don't like the loss of the paragraph breaks below (original lines 30 and 37).  I'd also like to see a blank between the new lines 32 and 33.

Would you be willing to add those (back) in?

paultpearson: Yes, I'd be happy to do that.  Those blank lines must have been munched in the process of copying and pasting from one text editor to another, or inadvertently by me.

paultpearson: I'm going to close this pull request, make the changes, and submit another pull request.

dpvc: I don't think you need to do that.  You can just add commits to this one.  If you add them to this branch, they will become part of this pull request.  That will keep the comments together with the final pull request.

PR #294, opened 2017-01-13
Fix issue where test_at is ignored if upToConstant is used

dlglin:
This fixes an issue where if the upToConstant flag is set when checking an answer, the test_at option was being ignored.


dpvc:
This looks good.  Thanks for fixing it!

PR #298, opened 2017-05-25
Add Value::maketext() subroutine to Value::WeBWorK.pm

mgage:
Use as 
Value::maketext(…text to be translated …. )  

inside pg/lib   .pm files

### DIFF
--- a/lib/Value/WeBWorK.pm
+++ b/lib/Value/WeBWorK.pm
@@ -73,6 +73,11 @@ sub Value::Formula::PGseedRandom {
 }
 sub Value::Formula::PGgetRandom {shift->{PGrandom}->random(@_)}
 
+sub Value::maketext {
+	my $envir =eval('$main::rh_envir');
+	my $loc = $envir->{language_subroutine};
+	&{$loc}(@_);
+}
 #####################################################
 #
 #  Initialize contexts with WW default data


lib/Value/WeBWorK.pm
```diff
@@ -73,6 +73,11 @@ sub Value::Formula::PGseedRandom {
 }
 sub Value::Formula::PGgetRandom {shift->{PGrandom}->random(@_)}
 
+sub Value::maketext {
+	my $envir =eval('$main::rh_envir');
```
dpvc: I'm not sure I understand the need for `eval`.  Is that really necessary?

mgage: I think so.  The Value::WeBWorK.pm file is evaluated at compile time when the child is launched.  Each request is handled within a new Safe compartment which changes the value of $main (in addition the contents of the environment variable $rh_envir is different for every request). 

This is also how Davide handled the initialization of $envir for context  a few lines above in WeBWorK.pm
---------------------------------------
sub Parser::Context::copy {
  my $self = shift;
  my $context = Value::Context::copy($self,@_);
  return $context unless $Parser::installed;  # only do WW initialization after parser is fully loaded
  return $context if $context->{WW} && scalar(keys %{$context->{WW}}) > 0;
  my $envir = eval('\\%main::envir');
  return $context unless $envir && scalar(keys(%{$envir})) > 0;
  my $ww = $context->{WW} = {}; push @{$context->{data}{values}}, 'WW';
  return $context if $Value::_no_WeBWorK_; # hack for command-line debugging
  foreach my $x (@wwEvalFields) {$context->{WW}{$x} = $envir->{$x}}
  $context->flags->set(
     tolerance      => $ww->{numRelPercentTolDefault} / 100,
     zeroLevel      => $ww->{numZeroLevelDefault},
     zeroLevelTol   => $ww->{numZeroLevelTolDefault},
     num_points     => $ww->{functNumOfPoints} + 2,
     max_adapt      => $ww->{functMaxConstantOfIntegration},
     useBaseTenLog  => $ww->{useBaseTenLog},
  );
  $context->{format}{number} = $ww->{numFormatDefault} if $ww->{numFormatDefault} ne '';
  $context;
}

mgage: We should hold off accepting this pull request until we decide how to handle localization within MathObjects.  As Davide points out in  https://github.com/openwebwork/pg/pull/300   there may be a better way.

dpvc: I think the `eval` in the code you quote is used for error trapping (in case `%main::envir` doesn't exist -- I do testing in perl outside of WeBWorK proper).  I don't think the `eval()` has to do with the Safe compartment issues, here.  But it has been a long time since I worked on this part of the code.

mgage:
I'm going to remove this pull request for now.  We'll revisit this when we decide on the best way to translate the phrases inside MathObjects.

PR #300, opened 2017-05-25
add maketext calls in lib/Value/Interval.pm

heiderich:
This uses the subroutine Value::maketext() added by @mgage in #298

PR #301, opened 2017-05-25
Revert "add maketext calls in lib/Value/Interval.pm"

mgage:
Reverts openwebwork/pg#300


mgage:
Can add the Value::maketext() calls once the PR defining the Value::maketext() has been added.

PR #302, opened 2017-06-04
Develop multi answer fix

mgage:
This is companion pull request to webwork2 PR #767. It addresses bugs that occur in multiAnswer and matrix answer evaluators.  Many of the bugs arose because of the change to using the answer evaluator as the primary the primary object for a given answer —rather than the answer blank. (This helps regularize the increasing number of answer evaluators that use more than one answer blank — e.g. matrices and multiAnswer.  At the moment I don’t know of any answer evaluators that share an answer blank — if that comes up there are mechanisms to deal with it on a case by case basis — but no organized protocol.)

### DIFF
--- a/lib/PGcore.pm
+++ b/lib/PGcore.pm
@@ -491,6 +491,8 @@ sub record_array_name {  # currently the same as record ans name
 	$label;
 
 }
+
+
 sub extend_ans_group {         # modifies the group type
 	my $self = shift;
 	my $label = shift;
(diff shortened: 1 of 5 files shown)


mgage:
I believe this PR is ready for extensive checking.  It should be checked on questions that call multiAnswer() and check matrices.  I have attached the set definition file containing the questions I used for checking. You will need to change the extension from .txt to .def before uploading it to your webwork course.
[setmultiAnswer.txt](https://github.com/openwebwork/webwork2/files/1055342/setmultiAnswer.txt)

PR #303, opened 2017-06-22
Cellspacing

pstaabp:
This is to replace PR #284 with a little different CSS. 

Also apologies about all of the white space changes.   There doesn't seem to be a quick setting on my text editor to ignore those changes...

### DIFF
--- a/macros/PGmatrixmacros.pl
+++ b/macros/PGmatrixmacros.pl
@@ -481,13 +481,13 @@ sub dm_tth_delimeter {
         my ($top, $mid, $bot, $extra);
         my ($j, $out);
 
-        if($char eq "(") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
-        elsif($char eq ")") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
-        elsif($char eq "|") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
-        elsif($char eq "[") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
-        elsif($char eq "]") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
-        elsif($char eq "{") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
-        elsif($char eq "}") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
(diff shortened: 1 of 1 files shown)


macros/PGmatrixmacros.pl
```diff
@@ -481,13 +481,13 @@ sub dm_tth_delimeter {
...
         my ($j, $out);
 
-        if($char eq "(") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
-        elsif($char eq ")") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
-        elsif($char eq "|") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
-        elsif($char eq "[") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
-        elsif($char eq "]") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
-        elsif($char eq "{") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
-        elsif($char eq "}") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
+        if($char eq "(") { ($top, $mid, $bot, $extra) = ('�','�','�','�');}
```
Alex-Jordan: Peter, it looks like your text editor didn't encode these accented characters correctly.

Alex-Jordan:
You could try again, and use  `git add -p` to only add the one line of css change.

pstaabp:
I think I have it now.  I actually figured out where to set my editor to do that.  It appears that there are some strong feelings out there about removing trailing white space (and when not too).

Alex-Jordan:
One place in WeBWorK where removing trailing white space would make a difference is in a problem file that uses PGML, where trailing white space sometimes affects the markup.

Someday I would like to "clean up" the WeBWorK code. Addressing whitespace, indentation, etc. Removing obsolete code and comments. I think it would make a difference with getting new people involved. Less of an intimidation factor if the code is easier to read.

pstaabp:
I agree with the cleaning up part.  I have ideas on that (especially with webwork 3) if you want to discuss further.

Thanks for reminding me about PGML.  I usually use the built-in editor to work on problem writing and not this text editor.

mgage:
The matrices in the problem mentioned and a number of other randomly chosen problems containing matrices all look fine.  I will pull this.

PR #304, opened 2017-06-26
fix bug 3845 as proposed in the bug report; add maketext calls

heiderich:
- fix bug 3845 as proposed in the bug report: http://bugs.webwork.maa.org/show_bug.cgi?id=3845
- add maketext calls

Warning: I do not oversee the consequences of these changes, in particular the first one.
But on [https://perldoc.perl.org/functions/defined.html](https://perldoc.perl.org/functions/defined.html) we find:

> Use of defined on aggregates (hashes and arrays) is deprecated.


mgage:
This fixes bug 3845 and the maketext changes are straightforward.

PR #306, opened 2017-07-01
Fix typos, and make Unions report more uncombined sets

dpvc:
This fixes a couple of typos in `AnswerCheckers.pm`, makes sure that the derivative of a fraction copies the denominator when squaring it (otherwise the derivative shares a tree node with the original), and fixes an error in `isReduced()` for Union objects where unions like `[0,1) U {1}` would be reported as reduced (rather than requiring it to be `[0,1]`).

The latter can be checked using

```
Context("Interval")->flags->set(
  reduceUnions => 0,
  reduceSets => 0,
);

($reduced,$reason) = Union("[0,1) U {1}")->isReduced();
TEXT("Reduced = $reduced",$BR,"Reason = $reason");
```

Without the patch, this should report "Reduced = 0, Reason = uncombined sets", while without the patch, it should be "Reduced = 1, Reason = " with no reason.

### DIFF
--- a/lib/Parser/Differentiation.pm
+++ b/lib/Parser/Differentiation.pm
@@ -100,7 +100,8 @@ sub Parser::BOP::divide::D {
         $BOP->new($equation,'*',
           $self->{lop}->copy($equation),$self->{rop}->D($x))
       ),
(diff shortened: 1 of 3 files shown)


mgage:
Passes the tests.

PR #307, opened 2017-07-01
Convert vectors to column matrices for some matrix operations.

dpvc:
For the MatrixReal1 macros that require column vectors, convert vectors to column matrices.  This is for `normalize()`, `solve_LR()`, and other solve macros.

To check, use

```
TEXT(join(', ',Matrix([1,2],[3,4])->normalize(Vector(1,2))));
TEXT(join(', ',Matrix([1,2],[3,4])->solve_LR(Vector(1,2))));
```

Without the patch, these will throw errors.  With the patch, they should succeed.

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -412,6 +412,14 @@ sub wwMatrixLR {
   return $self->{lrM};
 }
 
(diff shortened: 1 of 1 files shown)


mgage:
This passes the tests.

PR #308, opened 2017-07-02
Add differentiation to fractions, and add cmp_defaults

dpvc:
This adds the `D()` method for differentiation (the inherited one produces fractions of the form 0/0 which throw an error).  Also add `cmp_defaults()` so that if a fraction-valued formula is used in an answer checker, it will work.

To test the differentiation, use

```
Context("Fraction");
$f = Fraction("1/2");
TEXT($f->D);
```

With the patch, this should output 0; without it, it should produce an error.

To test the second method, use

```
loadMacros("PGML.pl","contextFraction.pl");
Context("Fraction");
$f = Formula("1/2");
BEGIN_PGML
[_____]{$f}
END_PGML
```

Without the patch you should get an error; with the patch, you should get no error (and an answer blank for the fraction);

### DIFF
--- a/macros/contextFraction.pl
+++ b/macros/contextFraction.pl
@@ -618,6 +618,8 @@ sub reduce {
 package context::Fraction::Real;
 our @ISA = ('Value::Real');
 
+sub cmp_defaults {Value::Real::cmp_defaults(@_)}
+
 #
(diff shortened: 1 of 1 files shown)


mgage:
Passes tests.

PR #312, opened 2017-07-15
Update version number

mgage:
Update version number for PG

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -5,10 +5,10 @@
 # initialize PGcore and PGrandom
 
 
-$main::VERSION ="WW2";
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Sort of related, and maybe this is not the best place to comment, but I was wondering if it's possible to handle the version number differently. Could we make it so that it is loaded with some perl module? So if you switch versions but forget to restart apache, that it would still display the older version?

mgage:
Good idea.  I think we can force the version files to be read when the ContentGenerator.pm is compiled which accomplish that trick.  Hang on to the idea and we can put it into develop

PR #313, opened 2017-07-20
add full word options for some units

Alex-Jordan:
At PCC we have customized as you see in this commit, after there was enough demand to have all these units with full word options.

As part of my cleanup of how I manage the PCC WeBWorK server, I've partitioned these off into a separate branch and maybe these changes are things worth merging everywhere. But if whole words are not the direction to take the units parser, we can reject this.


Alex-Jordan:
OK, responding to feedback in PR #313:

I added 'sec', 'second', and 'seconds' as suggested. That hadn't occurred to me because second is a fundamental unit, unlike the others.

The other comment that was relevant to this PR was about 'foot' not being recognized. That kind of thing happens any time the parser doesn't identify the end of the expression as a unit. For instance, "5 m/widget" will lead to the message that "variable m is not recognized". And if more than one character long, like "5 met/widget", the message will be "met is not recognized". So what you were seeing wasn't about "foot" not being recognized as a unit. The parser wasn't even trying to recognize it as a unit in that situation.

Otherwise the testing passes as it did when you tested earlier. There shouldn't be any surprises here. When you look at these new unit words, none of them can be split into strings that are made up only of previously existing units, which could be a problem in theory.

Would you like these added to that units test in `pg/t/`? I can add more lines like I did for `%`, but as I noted in the other PR thread, I'm not sure how to check that what I added works. Actually running the test isn't working for me.

mgage:
The new additions all pass.  I'll upload two  a new pull request with a .pg test file and a  modified .t test file for units. 

The problem with the parser error message (which has been there forever) is not resolved here. 

The problem of how to resolve Hz remains somewhat open. 2pi rad/sec is an angular velocity
but it does appear that Hz is sec^-1 for most authorities. Possibly this can be resolved but it will require checking a lot of problems.  One consequence of our definitions of units is that 
% and  .01 rad/rad   are equal as numbers with units. Perhaps we can set cycle equal to 1 and 
define rad and degree in terms of cycle??

PR #314, opened 2017-07-20
New units

Alex-Jordan:
These are units used in some custom PCC library and ORCCA (a PreTeXt textbook) problems. I suspect `mph` is not controversial.

The other one here is `%`. I know there is `contextPercent.pl` for handling answers like `5 %`. But there's nothing for handling answers like `5 %/s` and that's what adding `%` does here.  And it adds it in such a way that it would be equivalent to `0.05 Hz`.

Test with a problem that makes use of `$a=NumberWithUnits("5 %")`. Test using `$a` in `TEXT`, `PGML`, as an answer, seeing what the answer appears as in the correct answer field, and all these things in hard copy output too.


mgage:
I've tested this and PR #313 together.  There is some more work that needs to be done. 

1. Do you want to include sec and seconds as well as s for seconds ( you have done that for hours and minutes). 
2. At present if you type  feet/sec  you get an error that feet is not defined. feet/s is correctly recognized so this is an issue with interpreting incorrect units -- possibly not easy to fix.
3. 5 %/s is not equivalent to .05 Hz -- the units aren't matching  -- some of this might
be a problem with Hz    5 Hz and 5 s^-1 are the same but 5/s is not. 
4. Here is some data:

5. I'm not convinced the definition for steradian  sr ("radian squared") is correct. Despite
the name it is supposed to be a solid angle that subtends region of area 1 on a sphere of radius 1
(or r^2 on a sphere of radius r). https://en.wikipedia.org/wiki/Solid_angle.  Making this equal 
to rad^2 doesn't seem right somehow.  

There is a file  pg/t/test_units.t  which is a unit test for the Units.pm module. (  heh, heh  ).
(Just run perl pg/t/test_units.t from the command line)
Could you add tests for the new units to that?

This the text of the .pg file that I created for testing is below. See if you get the same results 
I did. 

```
## DESCRIPTION
## Functions: Input and Output


## DBsubject(WeBWorK)
## DBchapter(TEST)
## DBsection(Units)
## Date(01/01/10)
## Institution(U of R)
## Author(Gage)
## Level()
## MO(1)

## KEYWORDS('test','units',)

DOCUMENT();

loadMacros(
"PGstandard.pl",
"MathObjects.pl",
"AnswerFormatHelp.pl",
"PGML.pl",
"parserNumberWithUnits.pl",
"parserFormulaWithUnits.pl",
"PGcourse.pl",
);

TEXT(beginproblem());

##################################
#  Setup

Context("Numeric");


$one_mph = NumberWithUnits(1,"mph");
$x2_mph  = FormulaWithUnits("x^2", "mph");

$hz_0_05 = NumberWithUnits(.05, "Hz");
$five_percent_per_second = NumberWithUnits(5, "%/s");
$five_percent = NumberWithUnits(5, "%");
$one_degree = NumberWithUnits(1, "deg");
$one_radian = NumberWithUnits(1, "rad");
$one_hour = NumberWithUnits(1,"hours");
$one_inch = NumberWithUnits(1,"inches");
$one_foot = NumberWithUnits(1,"feet");
$one_minute = NumberWithUnits(1,"minute");
$one_second = NumberWithUnits(1,"s"); #only s works
$one_cup = NumberWithUnits(.236588 ,"L");
$one_footpersecond = NumberWithUnits(1,"feet/s");
$x2_footpersecond = FormulaWithUnits("x^2","feet/s");


#####################################
#  Main text

Context()->texStrings;
BEGIN_PGML
These units are equivalent: 
* deg, degree, degrees
* rad, radian, radians,
* in, inch, inches
* ft, feet, foot
* min, minute, minutes
* h, hr, hour, hours
* s, sec,??
* cup, cups,  0.000236588 m^3 or .236588 L


* sr       ( steradian or square radian a measure of solid angle.  1 sr = rad^2 ?) The actual definition is the central solid angle of a patch on a sphere of radius r whose area is r^2.  A full sphere measures 4pi steradian's.

The definition here is  1 sr = 1 rad^2   --- is that really right?

TESTS:

These should be equivalent:   
* Enter .05 Hz in %/s [_________]{$hz_0_05}   
* Enter 5 %/s as Hz [__________]{$five_percent_per_second}   
* Enter 5 % as  [__________]{$five_percent}   
* Enter 1 mph [__________]{$one_mph}   
* Enter x^2 mph [__________]{$x2_mph}

* Enter one degree: [__________]{$one_degree}   
* Enter one radian: [__________]{$one_radian}   
* Enter one hour:   [__________]{$one_hour}   
* Enter one inch:   [__________]{$one_inch}   
* Enter one foot:   [__________]{$one_foot}   
* Enter one minute: [__________]{$one_minute}   
* Enter one second: [__________]{$one_second}   
* Enter one cup:    [__________]{$one_cup}   
* Enter one ft/sec  [__________]{$one_footpersecond}   
* [`x^2`] ft/sec    [__________]{$x2_footpersecond}   
END_PGML


ENDDOCUMENT();



```

Alex-Jordan:
OK, I'll take a look at all this. The steradian def is not something I added. That has how it has been for a long time, but probably no one is using steradians. I just added a space on that line to get the assignments aligned.

mgage:
I've uploaded the screen shot of the answer table again:

Alex-Jordan:
Working on this now. 

Personally I would define Hz as "so many times per second", which is why I would have had 0.05Hz equivalent to 5%/s. But Hz is defined here as 2pi rad/s, and I see the argument there. So for testing, 5%/s should be equivalent to 0.05/(2pi) Hz/rad.

I don't want to change how Hz is defined. Partly because I see the other viewpoint, and partly because OPL problems may rely on the current Hz definition.

Something like "5/s" is not recognized as a unit, but that issue is beyond the scope of what this commit is supposed to do. The parser just doesn't recognize that form. "5/m" etc is also not recognized.

mgage:
You could create a unit called cycles.  1 Hz   was  one cycle per second in my ham radio days when I was in high school. :-)

Alex-Jordan:
"rev" and "cycles" are already in there, equivalent to "2pi rad". I think I'll just leave all that stuff (Hz, cycles, etc) alone. And there will be no expectation that "% / s" is dimensionally equivalent to "Hz". 

This all stems from how the Units module is treating "rad" as on par with meters, seconds etc., as a fundamental unit. But "rad" is kind of an artificial, dimensionless unit (a ratio of two lengths). So in a sense equivalent to just a plain number. But it's too entrenched to think about changing it.

Later today I should have updates up for this one and also PR #313. And I'm going to seek out something else I could review.

Alex-Jordan:
Your sample problem has `Context()->texStrings;` before PGML starts. Can I confirm that should not be there? (But should be there before a BEGIN_TEXT?)

Alex-Jordan:
There is nothing unusual with mph, and your testing suggests it works as expected. So I'm focusing on `%` here.

[Here](http://spot.pcc.edu/~ajordan/new-units.pg) is a problem file for testing the percent character in NumberWithUnits. It is documented with what it is supposed to reveal. In summary, now that I added the most recent commit, things work well for `%` in MathJax, images, and tex output. Both in the problem statements and in showing correct answers. There are two configurations where it doesn't work, but those configurations also fail for units in general, and not because of `%`.

I think I added a test line for percent to `test_units.t` as requested, but when I try `perl pg/t/test_units.t` I get an error:
`Error:  Can't locate Units.pm in @INC...`
so I can't confirm this works. 

Before, you reviewed this one together with #313. If you would like to review this one by itself, I think it's ready now. I'll work on #313 next, but I'm not sure I'll get to it for a few days now.

Alex-Jordan:
Just noting that aside from tests in `pg/t/` that I can't confirm work, this one and #313 are ready for another review.

mgage:
See comments in merge for #313 which is a companion for this. I will shortly issue a pull request with test files for these units.

PR #316, opened 2017-07-20
allow omission of height from image subroutine and preserve aspect ratio

Alex-Jordan:
Prior to this commit, when calling the `image()` subroutine, omitting a height specification causes a default of 100 (px) to be used in HTML output. This changes the behavior so that when a height is not specified, it is not specified in the HTML `IMG` tag either. So the specified width (or the default width of 100) will be combined with the image's actual aspect ratio by the browser to define the width that we see.

I tested with:
```
$g = init_graph(-2,-2,2,2,axes=>[0,0],size=>[300,200]);
add_functions($g, "-x for x in <-2,2> " .
    "using color:blue and weight:2");

BEGIN_TEXT
\{image(insertGraph($g))\}

\{image("some-gif.gif")\}

\{image("some-jpg.jpg")\}

END_TEXT
```
where the first image is a dynamically created png, and the second two images were uploaded adjacent to the .pg test file. Best to use images that do not have 1:1 aspect ratios. Then play around with adding `width=>some-number` and see if the width is respected and the aspect ratio is maintained. Then also specify height and see if that is respected.

Test across browsers.

No need to test hard copy output because the code doesn't touch that.

This change in behavior should not affect OPL problems because they _should_ have been specifying a height up to now. The only way this would change OPL problems is if there is a problem that wanted the 100 px height and was obtaining it by letting the default take hold instead of explicitly asking for 100px. And even then, there will be no bad effect unless the problem author intentionally wanted a bad aspect ratio.

The purpose of this commit is to allow authors to only specify a width when using `image()` and not worry about height.

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2689,7 +2689,7 @@ sub image {
 	my %in_options = @opt;
 	my %known_options = (
 		width    => 100,
-		height   => 100,
+		height   => '',
 		tex_size => 800,
 		extra_html_tags => '',
 	);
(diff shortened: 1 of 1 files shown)


mgage:
This mostly works as expected.  Setting neither height nor width gives a width of 100 and preserves the native aspect ratio of the image. Setting width, without specifying height  changes the size of the image preserving the aspect ratio to the specified width.  Setting height, without setting width, uses the default size (100) for the width and changes the height -- does not preserve the aspect ratio. (this is with chrome -- some of this stuff may be browser dependent.)

mgage:
Seems like it does what is needed for PreTeXt.  and doesn't mess with the expected defaults. (The expected defaults were not well defined in any case.)

PR #317, opened 2017-07-27
Pg 2.13

mgage:


dpvc:
Mike, the last commit in this PR (73356d1) is not allowed by the git-flow model.  You never pull master into another branch; master is only ever a target of a merge, never a source.  (The only exception to this is hot-fixes, which originate from a branch off master and go back to master, but also get pulled into develop, so technically, this is pulling master into develop, but that is an accepted exception).

I suspect that this might have been because the branch wouldn't merge cleanly, and you were trying to fix that?  I think the error was actually mine, when I merged #294, which was targeted at master rather than develop, and I didn't catch it.  I should have retargeted that at the time, but missed it.  Had I caught it then, I could have removed that commit and redone it to develop, but at this point, the solution would have been to make a new commit to develop from the original branch.  That should have brought things up to date in develop so the release branch could be merged cleanly (at least, I think so).

In any case, for future reference, merging master into develop or a release branch is not recommended.

mgage:
Yeah.  I figured that out when I was about to merge things back in to develop.  I think so far the merge into master is innocuous because because master is merged in to PG-2.13 and then immediately merged back into master.  

I'm going to close this pull request and issue a new one from PG-213a -- which is a branch i created from the node just before master was merged into PG-213.  We'll see if that works.

PR #321, opened 2017-08-02
Add missing parentheses around vectors in IJK mode.

dpvc:
When vectors are printed in IJK mode as part of a formula, parentheses may be needed around them, as in `2(i+3t-k)`, but MathObjects isn't inserting these.  This PR resolves that issue.

To test, use

```
Context("Vector")->flags->set(ijk=>1);
TEXT(Compute("2<1,x,3>"),$BR);
TEXT(Compute("2<0,-x,0>"),$BR);
TEXT(Compute("2<0,x,0>"),$BR);
```

Without the patch, you will get `2*i+xj+3k`, `2*-xj` and `2*xj`.  With the patch, you will get (the correct results) `2*(i+xj+3k)`, `2*(-xj)` and `2*xj`.

See the associated [bug report](http://bugs.webwork.maa.org/show_bug.cgi?id=3702).

### DIFF
--- a/lib/Parser/List/Vector.pm
+++ b/lib/Parser/List/Vector.pm
@@ -27,6 +27,7 @@ sub _check {
 sub ijk {
   my $self = shift; my $context = $self->context;
   my $method = shift || ($context->flag("StringifyAsTeX") ? 'TeX': 'string');
+  my $precedence = shift || 0;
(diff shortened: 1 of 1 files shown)


mgage:
This works as described and resolves bug3702.

PR #322, opened 2017-08-02
Use proper closing markers for inequalities using 'and x != ...'

dpvc:
When displaying some inequalities, the type of the right-hand endpoint (open or closed) can be displayed incorrectly.  See the original [forum post](http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4162) for details.  Note that the display is wrong, but the internal representation is correct, and answers will be marked correctly even with the current code.

To test the patch, use

```
loadMacros("contextInequalities.pl");
Context("Inequalities");
TEXT(Compute("0 <= x <= 2 and x != 1"))
```

Without the patch, you will see the incorrect result `0 <= x < 2 and x != 1` (the `x <= 2` has become `x < 2`).  With the patch, you should see `0 <= x <= 2 and x != 1`.


mgage:
This works as advertised.

This tests pull request #322 and corrects a bug reported on the forum on July 31, 2017

PR #323, opened 2017-08-09
introduce BLTR and ELTR for "left to right" environments

heiderich:
Purpose: In HTML pages with global dir="rtl" setting (for instance in an
Arabic or Hebrew environment) one may still want mathematical formulas
to be printed from left to right. The pull request https://github.com/openwebwork/webwork2/pull/802 provides a theme for a global right-to-left environment.

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -402,7 +402,7 @@ sub NAMED_ANS_RULE {
 
 	    # Note: codeshard is used in the css to identify input elements 
 	    # that come from pg
-		HTML => qq!<input type=text class="codeshard" size=$col name="$name" id="$name" aria-label="$label" value="$answer_value"/>\n!.
+		HTML => qq!<input type=text class="codeshard" size=$col name="$name" id="$name" aria-label="$label" dir="auto" value="$answer_value"/>\n!.
 		              $add_html. # added for dragmath
                         qq!<input type=hidden  name="$previous_name" value="$answer_value"/>\n!,
 		
@@ -488,7 +488,7 @@ sub NAMED_ANS_RULE_EXTENSION {
 	MODES(
 		TeX => "\\mbox{\\parbox[t]{${tcol}ex}{\\hrulefill}}",
 		Latex2HTML => qq!\\begin{rawhtml}\n<INPUT TYPE=TEXT SIZE=$col NAME="$name" id="$name" VALUE = " ">\n\\end{rawhtml}\n!,
-		HTML => qq!<INPUT TYPE=TEXT CLASS="codeshard" SIZE=$col NAME = "$name" id="$name" aria-label="$label" VALUE = "$answer_value">!.
+		HTML => qq!<INPUT TYPE=TEXT CLASS="codeshard" SIZE=$col NAME = "$name" id="$name" aria-label="$label" dir="auto" VALUE = "$answer_value">!.
                         qq!<INPUT TYPE=HIDDEN  NAME="previous_$name" id="previous_$name" VALUE = "$answer_value">!
 	);
 }
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -77,6 +77,8 @@ my ($PAR,
 	$EUL,
 	$BCENTER,
 	$ECENTER,
+	$BLTR,
+	$ELTR,
 	$HR,
 	$LBRACE,
 	$RBRACE,
@@ -144,6 +146,8 @@ main::PG_restricted_eval( <<'EndOfFile');
 	$main::EUL              = EUL();
 	$main::BCENTER          = BCENTER();
 	$main::ECENTER          = ECENTER();
+	$main::BLTR          = BLTR();
+	$main::ELTR          = ELTR();
 	$main::HR				= HR();
 	$main::LBRACE			= LBRACE();
 	$main::RBRACE			= RBRACE();
@@ -197,6 +201,8 @@ EndOfFile
 	$EUL                 = EUL();
 	$BCENTER             = BCENTER();
 	$ECENTER             = ECENTER();
+	$BLTR             = BLTR();
+	$ELTR             = ELTR();
 	$HR				     = HR();
 	$LBRACE			     = LBRACE();
 	$RBRACE			     = RBRACE();
@@ -1512,6 +1518,8 @@ sub MODES {
 	$EUL    			EUL()  				end underlined type
 	$BCENTER    		BCENTER()   		begin centered environment
 	$ECENTER    		ECENTER()  			end centered environment
+	$BLTR    		BLTR()   		begin left to right environment
+	$ELTR    		ELTR()  			end left to right environment
 	$HR					HR()				horizontal rule
 	$LBRACE				LBRACE()			left brace
 	$LB					LB ()				left brace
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Would you educate me more about this? Suppose I am a rtl language user, operating a WeBWorK course. Why would it be better to address the math formula direction in the problem file code (one problem file at a time, one formula at a time), instead of in some global (WW config or MathJax config) setting?

Are rtl math formulas even possible with MathJax? And with the LaTeX processor that makes the image pngs? (I'm just in the dark with rtl languages, so forgive me if I'm asking silly questions.)

Do you have a small example problem file in some rtl language you could post? Something demonstrating some math that you would want to go rtl, and other math you would want to go ltr?

heiderich:
The main use-case for the moment is Arabic. I am not speaking it myself, but apparently normal text is written from right to left, while mathematical formulas are written from the left to right and I am not aware of any desire to typeset mathematics from the right to left. Within LaTeX it seems that formulas are always displayed from left to right (at least with default settings). So there is no need to specify it. But there are situations where formulas are split into parts, as in the following PG code snippet:

`[`x`] = [____________]{"1"}`

With rtl a global setting this would be displayed as

___ = x

It might be acceptable, but apparently this is not desirable. In WeBWorK formulas are often not restricted to single math-mode blocks, but include other HTML entities (such as the input field in this case). It seems to be desirable to typeset these formulas entirely from left to right. This is the motivation for this pull request.

It follows a translation of a problem from the OPL to Arabic due to @kamalsaleh. He could certainly explain the issues much better I.

```
## Translator(Kamal Saleh)
## DESCRIPTION
## Linear Algebra
## ENDDESCRIPTION

## Tagged by tda2d

## DBsubject(Lineare Algebra)
## DBchapter(Lineare Gleichungssysteme)
## DBsection(Matrix-Vektor-Form)
## Date(July 2013)
## Institution(TCNJ and Hope College)
## Author(Paul Pearson)
## Level(4)
## MO(1)
## KEYWORDS('matrix' 'equation')

loadMacros(
"PGstandard.pl",
"MathObjects.pl",
"PGcourse.pl",
);

TEXT(beginproblem());

Context('Matrix');

foreach $i (1..2) { 
	$u[$i] = non_zero_random(-5,5,1);
	$v[$i] = non_zero_random(-5,5,1);
}
$m = random(-1,1,2) * random(2,5,1);
$n = random(-1,1,2) * random(2,5,1);

$U = Matrix([[$u[1]], [$u[2]]]);
$V = Matrix([[$v[1]], [$v[2]]]);

$answer = Matrix([ $m*$u[1]+$n*$v[1], $m*$u[2]+$n*$v[2] ])->transpose;

Context()->texStrings;
BEGIN_TEXT
لتكن
\( A \) 
 مصفوفة تطبيق خطي أبعادها
\( 3 \times 2 \) 
و أعمدتها مستقلة خطياًًًََُ ً
.
ليكن
\( \vec{u} = $U \) 
و
\( \vec{v} = $V \)
شعاعين يحققان المعادلتين
\( A\vec{u} = \vec{a} \)
و
\( A\vec{v} = \vec{b} \) 
. 
أوجد حل 
\( \vec{x} \) 
للمعادلة
\( A\vec{x} = $m \vec{a} + $n \vec{b} \).
$BR 
$BR
$BLTR
\(\vec{x} = \) \{ $answer->ans_array(20) \}
$ELTR
.
END_TEXT
Context()->normalStrings;

ANS($answer->cmp);
;
ENDDOCUMENT();
```

kamalsaleh:
Exactly as Florian said, in Arabic all mathematical formulas and equations are written in Latin letters, exactly as in English mathematical texts. This is why we may need to write locally from left-to-right in the global right-to-left context.

heiderich:
I added the attribute `dir="auto"` to input fields. This is a priori independent of the left to right environment, but it is probably relevant in the same environments.

Alex-Jordan:
Alright,  I understand the issue better, and I see the related pull requests to webwork2. Trying to work through some open pull requests this week. I thought about these ones this evening, and here are my thoughts. Curious what @mgage (with localization experience) and @dpvc (with styling experience) think about all this too.

In https://github.com/openwebwork/webwork2/pull/802, you have a new theme for rtl in the HTML output. I think that is not a good long term approach, because whenever math4 gets updated, someone will need to remember to update this other theme too. Could `math4-overrides.css` be used instead?

In https://github.com/openwebwork/webwork2/pull/810, there is one very specific place where you are adding `dir="auto"`. Could the same thing be achieved using the CSS `direction` property and a very specific selector? Then we could avoid hard coding more into that table element, and instead put the right selector into `math4.css`.

My feeling at the moment is that we'd want a "bigger" strategy to support rtl languages than what is in these pull requests. For example, I see that Hebrew is a language option in Course Configuration, (but not yet Arabic). When I toggle to Hebrew (which I have no ability to read) the CSS is still styling text using left-alignment, and I assume right-alignment would be preferable. And what about hard copies? Do they print right to left? Are we using xelatex that would support characters from these languages?

For a "bigger" strategy, I'm imagining the user sets the language in course config to `ar-lb`, `heb`, or some other rtl language, and can keep the theme as `math4`. Then conditionally based on the language choice, page headers toss in some extra CSS lines at the end to change direction, alignment, and maybe other things. Some central place keeps a list of the supported rtl languages. Again based on the language choice, things get put into the hardcopy header to support that language and its direction. Probably everything I am saying should come after the utf-8 support is merged.

I'm describing a long term goal, and I worry that these current pull requests will interfere or muddy the transition. With this PR, the `dir=auto` give me pause. `$BLTR` and `$ELTR` give me pause because there is no provision for what they do in the hard copy. (Also, I'm not a fan of the $BFOO,$EFOO macros because they don't enforce parity. I've seen lots of problems with $BFOO, but $EFOO is left off. No one notices at first because HTML is forgiving. But the hard copy is broken. Is a subroutine inside `\{`...`\}` maybe better?)

While a longer term plan is on a back burner, could `$BLTR` and `$ELTR` be put into some new macro library instead? Could everything being done with `dir` be handled with CSS `direction` in `math4-overrides.css`?

mgage:
The companion PR in webwork2 adds math4-ar to the list of possible templates.  If math4-ar is not selected there is no effect on existing problems.  The most obvious change of selecting math4-ar is right justification for all of the pages. Since we don't have arabic translations of most webwork2 pages yet it's hard to tell whether this is the desired effect.  

We will also need to merge the PR related to utf8 before we will be able to view the arabic script. These files can serve as good tests for the utf8 changes as we start to review and merge them.

@Alex-Jordan points out that in the long run cloning math4 might require more work as each change in math4 would need to be duplicated. This is true, but in the short term it's useful to have this file separate so that it can be experimented with and updated without fear of affecting math4 and existing courses.   Once it's determined what a math4_ar (or math4_RL) template should look like we can try to overlay the css of math4 to accommodate right to left languages.

In the meantime this these files do no harm and give us something to experiment with. 

I will submit a copy of the arabic problem above to my unit test pull request.

taniwallach:
I am starting work on bringing WW into use in Hebrew in my university. Our needs will depend on getting the UTF8 support and RTL support working, so I have quite some interest in the needed features.

Here are some thoughts and comments on the discussion above:

Webwork2 PR #850 [https://github.com/openwebwork/webwork2/pull/850] is an attempt to provide global language and RTL changeover at the HTML level by setting the values according the the course language setting. For Hebrew (and Arabic) the PR will trigger the global RTL direction. It may remove the need to have a separate RTL template like math4-ar in the long run, but we may need to find a mechanism to apply CSS overrides conditionally (only to RTL languages) to fully avoid the need for a RTL template in the long run. 

However, a global setting (such as math4-ar and PR 850 apply) is certainly not perfect. It forces existing LTR material to appear in RTL mode in an RTL course. For example when browsing English problems in the OPL (looking for candidates to translate, or even assigning an English problem in a non-English course) causes trouble, as the text is not meant to be handled under the RTL direction.

Thus, we also need to allow problems to set their language and direction in a manner which override the global HTML setting. I started work on trying to do this also, and hope to have initial code ready soon.

The matter of interleaving LTR and RTL material inside mixed text is somewhat complicated, and will need careful work. As noted, math in Hebrew and Arabic get written LTR, as can words/phrases in English or other LTR languages. The various LaTeX packages (babel, polyglossia) which support Hebrew and Arabic use various methods of swapping direction both inside a paragraph primarily comtaining text of the other direction (in HTML "span" is used to do this) and at the paragraph level. 

Babel (older) and Polyglossia (newer and only for xelatex) do this differently, and it is likely that getting the proper types of macros and settings to enable hardcopy support will be somewhat complicated, and will require conditionals or special LaTeX macros so that the code will run under different sets of packages and LaTeX engines. As a first stage, we may need try to get things working with a custom "hardcopy theme" linked to a specific set of packages. I hope to be able to create something which works well for Hebrew using polyglossia and xelatex.

@Alex-Jordan  - I do agree that it probably makes sense to move all PG macros which are needed to support RTL language needs into a separate Perl file. 

I agree that it makes sense to move the recently added BLTR and ELTR (which seem to are "span" level in HTML) into a new macro file. I think that they also might better be renamed in a manner to indicate that they are for "in paragraph" (span style) use. Adding macros to allow a LTR paragraph inside a RTL problem is another thing to consider and implement. It may even be necessary to allow nesting of RTL spans inside LTR spans inside a RTL paragraph to get the intended text to display properly. As I work on getting some material translated to Hebrew, I will discover to what extent there are such needs.

taniwallach:
An additional remark about what @Alex-Jordan asked above. CSS is not a sufficient approach to handle RTL needs, as some users who depend on certain types of accessibility support tools will not necessarily have CSS being parsed. Thus, the setting of the text direction and language code should be done at the level of HTML tags.

https://github.com/openwebwork/pg/pull/384 has code to try to detect parity errors for the begin/end formatting variables. It is a reworked version of the code in the close https://github.com/openwebwork/pg/pull/383 as the fiirst attempt would have broken certain uses outside of `BEGINBTEXT`/`END_TEXT` type blocks. Thanks to @dpvc for the critical feedback about the first version of the code.

PR #324, opened 2017-08-25
Modify warning messages 

mgage:
Modifications in warning messages for extending answer rules and for persistent…storage

### DIFF
--- a/lib/PGcore.pm
+++ b/lib/PGcore.pm
@@ -527,7 +527,7 @@ sub store_persistent_data {  # will store strings only (so far)
 	my $self = shift;
 	my $label = shift;
 	my @content = @_;
-	$self->internal_debug_message("PGcore::store_persistent_data: storing $label in PERSISTENCE_HASH");
+	# $self->internal_debug_message("PGcore::store_persistent_data: storing $label in PERSISTENCE_HASH");
 	if (defined($self->{PERSISTENCE_HASH}->{$label}) ) {
 		warn "can' overwrite $label in persistent data";
 	} else {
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
Is there anything I can do to test this? What's the reason that the internal_debug_message is commented out?

mgage:
The internal_debug_message was simply an alert that the persistent storage method was being used instead of storing the data in the html in a fake answer blank.  It was there for a while so that if something went wrong I would be reminded of where to look for errors.  I've commented it out for general use since the warning (which is really just an alert) won't be a bother in production use.

There is not really much else to check other than to make sure that the new code compiles.  These things will only come up if you are doing multiplAnswers of some type and not using the built in methods.  The old version of draggableProofs for example had warning errors about "answer_group" missing -- the current version has fixed that, but you can see the change I made earlier this fall. The change here is to make it clearer what the fix is. (draggableProofs is  in OPL/macros/MC )

lib/PGcore.pm
```diff
@@ -527,7 +527,7 @@ sub store_persistent_data {  # will store strings only (so far)
 	my $self = shift;
 	my $label = shift;
 	my @content = @_;
-	$self->internal_debug_message("PGcore::store_persistent_data: storing $label in PERSISTENCE_HASH");
+	# $self->internal_debug_message("PGcore::store_persistent_data: storing $label in PERSISTENCE_HASH");
```
dsteinmo: I wouldn't be much of an industrial dev if I didn't call out that keeping around dead code in comments is generally an anti-pattern. I found a decent discussion on this here: https://blog.kentcdodds.com/please-don-t-commit-commented-out-code-53d0b5b26d5f

tl;dr: It confuses the next developer to come along, and doesn't really provide any context. All the context is kept around in revision control though, so deleting it is safe and git handles all the archiving.

mgage: Thanks for the pointer to the blog.  It's quite informative.  
I'll remove the commented out code.  
Some of my habits were developed long before we were using git, or even CVS to version code.  :-)

PR #326, opened 2017-09-18
Tidying up addition of customizeLaTeX

jwj61:
New file for customizable LaTeX constructions.  Instructors can override the default in a PGcourse file (usage of these constructs will already load PGstandard so they get the original definitions, and PGcourse so they can be overridden).

### DIFF
--- a/macros/customizeLaTeX.pl
+++ b/macros/customizeLaTeX.pl
@@ -9,6 +9,10 @@ The functions are loaded by default.  Any/all can be overridden
 in your course's PGcourse.pl
 =cut
 
+sub _customizeLaTeX_init {
+
+} #prevents this file from being loaded twice.
+
 ##### Set theory macros
 sub set_minus{ 
 	#return "\\setminus";
@@ -17,7 +21,7 @@ sub set_minus{
 
 #####  Logic macros
 sub negate { 
-	return "\\mathbin{\\sim}";
+	return "{\\sim}";
 	#return "\\lnot";
 };
 
--- a/macros/customizeLaTeX.pl
+++ b/macros/customizeLaTeX.pl
@@ -1,0 +1,84 @@
+=head1 NAME
+
+customizeLaTeX.pl - Defines default LaTeX constructs for certain mathematical
+                    ideas.
+
+=head1 DESCRIPTION
+
+The functions are loaded by default.  Any/all can be overridden 
+in your course's PGcourse.pl
+=cut
+
+##### Set theory macros
+sub set_minus{ 
+	#return "\\setminus";
+	return '-';
+};
+
+#####  Logic macros
+sub negate { 
+	return "\\sim";
+	#return "\\lnot";
+};
+
+sub implies {
+	return "\\Rightarrow";
+}
+
+##### Algebra macros
+
+sub cyclic { 
+
+        my $n = shift;
+
+        # leave one of the following return commands uncommented, depending on what notation you want to use for finite cyclic groups (e.g., Z/nZ)
+        
+        # display order n cyclic group as Z_n
+        return "\\mathbb{Z}_{$n}",
+        
+        # display order n cyclic group as C_n
+        # return "C_{$n}";
+        
+        # display order n cyclic group as Z/nZ
+        # return "\\mathbb{Z}/{$n}\\mathbb{Z}";
+
+};
+
+sub dihedral { 
+
+        my $n = shift;
+        
+        # if you want to display dihedral groups as D_n (for instance, D_4 is the dihedral group of order 8), then leave this subroutine unmodified
+        
+        
+        # if you want to display dihedral groups as D_{2n} (for instance, D_8 is the dihedral group of order 8), then uncomment this set of if/else statements. The regular expression conditionals are to make sure it handles different types of arguments correctly.
+        # if( "$n" =~ m/^\s*(\d+)\s*$/ )
+        # {
+                # $n = 2 * $1;
+        # }
+        # elsif( "$n" =~ m/^\s*(\w+)\s*$/ )
+        # {
+                # $n = "2$1";
+        # }
+        # else
+        # {
+                # $n = "2($n)";
+        # }
+                
+        return "D_{$n}";
+        
+};
+
+sub quaternions {
+
+        # if you want to display the Quaternion group as Q_8, then leave this subroutine unmodified
+        
+        return "Q_8"
+        
+        # Alternatives
+        
+        # return "H_8"
+        # return "Q"
+};
+
+1;
--- a/macros/customizeLaTeX.pl
+++ b/macros/customizeLaTeX.pl
@@ -21,13 +21,21 @@ sub set_minus{
 
 #####  Logic macros
 sub negate { 
-	return "{\\sim}";
+	return "\\mathbin{\\sim}";
 	#return "\\lnot";
 };
 
 sub implies {
 	return "\\implies";
 	#return "\\Rightarrow";
+}
+
+##### Linear algebra macros
+
+sub vectorstyle {
+	my $v = shift;
+	#return "\\vec\{$v\}"
+	return "$v";
 }
 
 ##### Algebra macros
@@ -48,6 +56,12 @@ sub cyclic {
         # return "\\mathbb{Z}/{$n}\\mathbb{Z}";
 
 };
+
+# Macro to display the ring Z/nZ
+sub ZmodnZ {
+	my $n = shift;
+	return "\\mathbb{Z} / $n \\mathbb{Z}";
+}
 
 sub dihedral { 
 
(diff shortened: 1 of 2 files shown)


macros/customizeLaTeX.pl
```diff
@@ -0,0 +1,84 @@
...
+
+##### Set theory macros
+sub set_minus{ 
+	#return "\\setminus";
+	return '-';
+};
+
+#####  Logic macros
+sub negate { 
+	return "\\sim";
```
Alex-Jordan: `\sim` is a LaTeX binary relation, and so something like `x\vee\sim y` doesn't look right. There will be too much space between the `\sim` and the `y`. I think (but am not sure) that the "right" thing to use is `\mathop{\sim}`.

macros/customizeLaTeX.pl
```diff
@@ -0,0 +1,84 @@
...
+	#return "\\lnot";
+};
+
+sub implies {
+	return "\\Rightarrow";
```
Alex-Jordan: `\implies` would be better than `\Rightarrow`. Here's a discussion: https://tex.stackexchange.com/questions/47063/rightarrow-vs-implies-and-does-not-imply-symbol

Alex-Jordan:
Can you say more about how this would be used? For instance, you have defined the perl subroutine `set_minus`. How would an author use that? If I'm writing the text of a problem , I might write `..we have \(A\setminus B\)...` or in PGML I might write ``...we have [`A\setminus B`]``. Can you say how the subroutine `set_minus` helps? 

I'm wondering how I even use it if I'm in math mode. Is this something you use external to the math mode like `$AnB = "A set_minus() B";` and then later in math mode you just invoke `$AnB`?

macros/customizeLaTeX.pl
dpvc: No, `\mathop{\sim}` will make it act like `\sum`.  You want it to act like `-` so use `\mathbin{\sim}`

jwj61:
The basic idea here is to collect LaTeX math macros where some people like one version and other people like another.  For example, I like a minus sign for set difference, but other people like backslash.  If a problem accesses the symbol through the macro, then instructors can override it in PGcourse.

In a pg problem, I would do something like

$implies = implies();

Then between BEGIN_TEXT/END_TEXT, I can write \( P $implies Q \), which is not too cumbersome.  I don't know how this would affect PGML authors.

Alex-Jordan:
A PGML author would need to write ``[` P [$implies] Q `]`` with the inner brackets.

Is this something that the OPL curators would retrofit problems with? I'm pessimistic about most (not all) problem authors being/staying conscious of allowing others to customize their problems. Some may not even realize that some of the alternate conventions exist. But if the OPL curation process included inserting this when applicable, that could have a big impact. And of course some authors would use it directly.

jwj61:
I expect this to be retrofitted to problems in a piecemeal fashion.  Someone else has already started doing this.  If we ever thought we got to 100% compliance for existing problems, then we would start worrying about new additions.  Since many people look to existing problems as templates for new ones, the more retrofitting to take place, the more likely people are to pick up on it in the future.

Alex-Jordan:
Coming back to this. Small things: I think there is a comma in the subroutine for cyclic that should be a semicolon. And do you want to change the defaults of `\Rightarrow` to `\implies`? And `\sim` to `\neg`? Or if you feel those are better defaults, could you say why?

Bigger thing. Subroutines seem like overkill for this. And as this grows, it would add more subroutines that are loaded essentially by default (because of inclusion in PGbasicmacros.pl), and some of those subroutine names may start to conflict with subroutine names in other macro libraries or written in problem preambles. Like if a macro library author wanted to make a `sub negate` that was for some totally different purpose.

Just an idea, but what if this was all done with a single hash, defined in `PGbasicmacros.pl`? Like:
```
%CLTX = (
  setminus=>'-',
  #setminus=>'\setminus',

  negate=>'\neg{}',
  #negate=>'\mathbin{\sim}',

  implies=>'\implies{}', 

  cyclic=>sub {
      my $n = shift;
          # Leave one of the following uncommented.
          return '\mathbb{Z}_{' . $n . '}';
          #return "C_{$n}";
          #return '\mathbb{Z}/{' . $n . '\mathbb{Z}';
  },

  dihedral=>sub {
      my $n = shift; 
      # If you want to display dihedral groups as D_n 
      # (for instance, D_4 is the dihedral group of order 8), 
      # then comment all lines except the last.
      # If you want to display dihedral groups as D_{2n} 
      # (for instance, D_8 is the dihedral group of order 8), 
      # then uncomment all lines.
      # The regular expression conditionals are to make sure 
      # it handles different types of arguments correctly.
      if( "$n" =~ m/~~s*(~~d+)~~s*$/ )
          {$n = 2 * $1;}
      elsif( "$n" =~ m/^~~s*(~~w+)~~s*$/ )
          {$n = "2$1";}
      else {$n = "2($n)";}
      return "D_{$n}";
  },

  quaternions=>'Q_8',
  #quaternions=>'H_8',
  #quaternions=>'Q',
  );
```

Then you could do like:
```
$implies = $CLTX{implies};
$myDn = $CLTX{dihedral}(3);

BEGIN_TEXT
\( P $implies Q \)
$PAR
\($myDn\)
END_TEXT

BEGIN_PGML
[` P [$implies] Q`]

[`[$myDn]`]
END_PGML
```

Or inline like:

```
BEGIN_TEXT
\( P \{$CLTX{implies}\} Q \)
$PAR
\(\{$CLTX{dihedral}(3)\}\)
END_TEXT

BEGIN_PGML
[` P [$CLTX{implies}] Q`]

[`[$CLTX{dihedral}(3)]`]
END_PGML
```

Of course, this is a bit more cumbersome to use with the "CLTX" (or whatever name is best). But whatever the hash's name is would be the only thing to conflict with.

jwj61:
Thank you for taking a look at this.

- the comma was a typo, now fixed
- the goal is not to change how a specific latex symbol is rendered, but to take some conceptual things which mathematicians notate differently, and let them customize it.  I think it is pointless to tell people they need to use \implies as opposed to \Rightarrow for implications, or vice versa.  We just let them choose.  I can imagine that the standard in some countries is different than in others, or that a formal logic course may set different priorities than a "how to do proofs course".
- I have no axe to grind as far as what should be the defaults.  I think I changed them to the ones you like.
- I see you point about the namespace, but it doesn't really worry me.  WeBWorK is 20 years old, pretty stable, and any additions that come as object methods are immune.
- That said, I am not totally opposed to doing it as a hash.  As you point out, that is a little less convenient for authors.  Also, the abstract algebra functions are already in use.  Either those pg files would have to be changed, or we leave abstract algebra out of this central place for customizations.

macros/customizeLaTeX.pl
```diff
@@ -0,0 +1,85 @@
...
+
+customizeLaTeX.pl - Defines default LaTeX constructs for certain mathematical
+                    ideas.
+
+=head1 DESCRIPTION
+
+The functions are loaded by default.  Any/all can be overridden 
+in your course's PGcourse.pl
+=cut
+
```
mgage: sub _customizeLaTeX_init {

}  #prevents this file from being loaded twice.

mgage: add this snippet to keep the file from being loaded twice.

jwj61:
I added the initialization code.

I also changed the spacing for sim which used as logical not (based on discussion from the web).

macros/customizeLaTeX.pl
```diff
@@ -0,0 +1,103 @@
...
+sub implies {
+	return "\\implies";
+	#return "\\Rightarrow";
+}
+
+##### Linear algebra macros
+
+sub vectorstyle {
+	my $v = shift;
+	#return "\\vec\{$v\}"
```
mgage: I rewrote this as 

sub vectorstyle {
	my $v = shift;
	#return "\\vec\{$v\}";
	 return "\\vec{$v} ";
	#return "$v";
}

so that the default value shows some change in the style of the vector.  As far as I can tell the backslash before the braces is unnecessary although it does no harm in PGML.

macros/customizeLaTeX.pl
```diff
@@ -0,0 +1,103 @@
...
+        # {
+                # $n = "2($n)";
+        # }
+                
+        return "D_{$n}";
+        
+};
+
+sub quaternions {
+
```
mgage: should this be quaternion     for consistency with dihedral, cyclic   etc. ?

macros/customizeLaTeX.pl
mgage: there are double backslashes before vec. (the HTML wipes them out).  There were single backslashes before the braces in the 3rd line.

jwj61:
I changed the default for vector.

I left quaternions for two reasons.  First, this came directly from the UMass Amherst algebra macros, so I would rather stay consistent with what is already in use.  

Second, it is not quite parallel with cyclic and dihedral.  Those two are families of groups whereas quaternions refers to a single group.  One might refer to "the quaternion group" or "the quaternions", so the plural version seems ok.

mgage:
Sounds good to me.

PR #327, opened 2017-09-29
PreTeXt support

Alex-Jordan:
This PR adds a `PTX` display mode, which makes XML source for a PreTeXt document. (http://mathbook.pugetsound.edu/)
There is a companion PR to webwork2.

The purpose is for WeBWorK to render a problem (with a seed) into XML content that could be pasted into a PTX document. Or, more likely, harvested by some script and inserted into a PTX document. To appreciate what this means before starting a review, you might visit [this eBook](http://spot.pcc.edu/math/orcca/section-arithmetic-with-negative-numbers.html) and scroll to the bottom where the exercises come from OPL problems. Those exercises have simultaneously made it into this [PDF version](http://spot.pcc.edu/~ajordan/orcca.pdf) of the same book. And this [demonstration WeBWorK course](https://webwork.pcc.edu/webwork2/orcca-demonstration/) shows the same exercises as live, randomized WeBWorK questions.

Nothing here should interfere with any of the ways that WeBWorK is currently used, so testers should have that in mind and raise an alert if they discover something changes with regular WeBWorK usage.

[Here is a .pg file](http://spot.pcc.edu/~ajordan/pretext.pg) I used for testing. To test, you need to first set up a course as laid out [here](http://mathbook.pugetsound.edu/doc/author-guide/html/section-82.html). Then place the .pg file in that course's templates folder. Then use a web browser to visit:
https://your.ww.server/webwork2/html2xml?&answersSubmitted=0&sourceFilePath=pretext.pg&problemSeed=1234&displayMode=PTX&courseID=anonymous&userID=anonymous&course_password=anonymous&outputformat=ptx
Then view the source (not what the web browser renders it as). What you should be seeing is valid PreTeXt XML code. (You may not know how to check that it is valid, but comments are welcome regarding the file headers and footers.)

There's a lot to look at here in the code, but most of it is easy to review quickly. The bulk of these edits just add a `PTX` key to places where `MODES` is used, or other places where `$displayStyle` is checked as this or that type. So a strategy for reviewing this code is to go through the diff and see that each change is only effective when `$displayMode eq 'PTX'`. Beyond that, here are noteworthy file edits:

In Translator.pm, a STATEMENT macro is used that is analogous to the HINT and SOLUTION macros. It's definition is in PGbasicmacros.pl. This should not be noticeable for existing uses of WW, but the explicit demarcation of where non-hint, non-solution content is printed was important for PTX.

`PGML.pl` has an entire section devoted to PTX, analogous to its HTML and TeX sections.

`PGbasicmacros.pl` has a cleanup subroutine to format output.

### DIFF
--- a/lib/PGalias.pm
+++ b/lib/PGalias.pm
@@ -265,7 +265,7 @@ sub make_alias {
 		    or $ext eq 'js'
 		    or $ext eq 'nb'
 		    ) {
-		if ($displayMode =~ /^HTML/ ) {			 
+		if ($displayMode =~ /^HTML/ or $displayMode eq 'PTX') {
 			 $adr_output=$self->alias_for_html($aux_file_id, $ext);
 		} elsif ($displayMode eq 'TeX') {
 			################################################################################
--- a/lib/WeBWorK/PG/Translator.pm
+++ b/lib/WeBWorK/PG/Translator.pm
@@ -1858,8 +1858,8 @@ sub default_preprocess_code {
 	$evalString =~ s/\n\h*END_PGML_HINT[\h;]*\n/\nEND_PGML_HINT\n/g;
 	$evalString =~ s/\n\h*END_SOLUTION[\h;]*\n/\nEND_SOLUTION\n/g;
 	$evalString =~ s/\n\h*END_HINT[\h;]*\n/\nEND_HINT\n/g;
-	$evalString =~ s/\n\h*BEGIN_TEXT[\h;]*\n/\nTEXT\(EV3P\(<<'END_TEXT'\)\);\n/g;
-	$evalString =~ s/\n\h*BEGIN_PGML[\h;]*\n/\nTEXT\(PGML::Format2\(<<'END_PGML'\)\);\n/g;
+	$evalString =~ s/\n\h*BEGIN_TEXT[\h;]*\n/\nSTATEMENT\(EV3P\(<<'END_TEXT'\)\);\n/g;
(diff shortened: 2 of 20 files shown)


mgage:
@Alex-Jordan -- this PR has conflicts now in PGbasicmacros.pl

Also it appears to be based on an earlier version of develop.  Would it be possible for you to rebase it 
on the current version of develop so that not so many commits are duplicated?  Just for example
customlatex file was committed yesterday and appears also in this pull request.

@dpvc  is there an easy way to do this?  Is it necessary for clean merge lines?

Alex-Jordan:
Hmm, I rebased off of develop, and then this thing blew up to 35 commits, 41 files changed. I'll investigate more.

Alex-Jordan:
OK, now I think it's cleaner. How does it look to you?

dpvc:
> @dpvc is there an easy way to do this? Is it necessary for clean merge lines?

It looks like Alex has rebased it properly, so it is now clean.  I like it when the merges don't involve merges from other branches, as that makes it hard to tell what is being combined, and it is possible that that will include other commits of the other branch is not yet fully merged.  I suspect that is what happened here, but I can't tell, since the original branching is no longer in the diagram.

In any case, it looks good now.

dpvc:
PS, I made a quick look through the changes, and don't see any red flags.

mgage:
I've been able to check that this does no harm to the standard use of webwork.  
I haven't done extensive testing with PreTeXt but if there are problems or if improvements
need to be made that can be done in later PR.

PR #328, opened 2017-10-08
Develop pod docs

mgage:
This is a preliminary reworking of the POD documentation for PG files which use matrices in some way.  It's a first step toward unifying the documentation for the various types of matrices which are available in PG.

Test that the files compile and don't blow up when being used for problems in linear algebra and multivariable calculus.  

Typo corrections and suggestions for better wording would be welcome but are not crucial. In addition if you notice any factual inaccuracies about how these functions and methods work
please let me know.


mgage:
Replaced by pull request #338

PR #329, opened 2017-10-08
Develop sagefix

mgage:
Changed to sage.pl to accommodate the new way of handling html fragments and to mesh with the new handling of answer blanks.


mgage:
Test this file against a collection of sage interacts (such as the ones that John Travis has developed) and make sure that they work with the develop version of WeBWorK and the current versions of SageMath.

mgage:
John Travis's sage interacts can be found in the OPL at https://github.com/openwebwork/webwork-open-problem-library/tree/master/OpenProblemLibrary/MC/Sage.  
They can be reached using the library browser and clicking on the "OPL Directory" button.

gajennings:
I checked most of John Travis's sage interacts.  All exercises worked except MC/Sage/Trig/Solve_Trig_Expression2.pg which reports an error.
The "Experimental stuff" in sage.pl also work, if one corrects the typo on line 137.
Suggestions (line numbers refer to sage.pl):
line 137 fix typo -  $top_test should be $top_text
lines 127-129 add an item - hide: ["permalink"] - to the list of configuration options to hide the "share" button
line 23 comment out - SageAnswerValue is not used anywhere
line 27 comment out - accepted_tos is not used, and does not appear as a configuration option for the sage cell server.
lines 122-125 comment out?  These lines don't seem to add any new features to what's already there.

mgage:
"line 27 comment out - accepted_tos is not used, and does not appear as a configuration option for the sage cell server."

This is a change in how the sagemathcell operates:
see https://groups.google.com/forum/#!topic/sage-cell/UOSKrJjNHc4 

```
...
    That appears to have solved my issue. Thanks!
    Will terms of service stay off or is this a temporary change?

Most likely remain off - they were introduced due to UW lawyers, but we no longer host there.
```

mgage:
More sageMathCell references:
https://sagecell.sagemath.org/static/about.html?v=98b56535a5f3e54e272938b62c79287c
and for more detail:
https://github.com/sagemath/sagecell/blob/master/doc/embedding.rst

mgage:
I've partially resolved the CSS styling issues when a sage interact is used.  The sage interact seems to reload an old version of bootstrap.js or in some other way interferes with the bootstrap function calls.
In particular the functions $.fn.popover  and $.fn.modal  are not available.  

A companion PR to webwork2  (june1_fix) modifies math4.js and problem.js so that these functions are used only if they are defined on the page.  This means that tooltips for the attemptResults table work, but have the old styling before we started using bootstrap styling.  

Trying to avoid this by loading the sage interact into an iframe and communicating with it through the iframe might be a good idea, but is too big a project to attempt at the moment.

PR #330, opened 2017-10-14
Fix checks for scalar multiplication to not have to stringify the val…

dpvc:
Fix checks for scalar multiplication (for points, vectors, and matrices) to not have to stringify the value if it is already a Real or Complex number.  This resolves [bug 3991](http://bugs.webwork.maa.org/show_bug.cgi?id=3991).  The issue was that when `Context()->texStrings` is in effect, the string match that was used to decide if the coefficient is a number would match against the TeX version of the number.  When the number is large enough for scientific notation (i.e., more than 6 digits), then the TeX version of scientific notation would not match as a number, and would give an error.

To test the patch, use

```
Context("Matrix")->texStrings;
$P = Point(1,2,3);
$V = Vector(1,2,3);
$M = Matrix([1,2],[3,4]);
$a = Real(1230000);
$b = Real(123000);

sub TEST {
  my $test = shift;
  my ($result, $error) = PG_restricted_eval($test);
  TEXT($error? "$test: $error" : "$test: OK", $BR);
}

TEST('$a*$P');
TEST('$P/$a');
TEST('$a*$V');
TEST('$V/$a');
TEST('$a*$M');
TEST('$M/$a');

TEXT($HR);

TEST('$b*$P');
TEST('$P/$b');
TEST('$b*$V');
TEST('$V/$b');
TEST('$b*$M');
TEST('$M/$b');
```

Without the patch, all the lines involving `$a` should produce an error, but the ones with `$b` should be marked `OK`.  With the patch, all the lines should be `OK`.

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -209,7 +209,7 @@ sub mult {
   #
   #  Constant multiplication
   #
-  if (Value::matchNumber($r) || Value::isComplex($r)) {
+  if (Value::isNumber($r)) {
     my @coords = ();
     foreach my $x (@{$l->data}) {push(@coords,$x*$r)}
     return $self->make(@coords);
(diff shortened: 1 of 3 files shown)


mgage:
Works fine.  Thanks very much Davide.  That removes an annoyance I've been working around.

PR #331, opened 2017-10-26
Fix reduction rule for negative powers

dpvc:
This PR fixes the reduction rule for `x^(-1)` to not reduce matrices, which caused errors, and extends it to `x^(-a)` that handles any negative constant power.

To test, use

```
Context("Matrix");
Context()->constants->add(A=>Matrix([1,2],[3,4]));
Context()->variables->add(M=>Matrix([1,2],[3,4]));
TEXT(Formula("A^(-1)")->reduce,$BR);
TEXT(Formula("M^(-1)")->reduce,$BR);
TEXT(Formula("[[x,2],[3,x]]^(-1)")->reduce,$BR);
TEXT(Formula("x^(-1)")->reduce,$BR);
TEXT(Formula("x^(-2)")->reduce,$BR);
```

Without the patch, the first three reductions will produce an error about only being able to divide by a number.  With the patch, all three should succeed, and you should get

```
A^(-1)
M^(-1)
[[x,2],[3,x]]^(-1)
1/x
1/(x^2)
```

as the output.

Note that this does change the name of the reduction rule from `x^(-1)` to `x^(-a)`.  I've checked the macro files and the OPL and don't see any uses of reduction rule `x^(-1)`, so I think this should not be a problem (though some Wiki pages may need to be updated).

This arose from [this issue](http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4276).

### DIFF
--- a/lib/Parser/BOP/power.pm
+++ b/lib/Parser/BOP/power.pm
@@ -49,8 +49,12 @@ sub _reduce {
     if (($self->{rop}{isZero} && !$self->{lop}{isZero} && $reduce->{'x^0'}) ||
 	($self->{lop}{isOne} && $reduce->{'1^x'}));
   return $self->{lop} if $self->{rop}{isOne} && $reduce->{'x^1'};
-  if ($self->{rop}->isNeg && $self->{rop}->string eq '-1' && $reduce->{'x^(-1)'}) {
-    $self = $self->Item("BOP")->new($equation,'/',$self->Item("Number")->new($equation,1),$self->{lop});
+  if ($self->{rop}->isNeg && $self->{rop}{isConstant} &&
+      $self->{lop}->typeRef->{name} ne "Matrix" && $reduce->{'x^(-a)'}) {
+    my $copy = $self->copy($equation);
+    $copy->{rop} = $self->Item("Number")->new($equation,-($copy->{rop}->eval));
+    $self = $self->Item("BOP")->new($equation,'/',$self->Item("Number")->new($equation,1),
+                                    ($copy->{rop}->string eq '1' ? $copy->{lop} : $copy));
     $self = $self->reduce;
   }
   return $self;
@@ -58,7 +62,7 @@ sub _reduce {
 
 $Parser::reduce->{'x^0'} = 1;
 $Parser::reduce->{'1^x'} = 1;
-$Parser::reduce->{'x^(-1)'} = 1;
+$Parser::reduce->{'x^(-a)'} = 1;
 $Parser::reduce->{'x^1'} = 1;
 
 


Alex-Jordan:
In the example, `A` is a matrix constant. Can `A` be a just a Matrix math object? Or a variable in a Formula for Matrices? Is anything supposed to happen with `Formula("[[1,2],[3,4]]^-1")`?

Also does this just leave things as `A^{-1}`, regardless of whether `A` is invertible (or even square)? And it doesn't try to "reduce" by calculating the inverse?

I think I know the answers to these questions, just confirming before I start testing.

dpvc:
> Can A be a just a Matrix math object?

If you mean an explicit Matrix in a formula, like `Formula("[[1,2],[3,4]]^(-1)")` as in your third question, then not that since the matrix is constant, the inverse will be taken automatically (even without `reduce()`) unless `reduceConstants=>0` is set in the flags (since operations where both operands are constant will be performed automatically).  But it will affect `Formula("[[x,2],[3,x]]^(-1)"`, which would produce an error without the patch, but will remain unchanged with it.

If added this test to the suggested test cases above.

> Or a variable in a Formula for Matrices?

Yes, it should handle that now properly as well.  For example,

```
Context("Matrix")->variables->add(M=>Matrix([[1,2],[3,4]]);
Formula("M^(-1)")->reduce;
```

should now work (it will remain the formula 'M^(-1)') whereas before it would produce an error.  I've added that to the test cases as well.

> Is anything supposed to happen with `Formula("[[1,2],[3,4]]^-1")`

It should produce the inverse matrix regardless of the setting of the reductions (see above).

> Does this just leave things as A^{-1}, regardless of whether A is invertible (or even square)? 

```
Context("Matrix")->constants->add(A=>Matrix([[1,2,3],[3,4,5]]));
Formula("A^(-1)");
```

(and the other similar ones for matrix variables or explicit matrices) will produce an error (though it is an inappropriate message), so you can never get to the reduce step.  I will look into improving the message in a separate PR.

> It doesn't try to "reduce" by calculating the inverse?

For the constant `A`, the inverse is not taken (it remains `A^(-1)`).  This is consistent with other constants not being replaced (e.g., `Formula("pi^-1")->reduce` yields `1/pi` not `0.31831`).  For the variable `M`, again it is just `M^(-1)` since the value of `M` is not known until the formula is evaluated.  For an explicit matrix formula like `Formula("[[x,2],[3,x]]^(-1)")->reduce()`, is also doesn't change (since the invertibility of the matrix isn't known until `x` is known).  But for a constant matrix like `Formula("[[1,2],[3,4]]^(-1)")`, the inverse will be taken immediately (unless `reduceConstants` is set to `0`, and even if `reduceConstants` *is* zero, `reduce()` will override that and produce the inverse).

In the case where the inverse does not exist, there is an error message (but again, a bad one, so I will look into improving that in a separate PR).

mgage:
I'll add a test problem    t/pg_test_problems/matrix_inverse.pg

drdrew42:
I've updated the wiki page for this reduction rule, and edited my own as-yet-unsubmitted problems to reflect the changes. It was quite the surprise when certain previously-manageable problems began rendering in horribly-complicated fashion.

This may affect more problems in the OPL than expected, since previously x^(-a) would not be affected by any of the reduction rules, and now it will be. 

It is not surprising, then, that searching for `noreduce( 'x^(-1)' )` yielded no hits - given that authors expected that x^(-a) was not subject to reduction rules.

PR #333, opened 2017-12-27
Improve error messages for non-invertible matrices

dpvc:
This PR fixes several bad error messages about non-invertible matrices.  To test it, use

```
Context("Matrix");
Context()->constants->add(
  A => Matrix([[1,2,3],[4,5,6]]),
  B => Matrix([0,0],[0,0]),
);
Context()->variables->add(
  M => Matrix([[1,2,3],[4,5,6]]),
  N => Matrix([[1,2],[3,4]]),
);

Formula("A^(-1)");
Formula("M^(-1)");
Formula("[[1,2,3],[4,5,x]]^(-1)");
Formula("[[1,2,3],[4,5,6]]^(-1)");

Formula("B^(-1)")->eval;
Formula("N^(-1)")->eval(N => Matrix([[0,0],[0,0]]));
Formula("[[0,0],[0,0]]^(-1)");
Formula("[[x,0],[0,x]]^(-1)")->eval(x => 0);
Matrix([[0,0],[0,0]]) ** -1;

Matrix([[0,0],[0,0])->inverse;
Matrix([[0,0],[0,0]])->solve_LR(Vector(1,1));
```

Without the patch, the first four should produce errors about 2x3 and 2x3 matrixes not being able to be multiplied, while the next five and final two should produce errors about matrices needing at least one entry.

With the patch, the first four should produce messages that only a square matrix can be inverted, the next five should say that the matrix is not invertible, and the last two should produce an undefined value and a list of three undefined values, respectively.

### DIFF
--- a/lib/Parser/BOP.pm
+++ b/lib/Parser/BOP.pm
@@ -176,6 +176,23 @@ sub checkMatrixSize {
 }
 
 #
+#  Check if a matrix is square
+#
+sub checkMatrixSquare {
+  my $self = shift;
+  my $m = shift; my $type = $m->{entryType};
+  if ($type->{entryType}{name} eq 'Number') {
+    my ($r,$c) = ($m->{length},$type->{length});
+    if ($r == $c) {
+      my $rowType = Value::Type('Matrix',$r,$Value::Type{number},formMatrix=>1);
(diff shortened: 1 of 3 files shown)


mgage:
verified that the error messages are appropriate

PR #334, opened 2017-12-27
Add support for ions and states to the Reaciton context

dpvc:
This PR adds support for ions and states to the Reaciton context, and updates the heavy atoms to their official names.

With this patch, you should be able to use things like

```
Compute("Ag^{+1}(aq) + Cl^{-1}(aq) --> AgCl(s)");
```

See [this forum post](http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4307) for more details, where Eric Stroyan has done some checking of the context.

### DIFF
--- a/macros/contextReaction.pl
+++ b/macros/contextReaction.pl
@@ -23,6 +23,19 @@ reactions.  For example:
 
 	$R = Formula("4P + 5O_2 --> 2P_2O_5");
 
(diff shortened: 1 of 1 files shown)


mgage:
This typesets correctly (as far as I can tell) in both HTML and TeX mode.

PR #336, opened 2017-12-30
Add new test files for units.

mgage:
Add tests to pg/t to test the new units.  

The t/test_units.t file was updated and a file t/pg_test_problems/test_units1.pg was added. 

Update the definition of parsec to include more accuracy -- it still doesn't pass the test_units.t  file.

The definition of steradian is still iffy.

### DIFF
--- a/lib/Parser/BOP.pm
+++ b/lib/Parser/BOP.pm
@@ -176,6 +176,23 @@ sub checkMatrixSize {
 }
 
 #
+#  Check if a matrix is square
+#
+sub checkMatrixSquare {
+  my $self = shift;
+  my $m = shift; my $type = $m->{entryType};
+  if ($type->{entryType}{name} eq 'Number') {
+    my ($r,$c) = ($m->{length},$type->{length});
+    if ($r == $c) {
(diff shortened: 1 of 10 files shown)


mgage:
This adds unit tests which test modifications from pull request #313 and #314

mgage:
This adds files that can be used to test modifications made in PR #333 to #335.
These are individual .pg files that can be run using webwork2/clients/sendXMLRPC.pl or 
which can be placed in a problem set and run in a webwork course.

The file t/pg_test_problems/matrix_inverse2.pg   is meant to test errors -- so this file will fail with an error message.  By choosing which lines to comment out in the problem you can test different error messages, but unfortunately you can only test one at a time.

mgage:
To test:

Run each of these files using sendXMLRPC.pl or placing them in a webwork problem set
 t/pg_test_problems/chem_react1.pg
t/pg_test_problems/chem_react2.pg
t/pg_test_problems/matrix_inverse2.pg (this one produces errors)
t/pg_test_problems/matrix_inverse.pg
 t/pg_test_problems/test_units1.pg

this test can be run from the command line:
perl  t/test_units.t

mgage:
This pull request has been replaced by pull request #337

PR #338, opened 2017-12-30
Develop 328

mgage:
This pull requests is a cleaner version of pull request 328.  Only the pod documents are modified.


mgage:
This PR replaces PR#328 which added some files in addition to changing POD documentation.  This PR should only change POD documentation for matrix files. 

To test -- read through the documentation.  You can request changes via the review mechanism if some portions of the documentation are incorrect or unclear.

lib/Matrix.pm
```diff
@@ -68,6 +74,17 @@ sub _stringify {
...
+=head3 Accessor functions
+	
+	(these are deprecated for direct use.  Use the covering Methods
+	 provided by MathObject Matrices instead.)
+	 
+	L($matrix) - return matrix L of the LR decomposition
+	R($matrix) - return matrix R of the LR decomposition
+	PL($matrix) return 
+	PR($matrix 
+
```
dpvc: Is there something missing in the above two lines?

lib/Value/Matrix.pm
```diff
@@ -3,6 +3,109 @@
...
+	inverse 
+	is_symmetric
+	decompose_LR 
+	dim 
+	norm_one 
+	norm_max 
+	kleene 
+	normalize 
+	solve_LR (also solve())
+	order_LR (also order()
```
dpvc: Missing a close paren.  :-)

macros/MatrixReduce.pl
```diff
@@ -14,89 +14,105 @@ =head1 SYNOPSIS
...
+After selecting the Fraction context using Context('Fraction')->parens->set("[" => {formMatrix => 1}), C<$A = apply_fraction_to_matrix_entries($A);> applies Fraction() to all of the entries of $A, which makes subsequent matrix algebra computations with $A use fraction arithmetic.
 
 =item Get the reduced column echelon form: C<$Areduced = rcef($A);>
 
-=item Change the value of a matrix entry: C<change_matrix_entry($A,[2,3],50);> changes the [2,3] entry to the value 50.
+=item Change the value of a matrix entry: C<change_matrix_entry($A,[2,3],50);> 
+
+changes the [2,3] entry to the value 50.
 
 =item Construct an n x n identity matrix: C<$E = identity_matrix(5);>
```
dpvc: Note:  there is already a MatahObject method for that:  `Value::Matrix->I(5)`.

macros/PGmatrixmacros.pl
```diff
@@ -775,11 +720,42 @@ =head4 create2d_matrix
...
+
+$mathobject_matrix->value outputs an array (usually an array of array references) so placing it inside
+square bracket produces and array reference (of array references) which is what lp_display_mm() is
+seeking.
+
+=cut
+
+sub convert_to_array_ref {
+	my $input = shift;
+	if (ref($input) eq 'Value::Matrix' ) {
```
dpvc: This won't catch subclasses of `Value::Matrix`, so perhaps `if (Value::classMatch($input,"Matrix") {` would be better (this will return false if `$input` is not a MathObject at all, so should be safe).

dpvc:
I made a few comments.  Looks good otherwise.  I haven't actually run anything, though.  If you need me to do that level of testing, let me know, and I'll see if I can work on it next week.

lib/Matrix.pm
mgage: Good catch.  I'd also missed a =cut which would have wiped out the definition of the L and R 
decomposition functions. Here is the whole segment in context:
-----------------
=head3 Accessor functions
	
	(these are deprecated for direct use.  Use the covering Methods
	 provided by MathObject Matrices instead.)
	 
	L($matrix) - return matrix L of the LR decomposition
	R($matrix) - return matrix R of the LR decomposition
	PL($matrix) - return permutation matrix
	PR($matrix) - return permutation matrix
	Original matrix is  PL * L * R *PR = M 
	
Obtain the Left Right matrices of the decomposition 
and the two pivot permutation matrices
the original is M = PL*L*R*PR

=cut

lib/Value/Matrix.pm
mgage: OK.  I fixed these and added some clarifying remarks.  For a more complete description one still needs to consult the MatrixReal1.pm documentation.

macros/MatrixReduce.pl
mgage: True, but I believe it was added after MatrixReduce.pl was created. I've added a deprecated statement to this for now.  Eventually I'd like to consolidate functions in MatrixReduce, and in Tableau and in Peter @Selinger 's linear algebra macros. Some of them should go into Value::Matrix -- the others should be gathered into a small number of macro files.

macros/PGmatrixmacros.pl
mgage: Accepted change.  Thanks.

mgage:
I don't think you need to do extensive testing -- other than making sure it compiles. I'm doing this to try to catalog all the matrix related commands that currently exist -- writing up at least partial documentation along the way.  The next step will be to consolidate and refactor the overlapping commands so that they are easier to maintain and to use -- the step will require additional testing. Meantime I think the upgraded documentation can be pulled in to develop so I don't forget what has been done so far. -:)

selinger:
Michael,

I haven't yet consolidated my linear algebra macros (and problem
collection) because the course is still ongoing. When the course is
finished, I am hoping to organize the macros in a more coherent way
and to add the problems to the library. At that time I'll ask you
about the best way of doing these things.

-- Peter

Michael Gage wrote:

mgage:
Cleaned up the conflict with MatrixReduce.pl

PR #343, opened 2018-02-06
correct typo in macros/contextLimitedNumeric.pl

heiderich:


dpvc:
Thanks for spotting that.

PR #344, opened 2018-02-11
Matrix bugs

selinger:
This is my first pull request; I am starting with something uncontroversial just to see how the process works. It fixes some real bugs though.

Summary: this fixes a bug in the functions rref, elem_matrix_row_mult, row_mult, apply_fraction_to_matrix_entries, and possibly others, in MatrixReduce.pl and similar places. These functions do not work correctly in case of a 1xm-matrix. The bug is caused by calling Matrix(@m) instead of Matrix(\\@m). While the Matrix constructor can either accept an array reference or an array, the behavior is not correct when the matrix happens to have a single row.

In other words, while Matrix([[1,2,3],[4,5,6]]) and Matrix([1,2,3],[4,5,6]) are equivalent, Matrix([[1,2,3]]) and  Matrix([1,2,3]) are not equivalent (and the latter gives the wrong result, namely, a vector instead of a matrix).

Therefore, one should never call Matrix(@m) programmatically when the length of @m could be 1.
One should always call Matrix(\\@m). Even in cases where the length of @m is known (e.g., the function GL2Z), the form Matrix(\\@m) is preferable, because someone might repurpose the code for a different @m.

This patch fixes the issue.

### DIFF
--- a/macros/MatrixCheckers.pl
+++ b/macros/MatrixCheckers.pl
@@ -144,7 +144,7 @@ sub concatenate_columns_into_matrix {
   for my $column (@c) {
     push(@temp,Matrix($column)->transpose->row(1));    
   }
-  return Matrix(@temp)->transpose;
+  return Matrix(\@temp)->transpose;
 
 }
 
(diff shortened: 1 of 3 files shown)


selinger:
P.S.: I tested all of the modified functions.

paultpearson:
Good catch.  I see no reason not to merge.

dsteinmo:
The bug is well-explained in the initial comment and it is clear after reading it why the pass-by-reference is necessary. I am also happy with this.

dpvc:
Looks OK to me.

mgage:
Thanks very much everybody.  It helps a lot to have several people who can help review before the merge.

PR #347, opened 2018-03-28
Bugfix: macros should not set global variables $n, $i, $j, $s

selinger:
Several macros in macros/MatrixReduce.pg overwrite global variables such as $n, $i, $j, $s. This is obviously a bad idea as these variables might be used for different purposes by problem authors.


mgage:
This should have been pulled earlier.  Sorry.

PR #348, opened 2018-04-02
Ln e reduction

dpvc:
This introduces a reduction rule for `ln(e)`, `log(e)` (when `useBaseTonLog` is false).

You can test this by using

    TEXT(Formula("ln(e)")->reduce,$BR);
    TEXT(Formula("log(e)")->reduce,$BR);

Without the patch, the formula will be unchanged.  With the patch, you should get the number 1 from both of them.

### DIFF
--- a/lib/Parser/Function/numeric.pm
+++ b/lib/Parser/Function/numeric.pm
@@ -52,6 +52,16 @@ sub log   {
   CORE::log($_[0]);
 }
 
+sub _reduce {
+  my $self = shift;
+  return $self unless $self->context->flag('reduceConstantFunctions');
+  my $base10 = $self->context->flag('useBaseTenLog');
+  my $s = $self->string;
+  return $self->Item('Value')->new($self->{equation}, [$self->eval])
+    if $s eq 'ln(e)' || $s eq 'log10(10)' || $s eq ($base10 ? 'log(10)' : 'log(e)');
+  return $self;
+}
+
 #
 #  Handle absolute values as a special case
 #
--- a/lib/Parser/Function/numeric.pm
+++ b/lib/Parser/Function/numeric.pm
@@ -58,7 +58,7 @@ sub _reduce {
   my $base10 = $self->context->flag('useBaseTenLog');
   my $s = $self->string;
   return $self->Item('Value')->new($self->{equation}, [$self->eval])
-    if $s eq 'ln(e)' || $s eq 'log10(10)' || $s eq ($base10 ? 'log(10)' : 'log(e)');
+    if $s eq 'ln(e)' || ($s eq 'log(e)' && $base10);
   return $self;
 }
 


Alex-Jordan:
Some questions, really just out of curiosity. Please don't read them as suggestions.

Does `reduceConstantFunctions=>1` not apply to `ln(e)` because the `e` is a math objects constant?

Is there a situation where you would want `reduceConstantFunctions=>0`, but you would also hope to see things like `log(10)` reduce? (When log is base 10.)

Is this where you'd like to draw the line with reduction rules? As in, not try to have one for `ln(e^(expression))`?

dpvc:
> Does `reduceConstantFunctions=>1` not apply to `ln(e)` because the `e` is a math objects constant?

Right.  This makes it apply to it, just like `log10(10)` already does.

> Is there a situation where you would want `reduceConstantFunctions=>0`, but you would also hope to see things like `log(10)` reduce?

Maybe, but there is no provision for that, just as there is no provision to get `2*3` to reduce when you have `reduceConstants=>0`.  That is essentially what these flags control.

> Is this where you'd like to draw the line with reduction rules? As in, not try to have one for `ln(e^(expression))`?

I guess that would not be too hard.  I would also want to handle `ln(exp(...))` in that case.  I will look into it...

Alex-Jordan:
I think `sin(asin(expression))`, `asin(sin(expression))`, and `exp(ln(expression))` all raise doinain issues and shouldn't be simplified to `expression` unless there is a really good check that `expression` outputs into an acceptable range.  It just happens that with, `ln(exp(expression))` these issues don't matter since the domain of `exp` is all of R, and `ln` and `exp` are true inverses unlike `sin` and `asin`.

For myself, I don't have any feelings about the `x^2` ideas. 

Thanks for the `ln(exp(expression))` reduction. I think I've seen places come up in other output from the `D` method where this will be nice.

There is already the `noreduce` mecahnism, is that what you mean? I typically load them all (whatever the context uses) and then use `Context()->noreduce(...)` to undo a few here and there.

dpvc:
> I think `sin(asin(expression))`, `asin(sin(expression))`, and `exp(ln(expression))` all raise doinain issues

Good point.  I wasn't planning to do `asin(sin(...))` since `asin(sin(x))` need not be `x`, whereas `sin(asin(x))` does equal `x` everywhere it is defined (but I neglected to consider that last phrase).  So good, I don't have to take time to implement that.  :-)

> There is already the `noreduce` mecahnism, is that what you mean?

No, I meant have something like `reduceUsing()` that you pass a list of reduction rules to apply.  None of the others get applied, only the ones you explicitly request.  (That is, this is a whitelist rather than a blacklist).  That way you could get specific reductions that you need without causing unexpected results.  But perhaps it is not needed.

heiderich:
> I suppose this means I should do similar things for sin(asin(x)), and so on. How about sqrt(x^2) going to |x|? Should |x^2| go to x^2? I think it is probably best not to get overzealous about it.

What does pg know about the arguments? None of the following equalities

`ln(e(x)) = x`
`sqrt(x^2) = |x|`
 `|x^2| = x^2`

hold for every complex number x. If nothing is known about the argument, I would opt against these reductions.

> It just happens that with, ln(exp(expression)) these issues don't matter since the domain of exp is all of R, and ln and exp are true inverses unlike sin and asin.

It is true that `ln` is an inverse of `exp` when you consider `exp` as a function from the real numbers to the positive real numbers. This is however false when you consider `exp` as a function on the complex numbers.

dpvc:
> None of the ... equalities ... hold for every complex number x.

Good point.  Perhaps the solution is to remove the `ln(e^x)` reduction rule from the complex contexts.  (Each context has its own set of rules, so they can be different from context to context.)

mgage:
This works fine for real values.  I will leave the modifications for complex values for a future PR.

PR #351, opened 2018-04-25
add maketext calls in macros/problemRandomize.pl

heiderich:


mgage:
Tested it on one of MC prealgebra questions.

PR #353, opened 2018-05-03
implement PGML explicit disc

Alex-Jordan:
In PGML, currently both `*` and `-` effectively create a filled-in-disc as a bullet in a list. But neither options leads to that getting written as an attribute to the `ul` (in the same way that `type="circle"` and `type="square"` get written as attributes). This means that the CSS that controls bullet styling can override the author's choice to make a disc, which will generally happen if the list is a sublist. 

With the change, `*` makes an explicit disc bullet, and `-` remains generic.


dpvc:
I think these are the changes we discussed some time ago, and they look fine.  I approve.

PR #354, opened 2018-05-03
Add protecting <div> around the geogebra object.  

mgage:
This is a hot fix.

The geogebra javaScript modifies the style of the second enveloping &lt;div> containing the geogebra object.  Specifically it sets the height and width.  If the &lt;div> contains other material besides the geogebra object these values for the height and width will be incorrect causing overlapping outputs.

This fixes the display problems discussed in the forum at 
http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4401


mgage:
This a small fix, but it should compensate for apparent changes in the geogebra site code for geogebra HTML5 applets

PR #355, opened 2018-05-03
Geogebra hotfix

mgage:
This is a hot fix.

The geogebra javaScript modifies the style of the second enveloping &lt;div> containing the geogebra object. Specifically it sets the height and width. If the &lt;div> contains other material besides the geogebra object these values for the height and width will be incorrect causing overlapping outputs.

This fixes the display problems discussed in the forum at
http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4401


mgage:
This is a companion to the hot fix #354 issued for master. (I think that's why there are some additional commits mentioned besides the geogebra fix.)

PR #356, opened 2018-05-10
Add lang and direction to pg flags

taniwallach:
This is a pull request which is intended to allow WebWork to receive information about the (primary) language and text direction of a problem via the PG flags, which can be accessed by the ContentGenerator code to set HTML language and direction attributes for each specific problem.

The primary need for this is likely to be in courses in right-to-left (RTL)  languages such as Hebrew and Arabic where the main HTML direction will be RTL, but when viewing problems from the library browser to be selected for translation - those problems should be displayed in the LTR (left-to-right) direction. The opposite case of viewing RTL problems in a LTR course is likely to be pretty rare.

The capability of setting the language of a problem, would allow problems in multiple languages to be used in a single course, while setting the language of the block containing a problem to match that of the problem rather than the course-wide default. This is something desirable in terms of accessibility for screen-reader users, even if a typical user would not be aware of there being an issue.

Details: the functions SET_PROBLEM_LANGUAGE() and SET_PROBLEM_TEXTDIRECTION() were added to macros/PG.pl which set the flags expected by the code being written to process the data they pass on.

The flag values set in the problem are processed by a new utility function get_problem_lang_and_dir() which passed on an array of data used to set the relevant HTML tags in the various ContentGenerator modules. That code also checks a new configuration variable added to webwork2/conf/defaults.config which can force specific fixed behavior, set defaults to use when there are no PG set flags, or disable to addition of the attributes which would normally be done by the new code.

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -189,7 +189,7 @@ sub SET_PROBLEM_TEXTDIRECTION {
   if ( $requested_dir =~ /^ltr$/i ) {
     $PG->{flags}->{"textdirection"} = "ltr";
   } elsif ( $requested_dir =~ /^rtl$/i ) {
-    $PG->{flags}->{"textdirection"} = "trl";
+    $PG->{flags}->{"textdirection"} = "rtl";
   } elsif ( $requested_dir =~ /^auto$/i ) {
     $PG->{flags}->{"textdirection"} = "auto"; # NOT RECOMMENDED
   } else {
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
@mgage is the plan to do all the character support stuff for 2.14 and then language support stuff like this one in 2.15? Or is some of this meant to go into 2.14 as well?

PR #360, opened 2018-06-12
fix typos in PTX_cleanup in PGbasicmacros.pl

mgage:
This fixes some typos in PTX_cleanup

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2207,11 +2207,11 @@ sub PTX_cleanup {
     #insert opening and closing p, to be removed later if they enclose an image, video or tabular
     $string =~ s/(<sidebyside[^>]*(?<!\/)>)/$1\n<p>/g;
     $string =~ s/(<\/sidebyside>)/<\/p>\n$1/g;
-    $string =~ s/(<sidebyside[^>]*(?<=\/)>)/</p>\n$1\n<p>/g;
+    $string =~ s/(<sidebyside[^>]*(?<=\/)>)/<\/p>\n$1\n<p>/g;
     #ditto for li
     $string =~ s/(<li[^>]*(?<!\/)>)/$1\n<p>/g;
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I'm taking a look at this. The lines at 2210 and 2214 look like they do not belong in the first place. I suspect that I had done lines 2218--2220, and then I copied up to these earlier places to modify. In 2218--2220, the intent is to alter *outside* of a "sidebyside" element. In the blocks at lines 2210 and 2214, then intent is to alter *inside*. But those specific lines are targeting self-closing tags, so it just makes no sense to do anything since there is no inside. That would also explain whey this has been in place and there have been no errors. Because the absence of the slash that you would be fixing here should have show up by now.

Could you adjust your branch with a commit to delete those two lines instead?

At line 2251, I could swear I have seen the effect of that code block perform as intended. Is it possible the backslash numbers work just as well as dollar sign numbers? Anyway, the cahnge should be fine. I will test soon. Except I sort of have a bigger fish to fry with this particular code block in the first place. I may just remove it. The issue is a PTX issue that I won't try to fully explain here. But this code block attempts to merge adjacent images into a single sidebyside, and that is going to produce illegal results if the images' percentage widths add up to more than 100%.

mgage:
The backslash numbers \1, etc. work as well as $1, . etc., but they generate several zillion

\1 better written as $1 at (eval 9251) line 2253, <MACROFILE> chunk 1.

messages which pollute the error log file.

mgage:
I added a small change to PGmatrixmacros.pl to fix a typo and type check error there. 

I'm going to pull this at this point.  Alex you can rewrite the PTX_cleanup macro when you are ready in a separate pull request

Alex-Jordan:
Mike did you see that I was able to apply a commit to your branch? This pull request already has the edits to PTX_cleanup that I would apply at this time. Unless there is something else that I missed that you think I should do.

If you'd just like to separate the issues into two distinct pull requests for management reasons, I could do that.

PR #366, opened 2018-07-22
Fix crash when preformatted text is used in a list

dpvc:
This PR fixes a problem reported in [this post](http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4427#p12826).

To test, use

```
loadMacros("PGML.pl");

BEGIN_PGML
1.  Item 1
    :   pre-formatted

2. Item 2
END_PGML
```

Without the patch, this will produce an error about not being able to use `combineTopItems` on an undefined reference.  With the patch, this should format as a list with a preformatted section within it.

### DIFF
--- a/macros/PGML.pl
+++ b/macros/PGML.pl
@@ -276,7 +276,7 @@ sub ForceBreak {
 
 sub Par {
   my $self = shift; my $token = shift;
-  $self->End;
+  $self->End(null, shift);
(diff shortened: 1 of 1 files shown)


mgage:
line this up as a hot-fix to 2.14

mgage:
This works.  Merged into 2.14

PR #373, opened 2018-10-27
Develop candidate

mgage:
This includes develop_international branch which was on mgage


mgage:
This is being replaced by #386

PR #374, opened 2018-10-31
Make sure ImplicitPlane retains its full precision

dpvc:
Make sure ImplicitPlane retains its full precision (i.e., don't go through Real to string conversion).

An implicit plane created from a coefficient array and constant is created by producing a string version of the plane that is then parsed.  If the array was of MathObject Reals, when stringified, they would reduce to 6 digits of precision, so the resulting plane might not be exactly the one intended.  Since this was used to check if the student answer is in fact a linear one, that could lead to an incorrect message that the students answer was non linear (due to a comparison failing because of the loss of precision).  See [this forum post](http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4515) for an example.

To test, you can use

```
loadMacros("parserImplicitPlane.pl");
Context("ImplicitPlane")->variables->are(x=>"Real", y=>"Real");
$f = ImplicitPlane([1/6,1],1/3);
TEXT(Formula($f->{tree}{lop})->eval(x=>1, y=>0)->value, $BR);
TEXT(Formula($f->{tree}{rop})->eval(x=>1, y=>0)->value, $BR);
```

With the patch, you should get 16 or 17 digits of output.  Without the patch, you should get only 6.

### DIFF
--- a/macros/parserImplicitPlane.pl
+++ b/macros/parserImplicitPlane.pl
@@ -116,8 +116,8 @@ sub new {
     $vars = [$vars] unless ref($vars) eq 'ARRAY';
     $type = 'line' if scalar(@{$vars}) == 2;
     my @terms = (); my $i = 0;
-    foreach my $x (@{$vars}) {push @terms, $N->{data}[$i++]->string.$x}
-    $plane = $formula->new(join(' + ',@terms).' = '.$d->string)->reduce(@_);
+    foreach my $x (@{$vars}) {push @terms, $N->{data}[$i++]->value.$x}
+    $plane = $formula->new(join(' + ',@terms).' = '.$d->value)->reduce(@_);
   } else {
     #
     #  Determine the normal vector and d value from the equation


dpvc:
PS, you can redirect to the 2.14 branch, if it is not too late for that.

mgage:
I’ll do that.   This seems safe enough.

On Oct 31, 2018, at 9:33 AM, Davide P. Cervone <notifications@github.com<mailto:notifications@github.com>> wrote:


PS, you can redirect to the 2.14 branch, if it is not too late for that.

dpvc:
I've made a couple of other PRs as well that you might want to consider for 2.14 as well.  Nothing off much consequence, which is why I haven't been pushing them earlier.

mgage:
Works fine.  sorry not to get it into 2.14.   by the way develop + new_develop_candidate_01_01_2019 is actually pretty stable (certainly stable enough for use by developers and small classes) if you are using English or latin languages and characters.  It probably won't behave consistently yet with chinese or arabic characters since we haven't switched the mysql database over to 4 byte characters.  

The biggest glitch is that the copy right symbol that was used in the course.conf model is not utf8 compatible.  It appears everywhere and sets up warning flags.  There is a script
webwork2/bin/fix_copyright.sh that will change those copyright symbols to &copy;. 
 I've also
fixed the copyright symbol in the other files in webwork software.

PR #377, opened 2018-10-31
Don't try to convert MathObjects to complex numbers in Matrix->copy()

dpvc:
The (non-MathObject) Matrix object has a function to try to copy its entries as complex ones if they are complex, but it doesn't really check if they are complex, only that they are objects (assuming that that means they are (non-MathObject) Complex objects.  Since MathObject Matrix object use the older Matrix (and MatrixReal) objects to perform some computations (like inverse), if the MathObject Matrix has entries that are Fraction objects, for example, then the Matrix copy function will assume these are complex numbers.  This fix should prevent that.

To test, use

```
Context("Fraction");

BEGIN_TEXT
\(\{Matrix([Fraction(1,2),Fraction(2,1)],[Fraction(2,3), Fraction(3,2)])->inverse)\}\)
END_TEXT
```

With the patch, you should get a matrix with fractional entries.  Without the patch, you should get an error like 

```
Can't locate object method "Re" via package "context::Fraction::Fraction" at lib/Matrix.pm line 381
```

### DIFF
--- a/lib/Matrix.pm
+++ b/lib/Matrix.pm
@@ -377,9 +377,8 @@ sub new_from_col_vecs
 
 sub cp  { # MEG  makes new copies of complex number
 	my $z = shift;
-	return $z unless ref($z);
-	my $w = Complex1::cplx($z->Re,$z->Im);
-	return $w;
+	return $z unless ref($z) eq 'Complex1';
+	Complex1::cplx($z->Re,$z->Im);
 }
 
 =head4
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -186,7 +186,7 @@ sub numberMatrix {  #internal
   my @M = (); my $isFormula = 0;
   foreach my $x (@_) {
     $x = Value::makeValue($x,context=>$context);
-    Value::Error("Matrix row entries must be numbers") unless Value::isNumber($x);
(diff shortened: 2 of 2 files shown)


dpvc:
The issue has something to do with `Context()->texStrings` being set, as this works just fine:

```
Context("Fraction");

BEGIN_TEXT
\(\{Matrix([Fraction(1,2),5],[6,9])->inverse->TeX\}\)
END_TEXT
```

Since the inverse is being handled by RealMatrix, I'm wondering if it is stringifying things somewhere internally, which would cause the Fraction to produce its TeX version rather than its string version (when `texStrings` is set).

I'll have to look further.

dpvc:
Well, it was a MathObject problem after all.  The test for whether a matrix entry is a number didn't take fractions into account properly (and the same with several other locations where there are tests for numbers).  The `Value::isNumber` and related checks don't take fractions into account, but changing them causes other problems, since some uses require the result NOT to be a fraction.  it might be that only the uses in the Fraction context are the bad ones, but I'm afraid to make the change to `Value::isNumber` without a more thorough investigation.  So this is a fix for Matrix objects only for now.

mgage:
works for now. Let's try to remember to revisit Value::isNumber when there is time.

PR #380, opened 2018-12-03
Don't inherit non-inheritable properties when doing a union reduction.

dpvc:
This PR makes sure that the intervals used in a union that is being reduced don't inherit any of the non-inheritable properties.

To test, use

```
Context("Interval");

$I = Compute("(0,2]") + Compute("(-1,1)");
TEXT($I->{correct_ans_latex_string} ? "Inherited" : "Not Inherited");
```

Without the patch, you should see `Inherited`; with the patch you should see `Not Inherited`.

This comes originally from [this forum post](http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4516).


mgage:
This one works.

dpvc:
You seem to have closed this without merging.  Was that the intent?

mgage:
No.  That was a mistake.

mgage:
ok -- I've checked it again.  It works.

PR #381, opened 2018-12-20
Have PGML change tabs to four spaces at end of formatting

Alex-Jordan:
Prior to this change, PGML like this:

````
BEGIN_PGML
```
    Some code preceded
by four spaces
```
END_PGML
````

will internally replace the four spaces with a tab (recognizing an indent), and then the output (HTML, tex, PTX) will have a leading tab character. That might come out looking like 8 spaces in HTML, but in any case having a tab, so not faithful to the code that was input.

As a final step in PGML processing, this edit converts tab characters back to four spaces.

If I understand correctly, PGML starts out by converting all tabs to four spaces, then leading-four-spaces to a tab. So I don't think this would have unintended consequences because PGML is already controlling tabs. I've done some testing, but not all combinations of multiple depth indentations, lists and sublists, etc.

### DIFF
--- a/macros/PGML.pl
+++ b/macros/PGML.pl
@@ -1484,6 +1484,7 @@ sub Format {
   } else {
     $format = '<div class="PGML">'."\n".PGML::Format::html->new($parser)->format.'</div>'."\n";
   }
+  $format =~ s/\t/    /g;
   main::WARN_MESSAGE("==================","Errors parsing PGML:",@warnings,"==================") if scalar(@warnings);
   return $format;
 }

### CODE REVIEW

**Comment by dpvc** [comment]
I don't get the result you seem to be getting.  There are no tabs in the output for your example when I use it in a problem myself.  For example, try:

    loadMacros("PGML.pl");
    
    $x = PGML::Format(<<'END_PGML');
    ```
        Some code preceded
    by four spaces
    ```
    END_PGML
    
    TEXT($x =~ m/~~t/ ? 'tabs' : 'no tabs');

For me, I get the output "no tabs".  Do you get "tabs"?

The translation back from tabs is being done in the `PGML::Text` package, in the `pushText()` function, so you should not get tabs in the end.

**Comment by Alex-Jordan** [comment]
Experimentation shows the tab being there if there is a preceding line. Try:

````
BEGIN_PGML
A preceding line
```
    indent
```

END_PGML
````

**Comment by Alex-Jordan** [comment]
A separate question is, do we want to preserve tabs that the author puts in code? If tabs are just not going to be supported in code, I will mention that in the documentation (which I was working on when I saw this issue).

**Comment by dpvc** [comment]
OK, I get the tabs now, thanks.

If you want to keep tabs in the code blocks, then I think the real solution is to fix the handling of tabs internally, not post-process the output.  I have never liked the idea of doing regular expression substitutions on the results, as they are non-semantic and prone to bathing things that they shouldn't.  Since PGML already has an internal representation of what the structure is, it seems better to have it produce the correct output initially than to try to munge the results later.

Here is a set of diffs for PGML.pl that preserve tabs in code (and elsewhere), rather than converting them to 4 spaces.

```
diff --git a/macros/PGML.pl b/macros/PGML.pl
index e7971e1..8556f07 100644
--- a/macros/PGML.pl
+++ b/macros/PGML.pl
@@ -32,7 +32,7 @@ package PGML::Parse;
 
 my $wordStart = qr/[^a-z0-9]/;
 
-my $indent = '^\t+';
+my $indent = '^(?:\t|    )+';
 my $lineend = '\n+';
 my $linebreak = '   ?(?=\n)';
 my $heading = '#+';
@@ -67,8 +67,6 @@ sub new {
 
 sub Split {
   my $self = shift; my $string = shift;
-  $string =~ s/\t/    /g;                             # turn tabs into spaces
-  $string =~ s!^((?:    )+)!"\t"x(length($1)/4)!gme;  # make initial indent into tabs
   $string =~ s!^(?:\t* +|\t+ *)$!!gm;                 # make blank lines blank
   return split($splitPattern,$string);
 }
@@ -153,7 +151,7 @@ sub All {
   my $self = shift; my $token = shift;
   return $self->Begin($token) if substr($token,0,1) eq "[" && $BlockDefs{$token};
   for ($token) {
-    /\t/           && do {return $self->Indent($token)};
+    /^(?:\t|    )/ && do {return $self->Indent($token)};
     /\d+\. /       && do {return $self->Bullet($token,"numeric")};
     /[ivxl]+[.)] / && do {return $self->Bullet($token,"roman")};
     /[a-z][.)] /   && do {return $self->Bullet($token,"alpha")};
@@ -285,7 +283,9 @@ sub Par {
 sub Indent {
   my $self = shift; my $token = shift;
   if ($self->{atLineStart}) {
-    my $indent = $self->{actualIndent} = length($token);
+    my $tabs = $token;
+    $tabs =~ s/    /\t/g;  # turn spaces into tabs
+    my $indent = $self->{actualIndent} = length($tabs);
     if ($indent != $self->{indent}) {
       $self->End("indentation change");
       $self->{indent} = $indent;
@@ -855,7 +855,6 @@ sub pushText {
   my $self = shift;
   foreach my $text (@_) {
     if ($text ne "") {
-      $text =~ s/\t/    /g;
       push(@{$self->{stack}},$text);
     }
   }
```

Personally, I don't like having hard tabs in code (or elsewhere), which is one reason for the original conversion of tabs to spaces.

I think the changes to the way the initial tabs and spaces are handled in the diff above are good.  I'm not sure about losing the tab to space conversion in the rest of the text.  If we want that, then we have to decide if code blocks are an exception.  The answer to that will determine where the conversion of tabs to spaces needs to be done.

Alex-Jordan:
Thanks Davide. I tested that it no longer introduces the tab from the original issue, and also that it is preserving tabs in the author's source.

Inside a code block, I would vote to preserve author tabs just because who knows if they may be intentional and matter some time. If they are there, it seems like they would be very intentional, because simply hitting tab while in the WeBWorK problem editor navigates you outside of the editor. So they would have to pasted in if editing a problem that way. Of course if someone edits problems using some other editor, tabs come in. The OPL has over 5000 .pg files with tabs in them. I looked at a few and they were all tabs in the perl setup portion. Not sure how to multiline grep to see where there are tabs in actual PGML.

Outside of a code block, I might think it nicer to do as things were before, with all tabs being scrubbed away. But is it hard to do that while handling code differently? And if it is hard, is it really worth the effort?

I'll close this pull request, but we can keep discussing. 

(Tangentially related: What I'd really like is for all the tabs in webwork2 and pg to get replaced and code to become aligned sensibly.)

PR #383, opened 2018-12-20
Add code to to track balance of begin/end variable construct in macro in PGbasicmacros.pl and issue warnings on errors / final imbalance.

taniwallach:
Add code to to track balance of begin/end variable construct in macro in PGbasicmacros.pl and issue warnings on errors / final imbalance.
 * lib/PGcore.pm
   * Add a new flag warn_on_internalBalancing_errors and currently
     default to 1 (warnings on errors).
 * lib/WeBWorK/PG/Translator.pm
   * Put in calls to do the internalBalancing calls in the preprocess phase
     for the relevant variables. Must be done in the preprocess phase
     as the function which sets the value of these variables is evalualed
     once to define the variable, but no "code" is run when the variables
     are used in a PG file.
 * macros/PG.pl
   * Add the %internalBalancing hash to track the counters/status,
     initialize the local value of $warn_on_internalBalancing_errors
     for the problem, and initialize the value of the counters and
     status variables inside the hash.
   * Add the functions which adjust the counters/status variables, and issue
     warnings on errors and a debug message on valid changes, depending on the
     value of $warn_on_internalBalancing_errors.
   * In ENDDOCUMENT check on the counters/status variables, and warn
     if there is an imbalance error (if $warn_on_internalBalancing_errors > 0).
 * macros/PGbasicmacros.pl
   * Added a comment about this feature to the file, as the relevant constants
     are defined here.
   * Note: Other macros files may be adding HTML or LaTeX code which should be
     properly balanced. Hopefully they balance the tags themselves and do not use
     a begin/end construct which can lead to an imbalance.

### DIFF
--- a/lib/WeBWorK/PG/Translator.pm
+++ b/lib/WeBWorK/PG/Translator.pm
@@ -1851,6 +1851,35 @@ sub PG_answer_eval {
 # }
 sub default_preprocess_code {
 	my $evalString = shift//'';
+
+	# Put in calls to do the internalBalancing calls in the preprocess phase
+	$evalString =~ s/\$BBOLD\W/\$BBOLD \\{ internalBalancingIncrement("openBold"); \\}/g;
+	$evalString =~ s/\$EBOLD\W/\$EBOLD \\{ internalBalancingDecrement("openBold"); \\}/g;
+
+	$evalString =~ s/\$BITALIC\W/\$BITALIC \\{ internalBalancingIncrement("openItalic"); \\}/g;
+	$evalString =~ s/\$EITALIC\W/\$EITALIC \\{ internalBalancingDecrement("openItalic"); \\}/g;
+
+	$evalString =~ s/\$BUL\W/\$BUL \\{ internalBalancingIncrement("openUnderline"); \\}/g;
+	$evalString =~ s/\$EUL\W/\$EUL \\{ internalBalancingDecrement("openUnderline"); \\}/g;
+
+	$evalString =~ s/\$BCENTER\W/\$BCENTER \\{ internalBalancingIncrement("openCenter"); \\}/g;
+	$evalString =~ s/\$ECENTER\W/\$ECENTER \\{ internalBalancingDecrement("openCenter"); \\}/g;
+
+	$evalString =~ s/\$BLTR\W/\$BLTR \\{ internalBalancingIncrement("openSpan"); \\}/g;
+	$evalString =~ s/\$ELTR\W/\$ELTR \\{ internalBalancingDecrement("openSpan"); \\}/g;
+
+	$evalString =~ s/\$BM\W/\$BM \\{ internalBalancingTurnOn("inInlineMath"); \\}/g;
+	$evalString =~ s/\$EM\W/\$EM \\{ internalBalancingTurnOff("inInlineMath"); \\}/g;
+
+	$evalString =~ s/\$BDM\W/\$BDM \\{ internalBalancingTurnOn("inDisplayMath"); \\}/g;
+	$evalString =~ s/\$EDM\W/\$EDM \\{ internalBalancingTurnOff("inDisplayMath"); \\}/g;
+
+	$evalString =~ s/\$BEGIN_ONE_COLUMN\W/\$BEGIN_ONE_COLUMN \\{ internalBalancingTurnOn("inOneColumnMode"); \\}/g;
+	$evalString =~ s/\$END_ONE_COLUMN\W/\$END_ONE_COLUMN \\{ internalBalancingTurnOff("inOneColumnMode"); \\}/g;
+
+	$evalString =~ s/\$BLABEL\W/\$BLABEL \\{ internalBalancingTurnOn("inInputLabel"); \\}/g;
+	$evalString =~ s/\$ELABEL\W/\$ELABEL \\{ internalBalancingTurnOff("inInputLabel"); \\}/g;
+
 	# BEGIN_TEXT and END_TEXT must occur on a line by themselves.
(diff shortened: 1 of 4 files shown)


lib/WeBWorK/PG/Translator.pm
```diff
@@ -1851,6 +1851,35 @@ sub PG_answer_eval {
...
+	$evalString =~ s/\$BITALIC\W/\$BITALIC \\{ internalBalancingIncrement("openItalic"); \\}/g;
+	$evalString =~ s/\$EITALIC\W/\$EITALIC \\{ internalBalancingDecrement("openItalic"); \\}/g;
+
+	$evalString =~ s/\$BUL\W/\$BUL \\{ internalBalancingIncrement("openUnderline"); \\}/g;
+	$evalString =~ s/\$EUL\W/\$EUL \\{ internalBalancingDecrement("openUnderline"); \\}/g;
+
+	$evalString =~ s/\$BCENTER\W/\$BCENTER \\{ internalBalancingIncrement("openCenter"); \\}/g;
+	$evalString =~ s/\$ECENTER\W/\$ECENTER \\{ internalBalancingDecrement("openCenter"); \\}/g;
+
+	$evalString =~ s/\$BLTR\W/\$BLTR \\{ internalBalancingIncrement("openSpan"); \\}/g;
```
taniwallach: Note: The preprocessing substitutions for `$BLTR` and `$ELTR` are really not needed for release 2.14, as those new "formatting variables" are being added to PGbasicmacros.pl only as part of the UTF8 support code which has not yet been merge in. They were included, to avoid needed to add a different PR just to add these 2 substitutions at a later date either after the merge or into the PR with the current version of the internationalization code. 

I don't think the extra search/replace for the extra 2 strings should make any problem at present or add significant overhead which would not be necessary if those two variables do get defined as part of the UTF8 support additions.

dpvc: These substitutions assume that all uses of these variables are within `BEGIN_TEXT`/`END_TEXT` blocks, which isn't the case.  I think this is a bad idea that will cause problems for some use cases.  For example, I have done 

```
TEXT($BCENTER);
...
TEXT($ECENTER);
```
in some problems, and these changes will cause those problems to produce errors.  I would object to these changes being made.

Perhaps a better location for the change would be within the EVP3 function that processes the contents of `BEGIN_TEXT`/`END_TEXT` blocks, so that only the text within those will be affected.

taniwallach: > These substitutions assume that all uses of these variables are within `BEGIN_TEXT`/`END_TEXT` blocks, which isn't the case. I think this is a bad idea that will cause problems for some use cases. For example, I have done
> 
> ```
> TEXT($BCENTER);
> ...
> TEXT($ECENTER);
> ```
> in some problems, and these changes will cause those problems to produce errors. I would object to these changes being made.
> 
> Perhaps a better location for the change would be within the EVP3 function that processes the contents of `BEGIN_TEXT`/`END_TEXT` blocks, so that only the text within those will be affected.

@dpvc - **Thanks** you are certainly correct.

I had incorrectly assumed that these "formatting variables" would only be used inside `BEGIN_TEXT`/`END_TEXT` or similar blocks (`BEGIN_HINT` or `BEGIN_SOLUTION`) as I had though that the only real alternative was PGML which should not be using these constructs. I had not considered the possibility that problems might use these constructs directly using `TEXT($BCENTER)` or the like.

I will rework the code to move the substitutions into a very early stage of the `EV3P` function, where it should have the desired effect in places which this approach can try to "understand" without unforeseen adverse effects. 

Unfortunately, now that the case of these begin/end constructs being added to the output outside the `EV3P`  framework is something to consider - I realize that there may be "false positives" when the warning is triggered if balancing occurs outside `EV3P` and false negatives if an imbalance is triggered outside `EV3P`. Without deep parsing of the PG files, I doubt that such can be avoided, but the warning messages triggered in `ENDDOCUMENT()` will be amended to remark on the possibility of false positives.

Another weakness of this partial sanity check approach is that it cannot detect "out of order" pairings such as `$BUL $BBOLD $EUL $EBOLD`. But that too is something which would be very challenging to fix without a real parser doing detailed validation.

However, even with these limitations, I hope that an updated version of the (incomplete) sanity check for the balancing of these pairs will help limit the number of cases where an imbalance will break either the validity of the HTML code (needed to meet WCAG 2.0/2.1) or the LaTeX code (which will break PDF generation).

I also want to give credit to @Alex-Jordan  who raised the issue of the failure to enforce parity of these begin/end formatting variables in  https://github.com/openwebwork/pg/pull/323#issuecomment-353538792 which led me to try to add the sanity check code.

My motivation is related to plans which will add new methods of creating potential parity/nesting errors in the generated HTML (and probably other) output. The reason is that I am working to add a method to insert more flexible `<span>` and `<div>` tags into the HTML output that what `$BLTR` and `$ELTR` of https://github.com/openwebwork/pg/pull/323 were designed to do. The goal Is to allow:
  * setting the current text direction (what `$BLTR` and `$ELTR` of https://github.com/openwebwork/pg/pull/323 were intended to do),
  * changing the text language as need inline with `<span>` and at the block level with `<div>`,
  * allowing the code to provide class names to be included, to allow additional formatting via CSS where it may be desired.
  * Proving a manner to insert alternate "output" code for LaTeX mode, and hopefully also for PreTeXt mode.
For better or worse, it seems to me that the code to do this will also be using begin/end pairs (but as Perl functions rather than variables which just insert the output text for the correct context, so that the output can be built using parameters passed into the functions). As this will be new code, the balance tracking and sanity checks can be included in the code from the start.

Issue #289, opened 2016-11-29
Reducing perl warnings from existing problems


bldewolf:
We see a lot of perl output into our Apache logs from problem rendering.  To reduce this, I created a script that can headlessly render problems and collect their error output, if any.  The script is available in:
https://github.com/bldewolf/webwork-error-scanner

The scanner runs using checkouts of the webwork repos and only takes an hour or two to "render" every problem.

I've fixed a large amount of issues in these two branches:
https://github.com/bldewolf/pg/tree/fix-perl-warnings
https://github.com/bldewolf/webwork-open-problem-library/tree/fix-perl-warnings

There are still plenty of problems producing output but I haven't had time to work on it and I don't want the fixes to simply languish on my machine.  Should I make pull requests for what I have fixed so far?  The OPL branch has at least one very large commit, as I fixed a single class of problem in a lot of files.

Some of the errors are more complicated than simple perl style issues.  For example, NUM_CMP is misused in lots of places, with missing or incorrect arguments.  Fixing the problems to use NUM_CMP correctly is impractical, but fixing NUM_CMP to behave as it always has but without warnings is also a tricky task.

Anyway, this issue is more or less to make the team aware that this script exists and to ask for feedback on where to go from here.

mgage:
Hi Brian,

This is very nice.  I have a similar standaloneproblemeditor https://github.com/mgage/standaloneProblemRenderer  which you might like to compare
with yours.  Mine is a little more elaborate but I also use it to render batches of problems directly. 
Perhaps we can create a synthesis of your script and mine. Mine has been sitting around since before the semester started waiting until I had enough time to clean it up.  One thing it can do is render the problem, grab the expected answers, then rerender the problem while feeding in the expected answers. This tests whether the problem will accept its own correct answers (not always the case). In principle this could be used to test tolerances and many more unit tests for each pg problem.

Two things to note.  This standaloneProblemRenderer is directly coupled to PG and if you run it repeatedly, say on the same problem, you can increase the memory use of the process considerably. This means that PG has a memory leak (no surprise) but it also means that since the web is not involved there are tools to locate that memory leak and fix it.  It depends to some extent on the pg problem.  Some of the ones involving new MathObjects are particularly bad. -- not all of the links made when a module is created within PG are broken.  I'd really like to find that memory leak and plug it.  

[...]

PR #384, opened 2018-12-20
Add partial sanity checks for begin/end formatting constructs - revised code

taniwallach:
**Do NOT merge this - major rework is needed.**

---

This is a revised version of what had been attempted in https://github.com/openwebwork/webwork2/pull/851.
  * It was fixed after the feedback from @dpvc and following the suggested approach he indicated. 
    * Thanks Davide.
  * The code rewriting is done now only inside `EV3P` parsed blocks (and done somewhat more carefully) so the bugs which would have been triggered when the code rewriting was done inside `default_preprocess_code()` from `lib/WeBWorK/PG/Translator.pm` should not occur now.
  * The parity sanity check is far from perfect, but in my opinion still better than nothing.
    * Some comments on the currently unavoidable limitations of this parity sanity check code is in https://github.com/openwebwork/pg/pull/383#discussion_r243436461 .

Add code to to track balance of begin/end variables which are used as formatting constructs that open/close structures which require a parity check in HTML and/or in LaTeX.
  * Ex: `$BBOLD` and `$EBOLD`.
The variables of this sort which may lead to parity errors are defined in `PGbasicmacros.pl`.

The tracking is done by using substitutions early in the `EV3P()` parsing code forcing calls to adjust the counters/status settings. However, this only tracks usage processed by `EV3P`, namely inside `BEGIN_TEXT`/`END_TEXT`, `BEGIN_HINT`/`END_HINT`, and `BEGIN_SOLUTION`/`END_SOLUTION` blocks. Thus, there may be false positives and/or false negatives.

 * `lib/PGcore.pm`
   * Add a new flag warn_on_internalBalancing_errors and currently default to 1 (warnings on errors).
 * `macros/PG.pl`
   * Add the `%internalBalancing` hash to track the counters/status, initialize the local value of `$warn_on_internalBalancing`_errors for the problem, and initialize the value of the counters and
     status variables inside the hash.
   * Add the functions which adjust the counters/status variables, and issue warnings on errors and a debug message on valid changes, depending on the value of `$warn_on_internalBalancing_errors`.
  * In `ENDDOCUMENT` check on the counters/status variables, and warn if there is an imbalance error (if `$warn_on_internalBalancing_errors` > 0).
 * `macros/PGbasicmacros.pl`
   * Added a comment about this feature to the file, as the relevant constants are defined here.
   * Add a preprocessing phase to `EV3P()` to insert the calls to do the functions
     which adjust the internalBalancing counters/status variables. This must be
     done in a preprocessing phase as the function which sets the value of these
     variables are evaluated once to define the variable, but no "code" is run
     when the variables in a PG file.
   * Since the use is only traced inside `EV3P()` managed environments, false
     positives and negatives may occur if such formatting variables (or similar
     output code is pushed to the output stream) in a manner not tracked by the
     sanity check code added in this commit.
     * Without deep parsing of the PG files, I doubt that such can be avoided, but the warning messages 
        triggered in `ENDDOCUMENT()` will be amended to remark on the possibility of false positives.
   * Note: Other macros files may be adding HTML or LaTeX code which should be
     properly balanced. Hopefully they balance the tags themselves and do not use
     a begin/end construct which can lead to an imbalance.

Another weakness of this partial sanity check approach is that it cannot detect "out of order" pairings such as `$BUL $BBOLD $EUL $EBOLD`. But that too is something which would be very challenging to fix without a real parser doing detailed validation.

However, even with these limitations, I hope that an updated version of the (incomplete) sanity check for the balancing of these pairs will help limit the number of cases where an imbalance will break either the validity of the HTML code (needed to meet WCAG 2.0/2.1) or the LaTeX code (which will break PDF generation).

I also want to give credit to @Alex-Jordan  who raised the issue of the failure to enforce parity of these begin/end formatting variables in  https://github.com/openwebwork/pg/pull/323#issuecomment-353538792 which led me to try to add the sanity check code.

My motivation is related to plans which will add new methods of creating potential parity/nesting errors in the generated HTML (and probably other) output. The reason is that I am working to add a method to insert more flexible `<span>` and `<div>` tags into the HTML output that what `$BLTR` and `$ELTR` of https://github.com/openwebwork/pg/pull/323 were designed to do. The goal Is to allow:
  * setting the current text direction (what `$BLTR` and `$ELTR` of https://github.com/openwebwork/pg/pull/323 were intended to do),
  * changing the text language as need inline with `<span>` and at the block level with `<div>`,
  * allowing the code to provide class names to be included, to allow additional formatting via CSS where it may be desired.
  * Proving a manner to insert alternate "output" code for LaTeX mode, and hopefully also for PreTeXt mode.
For better or worse, it seems to me that the code to do this will also be using begin/end pairs (but as Perl functions rather than variables which just insert the output text for the correct context, so that the output can be built using parameters passed into the functions). As this will be new code, the balance tracking and sanity checks can be included in the code from the start.

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -1590,6 +1590,19 @@ sub MODES {
 	$PI					PI()				the number pi
 	$E					E()					the number e
 
+	Note: The begin/end constants are preprocessed at the start of EV3P() below
+	to add calls to functions which modify tracking variables inside the
+	%internalBalancing hash defined in macros/PG.pl, which are used to detect
+	imbalances in their use. Warnings are issued ONLY when the flag
+	warn_on_internalBalancing_errors is set to 1 or higher in lib/PGcore.pm.
+	When the flag is set to 2 or higher DEBUG_MESSAGE is used to report the
+	final values of the tracking variables at ENDDOCUMENT(). When the flag is set
+	to 3 or higher, DEBUG_MESSAGE will also be issued whenever the tracking
+	variables are modified. Since the sanity check code can only see the
+	parity changes which occur INSIDE blocks processed by EV3P(), there may
+	be false positives and/or false negatives.
+
+
 =cut
 
 
@@ -2128,6 +2141,45 @@ sub EV3P {
     %{$option_ref},
   );
   my $string = join(" ",@_);
+
+  # Preprocessing code to track internal balance (parity) of certain begin/end formatting
+  # variables by putting in calls to the Perl functions which track the status/counts for
+  # the internalBalancing sanity checks in a preprocess phase
+  # Note: This is being done after default_preprocess_code() from lib/WeBWorK/PG/Translator.pm
+  #       replaced "\\" by "\\\\" so we need 4 backslashes before the start/end braces below.
+
+    $string =~ s/\$BBOLD(\W)/\$BBOLD \\\\{ internalBalancingIncrement("openBold"); \\\\}\1/g;
+    $string =~ s/\$EBOLD(\W)/\$EBOLD \\\\{ internalBalancingDecrement("openBold"); \\\\}\1/g;
+
+    $string =~ s/\$BITALIC(\W)/\$BITALIC \\\\{ internalBalancingIncrement("openItalic"); \\\\}\1/g;
+    $string =~ s/\$EITALIC(\W)/\$EITALIC \\\\{ internalBalancingDecrement("openItalic"); \\\\}\1/g;
+
+    $string =~ s/\$BUL(\W)/\$BUL \\\\{ internalBalancingIncrement("openUnderline"); \\\\}\1/g;
+    $string =~ s/\$EUL(\W)/\$EUL \\\\{ internalBalancingDecrement("openUnderline"); \\\\}\1/g;
+
+    $string =~ s/\$BCENTER(\W)/\$BCENTER \\\\{ internalBalancingIncrement("openCenter"); \\\\}\1/g;
+    $string =~ s/\$ECENTER(\W)/\$ECENTER \\\\{ internalBalancingDecrement("openCenter"); \\\\}\1/g;
+
+    $string =~ s/\$BM(\W)/\$BM \\\\{ internalBalancingTurnOn("inInlineMath"); \\\\}\1/g;
+    $string =~ s/\$EM(\W)/\$EM \\\\{ internalBalancingTurnOff("inInlineMath"); \\\\}\1/g;
+
+    $string =~ s/\$BDM(\W)/\$BDM \\\\{ internalBalancingTurnOn("inDisplayMath"); \\\\}\1/g;
+    $string =~ s/\$EDM(\W)/\$EDM \\\\{ internalBalancingTurnOff("inDisplayMath"); \\\\}\1/g;
+
+    $string =~ s/\$BEGIN_ONE_COLUMN(\W)/\$BEGIN_ONE_COLUMN \\\\{ internalBalancingTurnOn("inOneColumnMode"); \\\\}\1/g;
+    $string =~ s/\$END_ONE_COLUMN(\W)/\$END_ONE_COLUMN \\\\{ internalBalancingTurnOff("inOneColumnMode"); \\\\}\1/g;
+
+    $string =~ s/\$BLABEL(\W)/\$BLABEL \\\\{ internalBalancingTurnOn("inInputLabel"); \\\\}\1/g;
+    $string =~ s/\$ELABEL(\W)/\$ELABEL \\\\{ internalBalancingTurnOff("inInputLabel"); \\\\}\1/g;
+
+    # The following two lines are related to new formatting variables added to development
+    # versions of PG by https://github.com/openwebwork/pg/pull/323 which use an HTML span to
+    # allow LTR text inside an RTL context.
+    $string =~ s/\$BLTR(\W)/\$BLTR \\\\{ internalBalancingIncrement("openSpan"); \\\\}\1/g;
+    $string =~ s/\$ELTR(\W)/\$ELTR \\\\{ internalBalancingDecrement("openSpan"); \\\\}\1/g;
+
+  # End preprocessing for internalBalancing sanity checks
+
   $string = ev_substring($string,"\\\\{","\\\\}",\&safe_ev) if $options{processCommands};
   if ($options{processVariables}) {
     my $eval_string = $string;
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2147,7 +2147,9 @@ sub EV3P {
   # the internalBalancing sanity checks in a preprocess phase
   # Note: This is being done after default_preprocess_code() from lib/WeBWorK/PG/Translator.pm
   #       replaced "\\" by "\\\\" so we need 4 backslashes before the start/end braces below.
+  my $warn_on_internalBalancing_errors = PG_restricted_eval(q!$main::warn_on_internalBalancing_errors!);
 
+  if ( $warn_on_internalBalancing_errors ) {
     $string =~ s/\$BBOLD(\W)/\$BBOLD \\\\{ internalBalancingIncrement("openBold"); \\\\}\1/g;
     $string =~ s/\$EBOLD(\W)/\$EBOLD \\\\{ internalBalancingDecrement("openBold"); \\\\}\1/g;
 
@@ -2177,7 +2179,7 @@ sub EV3P {
     # allow LTR text inside an RTL context.
     $string =~ s/\$BLTR(\W)/\$BLTR \\\\{ internalBalancingIncrement("openSpan"); \\\\}\1/g;
     $string =~ s/\$ELTR(\W)/\$ELTR \\\\{ internalBalancingDecrement("openSpan"); \\\\}\1/g;
-
+  }
   # End preprocessing for internalBalancing sanity checks
 
   $string = ev_substring($string,"\\\\{","\\\\}",\&safe_ev) if $options{processCommands};
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2150,35 +2150,35 @@ sub EV3P {
   my $warn_on_internalBalancing_errors = PG_restricted_eval(q!$main::warn_on_internalBalancing_errors!);
 
   if ( $warn_on_internalBalancing_errors ) {
-    $string =~ s/\$BBOLD(\W)/\$BBOLD \\\\{ internalBalancingIncrement("openBold"); \\\\}\1/g;
-    $string =~ s/\$EBOLD(\W)/\$EBOLD \\\\{ internalBalancingDecrement("openBold"); \\\\}\1/g;
+    $string =~ s/\$BBOLD(?=\W)/\$BBOLD \\\\{ internalBalancingIncrement("openBold"); \\\\}/g;
+    $string =~ s/\$EBOLD(?=\W)/\$EBOLD \\\\{ internalBalancingDecrement("openBold"); \\\\}/g;
 
-    $string =~ s/\$BITALIC(\W)/\$BITALIC \\\\{ internalBalancingIncrement("openItalic"); \\\\}\1/g;
-    $string =~ s/\$EITALIC(\W)/\$EITALIC \\\\{ internalBalancingDecrement("openItalic"); \\\\}\1/g;
+    $string =~ s/\$BITALIC(?=\W)/\$BITALIC \\\\{ internalBalancingIncrement("openItalic"); \\\\}/g;
+    $string =~ s/\$EITALIC(?=\W)/\$EITALIC \\\\{ internalBalancingDecrement("openItalic"); \\\\}/g;
 
-    $string =~ s/\$BUL(\W)/\$BUL \\\\{ internalBalancingIncrement("openUnderline"); \\\\}\1/g;
-    $string =~ s/\$EUL(\W)/\$EUL \\\\{ internalBalancingDecrement("openUnderline"); \\\\}\1/g;
+    $string =~ s/\$BUL(?=\W)/\$BUL \\\\{ internalBalancingIncrement("openUnderline"); \\\\}/g;
+    $string =~ s/\$EUL(?=\W)/\$EUL \\\\{ internalBalancingDecrement("openUnderline"); \\\\}/g;
 
-    $string =~ s/\$BCENTER(\W)/\$BCENTER \\\\{ internalBalancingIncrement("openCenter"); \\\\}\1/g;
-    $string =~ s/\$ECENTER(\W)/\$ECENTER \\\\{ internalBalancingDecrement("openCenter"); \\\\}\1/g;
+    $string =~ s/\$BCENTER(?=\W)/\$BCENTER \\\\{ internalBalancingIncrement("openCenter"); \\\\}/g;
+    $string =~ s/\$ECENTER(?=\W)/\$ECENTER \\\\{ internalBalancingDecrement("openCenter"); \\\\}/g;
 
-    $string =~ s/\$BM(\W)/\$BM \\\\{ internalBalancingTurnOn("inInlineMath"); \\\\}\1/g;
-    $string =~ s/\$EM(\W)/\$EM \\\\{ internalBalancingTurnOff("inInlineMath"); \\\\}\1/g;
+    $string =~ s/\$BM(?=\W)/\$BM \\\\{ internalBalancingTurnOn("inInlineMath"); \\\\}/g;
+    $string =~ s/\$EM(?=\W)/\$EM \\\\{ internalBalancingTurnOff("inInlineMath"); \\\\}/g;
 
-    $string =~ s/\$BDM(\W)/\$BDM \\\\{ internalBalancingTurnOn("inDisplayMath"); \\\\}\1/g;
-    $string =~ s/\$EDM(\W)/\$EDM \\\\{ internalBalancingTurnOff("inDisplayMath"); \\\\}\1/g;
+    $string =~ s/\$BDM(?=\W)/\$BDM \\\\{ internalBalancingTurnOn("inDisplayMath"); \\\\}/g;
+    $string =~ s/\$EDM(?=\W)/\$EDM \\\\{ internalBalancingTurnOff("inDisplayMath"); \\\\}/g;
 
-    $string =~ s/\$BEGIN_ONE_COLUMN(\W)/\$BEGIN_ONE_COLUMN \\\\{ internalBalancingTurnOn("inOneColumnMode"); \\\\}\1/g;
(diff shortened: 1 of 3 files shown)


taniwallach:
A sample PG file which will trigger several warning messages follows.
  1. It has an `$EITALIC` before a `$BITALIC` was used.
  2. It opens `$BM`, `$BBOLD`, `$BUL` but fails to close them.

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
);

TEXT(beginproblem());

Context("Numeric");

Context()->texStrings;
BEGIN_TEXT
Type 1234 in the answer: \{ ans_rule(5) \} 

$BR

EITALIC
$EITALIC
$BR

$BITALIC{Hello}$EITALIC{goodbye}
$BR

BM
$BM
$BR

BBOLD
$BBOLD
$BR

BUL
$BUL
$BR


END_TEXT
Context()->normalStrings;

ANS(Compute("1234")->cmp() );

ENDDOCUMENT();
```

With the code from this PR the warnings included below were issued.
  * Note: the code was run in a development environment with some other changes to these files, so the line numbers may be somewhat different than on a pristine merge of this PR into the current 2.14 CR code.

```
PG question processing error messages
PG warning messages

------
---- main /opt/webwork/pg/macros/PG.pl 228 ------
received a request to DECREMENT the internalBalancing counter called openItalic when it did not have a POSITIVE value.
__________________________
WeBWorK Warnings
WeBWorK has encountered warnings while processing your request. If this occured when viewing a problem, it was likely caused by an error or ambiguity in that problem. Otherwise, it may indicate a problem with the WeBWorK system itself. If you are a student, report these warnings to your professor to have them corrected. If you are a professor, please consult the warning output below for more information.

Warning messages
The internalBalancing counter called inInlineMath is not 0 when ENDDOCUMENT() was called, so there is some INBALANCE in the use of the relevant commands in this PG file. The current value is 1 at [PG]/macros/PG.pl line 446
The internalBalancing counter called openBold is not 0 when ENDDOCUMENT() was called, so there is some INBALANCE in the use of the relevant commands in this PG file. The current value is 1 at [PG]/macros/PG.pl line 446
The internalBalancing counter called openUnderline is not 0 when ENDDOCUMENT() was called, so there is some INBALANCE in the use of the relevant commands in this PG file. The current value is 1 at [PG]/macros/PG.pl line 446
Note: These internalBalancing warnings may be false positives if the relevant pairings were added outside of BEGIN_TEXT/END_TEXT BEGIN_HINT/END_HINT BEGIN_SOLUTION/END_SOLUTION blocks, ex. by direct use of TEXT(). at [PG]/macros/PG.pl line 451
```

Alex-Jordan:
Is there a way to test compile all problems in the OPL to see which ones trigger an imbalance warning with this?

taniwallach:
> Is there a way to test compile all problems in the OPL to see which ones trigger an imbalance warning with this?

I think that scripting such tests using `sendXMLRPC.pl` should be quite possible, as that will run a rendering of the PG code, which triggers the warnings and includes them in HTML output.  

As a quick check:
  * I set up a daemon course in my docker environment (and left the WW docker container running),
  * I set up the credentials files to use it
  * Note about the command line used below: Perl needed the setting to include the WW lib directory in the include path. I don't recall having this problem in the past, but did not have time to debug this now.
  * I ran (inside my `webwork2` directory from outside of Docker):
```
perl -I ./lib clients/sendXMLRPC.pl -vh /home/tani/NOTES/WebWork/sampleImbalance.pg
```

The console output included the warning messages set by the new code, as expected.

For the purpose of scanning for such issues, it could be that we would want to add a new "mode" to `sendXMLRPC.pl` which would only report the warnings to the console, and not include the regular output in what is send to the console.

@mgage - What do you think? Do you have any or know of any existing scripts which scan and run `sendXMLRPC.pl` on all pg files in a given directory?

taniwallach:
I have an idea to improve the sanity checks. 

I realized that the "out of order" issues are not difficult to detect it the code also tracks the current "stack" of begins.

The idea is to use a Perl array as a stack to track the sequence of open "begins", so then a "close" can check that the last open is what it is trying to close. That will help detect order issues like
```
$BUL $BBOLD $EUL $EBOLD
```
which should really be
```
$BUL $BBOLD $EBOLD $EUL 
```
so that validly nested HTML markup will be generated.

I'm working on an update to the code, and will put the revised code in the same branch, so it joins the current PR soon (probably only on Sunday or Monday).

macros/PGbasicmacros.pl
```diff
@@ -2128,6 +2141,45 @@ sub EV3P {
...
+    $string =~ s/\$ELABEL(\W)/\$ELABEL \\\\{ internalBalancingTurnOff("inInputLabel"); \\\\}\1/g;
+
+    # The following two lines are related to new formatting variables added to development
+    # versions of PG by https://github.com/openwebwork/pg/pull/323 which use an HTML span to
+    # allow LTR text inside an RTL context.
+    $string =~ s/\$BLTR(\W)/\$BLTR \\\\{ internalBalancingIncrement("openSpan"); \\\\}\1/g;
+    $string =~ s/\$ELTR(\W)/\$ELTR \\\\{ internalBalancingDecrement("openSpan"); \\\\}\1/g;
+
+  # End preprocessing for internalBalancing sanity checks
+
```
dpvc: All the other function in `EV3P()` are controlled by the `%options` hash above.  I'd recommend addition an option for this as well, so that it can be turned on or off as needed.  You can initialize it from the `warn_on_internalBalance_errors` environment variable if you want.  I would also move these out to a separate function (like `EV3P_parser()` below).

I'm also wondering if you could use a zero-width lookahead pattern `(?=\W)` rather than `(\W)`, as that would not require replacing `\1` in the replacement (apparently parenthesis sub matching is expensive).  This will also allow you to match two variables in a row, e.g., `$BBOLD$BITALIC`, which your current form will not match (because the first match will be for `$BBOLD$`, and after the replacement, the rest of the string is `ITALIC` and so won't match your pattern).  That means your current version might not properly count some situations, and will report errors incorrectly.

dpvc (changes requested):
I think your idea about using a stack to make sure the begin/end calls match correctly.  You would not need to use the individual counts at all, in that case, as the stack itself holds the counts.  That would seem to simplify the code considerably, as well as handing the nesting properly.  That seems to be the way to go.

macros/PGbasicmacros.pl
taniwallach: @dvpc - Thanks. Any additional feedback is welcome.

The new commits address most of this:
  * Added the stack (to track open tags), and using it to detect out-of-order issues and verify valid nesting.
  * Moved the reg-exp string replacement processing from EV3P() into a new subroutine called EV3P_parity_preprocessor().
  * Run the preprocessing only when a conditional passes, now using the value of `run_parity_preprocessor` which has been added to the options hash.
    * An earlier commit from today tested the $warn_on_internalBalance_errors variable directly, but as you said, it may be convenient to control this behavior using the options hash.
  * Changed to the zero-width look-ahead pattern, as suggested. 
    * Thanks for pointing out this technique, I have not seen it in the past.
    * I think the only cases the old code would have missed were when the same tag appeared twice in a row like $BBOLD$BBOLD as each tag is processed in a different regexp replacement line.

Notes:
  * For now I left in the use of status variables/counters:
    * It could be that there is no longer a real need for them.
    * However, it seems a simple method to save and use the current status to detect nesting of things which are not expected to nest (one-column mode, inline math mode, display math mode, being inside an input label).
    * It also detects balance errors without scanning the stack, which becomes necessary if case of improper nesting.
  * It seems that the 4 special functions could be replaced by 2: a "Push" and a "Pop" function which could depend on the type "inFoo" and "openFoo" to adjust some behavior based on whether nested use is expected or not. However, I think that 4 simple functions are probably easier to understand than 2 more complicated functions.
  * Some of the "mess" in the special functions is related to checking that the "type" is supported and tracked, in case someone would try to use these functions directly with a non-tracked type.

dpvc: > I think the only cases the old code would have missed were when the same tag appeared twice in a row like `$BBOLD$BBOLD` as each tag is processed in a different regexp replacement line.

Right.  Sorry, I had been considering asking the the regular expressions be combined, but decided against it.  I had that in mind, I guess.

> it seems a simple method to save and use the current status to detect nesting of things which are not expected to nest (one-column mode, inline math mode, display math mode, being inside an input label).

I'm not sure what you mean for inline and display math modes.  These are processed as a chunk (not as separate begin and end values), and not using variable substitution, so I'm not sure how this would work.

---

I have three additional comments:

1.  Your code should not be run when both `processCommands` and `processVariables` are both true.  If `processCommands` is false, then you parity commands will not be processed, and you will be inserting text that will not be removed (so will be changing the output), and if `processVariables` is false, then there it is incorrect to balance these values.  So the condition should include all three options.

2.  Your replacement string introduced extra spaces into the output text (between the variable and the `\{...\}`).  These may be significant, for example in `$BBOLD something$EBOLD,` would produce "**something** ," (with a space between the `something` and the `,`.  This not only is not what was written, it also introduces unwanted breakpoints, so you could get "**something**" on one line, with the comma at the beginning of the next line.

3.  Finally, your regular expression doesn't take braces into account.  I often use `This is ${BBOLD}bold$EBOLD.` when the variable must be used without a space following it.  For this example, you would miss the `${BBOLD}` and would report a balance error with the `$EBOLD`.  (Usually I will brace both of them, but this example is perfectly legal).  So I think the regular expression needs to be more complicated (sigh).

taniwallach:
> Is there a way to test compile all problems in the OPL to see which ones trigger an imbalance warning with this?

@Alex-Jordan - There seems to be a nice tool by @bldewolf to do a "warning scan" of a large set of problems (like the OPL). I ran into the tool almost by accident when looking at the issue list for pg. It seems that Brian's tool is probably a far more efficient approach for a large scan than using `sendXMLRPC.pl`. I have not yet tried it out at all or with the tests this PR is adding, and intend to wait to try it out until after I fix all the issues @dvpc raised in the current PR.

The tool is https://github.com/bldewolf/webwork-error-scanner and is referred to in https://github.com/openwebwork/pg/issues/289 where @mgage compares it to something he was working on.

@dlglin @jwj61  - I think we may want to add some mention of this tool to the Wiki and the OPL maintainers standard toolkit or processes. If we could extend Brian's tool to compare current run warnings to warnings from a prior run, there would be a mechanism to selectively ignore warnings which were "set aside" for whatever reason.

mgage:
The -c mode just checks that the problem "compiles".  The -cC mode checks that the problem compiles and that it's own "correct answers" satisfy the answer checkers.   The results are simply placed in the file (each file is marked correct, incorrect, or iffy). @bidewolf's script was developed independently and might well be quicker. It attempts to do less and is not quite as flexibile as sendXMLRPC.pl.  SendXMLRPC.pl could probably be optimized further as well. 

Notice also that sendXMRPC.pl will read from a directory if that is the input argument, and if I remember correctly it will process .pg files in the subdirectories as well.  It will only process files with the suffix .pg.

If you use the --list flag then the argument is a file which contains a list of the paths to the .pg problems to be rendered. (This is aimed for use as a unit test among other things, where the file contains a list of "representative" .pg problems that cover the known issues that need to be checked.)

The standaloneProblemEditor has the same front interface as sendXMLRPC.pl but connects directly to PG to render the problem.  It still needs a good deal more work.  It is useful because it catches more errors which are not transmitted back by a webserver. On the other hand you cannot check a large number of problems because of memory leaks.  I intend to use it to track down the last of the memory leaks in PG at some point.

Finding a better place to keep and advertise scripts is a good project.

macros/PGbasicmacros.pl
dpvc: [The following is from an email exchange, posted here to keep the history in place.]

> I think I understand what is needed in terms of items 2 and 3. I think it may be simpler to add an extra regexp with the braces then to create one which can handle the case of with and without braces at once. I'll need to do some review from the Camel books and some Googling+reading. 

One could do separate braced and unbraced versions, but I don't think thats necessary.  You could use

	/\$BBOLD(?=\W)/

to

	/\$(?:BBOLD|\{BBOLD\})(?=\W)/

which seems not too bad to me, and probably better than two separate searches.

> I also noticed that EV3P merges the array of "lines" it is sent into one single long string just before the location where I put in the call to the new EV3P_parity_preprocessor.  It may make sense do all the preprocessing to each line in the array, one at a time, before the lines are merged into one long string. The each line can be tested for matches to a "group" before running the regexp replacements on it (when needed). Ex. testing for say BOLD before running the BOLD regexp replacements on a given line. I'm not sure if this would be more efficient or not, but it seems likely that it should be more efficient, as "many" lines may not need any processing, and can be skipped after a single search rather than several searches by the various regexp operations. Any thought on this?

I have not done any testing, but I think you will find that doing single line expressions will be less efficient.  The regex engine is quite efficient even on large pieces of text, especially for simple patterns that are mostly runs of text, like yours.  As I recall, the regular expression string search complexity is linear in the size of the string used, which means it is roughly the same search time to run the replacement once on the combined strings as it would be to run it separately on the individual strings; but there is a setup cost per substitution (setting up the string to be searched, as I recall) of O(2^m), where m is the size of the search expression, so doing multiple (single-line) searches is more expensive than doing one combined search.

I also don't think that doing a pre-search for BOLD is the way to go.  You are trading off an extra search initially against the two searches for the begin-end pair if it does match.  So when there is NO match, you win, but when there IS a match, you end up with an extra search.  I'm not sure it is worth the extra effort.

You can reduce the number of searches by using an executable replacement via something like

	s/(\$([BE])(BOLD))(?=\W)/EV3P_begin_end($2, $3, $1)/e;

with something like (untested):

	sub EV3P_begin_end {
	  my ($be, $type, $var) = @_;
	  return ($be eq 'B' ? 
	    "$var\\{internalBalancingIncrement('$type');\\}" :
	    "$var\\{internalBalancingDecrement('$type');\\}");
 	};

Of course, you would have to change the defined types to be "BOLD", etc, rather than "openBold" and the others that you are using (or you could have a hash that translated from BOLD to openBold, and so on).  This lets you do both the begin and end checking in one substitution, which makes the checking for BOLD unnecessary (since this substitution already does that).

You could even go so far as to do something like

	s/(\$(?:([BE])(BOLD|ITALIC|UL|CENTER|D?M|LABEL|LTR)|(BEGIN|END)_(ONE_COLUMN)))/EV3P_begin_end($2||$4, $3||$5, $1)/e;

and

	sub EV3P_begin_end {
	  my ($be, $type, $var) = @_;
	  return ($be eq 'B' || $be eq 'BEGIN' ? 
	    "$var\\{internalBalancingIncrement('$type');\\}" :
	    "$var\\{internalBalancingDecrement('$type');\\}");
 	};

This would let you handle all the begin/end elements in one substitution, which would be more efficient (even though the regex is more complicated).  Because all are in one search, the EV3P_begin_end() function will be called for each variable in order, and so you can actually do the balancing right in that routine itself, rather than inserting commands to be executed later.  For example

	sub EV3P_begin_end {
	  my ($be, $type, $var) = @_;
	  if ($be eq 'B' || $be eq 'BEGIN') {
	    internalBalancingIncrement($type);
	  } else {
	    internalBalancingDecrement($type);
	  }
	  return $var;
 	};

This means the string is never actually changed (there replacement string is the same as the original), but the balancing functions are called at the time the substitution is processed.  So more efficient that inserting commands to be processed individually later.  Also, you would no longer need to make the processing dependent on the processCommands option.

You can add support for the braces as well.  Something like:

	$EV3P_pattern = qr/(?:([BE])(BOLD|ITALIC|UL|CENTER|D?M|LABEL|LTR)|(BEGIN|END)_(ONE_COLUMN))/;

	s/(\$(?:\{$EV3P_pattern\}|$EV3P_pattern))(?=\W)/EV3P_begin_end($2||$4||$6||$8, $3||$5||$7||$9, $1)/eo;

with the EV3P_begin_end() directly above.  Of course, this uses a lot of capturing parentheses, but it could be done with only one:

	$EV3P_pattern = qr/(?:[BE](?:BOLD|ITALIC|UL|CENTER|D?M|LABEL|LTR)|(?:BEGIN|END)_ONE_COLUMN)/;

	s/(\$(?:\{$EV3P_pattern\}|$EV3P_pattern))(?=\W)/EV3P_begin_end($1)/eo;

	sub EV3P_begin_end {
	  my $match = shift;
	  my $type = $match;
	  $type =~ s/[${}]//g;                  # remove dollar sign and braces
	  my $be = substr($type,0,1);           # get first character
	  $type =~ s/^(?:BEGIN_|B|END_|E)//;    # remove begin/end character or string
	  if ($be eq 'B') {
	    internalBalancingIncrement($type);
	  } else {
	    internalBalancingDecrement($type);
	  }
	  return $match;
	}

If you use a hash to turn the $type into your original categories, you can simplify this:

	%EV3P_balance_items = (
	  '$BBOLD' => 'openBold',
	  '$EBOLD' => 'openBold',
	  '$BITALICS' => 'openItalics',
	  '$EITALICS' => 'openItalics',
	  ...
	);

	sub EV3P_begin_end {
	  my $match = shift;
	  my $type = $EV3P_balance_items{$match};
	  if ($match =~ m/^\$\{?B/) {
	    internalBalancingIncrement($type);
	  } else {
	    internalBalancingDecrement($type);
	  }
	  return $match;
	}

I think something like this will be more efficient than what you are currently doing or suggesting.


> I'm less sure that I understood item 1, and think my confusion may really just be that a small text correction is needed in your feedback comment. You wrote:
> 
>> Your code **should not be run** when both processCommands and processVariablesare both true. If processCommands is false, then you parity commands will not be processed, and you will be inserting text that will not be removed (so will be changing the output), and if processVariables is false, then there it is incorrect to balance these values. So the condition should include all three options.
> 
> I think your intention was that the new code **should only be run** if both these flags are true (in addition to the flag added just for this code), as either one of the flags you mentioned is false - some part of the intended processing will fail. Is that correct?

Yes, you are correct.  I had edited the line several times, and apparently left it in a bad state.  I've corrected the comment on line.

> About "inline" and "display" math, ... in the WW context, I really mean to address $BM $EM vs $BDM $EDM  which are macros in macros/PGbasicmacros.pl defined by the lines below.

I had forgotten about these (if I even knew about them), and have never seen them used in practice.  I searched the OPL and found only one problem that used them, so I think they are not worth nothing about.  New problems are unlikely to be written using them.

My comment was about the fact that \(...\) and \[...\] are processed as units, so balancing them as you are other variables isn't necessary.  But I forgot about $BM/$EM and $BDM/$EDM.

---

One more thing to consider:  this balancing really only needs to be done during the creation or editing of a problem, not all the time.  There should probably be a development mode that is enabled by the PG editor (and perhaps the library browser), and your processing would only be done when that is in effect.  Once the problem has been developed (and is correct), there is no need to do this balancing every time a student views the problem or submits an answer.  That is a waste of resources for no reason.

taniwallach:
As noted in the revision at the top of the first message - this is **not** ready. Major rework is needed before a revised version of this code can be considered for the devel branch.

mgage:
@taniwallach let me know when this pull request is ready to join develop.

taniwallach:
Needs to be revised if someone has time to pick up on this.

PR #385, opened 2018-12-31
Correct version setting in macros/PG.pl in two places.

taniwallach:
I think this is a small fix which may need to be in the new master branch of PG (ver 2.14).

It fixes the value of `$main::VERSION` in `macros/PG.pl` from "PG-2.13+" to "PG-2.14"  in 2 places.


mgage:
Thanks.  i missed those.

PR #387, opened 2019-01-03
preserve tabs in PGML

Alex-Jordan:
This is a replacement for #381.

It changes PGML to preserve tab characters that are in a PGML block. The use case is for "code" in PGML, which is delimited with three backticks. You maybe intentionally want tabs to remain in that. Before this change, all tabs are replaced with four spaces. 

An unwanted side effect of things before this commit is that with the following (that has no tabs):
````
BEGIN_PGML
A preceding line
```
    indent
```

END_PGML
````

will introduce a tab character into the output. The new handling of tabs doesn't do this.

### DIFF
--- a/macros/PGML.pl
+++ b/macros/PGML.pl
@@ -32,7 +32,7 @@ package PGML::Parse;
 
 my $wordStart = qr/[^a-z0-9]/;
 
-my $indent = '^\t+';
+my $indent = '^(?:\t|    )+';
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
@dpvc should confirm here. Looking back, I'm not sure if the diff he provided was for the repo or just offering a solution to locally apply to my server.

PR #389, opened 2019-01-09
Bizarro updates

Alex-Jordan:
The main thing here is that the new definitions of f and g make bizarro usable with complex numbers.

I also added a bizarro unary negation operator. 

There's a commit that removes whitespace from the file and simplifies the example custom checker.

### DIFF
--- a/macros/bizarroArithmetic.pl
+++ b/macros/bizarroArithmetic.pl
@@ -17,6 +17,7 @@ The bizarro arithmetic is basically defined as:
 
 where f and g are inverse functions defined on all of R, and are odd functions. There is 
 also bizarro-, bizarro*, bizarro/, and bizarro^. f(x) = x^3+3x is a good choice for f.
+This has been extended below to a choice for f that works with complex numbers too.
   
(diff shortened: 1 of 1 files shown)


macros/bizarroArithmetic.pl
```diff
@@ -86,21 +86,24 @@ =head1 DESCRIPTION
...
 
+#The inverse of f.
 sub g {
-  my $x = shift;
-  return (cuberoot(($x+sqrt(($x)**2+4))/2) + cuberoot(($x-sqrt(($x)**2+4))/2));
+  my $z = shift;
+  my $r = abs($z);
+  return 0 if ($r == 0);
+  #Note that in what follows, base of (1/3) exponent is always a positive real
+  return $z/$r * ((($r+sqrt(($r)**2+4))/2)**(1/3) - ((sqrt(($r)**2+4)-$r)/2)**(1/3));
```
dpvc: You could cache the value of `sqrt(($r)**2+4)` for more efficiency.

PR #393, opened 2019-01-20
wantarray issue

Alex-Jordan:
The issue is explained in the thread for d61e0c7. Before the change here, `$ans->$method(@options)` is the argument to `PGML::LaTeX()`, therefore an array. Within certain methods, wantarray is present, and detects that `$ans->$method(@options)` is supposed to be an array, and behaves differently. This edit makes `$ans->$method(@options)` used in scalar context, changing the reaction from wantarray. And this reverts the behavior to how it was before d61e0c7.

### DIFF
--- a/macros/PGML.pl
+++ b/macros/PGML.pl
@@ -1012,7 +1012,8 @@ sub Answer {
         unshift(@options,$item->{name});
         $method = "named_".$method;
       }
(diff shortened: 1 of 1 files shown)


mgage:
I rechecked the code and the reasoning.  did not do extensive testing.  Alex has already done that in locating the bug.  Looks good.

PR #394, opened 2019-01-28
Devel candidate 01 27 2019  no multilingual changes

mgage:
This contains the commits that do not directly affect multilingual capabilities.  I.e. there are no (new) references to utf8 or UTF-8.  There have been some references to utf8 in develop (and master) for some time but with the multilingual pull request #390 we hope to unify and complete these references so that the code will work with all character sets.  

In order to facilitate testing I have removed all (I think) changes which involved utf8 from #390 
and placed them in this pull requests.  

The two pull requests should now be independent of each other.


mgage:
I let these changes accumulate for longer than I should have.  This commit is larger than usual.  There remain annoying issues with matrices/tableaus which should be resolved.  Some of the functions added in tableau.pl and also other .pl files involving matrices should probably be moved into Value::Matrix -- but that can be done later.

PR #396, opened 2019-02-09
Fix powers of functions so they dont become limits in display mode.

dpvc:
This fixes a problem with functions like `sech(x)` where `sech^2(x)` would appear with the `2` over top of the `sech(x)` in display mode.

To test, use

```
$f = Compute("sech^2(x)");
$g = Compute("asech(x)");

Context()->texStrings;
BEGIN_TEXT
\[$f \hbox{ and } $g\]
END_TEXT
Context()->normalStrings;
```

Without the patch, the 2 and -1 will appear above the sech(x).  With the patch, they should appear as superscripts.

This comes originally from [this forum post](http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4550)


mgage:
works.

PR #397, opened 2019-02-13
Fix the MultiAnswer macro when the namedRules option is used.

drgrice1:
The problem is that the ANS_NAME method is called first when an answer rule
is created (in the ans_rule method), and then again when answer
evaluators are created (in the cmp method).  Both times the NEW_ANS_NAME
method is called and different answer names are recorded.

This fix caches the answer names when they are first created in the
ANS_NAME method, and then subsequent calls just return the recorded
name.

### DIFF
--- a/macros/parserMultiAnswer.pl
+++ b/macros/parserMultiAnswer.pl
@@ -411,22 +411,21 @@ sub setMessage {
 ######################################################################
 
 #
-#  Produce the name for a named answer blank
-#  (Use the standard name for the first one, and
-#   create the prefixed names for the rest.)
+#  Produce the name for a named answer blank.
+#  (When the singleResult option is true, use the standard name for the first
+#  one, and create the prefixed names for the rest.)
 #
 sub ANS_NAME {
-  my $self = shift; my $i = shift; my $name;
+  my $self = shift; my $i = shift;
+  $self->{answerNames} = {} if !defined($self->{answerNames});
+  return $self->{answerNames}->{$i} if defined($self->{answerNames}->{$i});
   if ($self->{singleResult}) {
-    if (!$self->{id}) {
-      $name = $self->{answerName} = main::NEW_ANS_NAME();
-      $self->{id} = $answerPrefix.$name;
-    }
-    $name = $self->{id}."_".$i unless $i == 0;
+    $self->{answerNames}->{0} = main::NEW_ANS_NAME() if (!$self->{answerNames}->{0});
+    $self->{answerNames}->{$i} = $answerPrefix.$self->{answerNames}->{0}."_".$i unless $i == 0;
   } else {
-    $name = main::NEW_ANS_NAME();
+    $self->{answerNames}->{$i} = main::NEW_ANS_NAME();
   }
-  return $name;
+  return $self->{answerNames}->{$i};
 }
 
 #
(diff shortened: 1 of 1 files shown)


macros/parserMultiAnswer.pl
```diff
@@ -411,22 +411,21 @@ sub setMessage {
...
-#  (Use the standard name for the first one, and
-#   create the prefixed names for the rest.)
+#  Produce the name for a named answer blank.
+#  (When the singleResult option is true, use the standard name for the first
+#  one, and create the prefixed names for the rest.)
 #
 sub ANS_NAME {
-  my $self = shift; my $i = shift; my $name;
+  my $self = shift; my $i = shift;
+  $self->{answerNames} = {} if !defined($self->{answerNames});
```
dpvc: This is not strictly necessary, since Perl will create the hash when it is first used.

macros/parserMultiAnswer.pl
```diff
@@ -411,22 +411,21 @@ sub setMessage {
...
+  return $self->{answerNames}->{$i} if defined($self->{answerNames}->{$i});
```
dpvc: The second `->` is not necessary, as it is implied after the first one.  So `$self->{answerNames}{$i}` is sufficient, and follows the style used in the rest of the file.  Although we don't have a style guide, it seems best to keep a consistent style within a single file.

macros/parserMultiAnswer.pl
```diff
@@ -411,22 +411,21 @@ sub setMessage {
...
   if ($self->{singleResult}) {
-    if (!$self->{id}) {
-      $name = $self->{answerName} = main::NEW_ANS_NAME();
-      $self->{id} = $answerPrefix.$name;
-    }
-    $name = $self->{id}."_".$i unless $i == 0;
+    $self->{answerNames}->{0} = main::NEW_ANS_NAME() if (!$self->{answerNames}->{0});
```
dpvc: You probably want `if !defined($self->{answerNames}{0}` (or better `unless defined($self->{answerNames}{0}`), since `!$self->{answerNames}{0}` is true in some cases when `$self->{answerNames}{0}` is defined (e.g., if it were an empty string, or the number 0).  This has been a chronic source of problems (and I am certainly guilty of doing this myself all over the place), and it is best to get in the habit of using `defined()` when that is what you really mean.

macros/parserMultiAnswer.pl
drgrice1: I didn't realize latter ->'s were implied.  Good to know.

macros/parserMultiAnswer.pl
drgrice1: I wasn't sure on that.  I guess I should have tested that first.

macros/parserMultiAnswer.pl
drgrice1: That certainly does improve readability to use unless.  I did know the difference between defined and without, but I didn't think it mattered here.  I guess it is better though to use defined in any case here.

drgrice1: Thanks for taking a look at this.

mgage:
I'll commit this to master as a hot-fix.

mgage:
Thanks @drgrice1

PR #399, opened 2019-03-06
Added check to see if Livegraphics defined

goehle:
This is a small change to fix a bug in ProblemSetDetail2.  To reproduce it: 

1.  Add a problem with LiveGraphics3D to a homework set `Library/Union/setMVlevelsets/levels-5/levels-5b.pg` for example. 
2.  Click the "Automatically render problems on page load" option and save settings. 
3.  The page should load with broken javascript and be non-functional. 

After this patch the page should load correctly, but the LiveGraphics3D will not actually run.   You should also check that LiveGraphics3D works when you actually view the problem.

### DIFF
--- a/macros/LiveGraphics3D.pl
+++ b/macros/LiveGraphics3D.pl
@@ -156,8 +156,10 @@ sub LiveGraphics3D {
 		    vars : $ind_vars,
     };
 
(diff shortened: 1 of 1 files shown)


mgage:
I checked this.   Thanks Goeff.

PR #405, opened 2019-05-23
Fix parserRadioButtons.pl to avoid an undesired warning when a radio button is used inside a MultiAnswer group with allowBlankAnswers turned on

taniwallach:
Fix for a bug seen with a radio button used in MultiAnswer with `allowBlankAnswers` set on. 

  * In that case, a warning was issued when the problem is loaded and no radio button is selected. 
  * Submitting an answer with a radio button selected does not issue the warning.
  * However, submitting an answer where no choice was made for the set of radio buttons triggers the warning.

The warning reports:
```
substr outside of string at line 375 of [PG]/macros/parserRadioButtons.pl
Use of uninitialized value $index in array element at line 376 of [PG]/macros/parserRadioButtons.pl
Use of uninitialized value $index in array element at line 377 of [PG]/macros/parserRadioButtons.pl
```

The **only** occur when `allowBlankAnswers` is set (to `1`), which was desired in the original question. The example below was pruned to be minimal.

The root problem is that labelText() is apparently sometimes called with either an empty second ("index") argument or one too short to allow the code
```
	$index = substr(shift,1);
```
to work (when a blank answer is allowed). 

The fix defers the call to substr until it is certain that it will work, and when it would fail - returns a fixed string "Bn".

A hopefully relatively minimal pg file which hits the issue (cut down from something which once had meaning to it) is the following:
```

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
  "parserRadioButtons.pl",
  "parserMultiAnswer.pl"
);

TEXT(beginproblem());

$radio1 = RadioButtons(
  ["No",
  "Yes"],
  "No" # correct answer
);

$multians1 = MultiAnswer($radio1)->with(
  allowBlankAnswers => 1,
  checker => sub {
      my ( $correct, $student, $self ) = @_;
      my ( $stu1rb  ) = @{$student};
      my ( $corr1rb ) = @{$correct};
      if ( $stu1rb == $corr1rb ) {
        return 1;
      } else {
        return 0;
      }
    }
);

Context()->texStrings;
BEGIN_TEXT
\{ $multians1->ans_rule \}
END_TEXT
Context()->normalStrings;

ANS( $multians1->cmp() );

ENDDOCUMENT();        # This should be the last executable line in the problem.
```

When the line
```
  allowBlankAnswers => 1,
```
is not commented out - the warning will be given, while when this line is commented out - no warning is given.

### DIFF
--- a/macros/parserRadioButtons.pl
+++ b/macros/parserRadioButtons.pl
@@ -370,7 +370,12 @@ sub labelFormat {
 #  to be displayed in the results table.
 #
 sub labelText {
-  my $self = shift; my $index = substr(shift,1);
+  my $self = shift;
+  my $index = shift;
+  #warn "index = |$index|";
+  return "Bn" if ( !defined($index) || $index eq "" || length($index) < 2 );
+  $index = substr($index,1);
+#  my $self = shift; my $index = substr(shift,1);
   my $choice = $self->{labels}[$index];
   $choice = $self->{orderedChoices}[$index] unless defined $choice;
   return $choice if length($choice) < $self->{maxLabelSize};


dpvc:
This is not the best solution, and, in fact, it is wrong in a couple of ways.  The `labelText()` method is supposed to return the text of the selected answer, and the problem you are reporting occurs when no answer is selected and `allowBlankAswers` is true, as `labelText()` was not designed to be called in that case.  The various selections have been assigned values of `B0`, `B1`, etc, and `labelText` takes a value of this form and returns the answer that was given that value.  I.e., if the value is `Bn`, it is the n-th radio button, so it returns the n-th answer string in the list

Your proposed fix checks if the value passed to `labelText()` is blank, and returns `Bn` in that case.  That is not an appropriate value, as it is not one of the answers, and if the user doesn't select a value but presses "Submit Answers" or "Check Answers", the answer will be shown as `Bn`.  I assume you were trying to convert the blank answer into one of the button values, but `Bn` is not such a value (the `n` would have to be an actual number), and even if it were, you would not want to return it, but should go on to the rest of the function so that it would be used to look up the corresponding answer text.  But that would also be wrong, since that would make it appear that some choice **was** selected, which it isn't.  it would be better to return and empty string, as that would at least not show any answer as selected.

But I think this is the wrong place to be making the change in any case.  The real problem is that `labelText()` is being called at all.  I think a better solution is to change 

https://github.com/openwebwork/pg/blob/dfef4c682ccc93305545671fde7bd9edfb83ed37/macros/parserRadioButtons.pl#L398

to be 

```
  if (defined $ans->{student_value} && $ans->{student_value} ne '') {
```

That is all that needs to be changed to resolve the issue.

taniwallach:
I am closing this PR. It has been replaced by https://github.com/openwebwork/pg/pull/406 which uses the fix recommended by @dvpc.

@dpvc - thanks for the prompt reply. Point taken. My primary interest and need was to avoid the warning message on the initial problem display, and your fix certainly fixes that problem. "Bn" using the letter "n" and not a number was meant to be a "place marker" for cases when the bug made trouble, and I did not think that the letter "n" would lead to any strange behavior in practice.

PR #407, opened 2019-05-25
PG-2.15-BETA

mgage:
This is a BETA version containing multilingual features (utf8 4 byte characters). There are database changes so test cautiously (backup up the database just in case).  If $ENABLE_UTF8MB4=0 you should be able to use this on existing sites, even with mysql databases earlier than 5.3.  If you set $ENABLE_UTF8MB4=1 then you will get a "wide character" error message from the database if it is too old. If it is new enough you will be able to use all characters including Chinese, emoji, Arabic and so forth. (Download the companion WeBWorK-2.15-beta as well.)


mgage:
oops -- fixed the conflicts the wrong way.

PR #410, opened 2019-06-04
fix typo: maketex -> maketext

heiderich:


taniwallach:
This only changes a comment, so seems safe to merge.

PR #411, opened 2019-06-07
Add weightedGrader.pl to the base pg library

taniwallach:
`weightedGrader.pl` file is described [in the Wiki documentation](http://webwork.maa.org/wiki/WeightedGrader) but is not yet part of the base PG library 

It is currently available in the OPL in 2 places:
  - `OpenProblemLibrary/macros/CollegeOfIdaho/weightedGrader.pl`
  - `OpenProblemLibrary/macros/Union/weightedGrader.pl`
but it seems that it should be included in the base PG library so that using it will not depend on having the OPL tree available.

### DIFF
--- a/macros/weightedGrader.pl
+++ b/macros/weightedGrader.pl
@@ -1,0 +1,331 @@
+loadMacros('unionUtils.pl');
+
+sub _weightedGrader_init {}; # don't reload this file
+
+######################################################################
+#
+#  A weighted grader that allows you to assign arbitrary percentages
+#  to the various answers in a problem.  It also allows you to indicate
+#  that answering one part correctly will give you credit for some
+#  other part(s).  This way, if there are several parts leading up to
+#  a "goal" answer, and the student produces the goal answer by
+#  some other means, he can be given full credit for the problem anyway.
+#
+#
+#  WEIGHTED ANSWERS:
+#
+#  Each problem is assigned a weight (the default is 1).  The
+#  student's score is then the sum of the weights for his correct
+#  answers divided by the total of the weights for all answers.  (To
+#  assign weights as percentages, use integers that add up to 100, eg,
+#  use 40 and 60 for the weights for two answers.)
+#
+#  There are two ways to assign weights.  The first is to use the
+#  WEIGHTED_ANS() routine (in place of ANS) to give an answer checker
+#  plus a weight.
+#
+#  Example:
+#
+#      WEIGHTED_ANS(std_num_cmp($ans1),2);
+#
+#  This assigns a weight of 2 to the corresponding numeric answer.
+#
+#  As with ANS(), WEIGHTED_ANS() can take more than one answer checker and
+#  weight.
+#
+#  Example:
+#
+#      WEIGHTED_ANS(
+#         std_num_cmp($ans1), 40,
+#         std_num_cmp($ans2), 60
+#      );
+#
+#  This assigns 40% to the first answer and 60% to the second answer.
+#
+#  The second way of assigning a weight is through the weight_ans()
+#  function.  This takes a single answer checker and weight and returns
+#  a new answer checker of the same type that has the desired weight.
+#  Thus
+#
+#      ANS(weight_ans(std_num_cmp($ans1),2));
+#
+#  is equivalent to the first example above.
+#
+#  The main purpose for weighted_ans() is so that weights can be used
+#  with UNORDERED_ANS(), or in other places where you want to use the
+#  weighted answer checker directly.  For example:
+#
+#     UNORDERED_ANS(
+#        weight_ans(std_num_cmp($ans1),40),
+#        weight_ans(std_num_cmp($ans2),60),
+#     );
+#  
+#  produces two answers whose order doesn't matter, but the student
+#  will get 40% for getting $ans1 and 60% for getting $ans2 (no matter
+#  what order they appear in).
+#
+#  Note that the blank_cmp() answer checker has a weight of 0 by
+#  default.  You can override this using weight_ans(); for example,
+#  weight_ans(blank_cmp(),1) makes the blank count the same as all
+#  the other answers.  (If there are two or more non-blank answers,
+#  then having the blanks with weight 0 will allow the observant
+#  student to deduce the number of blank answers from the percentage
+#  for a single correct answer, provided all the non-blank answers are
+#  equally weighted).
+#
+#  Once you have given weights to the answers, you also need to
+#  install the weighted grader.  Do this using the command
+#
+#      install_weighted_grader();
+#
+#
+#  HAVING ONE ANSWER PROVIDE CREDIT FOR ANOTHER:
+#
+#  You may want to have a correct answer for one problem automatically
+#  give credit for one or more other parts of the problem.  For example
+#  If several parts are used to lead up to the "real" answer to the
+#  problem, and the student produces that final answer without doing
+#  the intermediate parts (perhaps using some other method), then you
+#  may want to give the student full credit for the problem anyway.
+#  You can do so in the following way.
+#
+#  First, let us call the final answer the "goal" answer, and the
+#  answer that it would give automatic credit for the "optional" answer.
+#
+#  The optional answer blank must be a named answer, e.g.,
+#
+#    BEGIN_TEXT
+#      You get credit for this answer: \{NAMED_ANS_RULE('optional',10)\}
+#      When you answer this one: \{ans_rule(10)\}
+#    END_TEXT
+#    NAMED_ANS('optional',std_num_cmp(5));
+#
+#  Then for the goal answer, in place of ANS, use CREDIT_ANS, listing the
+#  optional answer as the second argument:
+#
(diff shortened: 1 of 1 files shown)


dpvc:
Because this file relies on `unionUtils.pl`, you may want to move that file as well, or at least transfer the needed definitions from there to `weightedGrader.pl`.

This was one of my earlier contributions, and it probably bears updating.  Some of the techniques are pretty clunky.

taniwallach:
I will examine the files and try to see which approach seems more reasonable.

About "clunky" we may be stuck with it to the extent that it works and is probably used by many problems which expect the current behavior until someone wants to consider updating it and can carefully manage any changes to minimize changes to the interface and current behavior.

drgrice1:
This would be nice to have in the PG base.  I don't think that unionUtils.pl should also be added.  That seems silly to me for two one liner functions.

Delete line 1
```perl
loadMacros('unionUtils.pl')
```
change line 290 from
```perl
	  if (trimString($answers{$credit_name}->{student_ans}) eq "");
```
to
```perl
	  if ($answers{$credit_name}->{student_ans} =~ s/^\s+|\s+$//gr) eq "");
```
and change line 310 from
```perl
	    isNumber($problem_state{recorded_score}));
```
to
```perl
	    $problem_state{recorded_score} =~ m/^[+-]?(\d+(\.\d*)?|\.\d+)([Ee][-+]?\d+)?$/);
```
and then this macro will work without unionUtils.pl.

The techniques used in the file may be "clunky", but the macro works well.  Weighting answers to me is an important thing to be able to do.  Many proprietary online homework systems don't have this capability.  So this is one simple way that WeBWorK can get an edge.

mgage:
@taniwallach do you have items that you want to add to this pull request #411 (weighted graders) before it is merged?

macros/weightedGrader.pl
```diff
@@ -0,0 +1,330 @@
...
+  #
+  #  Mark any optional answers as correct, if the goal answers
+  #  are right and the optional ones are blank.
+  #
+  foreach $ans_name (keys %answers) {
+    if ($credit{$ans_name} == 1 &&
+	defined($answers{$ans_name}->{credit})) {
+      foreach $credit_name (@{$answers{$ans_name}->{credit}}) {
+	$credit{$credit_name} = 1 
+	  if ( ($answers{$credit_name}->{student_ans} =~ s/^\s+|\s+$//gr) eq "");
```
drgrice1, line 289: This is what I initially put into my comment and works, but I realized that this is a bit silly.  Rather than using a substitution to trim whitespace from the beginning and end of the string and then comparing to the empty string, just use a match against /^\s*$/.  I updated my comment with this.

taniwallach: Thanks. The first code was also missing a parenthesis... and after pushing I noticed your change. I force pushed a file with the revised change.

taniwallach:
> @taniwallach do you have items that you want to add to this pull request #411 (weighted graders) before it is merged?

@mgage Sorry I left this hanging. I did not get back to figuring out what was really needed from `unionUtils.pl` and got caught up in other work... (Pressure sometimes helps free time :) to get things finished.)

I used @drgrice1's  approach to avoiding the dependency on `unionUtils.pl` (first the original approach, then the improved one using a "git commit --amend" and a forced push).

@drgrice1 - Thanks!

I quickly tested this by hacking my Docker setup to avoid mounting the OPL, and to include the modified version of `weightedGrader.pl` and then tested using a problem with weighed grading. The test problem worked properly when there was no OPL to provide the regular source for  `weightedGrader.pl`. Thus, I think the current PR with the patches suggested by @drgrice1 works fine on an OPL-less system, so now this is ready to merge.

Warning: If you try to test with Docker, it is important to mount a hacked up version of `docker-config/docker-entrypoint.sh` to avoid the Docker startup process from cloning the OPL when it is missing.

drgrice1:
I just noticed that this pull request is set to be merged into PG-2.15.  It should go into the develop branch instead.

taniwallach:
> I just noticed that this pull request is set to be merged into PG-2.15. It should go into the develop branch instead.

Thanks for noticing. I had started work on this when we were still working on 2.15...

It is not targeted to develop.

Alex-Jordan:
I don't object to this move.

I do have a lower level concern about weightedGrader.pl in the first place that I'll express in this thread. If I were redesigning WeBWorK from the ground up, I would separate this kind of thing from individual problem coding. Say you write a problem with two answer blanks (a) and (b). You want (a) to be worth 90% and (b) to be worth 10%. But another instructor wants them to be 50/50. And a third wants no credit at all unless (a) is correct, and then after that weights them at 70/30. On top of all that, the original problem writer may want one scoring scheme for that problem in a homework set and a different scoring scheme when it appears on a quiz.

How much weight to assign each answer blank is an instructional choice, and by coding it into the problem, that puts restrictions on other instructors and reduces the potential audience of other instructors who might want to adopt the problem (through the OPL, say).

A WeBWorK problem set assigns a weight to each problem. That's the philosophy I would take if rebuilding things from the ground up in some hypothetical world: make the scoring scheme an attribute of the problem set, not the problem. Even the scoring scheme for an individual problem would be a feature of the problem set that it is being used in, not of the problem itself.

I think some of these considerations are not so abstract with the efforts being made to process problems in new ways. For example Rederly, PreTeXt, Runestone, Edfinity. In those places it can get awkward trying to process a problem code faithfully with these instructional settings baked in.

dpvc:
@Alex-Jordan, I agree with you that there are things currently that are handled in the problem that rightly belong as part of the set.  Grade weights are one, but so are number of tries until a hint, randomization, style of display for vectors and other things (vector norms, i,j,k vectors, matrix transpose, and so on).  The weighted grader is something like 16 years old, when things were still pretty new (and I didn't know as much about WeBWorK, and before MathObjects, PGML, and such even existed).

I'd suggest that there might be a middle ground, where the problem has some meta-data that is passed up to the homework set as defaults, but that can be overridden by the professor.  That would give the professor the sense of what the author had in mind, but would allow the professor to set them as they see fit.

taniwallach:
@Alex-Jordan IMHO we need to differentiate between the the weights assigned to different problems in a set, which is relatively easy to handle and already well supported and between adjusting internal weights to different "parts" of a single problem.

I think that in "large" problems with many answer blanks (or other answerable objects) there is often a desire by a problem author to assign non-uniform weights to the different items. I certainly find myself doing this quite often. One common use is when didactic preliminary questions are included before the "main" question (ex. do the conditions of theorem X hold in this case, what is the Jacobian of the transformation being used, etc.).

Trying to provide a high-level mechanism to pass "item" weights into such questions would certainly be complicated and unwieldy, require some sort of support in the database to store/retrieve such weights by item "name" and involves creating a mechanism via which a problem would suggest default part weights to the "webwork2" CMS level and accepting override values back into the the "pg" renderer/grader level. That seems quite complicated to handle, and would need to be quite general in score, and thus would require providing the instructor with a very general tool to adjust such weights. I suppose a JSON dictionary of "internal item weights" could provide a parsable text based format to transfer data about such weights into/out of problems and the database, and the documentation could explain how to edit such data (unless a very general form-based editor was created to handle it).

Given the complications of such an approach, I suspect that keeping the current approach using `weightedGrader.pl` provides reasonable functionality to a problem author, even if his decisions may not be exactly compatible with how some other users would want the weights assigned.

Problem users who wish to assign different internal weights to a given question can create a locally modified version of the PG file and adjust the weights in their copy of the problem. Doing that is not all that complicated, and is not significantly  different that what is needed if minor syntax/phrasing/formatting changes are desired for "local" use.

Alex-Jordan:
Maybe this sounds naive, but I will outline how I think it would work.

Currently, a set def file has `value = 3` or something like that to store the weight of the problem as a whole. And that value is in somewhere in the database (maybe many places, repeated for users where it has been overwritten.)

That could be `value = '[0.5, 0.5, 2]'`. I assume there is a natural order to the answer blanks, not counting extensions, and they would be assigned their weights this way. If there is some mismatch between the number of answer blanks and the number of values, I would have it default to 1/n for each blank, where there are n answer fields (again, not counting extensions).

drgrice1:
I think this should be merged at this point.  This isn't something new, and this just makes it more visible and in the PG code base.

Alex:

I think your request is a good feature request to consider for future development.  Your suggestions will take some renovation of the system to make work.  The current value of a problem is of INT data type in the database.  Note that databases do not have list data types.  So the list idea won't work directly.

What about extended answers?  For example, a MultiAnswer problem.  That has separate parts that can be weighted also.  This is usually done by the custom checker routine's return value, and for single result MultAnswers can even be completely dealt with internal to the grader.

Trying to do this is almost certainly going to break compatibility with existing problems, or at least would most likely be quite difficult to achieve without doing so.

Alex-Jordan:
Yes, thanks everyone for reading my post(s). I did not mean to derail the merge, just plan the seed for thinking about this kind of separation (and more broadly, as Davide listed) in the future. I will merge this now.

drgrice1:
I agree entirely on the sentiments of your idea though.  It will be a challenge to implement.

PR #413, opened 2019-06-18
Fix a typo in lib/PGcore.pm

taniwallach:
Fix a typo in `lib/PGcore.pm`. Had `":uft8"` where `":utf8"` is needed.

Due to the error, the error message
```
  Unknown PerlIO layer "uft8" at /opt/webwork/webwork2/../pg//lib/PGcore.pm line 34.
```
would be reported (when errors were visible) such as when using `sendXMLRPC.pl`.

The sample below used `sendXMLRPC.pl` from inside a Docker container (where I could get it to work), and using the `--help` option, but the same error message was reported for other uses of `sendXMLRPC.pl`.

```
docker container exec -it webwork2_app_1 bash 

cd webwork2/clients
./sendXMLRPC.pl --help     
Unknown PerlIO layer "uft8" at /opt/webwork/webwork2/../pg//lib/PGcore.pm line 34.
"our" variable @path_list redeclared at ./sendXMLRPC.pl line 400.
"my" variable $credentials_string masks earlier declaration in same scope at ./sendXMLRPC.pl line 402.
"our" variable $UNIT_TESTS_ON redeclared at ./sendXMLRPC.pl line 567.
```

### DIFF
--- a/lib/PGcore.pm
+++ b/lib/PGcore.pm
@@ -31,7 +31,7 @@ use MIME::Base64();
 use PGUtil();
 use Encode qw(encode_utf8 decode_utf8);
 use utf8;
-binmode(STDOUT, ":uft8");
+binmode(STDOUT, ":utf8");
 ##################################
 # PGcore object
 ##################################


taniwallach:
BTW the line being corrected:
```
binmode(STDOUT, ":utf8");
```
is needed only if something in the code needs to "print" UTF-8 output to `STDOUT`.

It is not clear to me if there is any real use of this under `PGcore.pm` (maybe some sort of logging output?). If it is not needed - the line can be commented out. When the line is commented out, problems using UTF-8 seems to work fine for me inside "regular" courses.

This line seems to have been added in https://github.com/mgage/pg/commit/a561c0e75a0db16ab71bf89818ba421c7bf0d30c#diff-3b0e3731b9d4f96b63aea99a092b15c4
together with
``
use utf8;
```
which is used when UTF-8 encoded strings are in the Perl source file, which also may be unnecessary in `PGcore.pm`.

PR #415, opened 2019-07-28
Add problemUUID to the identifiers used to create $unique_id_seed

mgage:
This, and the companion pull request for webwork2 (https://github.com/openwebwork/webwork2/pull/976) add the value of the variable problemUUID to the list of items used to create a unique UUID for a problem. Its use case is for submissions via the webservice where the problem is not part of a homework set framework which help identify the problem and its auxiliary files uniquely. Specifying problemUUID=>3245 (for example) in the form or iframe request for the problem will do this. problemUUID replaces problemIdentifierPref as the name of the identifying variable being submitted although to insure backward compatibility the latter will still work.

### DIFF
--- a/lib/PGalias.pm
+++ b/lib/PGalias.pm
@@ -102,6 +102,7 @@ sub initialize {
 	$self->{externalGif2PngPath} = $envir->{externalGif2PngPath};
 	$self->{courseID}            = $envir->{courseName};
 	$self->{problemSeed}         = $envir->{problemSeed};
+	$self->{problemUUID}         = $envir->{problemUUID}//0;
 	
 	$self->{appletPath} = $self->{envir}->{pgDirectories}->{appletPath};
 	#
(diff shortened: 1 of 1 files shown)


heiderich:
BTW: @mgage Did you notice https://github.com/mgage/webwork2/pull/29?

lib/PGalias.pm
```diff
@@ -117,7 +118,8 @@ sub initialize {
 				  $self->{courseID},
 				  'set'.$self->{setNumber},
 				  'prob'.$self->{probNum},
-				  $self->{problemSeed}
+				  $self->{problemSeed},
+				  $self->{problemUUID},
```
taniwallach: The current approach always adds the `problemUUID` into the data used to create `$unique_id_seed`.
This will mean that old UIDs become invalid.
Would it make sense to involve the `problemUUID` only in the cases where it is expected to be needed (problems outside a regular set/problem context where `$self->{setNumber}` and `$self->{probNum}` have no meaningful value?

taniwallach: Making a change was determined to be unnecessary.
See the discussion at https://github.com/mgage/webwork2/pull/30

Issue #421, opened 2019-09-10
verbatim delimiter


Alex-Jordan:
Once upon a time (actually still in 2.14), in `lib/Value/String.pm`, there was this code:

```
#
#  Mark a string to be display verbatim
#
sub verb {shift; return "\\verb".chr(0x85).(shift).chr(0x85)}
```

The idea is the `\verb` LaTeX command is going to be used on a string answer, and it needs a delimiter character. Character 0x85, ASCII 133, was chosen because it would be crazy for a student to have that as part of a string answer they "typed". 

Then in 539406c, the character changed to 0x1F, ASCII 31, the "unit separator character". This brought it down into 7 bit ASCII, and Geoff's comment in the commit suggests this has something to do with the utf8 conversion.

So now we have string answers that use character 0x1F in their display. This is causing an issue with PreTeXt. When WW processes a problem with "PTX" display mode, it makes XML. For each answer of the problem, it makes a single XML element, with lots of attributes and values that correspond to the Perl answer hash's keys and values. For an example, see:

 https://webwork-ptx.aimath.org/webwork2/html2xml?courseID=anonymous&amp;userID=anonymous&amp;password=anonymous&amp;course_password=anonymous&amp;answersSubmitted=0&amp;displayMode=PTX&amp;outputformat=ptx&amp;problemSeed=8435&amp;sourceFilePath=Library/PCC/BasicAlgebra/Geometry/CylinderVolume10.pg

and view source, since your web browser is likely to try to read the XML as HTML.

So you can see how a string like `\verb<0x1F>foo<0x1F>` could end up inside a value for an attribute of one of these XML elements. The problem is that XML does not allow this character in an attribute value. (Well, there are varying standards for what is allowed, but even when this one is allowed, its use is discouraged, and anyway, its presence causes the python validator we use to declare this to be invalid XML.)

So. We want a character that a student will not be able to type with normal use of the keyboard. So nothing in ASCII 32--127. And we want a character that is valid for XML in an attribute value. So nothing in ASCII 00--31. So we have to leave 7-bit ASCII to meet both conditions. Is it possible to do this? Can a character be chosen somewhere else in utf8 and that be compatible with the utf8 conversion happening now?

dpvc:
How about ASCII 127 (U+007F, DELETE), since the DELETE character is hard to get into a student's answer string (pressing delete causes an action, rather than inserts the character)?

[...]

PR #422, opened 2019-09-14
refactor verb to use CR for delimiter

Alex-Jordan:
For issue #421.

### DIFF
--- a/lib/Value/String.pm
+++ b/lib/Value/String.pm
@@ -80,7 +80,16 @@ sub compare {
 #
 #  Mark a string to be display verbatim
 #
-sub verb {shift; return "\\verb".chr(0x1F).(shift).chr(0x1F)}
+sub verb {
+  shift;
+  my $s = shift;
+  $s =~ s/\r/ /g;
+  return "{\\verb\r$s\r}";
+  # Note this does not handle \n in the input string
+  # A future effort to address that should concurrently
+  # handle it similarly for HTML output.
+  # And something similar should be done for the ArbitraryString context
+}
 
 #
 #  Put normal strings into \text{} and others into \verb


lib/Value/String.pm
```diff
@@ -80,7 +80,16 @@ sub compare {
 #
 #  Mark a string to be display verbatim
 #
-sub verb {shift; return "\\verb".chr(0x1F).(shift).chr(0x1F)}
+sub verb {
+  shift;
+  my $s = shift;
+  $s =~ 's/\r/ /g';
```
dpvc: I don't think the single quotes should be there.  Does that actually work to perform a substitution?

lib/Value/String.pm
```diff
@@ -80,7 +80,16 @@ sub compare {
...
+  return "{\\verb\r$s\r}" unless $s =~ m/\n/;
+  my @tex = split(/\n/,$s);
+  foreach (@tex) {$_ = $self->verb($_) if $_ =~ m/\S/}
```
dpvc: I'm not sure the recursion is really necessary in this case.  The removal of `\r` has already been done, so there is no need to redo that or to perform the setup and teardown for a function call, here.  I would think

```
  foreach (@tex) {$_ = "\\verb\r$s\r" if $_ =~ m/\S/}
```

would be sufficient (also no need for the braces in this case, as the only think following `\verb` will be either `\\` or `\end{array}`).

lib/Value/String.pm
Alex-Jordan: When I do command line perl substitutions, I use single quotes and it works. Like `perl -pi -e 's/\r/ /g' *.txt`. But I will test (and change to double if it's more standard here.)

lib/Value/String.pm
Alex-Jordan: OK, that tracks.

Alex-Jordan:
I'm confused about the points you make distinguishing TeX and HTML output here. I feel like every instance of when `\verb` would be used here in TeX output, it is also used the same way in HTML output, with MathJax applied to it all with math wrapping delimiters. Like in a `correct_ans_latex_string`. Are you saying (perhaps among other things) that something additional should be done for `correct_ans` strings?

lib/Value/String.pm
dpvc: The quotes here are shell characters that are not part of the actual perl code.  The line should be

    $s =~ s/\r/ /g;

with no quotation marks of any kind.

Alex-Jordan:
Alright, I think I understand better now after re-reading and thanks for the careful explanation. This has had some mission creep in trying to handle the `\n`, when the original mission was just to replace the `0x1F` with `\r`. I'll judge the time I have today, and I may just retreat back to only bringing in the `\r`, and maybe this can make it into 2.15.

dpvc:
> Are you saying (perhaps among other things) that something additional should be done for `correct_ans` strings?

Yes, this is one of the places that `quoteHTML` is used, and also for `student_ans_strings`, and for some other output in parser radio buttons and popups.  The idea is that the TeX output and the HTML output should match as closely as possible, so if you are handling line breaks in TeX output, they should also be handled in HTML output so that correct and student answers that are shown as text (like in the Entered field, and in the pop-overs in the TeX output) show the same.

It isn't critical, but not doing it now means it will be more difficult to get it later, when we've forgotten what has happened.

Going back to just the `\r` handling is fine with me.

Alex-Jordan:
OK, I narrowed the scope of the commit to just doing the `\r` replacement. And rebased this from PG-2.15, and am now targeting PG-2.15 for the merge instead of develop.

Alex-Jordan:
Hmm, well, I put `\r` into use in our server to actually try it out with PreTeXt. Here is a simple problem with three string answer blanks. The first is a normal string. The second is a string with some XML control characters. The third has some LaTeX control character. Follow this link (and view source) to see the output for PreTeXt to take hold of:
view-source:https://webwork.pcc.edu/webwork2/html2xml?&amp;problemSeed=1&amp;answersSubmitted=0&amp;sourceFilePath=local/verbatim.pg&amp;displayMode=PTX&amp;courseID=orcca&amp;userID=orcca&amp;course_password=orcca&amp;outputformat=ptx

I'm finding that somewhere along the way, something is translating the `\r` to `\n`. The text in my web browser has `\n`. At least, it looks like a line break, and if I copy and past into a text editor and search for `\r`, I don't find it, but I do find `\n`. And web browser aside, when PreTeXt uses the python requests module to get this content, it is writing a file to my disk with `\n`, not `\r`.

Somewhere I read that lots of things automatically convert a`\r` to `\n` in transfer. Does anyone have insight into what stage might be doing this, and if I can prevent it? Or, since there are long term plans to address `\n` within the string itself, should I try to accept having `\n` as the effective `\verb` delimiter?

Alex-Jordan:
Did anyone see my last note about `\r` transforming to `\n`? I'm not sure this should have been merged until that mystery is looked into more. If you grab the content from the link I gave, do you get a file with `\r` or with `\n` in the verbatim delimiter position? If I `wget` that URL, the file I get has `\r` where `\n` should be. This could impact what PreTeXt does with this output, which was the reason behind changing away from character 1F in the first place.

dpvc:
I saw it, so was a bit surprised to see it merged.  I'm wondering if you could use something like `&#xD;` within the PreTeXt attributes rather than raw carriage returns?

taniwallach:
I possibly did not pay enough attention to https://github.com/openwebwork/pg/pull/422#issuecomment-532057271 but I do not think there is an issue on the WeBWorK side of things. From what I can see, WeBWorK is generating the `\r` we expect it to generate, and it seems to me that the issue is happening somewhere farther on in the chain of processing.

@Alex-Jordan Maybe check what the python requests module gives you if you access the "raw" byte `content` of the response instead of the `text` of the response.

Note: After some reading, it seems that Python 3 has "universal newline" handling which intentionally changes newlines, ex when reading a file. Have a look at:
  - https://www.python.org/dev/peps/pep-3116/
  - https://docs.python.org/release/3.2/library/functions.html#open
  - https://www.enigma.com/blog/the-secret-world-of-newline-characters
  - https://2.python-requests.org//en/master/api/?highlight=encod#requests.Response
  - https://www.edureka.co/blog/python-requests-tutorial/#z4
  - https://github.com/psf/requests/issues/1604

If necessary, I can revert the merge. if it does not serve the purpose for which it was intended.

---

When I inspect the HTTP response (using Chrome developer tools) and **save the response to a file**, I see carriage-returns (0x0D). However, if I try to **cut and paste it** becomes new-lines (0x0A). 

The file obtained from developer mode - save to file is:
[tst1.txt](https://github.com/openwebwork/pg/files/3623464/tst1.txt)
and has `md5sum` :  `4b7374a83936dea9e6a6d38003486e5c` 

Relevant parts from the output of `od -a -t x1 tst1.txt `:
```
0002060   l   a   t   e   x   _   s   t   r   i   n   g   =   "   {   \
         6c  61  74  65  78  5f  73  74  72  69  6e  67  3d  22  7b  5c
0002100   v   e   r   b  cr   h   a   s  sp   X   M   L  sp   s   p   e
         76  65  72  62  0d  68  61  73  20  58  4d  4c  20  73  70  65
0002120   c   i   a   l  sp   c   h   a   r   a   c   t   e   r   s  sp
         63  69  61  6c  20  63  68  61  72  61  63  74  65  72  73  20
0002140   &   q   u   o   t   ;   &   a   m   p   ;   '   ;   &   l   t
         26  71  75  6f  74  3b  26  61  6d  70  3b  27  3b  26  6c  74
0002160   ;   &   g   t   ;  cr   "  sp   c   o   r   r   e   c   t   _
         3b  26  67  74  3b  0d  22  20  63  6f  72  72  65  63  74  5f



0003260   l   a   t   e   x   _   s   t   r   i   n   g   =   "   {   \
         6c  61  74  65  78  5f  73  74  72  69  6e  67  3d  22  7b  5c
0003300   v   e   r   b  cr   h   a   s  sp   L   a   T   e   X  sp   s
         76  65  72  62  0d  68  61  73  20  4c  61  54  65  58  20  73
0003320   p   e   c   i   a   l  sp   c   h   a   r   a   c   t   e   r
         70  65  63  69  61  6c  20  63  68  61  72  61  63  74  65  72
0003340   s  sp   #   %   &   a   m   p   ;   \   ^   _   ~   |  cr   "
         73  20  23  25  26  61  6d  70  3b  5c  5e  5f  7e  7c  0d  22
```

Using `wget` on the URL you sent (inside double quotes), I also obtained the same file where the character just after the `\verb` is a carriage-return (0x0D).

---

Hardcopy `.tex` files are also being generated using carriage-returns (0x0D) and compile. Here is an excerpt from the `hardcopy.tex` file:
[from-hardcopy.txt](https://github.com/openwebwork/pg/files/3623497/from-hardcopy.txt)
```
tani@lxtani:/tmp$ od -a -t x1  from-hardcopy.txt 
0000000   {   \   v   e   r   b  cr   h   e   l   ^   l   o  cr   }  sp
         7b  5c  76  65  72  62  0d  68  65  6c  5e  6c  6f  0d  7d  20
0000020  nl
         0a
0000021
```

taniwallach:
@Alex-Jordan - Another issue, with the generated PreTexT output from the URL: 

Looking at the data generated I noticed that the closing `}` after the `\verb` object and before the double quote ending the string in the data file is missing.

However, in generated LaTeX files the closing `}` seems to be where it should be.

taniwallach:
@Alex-Jordan Search for "newline can be None" in https://www.python.org/dev/peps/pep-3116/.

If the issue is caused by Python's "universal newlines mode" then you would want to set `newline=''` to keep incoming newlines unchanged.
> If it is '', universal newline mode is enabled, but line endings are returned to the caller untranslated.

Alex-Jordan:
Thanks @taniwallach , with the help you gave here, I should be able to figure this out. Or get help from the some Python experts. Some notes:

1. On my server, I had manually typed the change for a quick experiment and omitted that brace. Fortunately that omission was not in the PR.
2. I tried changing the `\r` to `&#xD;`, but doing that, I end up with `&amp;#xD;`. It seems that something escapes `&` not paying attention to whether it is already being used as an escape. Actually now that I look closer in the HTML from my link earlier in the thread, I see things like `&amp;quot;`, `&amp;amp;`, `&amp;lt;`, and `&amp;gt;` (in places that PreTeXt currently makes no use of). So there may be some overzealous escaping by WW.
3. When I `wget` the content, I still end up with `\n` in the file, not `\r`. Are you on a flavor of Linux? I'm on a Mac, and I see in one of your links that "`\r` support is still needed for some OSX applications..."
4. At present, the PreTeXt toolkit aims to behave for both python 2.7 and 3. I'd only been using 2.7, but I get the same results with 3. At some point PTX may move to reliance on 3. 

I'm looking into it more, and I will probably call upon the PTX Python gurus to help.

Alex-Jordan:
> When I `wget` the content, I still end up with `\n` in the file, not `\r`.

I take it back. I'm not sure what I used to look for `\r` before. But now I see it in the file I "wgot" the other day. So that's a relief.

taniwallach:
> 1. On my server, I had manually typed the change for a quick experiment and omitted that brace. Fortunately that omission was not in the PR.

That is a big relief - as I was worried we would need to hunt for a but if code somewhere "swallowing" the closing brace.

> Thanks @taniwallach , with the help you gave here, I should be able to figure this out. Or get help from the some Python experts. Some notes:

:smile:  Sorry, but my Python knowledge is very close to zero, but I am pretty certain that a Python side solution to avoid the conversion will be found. Hopefully not one which is too tedious or ugly. If it is, then the Python people should be made aware of the need to allow bypassing that sort of "kindly help" when necessary.

> 2. I tried changing the `\r` to `&#xD;`, but doing that, I end up with `&amp;#xD;`. It seems that something escapes `&` not paying attention to whether it is already being used as an escape. Actually now that I look closer in the HTML from my link earlier in the thread, I see things like `&amp;quot;`, `&amp;amp;`, `&amp;lt;`, and `&amp;gt;` (in places that PreTeXt currently makes no use of). So there may be some overzealous escaping by WW.

Yes - probably somewhere... HTML does not like certain unescaped characters, and we may need to determine if there is a need to avoid that in special settings. That is probably a bit of work.

> 3. When I `wget` the content, I still end up with `\n` in the file, not `\r`. Are you on a flavor of Linux? I'm on a Mac, and I see in one of your links that "`\r` support is still needed for some OSX applications..."

> I take it back. I'm not sure what I used to look for \r before. But now I see it in the file I "wgot" the other day. So that's a relief.

Good, because if basic Unix tools are mangling data - that would be pretty evil. 

> 4. At present, the PreTeXt toolkit aims to behave for both python 2.7 and 3. I'd only been using 2.7, but I get the same results with 3. At some point PTX may move to reliance on 3.
> 
> I'm looking into it more, and I will probably call upon the PTX Python gurus to help.

I do recommend asking the Python gurus for the help... Bypassing the "universal newline" replacement code seems to require getting a control setting far enough down in the i/o or string parsing code.

Alex-Jordan:
The good news is that I think I have the PreTeXt side under control.

The bad news is that MathJax is not liking the `\r` in regular WeBWorK HTML. Or maybe, it does not like an `\n` that is getting swapped in place of `\r`. If you go to the same question I linked to before, but view with MathJax mode instead of PTX, and click to show the answers...

https://webwork.pcc.edu/webwork2/html2xml?&amp;problemSeed=1&amp;answersSubmitted=0&amp;sourceFilePath=local/verbatim.pg&amp;displayMode=MathJax&amp;courseID=orcca&amp;userID=orcca&amp;course_password=orcca&amp;outputformat=simple

... for me in Firefox I see a MathJax processing error.

taniwallach:
@dvpc - Your advice would be greatly appreciated.

Using `'\r` as the delimiter now seems to work well for @Alex-Jordan's PreTeXt side, and when I tested it in the generated TeX output things are also fine. Rather than try to "fix" MathJax or browser code to handle `\r` well, maybe we can select the delimiter for screen output differently, and then use something suitable in the screen context which would not have been acceptable in the TeX and PreTeXt contexts where the 8-bit limit, UTF-8 rules about the contents of bytes, and various other considerations led to the very restricted list of options.

Would it be reasonably possible to have TeX for PreTeXt and real TeX files use `\r` as the verbatim delimiter, and a different (Unicode) character fill this role for HTML + MathJax output?

Could a "global" PG variable be used to select the verbatim delimiter and set appropriately early on based on the "real" output mode, before the code actually got to the call to `Tex()`?

taniwallach:
Note: HTML treats `\r` specially - so it seems that using it for the verbatim delimiter inside HTML output is not a good choice.
  - https://www.w3.org/TR/html401/struct/text.html#h-9.3.2
    - HTML 4.01: as white-space and as a line break:
  - https://html.spec.whatwg.org/multipage/parsing.html#the-input-byte-stream
    - special treatment in the parsing process

 `\n` would not be any better (same issue) and @Alex-Jordan in https://github.com/openwebwork/pg/issues/421 reported that the third option in ASCII 1-32 which was under consideration was `\t` and TeX did not like `\t` as the verbatim delimiter.

taniwallach:
I was not able to convince CodePen to put a CR where I wanted one...
  - https://codepen.io/taniwallach/pen/WNeYzwJ
Attached is a small HTML test file:    [WNeYzwJ.html.txt](https://github.com/openwebwork/pg/files/3635541/WNeYzwJ.html.txt)

dpvc:
Using one delimiter for hardcopy and PTX and another for HTML seems like the way to go, here.  Something like

```
sub verb {
  shift;
  my $s = shift;
  $s =~ s/\r/ /g;
  $d = main::MODES(HTML => char(0x1F), TeX => char(0xD), PTX=> char(0xD));
  return "{\\verb$d$s$d}";
  # Note this does not handle \n in the input string
  # A future effort to address that should concurrently
  # handle it similarly for HTML output.
  # And something similar should be done for the ArbitraryString context
}
```

should do it.

Alex-Jordan:
The function used prior to all this was `chr`. Should the `char` in the line defining `$d` be `chr`?

dpvc:
Yes, sorry.  Wrong language.  OOPS!

Alex-Jordan:
I figured, you are too fluent :)  OK, I was only able to get this to work after making `$d` local and parroting something else in the same file, passing `eval` a string. Does it make sense that this is needed?

```
sub verb {
  shift;
  my $s = shift;
  $s =~ s/\r/ /g;
  my $d = eval ('main::MODES(HTML => chr(0x1F), TeX => chr(0xD), PTX=> chr(0xD))');
  return "{\\verb$d$s$d}";
  # Note this does not handle \n in the input string
  # A future effort to address that should concurrently
  # handle it similarly for HTML output.
  # And something similar should be done for the ArbitraryString context
}
```

dpvc:
Yes, it should have been a `my` variable (again, thinking about the wrong language).

And yes, the `eval()` is needed because this file is a `.pm` file that is loaded outside the safe compartment, and `eval()` gets that to run inside the safe compartment, where the context of the problem is in place.

taniwallach:
@dpvc - Thanks for all the help, and the explanation of why the `eval` is needed. I would not have known that.
@Alex-Jordan - Can you please make a new PR with the patch, targeting the PG-2.15 branch. Please reference it as a followup to patch this one, and mention the current state of testing you have done. I'll try to review and merge quickly. Could you also leave a comment in the code about why the `eval` is needed - so others who one day look at it will also know why.

taniwallach:
The patch in https://github.com/openwebwork/pg/pull/424 was merged in, so I think this change should now be fully completed.

PR #423, opened 2019-09-17
Disable autocorrect, autocapitalize, autocomplete, and spellcheck on text inputs

drgrice1:
This was requested in issue #996 in the webwork2 repository.
I added a little general clean up in the area of the modified code!

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -404,18 +404,17 @@ sub NAMED_ANS_RULE {
         }
 
         # end of addition for dragmath
-	
+
 	MODES(
 		TeX => "\\mbox{\\parbox[t]{${tcol}ex}{\\hrulefill}}",
-		Latex2HTML => qq!\\begin{rawhtml}<INPUT TYPE=TEXT SIZE=$col NAME=\"$name\" VALUE = \"\">\\end{rawhtml}!,
+		Latex2HTML => qq!\\begin{rawhtml}<input type=text size=$col name="$name" value="">\\end{rawhtml}!,
 
-	    # Note: codeshard is used in the css to identify input elements 
-	    # that come from pg
-		HTML => qq!<input type=text class="codeshard" size=$col name="$name" id="$name" aria-label="$label" dir="auto" value="$answer_value"/>\n!.
-		              $add_html. # added for dragmath
-                        qq!<input type=hidden  name="$previous_name" value="$answer_value"/>\n!,
-        PTX => '<fillin name="'."$name".'" characters="'."$col".'" />',
-		
+		# Note: codeshard is used in the css to identify input elements that come from pg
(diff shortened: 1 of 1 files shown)


mgage:
This clean up looks good to me.  We are no longer supporting dragMath but we can perform that cleanup later. I suspect that there are other locations that need cleanup for that. 

I did not run the code.

drgrice1:
I tried not to be too ambitious with code clean up.

It wouldn't take much to remove dragMath from the code.  This is the only place in the pg code that dragMath appears.  The only places it appears in the webwork2 code are in defaults.config, localOverrides.conf.dist, the French translation file (fr.po), and the dragmath.js file itself.

taniwallach:
Does the textarea in `NAMED_ANS_BOX` in this file need something similar?

What about textarea objects elsewhere, in particular in:
  - `macros/quickMatrixEntry.pl`
  - `macros/answerDiscussion.pl`
  - `macros/PGessaymacros.pl`

drgrice1:
I can add the html flags to the textarea's also, if you think that is appropriate.  It depends on the usage and philosophy.  To me it seems that autocorrect and spellcheck could be useful for a textarea. Particularly for an essay type answers.  Although that may depend on what the content of the essay is.

mgage:
Textarea might contain mathjax Tex.  Could spellcheckers handle that.


sent from my phone, Mike

drgrice1:
I did some basic testing in the essay question textarea both on a computer with a physical keyboard and on a device with a soft keyboard.  These things are not nearly intrusive in a textarea as in a text input. Spell check does not interfere with typing things enclosed in \( and \). Auto capitalization doesn't mess with that either. It does capitalize the first word in a sentence. For an easy type question I think these features could actually be a good thing. I am not sure about the matrix entry macro. That macro is not used in any problems in the open problem library. There is a test problem in the pg repository, but it doesn't work. 

The behavior will probably vary depending on the browser and soft keyboard implementation. If you want to play it safe I can disable these features for textarea's. Let me know what you want.

drgrice1:
Michael:  The reason that the quickMatrixEntry.pl test problem is failing is because of an unresolved merge conflict.  It is commit fe741695 that occurred on January 8 of this year, when you pulled in an updated tableau.pl file.

mgage:
Thanks Glenn.  I'll fix it when I get back from this conference (which will be tomorrow).  I also have some examples, beyond the test examples.  The tableau.pl macros themselves are on my site and are nearly ready to be polished and added but I didn't think I had time to do it properly for webwork-2.15.  It's from teaching linear optimization and simplex method a year ago and I'm not teaching it now -- so the immediate urgency for me dropped off a bit. :-)  

Thanks for finding the source of the bug.

mgage:
I submitted a pull request to fix the tableau.pl file separately.  It doesn't affect the changes in this pull request.

PR #424, opened 2019-09-21
differentiate verbatim delimiter selection

Alex-Jordan:
patches #422, after we found that using `\r` as the verbatim delimiter all the time was not workable given how browsers treat `\r`.


Alex-Jordan:
For testing, I made a problem with string answers, with and without certain dangerous characters.

HTML: the answers render as expected in the results table.
PDF: the problem is fine in WW hardcopy, showing the answers well if I request the PDF show correct answers.
PTX: it comes out fine, and PTX is able to process it with the `\r`.

PR #425, opened 2019-09-25
Fix errors caused by bad merge of tableau.pl

mgage:
Hopefully this fixes a garbled tableau.pl file (resulting from a failed merge) that Glenn Rice caught.

### DIFF
--- a/macros/tableau.pl
+++ b/macros/tableau.pl
@@ -356,27 +356,22 @@ sub phase1_solve {
 	[complementary_basis_set] = $self->primal_basis_to_dual(primal_basis_set)
 	[complementary_basis_set] = $self->dual_basis_to_primal(dual_basis_set)
 
-<<<<<<< HEAD
+
 ########################
 ##############
 # get_tableau_variable_values
 #
-
+# deprecated for tableaus - use $tableau->statevars instead
+#
 # Calculates the values of the basis variables of the tableau, 
 # assuming the parameter variables are 0.
 #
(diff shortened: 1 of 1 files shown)


macros/tableau.pl
```diff
@@ -356,27 +356,22 @@ sub phase1_solve {
 	[complementary_basis_set] = $self->primal_basis_to_dual(primal_basis_set)
 	[complementary_basis_set] = $self->dual_basis_to_primal(dual_basis_set)
```
drgrice1: I think there needs to be a "=cut" line here.

mgage: yep -- thanks.

drgrice1:
It is probably good to get this file back in working order.  The tableau_javascript.pg test problem works again with this.  There seem to be problems with the javascript in the quickMatrixEntry.pl file that causes "undefined" to appear in the textarea of the dialog for each row of the matrix.

It may not be worth the attention at this point.  None of the files in the open problem library use tableau.pl or quickMatrixEntry.pl.

Of course, there may be others using these files in problems not in the library.  So this fixed version should be pull in in any case.

mgage:
I agree the quickMatrixEntry javascript could use more work.  In addition to cosmetic problems I would like the quickMatrixEntry to hand a variety of matrix input styles.  As it is it handles matrices represented in rows with space separation between entries and a \n to indicate the end of the row. This is the format that websim uses.  I would like it to handle MatLab formats and Mathematica formats and perhaps other formats as well.

Issue #416, opened 2019-08-18
Accessibility: proposal to consider changing the HTML output of $BBOLD be <strong>


taniwallach:
As far as I understand, WCAG 2.0 accessibility compliance recommends/requires using the HTML `<strong>` tag to generate "strong emphasis" by a screen reader when necessary, while the older `<b>` tag for bold does not provide any screen-reader hint about emphasis.

Although it is not 100% true, much author use of **bold** will be to emphasize visually with an implicit "semantic" meaning. 

As such, I would like to recommend considering whether the existing  `$BBOLD` (and the corresponding `$EBOLD`) be changed to use `<strong>` instead of `<b>` by default.

Alternately, we could add `$BSTRONG` and `$ESTRONG` and recommend that problem authors make changes on their own.

If HTML `strong` does become the default interpretation of `$BBOLD` we could add a  `$BJUSTBOLD` or the like for when plain bold (`<B>`) is the "right thing" in HTML output.

To some extent, the same question may apply to `$BITALIC` giving HTML `<em>` instead of `<i>`, but I suspect that much "mathematical use" of italics is not for emphasis but to set of special terms, so that change might be less clear than that for bold.

@Alex-Jordan - you seem to be the local accessibility expert or have access to accessibility experts - what do you/they say?

---

References:
  - https://accessibility.psu.edu/boldfacehtml/
  - https://www.freecodecamp.org/forum/t/bold-and-strong-text/178376/3
  - https://stackoverflow.com/questions/2243074/is-it-ok-to-use-strong-in-place-of-b-blindly
  - https://stackoverflow.com/questions/2243074/is-it-ok-to-use-strong-in-place-of-b-blindly
  - https://html.spec.whatwg.org/#the-b-element
  - https://thatdevgirl.com/blog/bold-vs-strong
  - https://www.brucelawson.co.uk/2018/screenreader-support-for-text-level-semantics/

dlglin:
It’s worth noting that these constructions are only used in text mode.  Any typeset math would use standard LaTeX constructions to obtain bold or italics.  This further supports Tani’s observation that $BBOLD is primarily used to emphasize visually.  I think that this is a reasonable change.  For any edge cases where bold text is serving some other purpose, we could handle these as they are reported.

On Aug 18, 2019, at 3:52 AM, Nathan Wallach <notifications@github.com<mailto:notifications@github.com>> wrote:


[...]

PR #426, opened 2019-09-25
HTML output to use <STRONG> instead of <B> for BBOLD

taniwallach:
Replace `<B>` by `<STRONG>` as HTML output of `$BBOLD` to make HTML output more accessible to users who depend on a screen reader. Matching change to `$EBOLD`.

TeX output remains plain bold (PDF output from TeX has "failures" to meet the WCAG 2.0 accessibility standards.)

See discussion at: https://github.com/openwebwork/pg/issues/416 where @jwj61 and @dlglin approved the change in the default behavior.


taniwallach:
> Shouldn't `<i>` also be changed to `<em>` for the same reason?

I don't think so, as the typical use of italics in mathematical text is not for "emphasis", but to give a visual hint about "special terms" often those being defined, as I noted in the discussion at https://github.com/openwebwork/pg/issues/416. This differs from the typical use of italics in regular prose, where much (but not all) use may indeed be for "emphasis".

For better or worse, I think that a global change from `<i>` to `<em>`  would be semantically incorrect in many if not most cases. Maybe we need to find a better HTML tag to use for many cases, and create a new begin/end pair for such uses. If and when a decision is made on how best to handle this, in can certainly be done. Implementing it fully in WW/PG/OPL would require editing many PG files in the OPL, and that requires effort on a large scale.

drgrice1:
I see.  That makes sense.

PR #427, opened 2019-09-25
Add $BKBD and $EKBD

taniwallach:
Add `$BKBD` and `$EKBD` to create a simple "verbatim" type unit using the HTML `<KBD>` tag for HTML output. Browsers will usually render the output in a monospace font. For TeX and PreTeXt output, the code will use `\verb`.

No effort was made to allow the use of special characters inside the unit of content, so it should be used with care.

**Note** some cleanup was done to tab/space based layout in `macros/PGbasicmacros.pl` which was primarily modified to add the new macros/variables.

---

Sample test file:
```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "PGcourse.pl"
);

TEXT(beginproblem());

BEGIN_TEXT
${BKBD}This should be in a monospace font.${EKBD}
$PAR
This is in the regular font.
END_TEXT

ENDDOCUMENT();
```

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -79,6 +79,8 @@ my ($PAR,
 	$ECENTER,
 	$BLTR,
 	$ELTR,
+	$BKBD,
+	$EKBD,
 	$HR,
 	$LBRACE,
 	$RBRACE,
@@ -151,6 +153,8 @@ main::PG_restricted_eval( <<'EndOfFile');
 	$main::ECENTER          = ECENTER();
 	$main::BLTR          = BLTR();
 	$main::ELTR          = ELTR();
+	$main::BKBD                     = BKBD();
+	$main::EKBD                     = EKBD();
 	$main::HR				= HR();
 	$main::LBRACE			= LBRACE();
 	$main::RBRACE			= RBRACE();
@@ -209,6 +213,8 @@ EndOfFile
 	$ECENTER             = ECENTER();
 	$BLTR             = BLTR();
 	$ELTR             = ELTR();
+	$BKBD                           = BKBD();
+	$EKBD                           = EKBD();
 	$HR				     = HR();
 	$LBRACE			     = LBRACE();
 	$RBRACE			     = RBRACE();
@@ -1590,6 +1596,8 @@ sub MODES {
 	$ECENTER    		ECENTER()  			end centered environment
 	$BLTR    		BLTR()   		begin left to right environment
 	$ELTR    		ELTR()  			end left to right environment
+	$BKBD                   BKBD()                  begin "keyboard" input text
+	$EKBD                   EKBD()                  end "keyboard" input text
 	$HR					HR()				horizontal rule
 	$LBRACE				LBRACE()			left brace
 	$LB					LB ()				left brace
@@ -1666,6 +1674,8 @@ sub BCENTER { MODES(TeX => '\\begin{center} ',  Latex2HTML => ' \\begin{rawhtml}
 sub ECENTER { MODES(TeX => '\\end{center} ',  Latex2HTML => ' \\begin{rawhtml} </div> \\end{rawhtml} ', HTML => '</div>', PTX => ''); };
 sub BLTR { MODES(TeX => ' ',  Latex2HTML => ' \\begin{rawhtml} <div dir="ltr"> \\end{rawhtml} ', HTML => '<span dir="ltr">', PTX => ''); };
 sub ELTR { MODES(TeX => ' ',  Latex2HTML => ' \\begin{rawhtml} </div> \\end{rawhtml} ', HTML => '</span>', PTX => ''); };
+sub BKBD { MODES(TeX => '\\texttt{' , Latex2HTML => '', HTML => '<KBD>', PTX => ''); };
+sub EKBD { MODES(TeX => '}', Latex2HTML => '', HTML => '</KBD>', PTX => ''); };
 sub HR { MODES(TeX => '\\par\\hrulefill\\par ', Latex2HTML => '\\begin{rawhtml} <HR> \\end{rawhtml}', HTML =>  '<HR>', PTX => ''); };
 sub LBRACE { MODES( TeX => '\{', Latex2HTML =>   '\\lbrace',  HTML =>  '{' , HTML_tth=> '\\lbrace', PTX => '{' ); };  #not for use in math mode
 sub RBRACE { MODES( TeX => '\}', Latex2HTML =>   '\\rbrace',  HTML =>  '}' , HTML_tth=> '\\rbrace', PTX => '}' ); };  #not for use in math mode
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -18,7 +18,7 @@
 
 =head1 NAME
 
-	  PGbasicmacros.pl --- located in the courseScripts directory
+	PGbasicmacros.pl --- located in the courseScripts directory
 
 =head1 SYNPOSIS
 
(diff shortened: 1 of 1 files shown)


taniwallach:
The intended use is when explaining what to type as input.

References about the HTML `<KBD>` tag.
  - https://www.w3.org/TR/html50/text-level-semantics.html#the-kbd-element
  - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd
  - https://www.w3docs.com/learn-html/html-kbd-tag.html
  - https://www.w3schools.com/tags/tag_kbd.asp

dpvc:
I think `\verb` is the wrong thing to use in the TeX output.  I would recommend `\texttt` instead.  Note that `<kbd>` content can be long, and allows word wrapping, whereas `\verb` does not wrap (and can't contain newlines).  Also, multiple spaces in `<kdb>` would be compressed to a single space, and leaning and trailing spaces could be removed (if there are already spaces before or after it), but that would not be the case for `\verb`, so the spacing would be inconsistent between the two formats.  Finally, the use of `\verb` would complicate the use of other macros within the `$BKDB` and `$EKBD`; e.g., if you did `${BKBD}x${US}0${EKBD}` as you would expect to to to get `x_0` as the output, you would get that for HTML, but `x\_0` in hardcopy.  This is because `\verb` will treat its contents as verbatim, while the macros like `$US` do quoting of their contents so they can be used in non-verbatim situations.  This could be a source of confusion.

So I recommend `\texttt` rather than `\verb` be used.

taniwallach:
@dpvc - Accepted - I will make the change to use `\texttt` in a commit "soon".

@Alex-Jordan  - How would you like this handled in PreTeXt?

Alex-Jordan:
> @Alex-Jordan - How would you like this handled in PreTeXt?

PreTeXt tags are (intended to be) semantic and not for styling. So it's hard to answer without understanding the semantic purpose of `$BKBD`. If it is really just for making monospace font, there is no PTX counterpart for that, and it should just be `PTX => ''`. 

Otherwise, PTX has a `<pre>` tag which is for block level verbatim. (Yes, I know that is inconsistent with HTML `pre`, but it was built into PreTeXt very early before the finer points of HTML `pre` were understood.) And there is a `<c>` tag for inline verbatim. Are either of those appropriate for the intent of `$BKBD`?

taniwallach:
> I like that you tried to clean up the spacing of things here, but I don't agree with the way that you have done it. In short look at http://webwork.maa.org/wiki/Coding_Standards. What you have done does not adhere to those standards. I realize that page is not adhered to much in the WeBWorK code, but it should be, and that page should be made more complete.

Point taken. Will be addressed in a followup commit.

taniwallach:
> > @Alex-Jordan - How would you like this handled in PreTeXt?
> 
> PreTeXt tags are (intended to be) semantic and not for styling. So it's hard to answer without understanding the semantic purpose of `$BKBD`. If it is really just for making monospace font, there is no PTX counterpart for that, and it should just be `PTX => ''`.
> 
> Otherwise, PTX has a `<pre>` tag which is for block level verbatim. (Yes, I know that is inconsistent with HTML `pre`, but it was built into PreTeXt very early before the finer points of HTML `pre` were understood.) And there is a `<c>` tag for inline verbatim. Are either of those appropriate for the intent of `$BKBD`?

@Alex-Jordan :
  - HTML's `<kbd>` is meant to semantically indicate "user input" text, and many print formats traditionally used a mono-space font, and the browsers seem to have picked that up.
  - The intent is to have a nicer way of displaying samples of "input" the student may be expected to type, primarily when there is a desire to remind them of syntax issues. Until now, I have abused "verbatim" inside math mode to handle such text, but that is a very ugly approach. Does that fit with the intended use of PreTeXt's `<c>` tag? 
  - (I'm assuming mostly inline use, but people will end up using the feature however they want to.)

Alex-Jordan:
Probably PreTeXt should get a new tag for keyboard user input. 

PreTeXt verbatim options are probably not a good idea. Because something like `${BKBD}${NDASH} ${EKBD}` would be translated to `<c><ndash/></c>` and and it is against PTX schema to have tags inside `<c>`. That's a silly example, but you get the point.

So I say use `PTX => ''` and something maybe can be implemented later.

drgrice1:
Wow.  Nice work.  This is now probably the cleanest file in the entire webwork code base!

taniwallach:
The PR was revised to have 2 commits, and I recommend reviewers examine the commits separately.
  - The first commit is just a the code for `$BKBD` and `$EKBD` cleaned up to use `\texttt` in LaTeX and nothing for PreTeXt. This handles the comment by @dpvc and does the best possible at present for PreTeXt based on @Alex-Jordan's response above.
  - The second commit is essentially just formatting/whitespace changes, throught the file, to try to make the file reasonably compliant with http://webwork.maa.org/wiki/Coding_Standards. (@drgrice1 -  Thanks for the reminder.)
    - Wherever I could without major effort, initial indentation is using tabs (followed by spaces when lining up with text after non-tabs) and alignment further into a line is via spaces and not tabs.
  - Some minor non white-space code cleanup changes, with no functional change. The output of  `git diff -b --ignore-blank-lines c49274e0dd61d4f75660b574cc6dc9f825eb124b 809a49ca67f7ab116b73aad2ebe73d1857473ad0` follows:

```
diff --git a/macros/PGbasicmacros.pl b/macros/PGbasicmacros.pl
index 9a93658..3879e2a 100644
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2128,7 +2130,14 @@ sub ev_substring {
 	my $actionRef   = shift;
 	my ($eval_out,$PG_eval_errors,$PG_full_error_report)=();
     my $out = "";
-	while ($string ne "") { #  DPVC -- 2001/12/07 - original "while ($string)" fails to process the string "0" correctly
+                #
+                #  DPVC -- 2001/12/07
+                #     original "while ($string)" fails to process the string "0" correctly
+                #
+		while ($string ne "") {
+                #
+                #  end DPVC
+                #
 		    if ($string =~ /\Q$start_delim\E/s) {
 		   #print "$start_delim $end_delim evaluating_substring=$string<BR>";
 				$string =~ s/^(.*?)\Q$start_delim\E//s;  # get string up to next \{ ---treats string as a single line, ignoring returns
@@ -2141,7 +2150,8 @@ sub ev_substring {
 				$out = $out . $eval_out;
 		   #print "$start_delim $end_delim new substring_out=$out<BR><p><BR>";
 				$out .="$PAR ERROR $0 in ev_substring, PGbasicmacros.pl:$PAR <PRE>  $@ </PRE>$PAR" if $@;
-		} else {
+				}
+			else {
 				$out .= $string;  # flush the last part of the string
 				last;
 				}
@@ -2664,7 +2673,7 @@ sub nicestring {
 	if($coefs[$j] != 0) {
 	    if($coefs[$j] == 1) {
 		$ans .= ($others[$j]) ? "+ $others[$j]" : "+ 1";
-		} elsif ($coefs[$j] == -1) {
+	    } elsif($coefs[$j] == -1) {
 		$ans .= ($others[$j]) ? "- $others[$j]" : "-1";
 	    } else {
 		$ans .= "+ $coefs[$j] $others[$j]";

```

taniwallach:
> Wow. Nice work. This is now probably the cleanest file in the entire webwork code base!

Thanks. Lots of careful work with `vim` search and replace using `control-v` to allow replacing tabs, and searching for tabs preceded or followed by space characters... Maybe we can eventually come up with a "playbook" for how to script such changes mostly automatically, maybe in emacs where changes can be OK-ed one at a time.

drgrice1:
vim can ok the changes one at a time also.  Use s/.../.../gc
The c flag means confirm.
vim is the editor I use.

taniwallach:
@drgrice1 @Alex-Jordan @dpvc - Thanks for all your advice and help. Merging the final product.

PR #428, opened 2019-09-25
Add an ADD_CSS_FILE() method

taniwallach:
The pg side of an attempt to address the need to modify CSS code for right to left languages in a convenient manner. but will provide a mechanism to request a special CSS file at the problem level.

The webwork2 side of the code also provides a mechanism to add CSS files via `course.conf` (or even via `conf/localOverrides.conf`).

See: http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4624

---

Add an `ADD_CSS_FILE()` method, to request the inclusion of additional CSS files from `webwork2/htdocs/css/` in the HTML generate for this problem.

The `webwork2` side code to support this is added inhttps://github.com/openwebwork/webwork2/pull/1018.

---

Sample PG file:

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
  "parserRadioButtons.pl",
  "PGcourse.pl"
);

TEXT(beginproblem());

ADD_CSS_FILE("rtl.css");
ADD_CSS_FILE("rtl2.css");

SET_PROBLEM_TEXTDIRECTION("rtl");
SET_PROBLEM_LANGUAGE("he-IL");

Context("Numeric");

$radio1 = RadioButtons(
  [  {"1"=>"1"}, 
     {"2"=>"2"},
     {"3"=>"3"},
     {"4"=>"4"},
  ],
   "2" # correct answer
);

BEGIN_TEXT
\{ $radio1->buttons() \}
END_TEXT

ANS( $radio1->cmp() );

ENDDOCUMENT();
```

---

The following sample CSS file (expected at `webwork2/htdocs/css/rtl.css`) has code to move the radio buttons from the left side of the text of the option to the right side (where it would usually be wanted in a right-to-left language).

```
input[type="radio"], input[type="checkbox"] {
  float: right !important;
}
```

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -193,6 +193,16 @@ sub SET_PROBLEM_TEXTDIRECTION {
   }
 }
 
+# Request that the problem HTML page also include additional CSS files
+# from the webwork2/htdocs/css/ directory.
+sub ADD_CSS_FILE {
+  my $file = shift ;
+  if ( !defined( $PG->{flags}{extra_css_files} ) ) {
+    $PG->{flags}{extra_css_files} = [ "$file" ];
+  } else {
+    push( @{$PG->{flags}{extra_css_files}}, $file );
+  }
+}
 
 sub AskSage {
     my $python = shift;


dpvc:
What is the output going to be like in hardcopy?  (I admit I'm not up on what has been done for RTL output.)  Will the checkboxes be in the right place?  If not, are there plans to make that possible?

I've always been hesitant to allow direct access to formatting like CSS in the problems, as it is very difficult to mirror that in hardcopy (and potentially other output formats).  There is also the potential issue of CSS clashes when multiple problems are shown (e.g., in the library browser, and (I think) gateway quizzes).  One of the purposes of the various macros like `$BBOLD`/`$EBOLD` where to provide a layer of abstraction from the actual layout mechanisms so that they can be adapted to other output formats, and perhaps new formats that we don't even have yet.  This has proven to be a valuable practice in the past, and I'm a bit leary of having problems go directly to CSS, as that makes them less future-proof and more fragile.

For example, in your case couldn't `SET_PROBLEM_TEXTDIRECTION()` be used to insert the needed CSS to make this happen, and whatever is needed in TeX to make the appearance correct?

taniwallach:
Thanks for the feedback @dvpc.

The TeX output is already putting the bullets to indicate radio buttons where they belong.

The layout location of the radio buttons (and checkboxes, if I recall) in the HTML are controlled by formatting CSS code in `bootstrap.css` and other similar issues have come up elsewhere where CSS code is controlling the location of things in right-to-left content. Most things work fine once the page or DIV direction is `rtl`, but not the things which CSS is controlling.

My goal is to allow both course wide overrides of such existing CSS layout code to make it behave better in RTL languages like Hebrew, and to allow problems to individually request the loading of an additional "system provided" CSS file.

It would be possible to try to force CSS code into each problem one at a time by adding more code to `SET_PROBLEM_TEXTDIRECTION()` but that seems a less clean approach than a single course-wide override file, and editing a small number of special css files seems far simpler to manage and maintain than adding code to PG to push in CSS changes of this sort.

drgrice1:
Presumably a course that uses ltr text would also be using a ltr WeBWorK theme.  Isn't that what the math4-ar theme is for?  So maybe these style settings should be set in the math4.css file for that theme.  Perhaps I am misinterpreting the point of the math4-ar theme.

taniwallach:
@drgrice1 - Your question is in place, but is based on the premise that `math4-ar` and later more such minimally different themes is a long term part of the code we wish to maintain. I disagree with that premise.
  - As noted in the discussion at https://github.com/openwebwork/webwork2/pull/802 the math4-ar theme was only set up (so far) to set `lang="ar" dir="rtl"` in the root `<html>` tag of the 4 template files. Supporting additional languages in that approach would have required many parallel themes. Those tags are now set based on a course (or system) level configuration setting.
  - The need for a totally separate `math4-ar` theme was removed by https://github.com/openwebwork/webwork2/pull/850.
  - I think that `math4` was a reasonable temporary step along the way, but should eventually be removed.
  - Although some courses / languages may find a need for some customization of the CSS (and maybe also JavaScript) - the use of `math4-overrides.css` and `math4-overrides.js` are site-wide and not controllable at the course level. This PR and https://github.com/openwebwork/webwork2/pull/1018 are intends to provide mechanisms to support such course level CSS overrides. (JavaScript can be addressed in the future, if and when the need arises.)

drgrice1:
I agree that maintaining minimally different themes should not be needed.  What you say makes sense.

It is not technically true that customization of math4-overrides.js and math4-overrides.css must be site wide.  An instructor of a single course can copy the math4 theme (as a whole), make modifications, and use that theme for his course.  This can be accomplished without admin access.  Although this is really not the point here of course.

taniwallach:
> It is not technically true that customization of math4-overrides.js and math4-overrides.css must be site wide. An instructor of a single course can copy the math4 theme (as a whole), make modifications, and use that theme for his course. This can be accomplished without admin access. Although this is really not the point here of course.

If that is possible, the page at http://webwork.maa.org/wiki/Customizing_WeBWorK should be updated to explain that. The only thing I found quickly in the forums is http://webwork.maa.org/moodle/mod/forum/discuss.php?d=3978 which just remarks that the documentation is not up to date. Looking at the code, it seems that it is probably possible via `course.conf` by changing the location of the templates directory at the course level. However, I think that only course-level admin users can edit the `course.conf` file over the web.

In any case, the goal is to avoid needed to create/maintain extra themes when a small file of CSS code would suffice.

Relevant code: 
  - `getConfigValues()` in `lib/WeBWorK/ContentGenerator/Instructor/Config.pm` searches in `$ce->{webworkDirs}{themes}` and 
  - `content()` in `lib/WeBWorK/ContentGenerator.pm` also works with themes from under `$ce->{webworkDirs}{themes}`
  - A value for `$webworkDirs{themes}` is set in `conf/defaults.config`:
```
# Location of theme templates.
$webworkDirs{themes}     = "$webworkDirs{htdocs}/themes";
```

taniwallach:
I just put a note at http://webwork.maa.org/wiki/Customizing_WeBWorK that the page is out of date and a link to http://webworkgoehle.blogspot.com/2014/07/customizing-colors.html.

@dlglin - Someone who knows more may be able to edit the Wiki page more effectively.

mgage:
Shall we allow this modification for the time being?  I lean toward Davide's view that this is may not be the best long term fix and may need to be redone -- however I don't have immediate suggestions redoing and don't know the scope of the uses cases that need to be covered. That's something we could find out by accepting these changes for now, acknowledging that we should check back soon to make sure they are adequate. 

Is there general agreement to accept these for now? Immediately better suggestions?

taniwallach:
Let's me further clarify. There are already several techniques to allow problems to make use of "special" formatting, or to add content to the HTML header or body. The goal of this PR is to make it possible to load a system hosted CSS file in a convenient and easy to control manner, and not via ugly hacks, and to allow easy use of curated CSS modifications at the course and/or problem level.

I would be very happy to adopt or code a different approach, if someone has a better proposal.

Current techniques:
  - The existing `HEADER_TEXT()` function:
    - Can be used to add an entire `<link>` object to load a CSS file, but could not depend on the system settings for where "standard" HTML/CSS material is stored, so external hosting of the CSS file might be necessary.
    - The same approach could be used to include "inline styling code" in the header any required CSS code, per problem.
    - https://github.com/Edfinity/ww_ggb_applet uses such an approach for a special JavaScript load.
  - http://webwork.maa.org/moodle/mod/forum/discuss.php?d=3064
    - @dvpc wrote there " or could use `PGcourse.pl` to insert (into every problem that includes `PGcourse.pl`) some CSS that overrides the styles for `<code>` elements." and the same approach can be used to override other system provided CSS styling.
  - http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4462
    - Pushing `<style>` blocks into the main HTML output using `MODES(HTML=>"<style> css code </style>", TeX=>"")` as needed.

Issue #429, opened 2019-09-26
Context Fraction and non-terminating decimal reals


drdrew42:
Using Context(“Fraction”) and only the associated context default values, I have found that responses that use decimal approximations in response to an answer blank that is assigned a Fraction object for comparison are unable to achieve a valid result whenever the decimal form is non-terminating (regardless of how many significant figures are provided, and despite the default tolerance).

In such a case, no error is provided by default - but using the AnswerHashInfo option, an error appears in the `student_formula` portion of the answer hash as follows:
`Unable to determine stringify for this item Can’t locate object method “string” via package “context::Fraction::Real” at /opt/WeBWorK/pg/lib/Parser/Number.pm line 66`

I have attempted to add a variety of string methods to the context::Fraction::Real segment of contextFraction.pl to no avail. There’s still something I’m not grasping about perl inheritance, perhaps...

@dpvc any ideas?

```pg
Context(“Fraction”);
$ans = Fraction(1,3);
BEGIN_PGML
Try a decimal approximation of [`\frac{1}{3}`]: [_______]{$ans}
END_PGML
```

dpvc:
I've made a pull request to fix the issue.  It turns out to be due to how the fraction context promotes reals to fractions:  it makes a "fraction" as the real over 1, but it did so by taking the perl real from the student answer, not the MathObject.  That meant that it wasn't a fuzzy real, and so the comparison was being done using exact (perl real) comparisons.  The patch keeps the fraction using a MathObject, so when the comparison is made, it uses the fuzzy math that takes the tolerances into account.

You could take the patched `contextFraction.pl` from the pull request and put it in your `templates/macros` directory in order to get the updated version right away.

drgrice1:
I question the validity of this issue, and the pull request that Davide has submitted.  It seems to me that if you use the fraction context and the answer is 1/3, then 0.33333 should be incorrect.  In my opinion, you are using the wrong context for this problem.  You should do:

``Context(“Numeric”);
$ans = Compute("1/3");
BEGIN_PGML
Try a decimal approximation of [`\frac{1}{3}`]: [_______]{$ans}
END_PGML``

Of course to do this properly to not accept 1/3 as correct you would have to use a custom checker and something like bizarro arithmetic.

If Davide's pull request is accepted, then problems that are using context fraction (in what I think is the correct way to expect exact fractional answers) will no longer work correctly.

drdrew42:
[...]

PR #430, opened 2019-09-27
Make promote() use MathObject Real when converting a decimal.  #429

dpvc:
This allows comparison to use fuzzy tolerance when comparing fractions to explicit decimals.  So if the correct answer is 1/3 and the student enters .33333, this will be accepted (under standard tolerances).  But 33333/100000 will not, as explicit fractions require exact equality.

To test, use

```
Context("Fraction");
$f = Fraction(1,3);
TEXT($f->amp->evaluate(".33333")->{score} ? "Correct": "Wrong");
```

With the patch, this should output `Correct`, and without the patch, `Wrong`.

Resolves issue #429

### DIFF
--- a/macros/contextFraction.pl
+++ b/macros/contextFraction.pl
@@ -698,8 +698,8 @@ sub promote {
   my $x = (scalar(@_) ? shift : $self);
   if (scalar(@_) == 0) {
     return $x->inContext($context) if ref($x) eq $class;
-    return (bless {data => [$x->value,1], context => $context}, $class) if Value::isReal($x);
-    return (bless {data => [$x,1], context => $context}, $class) if Value::matchNumber($x);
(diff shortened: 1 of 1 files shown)


dpvc:
@mgage, Mike, I think there were some serious concerns about this PR expressed in the issue it references (#429).  I think it would be prudent to revert the merge until those are resolved.  I need to comment on them myself, but I was traveling this weekend, and got sick on my return, so haven't been able to take the time yet.  I hope to be able to say something later in the week.  Please see the discussion in the issue and see if you agree that the merge should be reverted for now.

mgage:
@dpvc -- when the issues are resolved you can resubmit the pull request.  I suggest submitting
it to the develop branch since I'm trying to close the 2.15 branches and merge them with master very soon.

dpvc:
@mgage -- OK, will do.

PR #438, opened 2019-10-27
Remove stray "}" from <script> url

drdrew42:
Fixes error: `Failed to load resource: A server with the specified hostname could not be found.` when using Sage Calculator in a problem. Error is caused by stray "}" in the jQuery URL.


Alex-Jordan:
I merged this without realizing it was into `master`.

Since it fixes a bug, I think it should be merged into `master`, but am I wrong about that? Secondly should a parallel PR be opened for `develop`?

taniwallach:
I think it was OK to merge it into master as a hotfix for a significant bug in the production code.
Yes, a PR into develop is also needed, but seems to be in https://github.com/openwebwork/pg/pull/442.

@Alex-Jordan - Thanks for giving the hot-fix attention when the rest of us were too busy to pay attention to it.

PR #439, opened 2019-10-28
Load PGnumericalmacros.pl

Alex-Jordan:
A couple of things in PGstatisticsmacros.pl use the romberg() function that is defined in PGnumericalmacros.pl. So this edit automatically loads PGnumericalmacros.pl when PGstatisticsmacros.pl is loaded.


macros/PGstatisticsmacros.pl
```diff
@@ -1,4 +1,6 @@
 
+# Some subroutines here use the romberg() funciton from PGnumericalmacros.pl
```
taniwallach: Typo in the comment: funciton ==> function

mgage:
I'll correct the spelling in #446.

PR #446, opened 2019-12-30
PopUp extended answer fix

drgrice1:
Add a missing call to INSERT_RESPONSE to the PopUp MENU method in the case that it is an extended answer.
    
Currently, if a PopUp answer is used as a later answer in a MultiAnswer problem with the singleResult option, then the answer for that part is not saved.  When answers are submitted the selected answer initially displays when the page reloads, but if you navigate away from the problem and come back, the answer to that part is no longer there.  The reason for this is that the answer name is never appended to the response group as is done in the NAMED_ANS_RULE_EXTENSION method.

### DIFF
--- a/macros/parserPopUp.pl
+++ b/macros/parserPopUp.pl
@@ -159,7 +159,7 @@ sub randomOrder {
 #
 sub menu {shift->MENU(0,@_)}
 sub MENU {
-  my $self = shift; my $extend = shift; my $name = shift;
+  my $self = shift; my $extend = shift; my $name = shift; my $size = shift; my %options = @_;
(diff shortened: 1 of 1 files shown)


mgage:
This looks ok to me. It is now the way I would have expected it to be written.   It works on all of the problems I tried.  Can you list the path of a problem that triggered the bad response?

drgrice1:
It was actually one of my own problems.  I attached a minimal example that triggers it though.
[test-popup-in-multians.zip](https://github.com/openwebwork/pg/files/4013439/test-popup-in-multians.zip)

PR #432, opened 2019-09-29
Implement TikZ images for problems.

drgrice1:
I am submitting this now so that others can take a look and at Michael Gage's request in issue #418.  There is a matching pull request to the webwork2 code (https://github.com/openwebwork/webwork2/pull/1030) that loads the module and sets the external convert program.

The general implementation is that a uniquely named working directory is created in the general webwork tmp location.  Then a tex file is written and pdflatex run.  Then convert is called to convert the pdf to a png, gif, or svg image.  The image data is read, and then written to disk by the insertGraph method.

I also added a simple macro that basically does the same thing the PGgraphmacros.pl macro does for images generated with GD in terms of setting up the file name.

I changed the output directory of that method from the gif directory to the images directory (inside the course temporary html location).  That part can be reverted if needed, but it seems silly to write what are almost always png images to a gif directory, and have a separate general image directory.

PR #448, opened 2020-01-14
Fix the getUniqueName method to give unique names in a single pg problem

drgrice1:
This was an obvious mistake from the TikZ image pull request.

### DIFF
--- a/lib/PGcore.pm
+++ b/lib/PGcore.pm
@@ -735,7 +735,7 @@ sub getUniqueName {
 	my $prob_name = "$self->{envir}{studentLogin}-$self->{envir}{problemSeed}" .
 		"-set$self->{envir}{setNumber}prob$self->{envir}{probNum}";
 	my $num  = ++$names_created{$prob_name};
-	my $resource = $self->{PG_alias}->make_resource_object($prob_name, $ext);
+	my $resource = $self->{PG_alias}->make_resource_object("$prob_name$num", $ext);
 	$resource->path("__");
 	return $resource->create_unique_id;
 }


taniwallach:
The current change is certainly necessary so different images will have different names, just as the older version of `macros/PGgraphmacros.pl` made use of `$imageNum` in the line:
```
my $resource = $main::PG->{PG_alias}->make_resource_object("image$imageNum","png");
```

---

However, I suspect that some additional work may be needed here. The subroutine `getUniqueName()` being fixed here which was recently added in https://github.com/openwebwork/pg/pull/432 . It is intended to give a unique file name, and the current code may not be sufficiently general for its purpose. Prior experience is that additional care is needed when creating such names, as not all use contexts have the same data available to them to differentiate different problems.

A new item called `problemUUID` was added in webwork/PG 2.15 to make sure that the unique ID's work well also in the `html2xml` framework, where is often no meaningful value for `setNumber` and 
`probNum` as there would be in the regular setting inside a course.

See: 
  - https://github.com/openwebwork/webwork2/pull/976
  - https://github.com/openwebwork/pg/pull/415

The code here has `$prob_name` depending on
  - `studentLogin`,
  - `problemSeed`,
  - `setNumber`
  - `probNum`

However, in the `html2xml` context that last 2 parameters will often fail to have meaningul values which serve the purpose of differentiating problems. 

In contrast, in `lib/PGalias.pg` we have `$unique_id_seed` depend also on
  - `psvn`,
  - `problemUUID`, and
  - `courseID`.

(`psvn` is an alternative to `problemSeed` for seeing the PRNG in cases where several problems in a set/context need to share parameters. It can be set in the `html2xml` context, and is provided in the regular context from the set_user table.)

I suspect that it might be necessary to add  `problemUUID` and probably also `psvn` to the definition of  `$prob_name` in the setting under discussion, unless it is certain that these values will be taken into account elsewhere in the path/filename generation (probably based on the definition of `$unique_id_stub` which is made during the `initialize()` method in `/lib/PGalias.pm` ). I did not trace the code flow to see if that will happen.

If the path already depends on the `courseID` somehow then that parameter may not be needed in the current setting; if not - it probably should also be added to the definition of `$prob_name` in `getUniqueName()`.

mgage:
For the historical record:

psvn (problem set version number) has a long, slightly bizarre history. It predates problemSeed.  The psvn is the same across the entire homework set.

This means that  when really necessary it can be used as a hack to populate several of problems with the same problemSeed so that successive problems
are closely related with the same constants.  This is not very robust so you need to use it cautiously.  Another approach to solve this issue is to use 
a scaffolded problem instead of successive semi-independent problems.

drgrice1:
It should be noted that getUniqueName just moves the code that was in PGgraphmacros.pl to PGcore.pm, and achieves the same ends as the code that was there.  What you are talking about is improving on that, and that may need to be done for ends that are desired at this point.  In any case this works as well as it did before with this change.

drgrice1:
I suggest that we merge this in so that it is in, and we can move on with the improvements that Nathan is suggesting.  I will do so if no one objects.

mgage:
This seems a reasonable fix for now.  I've pushed Tani's comments to Issues so that they don't get lost.

drgrice1:
Or would you rather, that I added problemUUID and courseID to the $prob_name variable first to somewhat address Nathan's concerns?

PR #436, opened 2019-10-08
Scaffold opening reflows MathQuill answer boxes

drgrice1:
When a scaffold opens iterate through the MathQuill answer boxes on the
page and call reflow on each one so that the contents are properly
displayed.

See http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4666

PR #454, opened 2020-02-25
Commits from master which are not in develop

taniwallach:
This is to pull 5 commits made to master into develop.
These commits contain several small fixes already in master, so the code should also be up-to-date in develop.
There are only 2 files changed based on those commits.


taniwallach:
Sorry - you are correct.

I just used GitHub's compare of branches.

I see the changes in develop from commits:
  -  https://github.com/openwebwork/pg/commit/b1e839ad95c4a487067a410c600689361b98594c#diff-ec688a0529a3964cd5ab6e67145adba3
  -https://github.com/openwebwork/pg/commit/d8811342ba58571a35d557640c2589e0197da6a3#diff-bc99f218dc5c213309fa03494df23eb4

Each "real" change was made by a different commit to master vs. develop. Apparently this is causing some issue with the "order" Git/GitHub is comparing files and causing these "fictive" differences.

I'm closing this, and if necessary we will need to resolve any issues when merging develop into master.

taniwallach:
I think I was bit by the same GitHub behavior as @Alex-Jordan was in https://github.com/openwebwork/pg/pull/436#issuecomment-560001360

PR #461, opened 2020-03-26
Fix erro in line 771 of PGmatrixmacros which made display_matrix() no…

mgage:
…t accept MathObjects.

Hot fix. This broke a lot of matrix problems when display_matrix() did not accept MathObject matrices.

PR #462, opened 2020-03-27
a fix for convert_to_array_ref  for develop.

mgage:
This is a companion to the hot fix #461 submitted to master.

@dpvc  I'm trying to understand the long term fix for this.  Should it be necessary to check
that $input satisfies Value::isValue($input) before calling Value::checkMath($input, "Matrix");?

I realize there might be an issue in case Value is subclassed, but I'm not sure what it is. 

Also i modified the check to make sure that $input is a hash to use Value::isHash() so that the definitions are uniform in the file.  Is that ok? or is that part of the problem.

### DIFF
--- a/lib/Value.pm
+++ b/lib/Value.pm
@@ -355,7 +355,7 @@ sub classMatch {
   my $class = class($self)//''; my $ref = ref($self);
   #my $isHash = ($ref && $ref ne 'ARRAY' && $ref ne 'CODE');
   #warn "hash $isHash $ref";
-  my $isHash = isHash($self);
+  my $isHash = Value::isHash($self);
   my $context = ($isHash ? $self->{context} || Value->context : Value->context);
   foreach my $name (@_) {
     my $isName = "is".$name;
--- a/lib/Value.pm
+++ b/lib/Value.pm
@@ -284,6 +284,10 @@ sub isHash {
 
 }
 
+
+# example: return Boolean:  Value->subclassed($self,"classMatch")
+# if $self has the method 'classMath' and 'Value' has the method 'classMatch'
+# and  the reference to these methods don't agree then the method 'classMatch' has been subclassed. 
 sub subclassed {
   my $self = shift; my $obj = shift; my $method = shift;
   my $code = UNIVERSAL::can($obj,$method);
@@ -352,9 +356,7 @@ sub Package {(shift)->context->Package(@_)}
 sub classMatch {
   my $self = shift;
   return $self->classMatch(@_) if Value->subclassed($self,"classMatch");
-  my $class = class($self)//''; my $ref = ref($self);
-  #my $isHash = ($ref && $ref ne 'ARRAY' && $ref ne 'CODE');
-  #warn "hash $isHash $ref";
+  my $class = Value::class($self)//''; my $ref = ref($self);
   my $isHash = Value::isHash($self);
   my $context = ($isHash ? $self->{context} || Value->context : Value->context);
   foreach my $name (@_) {
--- a/lib/Value.pm
+++ b/lib/Value.pm
@@ -353,7 +353,9 @@ sub classMatch {
   my $self = shift;
   return $self->classMatch(@_) if Value->subclassed($self,"classMatch");
   my $class = class($self)//''; my $ref = ref($self);
-  my $isHash = ($ref && $ref ne 'ARRAY' && $ref ne 'CODE');
+  #my $isHash = ($ref && $ref ne 'ARRAY' && $ref ne 'CODE');
+  #warn "hash $isHash $ref";
+  my $isHash = isHash($self);
   my $context = ($isHash ? $self->{context} || Value->context : Value->context);
   foreach my $name (@_) {
     my $isName = "is".$name;
--- a/macros/PGmatrixmacros.pl
+++ b/macros/PGmatrixmacros.pl
@@ -768,10 +768,12 @@ seeking.
 
 sub convert_to_array_ref {
 	my $input = shift;
-	if (Value::classMatch($input,"Matrix")) {
+	if (Value::isValue($input) && Value::classMatch($input,"Matrix")) {
 		$input = [$input->value];
+		warn "assuming a Value object";
 	} elsif (ref($input) eq 'Matrix' ) {
 		$input = $input->array_ref;
+		warn "assuming wwMatrix";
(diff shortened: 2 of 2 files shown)


mgage:
Test using this:

DOCUMENT();

loadMacros(
"PG.pl",
"PGbasicmacros.pl",
"PGanswermacros.pl",
"PGmatrixmacros.pl",
"MathObjects.pl",
"problemRandomize.pl"
);

TEXT(beginproblem());

$dim = 2;

Context ("Matrix");
$L1 = new Matrix($dim,$dim); #creates wwMatrix
$L1->assign(1,1,5);
$L2 = Matrix("[[2,5],[2,7]]"); #creates Matrix MathObject

BEGIN_TEXT

L1 {ref($L1)} {display_matrix($L1)}
$PAR
L2 {ref($L2)} {display_matrix($L2)}
$PAR
Notice that both old style (wwMatrix) and MathObject (Value::Matrix) should work now with no warnings. With the bug the wwMatrix was not properly processed.
END_TEXT

ENDDOCUMENT();

dpvc:
> Should it be necessary to check that `$input` satisfies `Value::isValue($input)` before calling `Value::checkMath($input, "Matrix");`?

It is not supposed to be necessary, but the problem turns out to be that MatrixReal1 objects are blessed arrays, so they have non-empty `ref()` but it is not `ARRAY`, so the `$isHash` check is wrong, as you determined.  I'm not sure why `Value::isHash()` is not used there (must be that it was written before the `isHash()` function was available, but both seem to be in the same commit, so I'm not sure what is up there).  In any case, I think you are correct to change the computation off `$isHash` and that should resolve the issue.  No need for `Value::isValue()` in this case.

> I realize there might be an issue in case Value is subclassed, but I'm not sure what it is.

I don't think there is an issue, and I don't know of a use-case for subclassing `Value` itself.

> Also I modified the check to make sure that `$input` is a hash to use `Value::isHash()` so that the definitions are uniform in the file. Is that ok? or is that part of the problem.

That is correct, and what I was going to suggest (I came to that conclusion independently myself before looking at your PR).

lib/Value.pm
```diff
@@ -353,7 +353,9 @@ sub classMatch {
   my $self = shift;
   return $self->classMatch(@_) if Value->subclassed($self,"classMatch");
   my $class = class($self)//''; my $ref = ref($self);
-  my $isHash = ($ref && $ref ne 'ARRAY' && $ref ne 'CODE');
+  #my $isHash = ($ref && $ref ne 'ARRAY' && $ref ne 'CODE');
+  #warn "hash $isHash $ref";
```
dpvc: Remove these comments (they just make things harder to read, and the old code is in the GitHub history, so no need to keep it here).

macros/PGmatrixmacros.pl
```diff
@@ -768,10 +768,12 @@ =head2 convert_to_array_ref {
 
 sub convert_to_array_ref {
 	my $input = shift;
-	if (Value::classMatch($input,"Matrix")) {
+	if (Value::isValue($input) && Value::classMatch($input,"Matrix")) {
```
dpvc: I would revert this file back to its original form.  No need to `Value::isValue($input)` with the correction above.  Also the `warn` lines below need to be removed.  So no actual changes to this file.

lib/Value.pm
mgage: ok. comments removed.

macros/PGmatrixmacros.pl
mgage: I agree on removing the warn lines.  

However removing Value::isValue($input)
makes the test case fail (see expanded test case below). 

I still think this glitch should be fixed in Value::classMatch or perhaps Value::class() so that Value::classMatch
returns true only if $input is of type Value.  even if there are other objects as in this case that have the same name e.g. "Matrix".  Was Value::class supposed to return true for anything except Value types?

mgage: I understand now that the subclassing refers to the classMatch method not to value.

mgage:
# extended test problem
# demonstrates that classMatch($input, "Matrix") matches a wwMatrix (Real1Matrix) object.

DOCUMENT();   

loadMacros(
  "PG.pl",
  "PGbasicmacros.pl",
  "PGanswermacros.pl",
  "PGmatrixmacros.pl",
  "MathObjects.pl",
  "problemRandomize.pl"
);

TEXT(beginproblem());

$dim = 2;

Context ("Matrix");
$L1 = new Matrix($dim,$dim); #creates wwMatrix
$L1->assign(1,1,5);
$L2 = Matrix("[[2,5],[2,7]]"); #creates Matrix MathObject
TEXT("L1: class is ", Value::class($L1), $PAR );
BEGIN_TEXT
L1: isHash \{Value::isHash($L1)? 1:0\}
$PAR
L1: ref is \{ref($L1)\}
$PAR
L1: class is \{Value::class($L1)\}
$PAR
L1: classMatch is \{Value::classMatch($L1, "Matrix")\}
$PAR
L1 \{ref($L1)\} \{display_matrix($L1)\}
$PAR
L2 \{ref($L2)\}  \{display_matrix($L2)\}
$PAR
Notice that both old style (wwMatrix) and MathObject (Value::Matrix) should work now with no warnings.  With the bug the wwMatrix was not properly processed. 
END_TEXT


ENDDOCUMENT();

dpvc:
I don't think you have pushed your changes.  I don't see any new commits, in any case.

dpvc:
OK, I see your point, now (I had been testing by hand and wasn't using your test files).

I think you will need to keep the `Value::isValue()` in the `display_macros()` function.

Although it would be best to make it part of the `Value::classMatch()` call, I'm afraid that it would likely break some macro files that define their own MathObject classes.  One is supposed to use the `isValue` property to mark them as Value objects if they aren't already a subclass of an existing MathObject, but I'm not sure that everyone (including me) has done that.  The `classMatch()`function is intentionally set up to allow a wide range of marches, so that if someone defines a package `my::Complex` it will be recognized as a MathObject Complex object.  

Requiring `Value::isValue()` would be a good idea, but it is likely to break things unexpectedly.  I haven't gone through the macro files to see what might be affected, but there are also private macro files that we don't know about that would also take a hit.

dpvc:
I've approved the current set of changes.

mgage:
Thanks Davide.

PR #464, opened 2020-04-08
Fix parserRadioButtons.pl when used as an extended response.

drgrice1:
For example, when used as a latter answer in a MultiAnswer question.  It was
incorrect to have the named_ans_rule_extension method of the
RadioButtons call NAMED_ANS_RADIO_EXTENSION for the first radio button.
That effectively did nothing as the response was not defined.
NAMED_ANS_RADIO still needs to be called, but that method needs to
call INSERT_RESPONSE for the answer group instead of RECORD_ANS_NAME.
Furthermore the NAMED_ANS_RADIO_EXTENSION method needed to use the same
answer group name.

Before this fix if a radio answer is a latter answer in a MultiAnswer
question with the singleResult option, the answers for the radio answer
were not saved.

This is the same issue that affected parserPopUp.pl, that was fixed in PR #446.

For a simple test problem use:
```
DOCUMENT();
loadMacros(
	"PGstandard.pl",
	"MathObjects.pl",
	"PGML.pl",
	"parserMultiAnswer.pl",
	"parserRadioButtons.pl"
);
TEXT(beginproblem());

$a = random(0, 10);
$radio = RadioButtons([['Part 1', 'Part 2', 'Part 3']], 0, labels => "ABC");

$ma = MultiAnswer($a, $radio)->with(
    singleResult => 1,
    checker => sub {
        return ($correct->[0] == $student->[0] && $correct->[1] == $student->[1]) ? 1 : 0;
    }
);
BEGIN_PGML
The first answer is [`[$a]`].

[_____]{$ma}

The second answer is part 1.

[_____]{$ma}
END_PGML
#warn $PG->pretty_print($PG->{PG_ANSWERS_HASH}, 'html', 8);
ENDDOCUMENT();
```
If you un-comment the warn statement, then without this patch you will see that there is no response for the radio answer, and with this patch there is.

The radio answer will still work as it did before in all other cases.

Note that I added general code clean up to PGbasicmacros.pl in the second commit.  View the first commit for the essential changes here.

### DIFF
--- a/macros/parserRadioButtons.pl
+++ b/macros/parserRadioButtons.pl
@@ -530,7 +530,7 @@ sub makeUncheckable {
 #  Create the radio-buttons text
 #
 sub BUTTONS {
-  my $self = shift; my $extend = shift; my $name = shift;
+  my $self = shift; my $extend = shift; my $name = shift; my $size = shift;
   my @choices = @{$self->{orderedChoices}};
   my @radio = ();
   $name = main::NEW_ANS_NAME() unless $name;
@@ -539,12 +539,11 @@ sub BUTTONS {
     my $value = "B$i"; my $tag = $choices[$i];
     $value = "%".$value if $i == $self->{checkedI};
     $tag = $self->labelFormat($self->{labels}[$i]).$tag if $self->{displayLabels};
-    if ($extend) {
-      push(@radio,main::NAMED_ANS_RADIO_EXTENSION($name,$value,$tag,
-	   aria_label=>$label."option $i "));
+    if ($i > 0) {
+		push(@radio,main::NAMED_ANS_RADIO_EXTENSION($name, $value, $tag,
+				aria_label => $label."option $i ", @_));
     } else {
-      push(@radio,main::NAMED_ANS_RADIO($name,$value,$tag));
-      $extend = 1;
+		push(@radio, main::NAMED_ANS_RADIO($name, $value, $tag, $extend, @_));
     }
   }
   #
--- a/macros/parserRadioButtons.pl
+++ b/macros/parserRadioButtons.pl
@@ -540,10 +540,10 @@ sub BUTTONS {
     $value = "%".$value if $i == $self->{checkedI};
     $tag = $self->labelFormat($self->{labels}[$i]).$tag if $self->{displayLabels};
     if ($i > 0) {
-		push(@radio,main::NAMED_ANS_RADIO_EXTENSION($name, $value, $tag,
-				aria_label => $label."option $i ", @_));
+      push(@radio,main::NAMED_ANS_RADIO_EXTENSION($name, $value, $tag,
+           aria_label => $label."option $i ", @_));
     } else {
-		push(@radio, main::NAMED_ANS_RADIO($name, $value, $tag, $extend, @_));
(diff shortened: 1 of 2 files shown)


macros/parserRadioButtons.pl
```diff
@@ -539,12 +539,11 @@ sub BUTTONS {
     my $value = "B$i"; my $tag = $choices[$i];
     $value = "%".$value if $i == $self->{checkedI};
     $tag = $self->labelFormat($self->{labels}[$i]).$tag if $self->{displayLabels};
-    if ($extend) {
-      push(@radio,main::NAMED_ANS_RADIO_EXTENSION($name,$value,$tag,
-	   aria_label=>$label."option $i "));
+    if ($i > 0) {
```
dpvc, line 542: The change from `$extend` to `$i > 0` is incorrect.  The initial `$extend` value comes from an argument to the `BUTTONS()` function, and it is not always 0.  The `named_ans_rule_extension()` method sets it to 1 initially, so `$extend` is tru even when `$i = 0` in that case.   Your modification would change how that is handled.  Note that the `named_ans_rule_extension()` method is used by the MultiAnswer object, so this is an actual use case.

drgrice1: I am sorry, but that is clearly not correct, and that is the whole point of this pull request and the problem here.  The confusion here is that there are two different types of response extension.  That is extending another answer (really `INSERT_RESPONSE` and `append_response`), or extending a radio button (really `EXTEND_RESPONSE` and `extend_response`).  The $extend argument that comes from `BUTTONS` is meant to extend another answer.  As such it should do what I have it set up to do.  That is pass the $extend argument on to `NAMED_ANS_RADIO` that is now properly coded to extend the answer group.

If it is put back to the way that it was before then when `NAMED_ANS_RADIO_EXTENSION` is called, that method then calls `EXTEND_RESPONSE`.  That method then does nothing because `$PG->{PG_ANSWERS_HASH}->{$ans_label}` is not defined.  So no response is extended, and no answers are recorded.  Failure ensues as later calls to `NAMED_ANS_RADIO_EXTENSION` and then `EXTEND_RESPONSE` also do nothing as `$PG->{PG_ANSWERS_HASH}->{$ans_label}` is still not defined.

With the way that it is now, `NAMED_ANS_RADIO` calls `INSERT_RESPONSE` if this is an $extend case, which calls `append_response`.  That adds a response to the response group for the radio answer.  Then the later calls the `NAMED_ANS_RADIO_EXTENSION` extend that response.

drgrice1: If you doubt my claims here, test this with the test problem in my initial comments for the pull request.  Uncomment the warn statement that I included.  Without this change you will see that the multianswer response has nothing for the radio answer.  While with this change there is a response added for the radio answer.

drgrice1: To save you the effort of testing, here are screenshots of the resulting response group without and with this pull request respectively:

without:
without-pr

with:
with-pr

dpvc: Sorry, I misunderstood the changes in `NAMED_ANS_RADIO()`, which now handles both initial buttons and extensions (I'm afraid I haven't followed all the changes to the answer blank infrastructure that are now in use).

mgage:
This seems to be settled.

Issue #443, opened 2019-12-12
UTF-8 trouble with parserPopup.pl in certain settings (may apply to other MathObjects with UTF-8 string answers)


taniwallach:
When `parserPopUp.pl` is used with non-English UTF-8 answer strings trouble occurs in some setting:

1. "images" display mode fails to work properly: when answers are submitted - the rendering of the answers table fails and the page fails to display.
2. html2xml fails to render problems. The problem seems to be that the `<methodResponse>` data includes data like
```
<member><name>correct_ans_latex_string</name><value><string>{\verb+JUNK DELETED+}</string></value></member>
```
where I replaced the contents of the `string` block as the generated data inside `\verb` is manged.

A sample response when trying to submit an answer to such a problem with images equation display mode in use is:
```
WeBWorK error
An error occured while processing your request. For help, please send mail to this site's webmaster (techdesk@mathnet.technion.ac.il), including all of the following information as well as what what you were doing when the error occured.

Thu Dec 12 13:08:07 2019

Warning messages
Error messages
Wide character in subroutine entry at /opt/webwork/pg/lib/WeBWorK/EquationCache.pm line 89.
Call stack
The information below can help locate the source of the problem.

in WeBWorK::EquationCache::lookup called at line 271 of /opt/webwork/pg/lib/WeBWorK/PG/ImageGenerator.pm
in WeBWorK::PG::ImageGenerator::add called at line 352 of /opt/webwork/webwork2/lib/WeBWorK/Utils/AttemptsTable.pm
in WeBWorK::Utils::AttemptsTable::previewAnswer called at line 250 of /opt/webwork/webwork2/lib/WeBWorK/Utils/AttemptsTable.pm
in WeBWorK::Utils::AttemptsTable::formatAnswerRow called at line 320 of /opt/webwork/webwork2/lib/WeBWorK/Utils/AttemptsTable.pm
in WeBWorK::Utils::AttemptsTable::answerTemplate called at line 350 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm
in WeBWorK::ContentGenerator::Problem::attemptResults called at line 1935 of /opt/webwork/webwork2/lib/WeBWorK/ContentGenerator/Problem.pm
in WeBWorK::ContentGenerator::Problem::output_summary called at line 155 of /opt/webwork/webwork2/lib/WeBWorK/Template.pm
[...]

PR #466, opened 2020-04-18
encode string as octets before passing to md5_hex()

mgage:
The string might have utf8 characters which md5_hex() can't handle. 

See https://perldoc.perl.org/Digest/MD5.html near the bottom of the page.

### DIFF
--- a/lib/WeBWorK/EquationCache.pm
+++ b/lib/WeBWorK/EquationCache.pm
@@ -87,7 +87,7 @@ sub lookup {
 	# Option 2 (the old default): remove all whitespace
 	# $tex =~ s/\s+//g;
 
-	my $md5 = md5_hex(encode_utf8($tex));
+	my $md5 = md5_hex(utf8::is_utf8($tex) ? encode_utf8($tex) : $tex);
 	
 	my $db = $self->{cacheDB};
 	unless($db) { return($md5 ."1"); }
--- a/lib/WeBWorK/EquationCache.pm
+++ b/lib/WeBWorK/EquationCache.pm
@@ -87,7 +87,7 @@ sub lookup {
 	# Option 2 (the old default): remove all whitespace
 	# $tex =~ s/\s+//g;
 
-	my $md5 = md5_hex(utf8::is_utf8($tex) ? encode_utf8($tex) : $tex);
+	my $md5 = md5_hex(encode_utf8($tex));
 	
 	my $db = $self->{cacheDB};
 	unless($db) { return($md5 ."1"); }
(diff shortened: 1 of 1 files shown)


lib/WeBWorK/EquationCache.pm
```diff
@@ -37,6 +37,7 @@ extending to the end of the line is also ignored.
 use strict;
 use warnings;
 use Digest::MD5 qw(md5_hex);
+use Encode qw(encode_utf8 decode_utf8 decode);
```
drgrice1, line 40: Why do you also include decode_utf8 and decode, when only encode_utf8 is used?

mgage: Fair enough.  I wasn't sure what would be needed.

taniwallach:
This is essentially the fix suggested by @dvpc in https://github.com/openwebwork/pg/issues/443 and it certainly makes sense that it would fix the issue.

mgage:
I think I like @dpvc 's fix better.  Less chance of an unexpected compatibility issue with old code and old cached equations:
`my $md5 = md5_hex(utf8::is_utf8($tex) ? encode_utf8($tex) : $tex);`

any objections, comments?

taniwallach:
When I enabled your fix on my server, the results table (where UTF-8 characters are used on popups) appears, but the "rendered" images are blank. There are several warnings:
```
Wide character in print at /opt/webwork/pg/lib/WeBWorK/PG/ImageGenerator.pm line 378
```
but getting a valid response page with the results table and the question shown rather than an error pages is already an improvement.

mgage:
Does restricting the encoding only to $tex strings that satisfy is_utf8 help?  
What images are you rendering?  Do you mean the images of the hebrew letters don't appear
in the popUp menu?

taniwallach:
> Does restricting the encoding only to $tex strings that satisfy is_utf8 help?

I did not check - but this is exactly the case where we do need to encoding into UTF-8 before calling MD5, as before the change the code would fail as this is a case where Unicode characters are in the answer. The "new" issues is in the image generation process itself (in ImageGenerator.pm).

> What images are you rendering? Do you mean the images of the hebrew letters don't appear
> in the popUp menu?

The popUp menu in the problem I tested has Hebrew works, and in the popup itself they do appear. However, in the "rendered" answer column of the results table they do not appear, but in the "plain text" column they do appear.

The problem I tested on has popups essentially that same as those in the example given in https://github.com/openwebwork/webwork2/pull/1089 which has:
```
$popup1 = PopUp(['נכון', 'לא נכון'], 'נכון');
```
and the generated image I got was a 1x1 PNG file with what seems to be a single point. The tooltip (if you click to get it) does show the correct Hebrew text of the answer selected.

As I wrote above, MathJax mode works better, and moreover the issue of the "missing" text in the image is not really critical - as the text shows up in the other column.

PR #467, opened 2020-04-27
WIP Fix Rserve connections

kitsook:
- if rserve_eval is called without first calling rserve_start, it
  creates a new connection without closing it.  This causes dangling
connections.


kitsook:
Closing this PR. `rserve_eval` is supposed to auto-create and keep the connection for multiple calls.  Needed to find out why the destructor is not called properly and close the connection when the http request ends.

PR #468, opened 2020-05-11
Block answerHints post-filter from running in Preview mode

taniwallach:
At present, the `AnswersHint` post-filter runs regularly in "Preview mode" so if any special messages would be triggered by the cases handled by AnswerHints, they would appear when "Preview My Answers" is clicked. That leaks "grading feedback" for a preview only request.

Since the "Preview My Answers" is only intended to report syntax type errors, the current behavior seems undesirable, and this PR will prevent it.
---
A pretty minimal sample problem which can be used to test behavior before/after this change is below. Entering -1 and submitting the answer should give 50% credit and the message "You seem to have a sign error", but before the change from this PR, that message is also given when "Preview My Answers" is clicked. 

preview-answerHint-bug

```
DOCUMENT();      

loadMacros(
   "PGstandard.pl",
   "MathObjects.pl",
   "answerHints.pl",
);

Context("Numeric");

$ans1 = Real(1);

Context()->texStrings;
BEGIN_TEXT
\( i^4 \) = \{ $ans1->ans_rule(3) \}
END_TEXT
Context()->normalStrings;

ANS( $ans1->cmp()->withPostFilter(AnswerHints(
   $ans1 => ["",replaceMessage=>0],
   (-1 * $ans1) => [ "You seem to have a sign error", score =>0.50]
)));

ENDDOCUMENT();
```

### DIFF
--- a/macros/answerHints.pl
+++ b/macros/answerHints.pl
@@ -111,6 +111,11 @@ sub AnswerHints {
     my $student = $ans->{student_value};
     Value::Error("AnswerHints can only be used with MathObjects answer checkers") unless ref($correct);
     return $ans unless ref($student);
+    my $isPreview = $ans->{isPreview};
+    if ( defined($isPreview) && ( ($isPreview) || ( $isPreview ne "") ) ) {
+      # In preview mode - we do not want to run this post filter, as it would leak feedback messages.
+      return $ans;
+    }
 
     while (@_) {
       my $wrongList = shift; my $message = shift; my @options;


dpvc:
I was looking for something else today, and found that I had a pending update for this issue myself.  It takes a slightly different tack, providing a configuration option that controls whether answer hints are processed (on a hint-by-hint basis) rather than always assuming we don't want to run the checks.   This is the diff from the current master copy:

``` diff
--- a/macros/answerHints.pl
+++ b/macros/answerHints.pl
@@ -59,6 +59,18 @@ if the student answer is the
 same type as the correct one.
 Default: 1
 
+=item C<S<< processPreview => 0 or 1 >>>
+
+1 means process student answers even
+during answer previews.  Usually, no
+hints are given durring previews, and
+only when answers are checked or submitted.
+The defaultcan be controlled on an individual
+message basis, or by adding
+C<answerHintsProcessPreview> to the C<cmp()>
+arguments, or in the context's flags.
+Default: 0
+
 =item C<S<< score => number >>>
 
 Specifies the score to use if
@@ -111,6 +123,11 @@ sub AnswerHints {
     my $student = $ans->{student_value};
     Value::Error("AnswerHints can only be used with MathObjects answer checkers") unless ref($correct);
     return $ans unless ref($student);
+    my $context = $correct->context;
+    my $hash = $context->{answerHash};
+    $context->{answerHash} = $ans;
+    my $processPreview = $correct->getFlag('answerHintsProcessPreview', 0);
+    $context->{answerHash} = $hash;
 
     while (@_) {
       my $wrongList = shift; my $message = shift; my @options;
@@ -119,11 +136,13 @@ sub AnswerHints {
         checkCorrect => 0,
         replaceMessage => 0,
        checkTypes => 1,
+        processPreview => $processPreview,
         score => undef,
         cmp_options => [],
         @options,
       );
-      next if $options{checkType} && $correct->type ne $student->type;
+      next if $options{checkTypes} && $correct->type ne $student->type;
+      next if !$options{processPreview} && $ans->{isPreview};
       $wrongList = [$wrongList] unless ref($wrongList) eq 'ARRAY';
       foreach my $wrong (@{$wrongList}) {
        if (ref($wrong) eq 'CODE') {
```

Your test for `isPreview` is more complex than mine, but if you want that extra testing, I would move that to the line in `pg/lib/AnswerChecker.pm` where the `isPreview` value is defined, as the intent is the `isPreview` should be all you have to check.

taniwallach:
@dvpc - Your approach makes this something configurable, so I'm closing this PR, and would appreciate your making a PR with your code.

I think I may have made the test more complicated, as I was worried that when isPreview gets a UTF-8 value it may not be handled well.

PR #469, opened 2020-05-12
Fix malformed utf8 copyright errors caused by the © character

drgrice1:
Switch from © to &copy; in all pg macro files where this had not been done already.  These can cause malformed utf-8 character errors for problems.  This was seen in problems that use the contextPolynomialFactors.pl macro.

This could probably be considered to be added to the master branch.  This should have been done when the switch to utf-8 was made.

### DIFF
--- a/macros/compoundProblem5.pl
+++ b/macros/compoundProblem5.pl
@@ -1,6 +1,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright © 2014 The WeBWorK Project, http://openwebwork.sf.net/
+# Copyright &copy; 2014 The WeBWorK Project, http://openwebwork.sf.net/
 # $$
 # 
 # This program is free software; you can redistribute it and/or modify it under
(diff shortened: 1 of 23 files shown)


mgage:
I wasn't able to trigger an error on my laptop installation using Library/CollegeOfIdaho/setAlgebra_05_06_Factoring/56IntAlg_29_Factor.pg
and a develop branch  before the patch was applied.  

However this is clearly a patch that should be applied.  There were no 
errors after the patch.

mgage:
I'm not sure that it's urgent that this be a hotfix.

Unless someone can give a specific example where this is causing trouble in the master 2.15 branch we can just wait until the is develop branch becomes 2.16

drgrice1:
I should note that I was not able to reproduce this using a user with professor (or above) permissions.  It only occurred with a student account.  It also only occurs in a course with a database that is fully upgraded to utf8mb4.  If I import a course that was originally created before the database upgrade and don't upgrade all tables, it does not occur.  That is most likely because the database is still really using the old character set.  To really upgrade the database, you need to go through each table and change its character set.  So try this with a newly created course on WW-2.15 using a student account.

drgrice1:
It is possible that a hot fix is needed. See issue #493.

taniwallach:
Hotfix is in https://github.com/openwebwork/pg/pull/500 and was merged

PR #471, opened 2020-05-17
Rewrite lcm() and gcf() to handle more than two arguments

Alex-Jordan:
This extends lcm() and gcf() to handle more than two arguments. For example, lcm(2,3,4)=12, or gcf(0,4,6,8)=2. The current versions ignore arguments beyond the second.

### DIFF
--- a/macros/PGauxiliaryFunctions.pl
+++ b/macros/PGauxiliaryFunctions.pl
@@ -25,9 +25,9 @@ loadMacros("PGcommonFunctions.pl");
 #  max(@listNumbers)
 #  min(@listNumbers)
 #  round($number)
-#  lcm($number1,$number2)
-#  gfc($number1,$number2)
-#  gcd($number1,$number2)  
+#  lcm(@listNumbers)
+#  gcf(@listNumbers)
+#  gcd(@listNumbers)
 #  isPrime($number)
 #  reduce($numerator,$denominator)
 #  preformat($scalar, "QuotedString")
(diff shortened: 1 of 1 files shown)


macros/PGauxiliaryFunctions.pl
```diff
@@ -107,72 +107,38 @@ sub Round {
 
 #least common multiple
 #VS 6/29/2000
+# made recursive to handle the gcf of any number of inputs 5/16/2020
```
drgrice1, line 110: There is no need to add comments explaining changes and dates.  The git repository shows the changes.  The dates that are here are from an old pre-git practice that should be discontinued.

macros/PGauxiliaryFunctions.pl
```diff
@@ -107,72 +107,38 @@ sub Round {
...
-	my $curr = $b;;
-
-	while($lcm == 0) {
-		$lcm = $curr if ($curr % $a == 0);
-		$curr += $b;
-	}
-
-	$lcm;
-
+        if (scalar @_ == 0) {warn 'Cannot take lcm of the empty set'; return;}
```
drgrice1, line 113: There is no need to use "scalar" here.  When comparing a list to a scalar it is converted to a scalar that is the number of elements in the list by default.

Alex-Jordan: This I do know, but I like using `scalar` for explicit code readability.

macros/PGauxiliaryFunctions.pl
Alex-Jordan: Thanks for reviewing this. OK, but should I delete the note that are already there? Whoever VS is, is probably no longer the person to consult with a question. (Assuming that is the reason for this kind of commentary historically.)

Alex-Jordan:
I changed the style of the conditionals in the code.

I had been under the impression that the array `(0)` evaluates to false as a boolean. But I now see it evaluates to true, and that makes me feel better about the style change I made.

Alex-Jordan:
Here is a problem to use for testing to see if the functions actually work. Tests with input arrays of length 0, 1, 2, 3. I can't think of a way to really test that doesn't involve human eyes sampling the output.

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
);

@a = (); 
TEXT("LCM of ".join(',',@a)." is ", lcm(@a), $BR);
TEXT("GCF of ".join(',',@a)." is ", gcf(@a), $PAR);

for $i (-6..6) {
  my @a = ($i);
  TEXT("GCF of ".join(',',@a)." is ", gcf(@a), $BR);
  TEXT("LCM of ".join(',',@a)." is ", lcm(@a), $PAR);
}

for $i (-4..4) {
  for $j (-4..4) {
    my @a = ($i,$j);
    TEXT("GCF of ".join(',',@a)." is ", gcf(@a), $BR);
    TEXT("LCM of ".join(',',@a)." is ", lcm(@a), $PAR);
  }
}

for $i (-4..4) {
  for $j (-4..4) {
    for $k (-4..4) {
        my @a = ($i,$j,$k);
        TEXT("GCF of ".join(',',@a)." is ", gcf(@a), $BR);
        TEXT("LCM of ".join(',',@a)." is ", lcm(@a), $PAR);
    }
  }
}

ENDDOCUMENT(); 
```

macros/PGauxiliaryFunctions.pl
```diff
@@ -106,73 +106,35 @@ sub Round {
...
-
-	}
-
-	$tempR;
+        do {warn 'Cannot take gcf of the empty set or an all-zero set'; return;} unless (@_);
+        my $a = abs(shift);
+        if ($a == 0) {return gcf(@_);}
+        return $a unless (@_);
+        my $b = abs(shift);
+        if ($b == 0) {return gcf($a,@_);}
```
drgrice1, line 130: There is no need for lines 127 and 130 above.  If $a is 0 and $a > $b (enforced by the next line below), then the while loop won't run, and the next iteration will remove off $a and the iteration after that will remove $b.  If $a > 0 and $b is 0, then the next line below will switch that and again the while loop won't run and the next iteration will remove $a.

drgrice1: I see.  Line 127 gives the warning about not taking the gcf of an all-zero set on the next iteration.  So that line is needed.  Line 130 is not though.  If it gets there then $a is not zero, and the next iteration will remove $b.

Alex-Jordan:
I guess I would call the gcf of a single positive integer (or a single positive integer with some zeros) to be that number, not undefined.

Did you mean with respect to lcm? If it is a single positive integer with some zeros, I would have the result be 0 as it should be returning right now. If it just the lcm of a single positive integer, then I would have it be undefined, which it does not return, and I'll work on that.

Sorry if I'm confused about the rest of it. Let me know if I misunderstood.

Alex-Jordan:
Er, wait. I want to talk that back. I think the lcm of just a single positive number should also just be that number. Clearly it's easy to get confused with this. I'll wait for more discussion.

drgrice1:
I am not sure what I was thinking when I wrote that. Clearly the gcf and lcm of a positive nonzero integer should be itself. The same is true if all other arguments are zero. Disregard that comment.

Alex-Jordan:
I removed what used to be line 130, added some code commentary, and changes some `if` lines to their `unless` equivalent.

Alex-Jordan:
If someone with merge permissions merges this, I will revisit #470.

enriqueacosta:
I think the definition of `gcd` needs to be updated to use the multi argument`gcf`. I am getting

```
gcd(56124,78432,585)=12     (wrong)
gcf(56124,78432,585)=3      (correct)
```

From the current definition of gcd, I think the issue is that it is just picking up the first two numbers:

```
#greatest common factor.
#same as gcf, but both names are sufficiently common names
# ^function gcd
# ^uses gcf
sub gcd {
        return gcf($_[0], $_[1]);
}
```

drgrice1:
I agree with Enrique.  If you update the gcf method to use multiple arguments, you should also update the gcd method to do the same.  This is as simple as changing the current gcd definition to
```perl
sub gcd {
    return gcf(@_);
}

Alex-Jordan:
@enriqueacosta nice catch! I'll fix that now.

Alex-Jordan:
This has `lcm` and `gcf` throwing an error when not passed any arguments. I'm wondering if I should leave that alone, change it so both return 0,  or change itso both return 1.

The GCF of the empty set is considered 0 in certain conventions following certain reasoning. See [this](https://math.stackexchange.com/questions/1755266/gcd-of-an-empty-set) for example. And then for a set S with at least 1 element, lcm(S) = prod(S)*gcd(S)^{1-|S|}. To extend that to the empty set, if the gcd is 0 and the empty product is 1, then the lcm would be 0.

Alternatively, the smallest positive integer that divides into every member of the empty set is 1. So maybe gcd(empty) should be 1. And then lcm(empty) would be 1 as well.

I think in the context of coding math problems for WeBWorK, if someone passes the empty set to one of these functions, it seems likely that they've made a typo or some other coding error. So I'm inclined to leave it the way it is. But since these alternatives have reasonable arguments, I wanted to ask this group if there are other opinions.

drgrice1:
I think it would be best to leave the gcf and lcm undefined for the empty set.

enriqueacosta:
Agree about leave\ing the gcf and lcm undefined for the empty set.

drgrice1:
I think this pull request is good now.  If there is no opposition, I will merge it in.

mgage:
go ahead and merge it Glenn.

Issue #472, opened 2020-05-17
Possible issue in how num_points and test_at are handled in lib/Value/Formula.pm


taniwallach:
According to the documentation at https://webwork.maa.org/wiki/Context_flags the value of `num_points` should be 
```
The number of random test points to use during function comparisons. This is in addition to any test_at points, but if test_points are given, no random points are given.
```

The generation of the test points seems to be in `lib/Value/Formula.pm`. In particular, the code in `sub compare` uses:
```
  my $points  = $l->{test_points} || $l->createRandomPoints(undef,$l->{test_at});
```
and `sub createRandomPoints` of `lib/Value/Formula.pm` seems to add the points from `test_at` into the array of points
```
    push(@{$points},@{$include});
```
before it runs a loop to add the random points, whose primary end condition is `scalar(@{$points}) < $num_points+$num_undef`.

As such, it seems that the points provided by `test_at` reduce the number of random test points used, unlike what was apparently intended. 

It seems to me that to obtain the behavior intended, the value of `$num_points` should be increased by `scalar(@{$include})` before starting the `while` loop.

@dpvc - Can you take a look?

dpvc:
Your suggestion seems reasonable.  I would recommend incrementing the `$num_points` in this conditional

https://github.com/openwebwork/pg/blob/5083fea21171623bb649b68ecaebd03176ff9af8/lib/Value/Formula.pm#L331-L334

which is where I suspect you already intended to do it.

PR #474, opened 2020-05-18
Fix sub createRandomPoints so that when test_at points are included, …

taniwallach:
Fix `sub createRandomPoints` so that when `test_at` points are included it still generates `num_points` random points in addition to the provided points, as intended. This fixes the issue reported in https://github.com/openwebwork/pg/issues/472 so that the `test_at` works as documented in https://webwork.maa.org/wiki/Context_flags .

### DIFF
--- a/lib/Value/Formula.pm
+++ b/lib/Value/Formula.pm
@@ -331,6 +331,7 @@ sub createRandomPoints {
   if ($include) {
     push(@{$points},@{$include});
     push(@{$values},@{$self->createPointValues($include,1,$cacheResults,$checkUndef)});
+    $num_points += scalar(@{$include});
   }
   my (@P,@p,$v,$i); my $k = 0;
   while (scalar(@{$points}) < $num_points+$num_undef && $k < 10) {


taniwallach:
The easiest way to see how this changes the behavior is to set `diagnostics=>1` inside a `cmp()` call where the relevant MathObject Formula has `test_at` set.  Then during grading a table about the test points and what occurred at each is shown.

A sample PG file which sets 8 `test_at` points and has that flag set is below. 

Before the patch, when graded - that problem would only test at the 8 given points.
After the patch - those 8 points and 5 random points are tested.

---
```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
);

Context("Numeric");
Context()->variables->are(r =>'Real', t =>'Real',p =>'Real');

$int  = Compute("(r^2)*sin(p)*cos(t)");

$int->{test_at} = [ [ 1, 1, 1],
                 [ 1, 1,-1],
                 [ 1,-1, 1],
                 [ 1,-1,-1],
                 [-1, 1, 1],
                 [-1, 1,-1],
                 [-1,-1, 1],
                 [-1,-1,-1],
               ];

#################################################################################

Context()->texStrings;
BEGIN_TEXT
Enter \( $int \) below:

$BCENTER
\{ $int->ans_rule(20)\}
$ECENTER

END_TEXT
Context()->normalStrings;

ANS( $int->cmp( diagnostics=>1 ) );

ENDDOCUMENT();
```

taniwallach:
@drgrice1 Do you mind doing a quick test and merging?

PR #475, opened 2020-05-18
random_coprime and random_pairwise_coprime macros

Alex-Jordan:
This should only be examined after #471 is resolved. It relies on the recursively defined `gcf()` from that PR.

I often need to generate some random relatively prime integers. Typically I resort to either ugly code, or an inefficient `do...until` loop. These subroutines seemed like a better approach and might be useful to others.

Say you want three coprime integers, maybe one nonzero from (-9..9), one in (10..19), and one in (2,3,5,7). Then you can use this like `random_coprime([-9..-1,1..9], [10..19], [2,3,5,7])`. The GCF of the output set will be 1. You might get `(2,13,2)`.

There is also `random_pairwise_coprime` which will not output something like `(2,13,2)`. (Both might output `(1,13,1)` though.)

I think it is important to note that the random selection is uniform over all possible tuples. Some other methods for doing this do not have even distributions.

At first I went to put this with the other random subroutines in PGbasicmacros. But it relies on gcf which is in PGauxiliaryfunctions, so that's where I landed to put this. If anyone has an argument to put it somewhere else, I'm happy to hear and move it.

### DIFF
--- a/macros/PGauxiliaryFunctions.pl
+++ b/macros/PGauxiliaryFunctions.pl
@@ -168,6 +168,11 @@ sub gcd {
 # random_coprime([1..9],[1..9]) to output maybe (2,9) or (1,1) but not (6,8)
 # random_coprime([-9..-1,1..9],[1..9],[1..9]) to output maybe (-3,7,4), (-1,1,1), or (-2,2,3) but not (-2,2,4)
 # random_pairwise_coprime([-9..-1,1..9],[1..9],[1..9]) to output maybe (-3,7,4) or (-1,1,1) but not (-2,2,3)
+# WARNING: random_coprime() will use a lot of memory and CPU resources if used with too many/too large arguments.
+# For example, random_coprime([-20..20],[-20..20],[-20..20],[-20..20],[-20..20]) involves processing 41^5 arrays.
+# Consider using random_pairwise_coprime() instead. Or breaking things up like:
+# random_coprime([-20..20],[-20..20]),random_coprime([-20..20],[-20..20],[-20..20])
+
 
 # ^ function random_coprime
 # ^uses gcd
--- a/macros/PGauxiliaryFunctions.pl
+++ b/macros/PGauxiliaryFunctions.pl
@@ -33,6 +33,12 @@ loadMacros("PGcommonFunctions.pl");
 #  preformat($scalar, "QuotedString")
 #
 
+# Generate random relatively prime tuple
+# (uses gcf(), so it's here, even though this isn't a "function")
+# random_pairwise_coprime($ar1, $ar2, ... )
+# random_coprime($ar1, $ar2, ... )
+
+
 =cut
 
 # ^function step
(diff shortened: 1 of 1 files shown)


drgrice1:
I tested this (on top of PR#471) with
```perl
DOCUMENT();

loadMacros("PGstandard.pl");

TEXT("random_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = ",
    random_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]), $BR);

TEXT("random_pairwise_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = ",
    random_pairwise_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]), $BR);

ENDDOCUMENT();
```
This gives the warnings
```
Use of uninitialized value $next in array dereference at line 181 of [PG]/macros/PGauxiliaryFunctions.pl
Use of uninitialized value $next in array dereference at line 218 of [PG]/macros/PGauxiliaryFunctions.pl
```
and the output
```
random_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = -7142
random_pairwise_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = -6197
```
So something isn't working right.

Alex-Jordan:
OK, I'm sorry for that. My workflow was to paste the subroutine definitions into a PG file and test them out, where it seemed to work. Then on a local (but not active) `pg` repo on my laptop, I pasted them and committed. I'll be more rigorous in future testing. I'll wait until #471 is resolved to continue anything here though.

Alex-Jordan:
I've revised this with an additional commit. Since I rebased from `develop` in the meantime, I force pushed. If you have a local copy of this branch, delete that first.

```
TEXT("random_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = ",
    join(',',random_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7])), $BR);
    
TEXT("random_pairwise_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = ",
    join(',',random_pairwise_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7])), $BR); 
```

produces

```
random_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = -3,14,3
random_pairwise_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = -2,11,5
```

using seed 2821, with no error messages.

enriqueacosta:
Hi @Alex-Jordan , I am not getting lists (just like what Glenn mentioned). Am I doing something wrong?
```
random_coprime([1,2],[1..3],[1..10]) = 21
random_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = 7195
random_pairwise_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = 7152
```

Alex-Jordan:
Glenn's sample did not insert anything between array entries. Your "7152" at the end is "7, 15, 2". Try the one that I posted that uses `join`.

Are you sure about the one producing "21" though? That first line should be producing an array with three numbers.

Alex-Jordan:
You might use this like:
```
($a,$b,$c) = random_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]);
$f = Formula("$a x^2 + $b x + $c")->reduce;
```

And then `$f` is guaranteed to have no common factor in its terms. And you put some restrictions on what `a`, `b`, and `c` can be.

In that example, `random_coprime()` is evaluated in array context, so it returns an array. If you do like:
```
$r = random_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]);
```
then it is evaluated in scalar context and `$r` is an array reference to a corresponding array.

enriqueacosta:
Thanks @Alex-Jordan , I had missed that! Now that I was able to check, I am getting a warning when trying to use with ranges that involve 0. For example,
```
@coeff=random_pairwise_coprime([1..9],[0..9],[0..9]);
```
is giving me 
```
Warning messages
*  Cannot take gcf of the empty set or an all-zero set at line 140 of [PG]/macros/PGauxiliaryFunctions.pl
*  Use of uninitialized value in numeric ne (!=) at line 241 of [PG]/macros/PGauxiliaryFunctions.pl
```
Same happens with `random_coprime([1..9],[0..9],[0..9])`.

Would this not be possible usage? (like in you example, generating a random quadratic so only coeff of x^2 is forced to be non-zero). I see no note in the documentation about being careful with zeros.

In general, I seem to be getting that warning message as soon as two of the three ranges can have a zero in them. For example, the warning is gone with `@coeff=random_pairwise_coprime([-1..9],[1..9],[1..9]);`.

Alex-Jordan:
Thanks @enriqueacosta . What is happening is there are several places where a gcf needs to be computed as part of checking that it is building a coprime set. If 0 is in two of the argument array references, there will be a call for `gcf(0,0)`. And at least as of #471, `gcf(0,0)` is undefined. I'll make a correction to the code for when this happens.

Alex-Jordan:
I added a commit and now `@coeff=random_pairwise_coprime([1..9],[0..9],[0..9]);` produces `(5,1,4)` with no errors (using seed 447.) Thanks for the thorough testing.

Note that if the algorithm ends up constructing an array where `0` is one of the entries, then the only numbers coprime to `0` are `1` and `-1`. So when using `random_pairwise_coprime`, it's probably unwise to include `0` in the arguments. But when using `random_coprime`, you might get out an array like `(5,0,4)` and that's fine.

enriqueacosta:
Ah! Good point about the zeros and random_pairwise_coprime, I had not thought about that.

Alex-Jordan:
@enriqueacosta OK, that makes sense. That example has 41*41*19*8*12>3,000,000 possible tuples. And as implemented, the algorithm needs to check all of their gcds. For me, this took 25 seconds.

I made an obvious edit to the `gcd` subroutine that sped things up, taking this down to 16 seconds. (I changed it to return 1 if the first argument is 1 rather than continuing recursively.)

Then I improved the efficiency of `random_coprime()`, so that it stops computing gcd's of tuples once it's established that the gcd of an earlier tuple from its chain is already 1. (It won't look at gcd(2,4,3,*) because it already found gcd(2,4,3) is 1.) This brought the example down to under 5 seconds.

As far as timing goes, nothing more occurs to me for how to make it more efficient. There is also a memory usage concern. Using your example, there will be a point where over 3 million array refs are being stored in memory, each one with 5 numbers in it. I added a warning to the commentary about this.

enriqueacosta:
Hi @Alex-Jordan, that improved performance noticeably, and I still see it behave as expected (nothing seems to have broken with the last commits). I am afraid I do not know enough Perl to be able to understand your code well enough to suggest performance improvements. The fact that the random selection is uniform over all possible tuples seems to be the right thing to do here.

Alex-Jordan:
I converted leading spaces to tabs across the entire file. Also did a few more edits to make this file conform more closely to https://webwork.maa.org/wiki/Coding_Standards.

Lately I've been writing a lot of deeply nested XML for PreTeXt, which explains why I was indenting by two.

drgrice1 (approved):
Looks much better.

You should check the text editor you use for per file type settings.  All of the good text editors do this.  Vim, emacs, notepad++, etc.  If you are editing an xml file, since deep nesting is common, set the tab stop to 2, and if you are editing perl or similar code where nesting usually doesn't go as deep, have your tab stop to 4 (or stay at 2 if you like).

Alex-Jordan:
PreTeXt discourages/bans tabs in source files. So I had the tab set to two, but also had it set to actually use spaces, not tabs.

drgrice1:
I see.  Sounds like PreTeXt developers could do some work on their parsers to fix that.

drgrice1:
I reviewed this a while ago.  I think this is ready to merge, but it would be good to have someone else take a look.

pstaabp:
I just took a look at this as well and looks good.  And I think I'll use this--I've been writing a few fraction problems for future elementary school teachers.

pstaabp (approved):

taniwallach:
Other than a minor comment that some of the changes apparently changed the output in a sample above - this seems well tested and ready to go. 
> TEXT("random_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = ",
>     join(',',random_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7])), $BR);
>     
> TEXT("random_pairwise_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = ",
>     join(',',random_pairwise_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7])), $BR);

I got (for seed :
```
random_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = -3,10,2
random_pairwise_coprime([-9 .. -1, 1 .. 9], [10 .. 19], [2, 3, 5, 7]) = -2,11,5
```

taniwallach (approved):
I did a quick test and it behaves as described and as expected.
I have not reviewed the code as it was already approved by 3 people.
I'm approving and merging.

macros/PGauxiliaryFunctions.pl
```diff
@@ -127,42 +127,162 @@ sub lcm {
...
+					for my $k (@{$i}) {do {$hasnonzero = 1; last;} if ($k != 0)};
+					do {push @{$newcandidates[1]}, [@{$i}, $j]; next} unless ($hasnonzero or $j != 0);
+					if (gcf($j,@{$i}) == 1) {
+						push @{$newcandidates[0]}, [@{$i}, $j];
+					} else {
+						push @{$newcandidates[1]}, [@{$i}, $j];
+					}
+				}
+			}
+			do {warn "Unable to find a coprime tuple from input"; return;} unless (@{$newcandidates[0]},@{$newcandidates[1]});
```
drgrice1, line 218: I have noticed that this causes a message `Useless use of a variable in void context` in the apache2 error log.  I am not sure what you intended here, but the `unless` conditional doesn't make sense and is what is causing the message.  The second element of the list (`@{$newcandidates[1]}`) is what is in the void context and is being ignored in the conditional.  Did you mean `unless (@{$newcandidates[0]} || @{$newcandidates[1]})`?

PR #479, opened 2020-05-28
Create UploadImages.pl

paultpearson:
Allow students to upload images (up to 21) with persistent storage in $inputs_ref.  Images are stored as base64 encoded dataURLs.  Basic functionality only (no removing images, rotating images, writing on top of images using a sketchpad, etc., yet).

PR #482, opened 2020-06-09
Test develop+upload

mgage:
This is a slightly cleaned up version of #479.  
Because it tries to store all of the uploaded images in the last_answer field it quickly becomes overloaded if there are more than 3 or 4 images.  

It's best if this gets rewritten using PGcore::store_persistent_data and an alternative storage location -- perhaps its own separate table. (PGcore::fetch_persistent_data needs to be written).


taniwallach:
There are new methods to store data... and they should be used for this functionality.

PR #483, opened 2020-06-10
Create GroupTableAnswer.pl macro

paultpearson:
Allows multiplication (group tables) to be created and graded with ease.  An editable html table is used to collect answers (instead of html form inputs) and javascript is used to get the student answers from the table and format them as a MathObject matrix on submit, and to update the table entries using the previous answers on load.

### DIFF
--- a/macros/GroupTableAnswer.pl
+++ b/macros/GroupTableAnswer.pl
@@ -39,7 +39,7 @@ are strings that get interpreted properly by the C<Matrix()> constructor.
         Context("Matrix");
         Context()->variables->are(e=>'Real',a=>'Real');
 
-        $ans = List("e,a,a^2,  a,a^2,e,  a^2,e,a");
+        $ans = Compute("e,a,a^2,  a,a^2,e,  a^2,e,a");
 
 Next, insert the group table into the text of the problem.
 
@@ -178,6 +178,8 @@ sub GroupTable {
             }
             var answers_string = answers_array.join();        
             document.getElementById('AnSwEr_$id').setAttribute('value', answers_string);
+
+            //sessionStorage.setItem("group",document.getElementById("div_$id").innerHTML);
         
         }, false);
         
@@ -187,13 +189,86 @@ sub GroupTable {
         # insert the html hidden input and check the answer
         my $ans_id = 'AnSwEr_' . $id;
         TEXT( NAMED_HIDDEN_ANS_RULE($ans_id) );
-        NAMED_ANS( $ans_id, $ans->cmp(ordered=>1) ); 
+        NAMED_ANS( $ans_id, $ans->cmp(ordered=>1, 'num_rows'=>$num_rows, 'num_cols'=>$num_cols)->withPostFilter(\&group_table_post_filter) ); 
 
         return $table_html . $js;
 
     }    # end HTML mode
 
 }    # end GroupTable
+
+
+
+sub group_table_post_filter {
+    # for answer checking, we use lists; however, we want the answer preview 
+    # and correct answer to be displayed as a table, which is what this post filter does.
+
+    my $ans_hash = shift;
+
+    my $nr = $ans_hash->{num_rows};
+    my $nc = $ans_hash->{num_cols};
+
+    #########################################
+    my $stu = $ans_hash->{student_ans};
+    my @stu_array = split(',' , $stu);
+    #warn scalar(@stu_array);
+
+    my $str = '';
+    foreach my $i (0..($nr-1)) {
+        $str .= join(','  , @stu_array[($i*$nc)..(($i+1)*$nc - 1)]) . '<br>';
+    }
+    $ans_hash->{student_ans} = $str;
+
+    #########################################
+    my $stu = $ans_hash->{preview_latex_string};
+    my @stu_array = split(',' , $stu);
+
+    my $str = '\begin{array}{';
+    $str .= "c" x $nc . '} ';
+    foreach my $i (0..($nr-1)) {
+        $str .= join(' & '  , @stu_array[($i*$nc)..(($i+1)*$nc - 1)]) . ' \cr ';
+    }
+    $str .= ' \end{array}';
+    $ans_hash->{preview_latex_string} = $str;
+
+    ######################################## 
+    my $stu = $ans_hash->{preview_text_string};
+    my @stu_array = split(',' , $stu);
+
+    my $str = '';
+    foreach my $i (0..($nr-1)) {
+        $str .= join(','  , @stu_array[($i*$nc)..(($i+1)*$nc - 1)]) . '<br>';
+    }
+    $ans_hash->{preview_text_string} = $str;
+
+    ########################################
+    my $cor = $ans_hash->{correct_ans_latex_string};
+    my @cor_array = split(',' , $cor);
+    
+    my $str = '\begin{array}{';
+    $str .= "c" x $nc . '} ';
+    foreach my $i (0..($nr-1)) {
+        $str .= join(' & '  , @cor_array[($i*$nc)..(($i+1)*$nc - 1)]) . ' \cr ';
+    }
+    $str .= ' \end{array}';
+    $ans_hash->{correct_ans_latex_string} = $str;
+
+    ######################################## 
+    my $cor = $ans_hash->{correct_ans};
+    my @cor_array = split(',' , $cor);
+
+    my $str = '';
+    foreach my $i (0..($nr-1)) {
+        $str .= join(','  , @cor_array[($i*$nc)..(($i+1)*$nc - 1)]) . '<br>';
+    }
+    $ans_hash->{correct_ans} = $str;
+
+    ########################################
+    # for debugging
+    #warn pretty_print($ans_hash);
+
+    return $ans_hash;
+}
 
 
 1;
--- a/macros/GroupTableAnswer.pl
+++ b/macros/GroupTableAnswer.pl
@@ -166,8 +166,8 @@ sub GroupTable {
 
             var i, j;
             var answers_array = [];
-            for (i = 1; i < 4; i++) {
-                for (j = 1; j < 4; j++) {
+            for (i = 1; i <= $num_rows; i++) {
+                for (j = 1; j <= $num_cols; j++) {
                     var cell = document.getElementById("$id").rows[i].cells[j].innerHTML;
       		    if (cell.length > 0) {
        			answers_array.push(cell);
--- a/macros/GroupTableAnswer.pl
+++ b/macros/GroupTableAnswer.pl
@@ -32,24 +32,35 @@ the html table cells.
 
 Load the macro file via C<loadMacros("GroupTableAnswer.pl")>.
 
-Begin by adding some named variables to the context and enter the group table 
-into a matrix, being sure to use quotes around entries in the matrix so that they 
-are strings that get interpreted properly by the C<Matrix()> constructor.
+Begin by adding some named constants to the context.  We use the Matrix context
+here, but we could have used the Permutation context, or for an abelian
+group we could even use the Numeric context.  We use parsed student
+answers so that student answers are printed by name rather than value.
 
         Context("Matrix");
-        Context()->variables->are(e=>'Real',a=>'Real');
 
-        $ans = Compute("e,a,a^2,  a,a^2,e,  a^2,e,a");
+        $I = Matrix([[1,0],[0,1]]);
+        $A = Matrix([[0,1],[1,0]]);
+        $B = Matrix([[0,1],[-1,-1]]);
+        $C = Matrix([[-1,-1],[0,1]]);
+        $D = Matrix([[-1,-1],[1,0]]);
+        $K = Matrix([[1,0],[-1,-1]]);
 
-Next, insert the group table into the text of the problem.
+        Context()->constants->are( I=>$I, A=>$A, B=>$B, C=>$C, D=>$D, K=>$K );
+        Context()->operators->undefine("+","-","inverse");
+        Context()->flags->set(
+            formatStudentAnswer=>'parsed'
+        );
+
+        $grouptable = Compute("I,A,B,C,D,K,  A,I,C,B,K,D,   B,K,D,A,I,C,   C,D,K,I,A,B,   D,C,I,K,B,A,   K,B,A,D,C,I");
 
         BEGIN_PGML
-        [@ GroupTable('grouptable_1', 'Group Table', '\circ', ['e','a','a^2'],['e','a','a^2'], $ans); @]*
-        END_PGML
+        # Group table input
 
-        BEGIN_TEXT
-        \{ GroupTable('grouptable_1', 'Group Table', '\circ', ['e','a','a^2'],['e','a','a^2'], $ans); \}
-        END_TEXT
+        Suppose [``I=[$I]``], [``A=[$A]``], [``B=[$B]``], [``C=[$C]``], [``D=[$D]``], [``K=[$K]``].  Complete the group table for the group defined by these six elements.
+
+        [@ GroupTable('grouptable_1', 'Group table', '\circ', ['I','A','B','C','D','K'], ['I','A','B','C','D','K'], $grouptable) @]*
+        END_PGML
 
 The first argument C<'grouptable_1'> is a unique identifier that gets used in the
 html source code.  If you have more than one group table in a problem, their
@@ -61,9 +72,9 @@ by people with visual impairments who use screen readers.
 The third argument C<'\circ'> is a TeX-formatted string for the operation in the group.
 If this string contains a backslash, be sure to enclose it in single quotes.
 
-The fourth and fifth arguments C<['e','a','a^2']> are the row and column headers.
+The fourth and fifth arguments C<['I','A','B','C','D','K']> are the row and column headers.
 
-The sixth argument C<$ans> is a MathObject List of answers.
+The sixth argument C<$grouptable> is a MathObject list of answers in row major format.
 
 =head1 AUTHOR
 
--- a/macros/GroupTableAnswer.pl
+++ b/macros/GroupTableAnswer.pl
@@ -187,9 +187,13 @@ sub GroupTable {
        		    }
                 }
             }
-            var answers_string = answers_array.join();        
+
+            //  create a string by joining elements of the array by commas, and then replacing any html tags using regex match and replace
+            //  https://stackoverflow.com/questions/1499889/remove-html-tags-in-javascript-with-regex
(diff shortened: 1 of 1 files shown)


paultpearson:
Here's an example pg file to test out the macro.

```
DOCUMENT();        
loadMacros(
"PGstandard.pl",
"PGML.pl",
"GroupTableAnswer.pl",
);
TEXT(beginproblem());

Context("Matrix");
Context()->variables->are(e=>'Real',a=>'Real');

$ans = Matrix([
['e','a','a^2'],
['a','a^2','e'],
['a^2','e','a'] ]);

BEGIN_PGML
# Group table input

Suppose [`a`] is an element of order [`3`].  Complete the group table for [`G = \langle a \rangle`]. 

[@ GroupTable('grouptable_1', 'Group table', '\circ', ['e','a','a^2'], ['e','a','a^2'], $ans) @]*
END_PGML

ENDDOCUMENT();
```

Alex-Jordan:
I tried this and it did not work for me. The server I used is on the most recent `develop` branch for `webwork2`, and then I checked out this branch for `pg`. I was viewing in Firefox.

 I tried the example pg file, and at least in HTML, it show the problem fine. But after I hit submit/check, I get WeBWorK warning messages.

```
Use of uninitialized value in join or string at line 231 of [PG]/macros/GroupTableAnswer.pl
Use of uninitialized value in join or string at line 242 of [PG]/macros/GroupTableAnswer.pl
Use of uninitialized value in join or string at line 253 of [PG]/macros/GroupTableAnswer.pl 
```
(Actually each of those happens 9 times.)

In the results table, in the "Entered" column, I get:
```
e<br>,a<br>,a^2<br>
a,a^2,e
a^2,e,a
```

This made me wonder about something. I put the cursor in cell (2,1) and hit `ENTER` after the `a`. (It seems like something a student might do.) This created a new line. Then I clicked check/submit and the results table has this in the "Entered" column:
```
e<br>,a<br>,a^2<br>
<div>a</div><div><br></div>,a^2,e
a^2,e,a
```

So a lot more is recorded as the student's answer than you would want. 

PDF hardcopy looks good, except that I looked at that after the above, and in the "Answer(s) submitted" area, I see `\begin{verbatim}e<br>,a<br>,a^2<br>,a,a^2,e,a^2,e,a\end{verbatim}`.

I reloaded the problem from scratch, and one thing I note now in the HTML is that when I put my cursor in a cell from the first row, it is centered. Everywhere else, it is flush left. Even without typing any answers, the inspector shows me those `br`s are there in the cells from the top row even before typing answers. Actually they are in there as `<br type="_moz">` and I see no "moz" in your file, so I wonder if Firefox is doing something here.

I tried in Chrome, and I do not get those `<br>`s.  If I hit enter within a cell and then sumbit, I do not get those `<div>`s _but_ then I do get `<br>`s. And in Chrome I still get those WeBWorK warning messages, although I get fewer of them.

paultpearson:
Hi Alex,

Thanks for testing this out.  I am bummed out that it's not working properly and that there may be browser-specific issues that would need resolution before it might work well.  Of course, we can implement client- or server-side string substitutions that will remove unwanted characters, but figuring out what strings are inserted and need to be removed might be tricky.

I don't know if this is related, but I'll mention it here anyway. The error message `Use of uninitialized value in join or string at line...` was exactly the error message I was getting using `GroupTable()` inside a section of a scaffold on Google Chrome.  Related to this, I filed a [bugzilla bug report 4430](https://bugs.webwork.maa.org/show_bug.cgi?id=4430) saying that I thought the root cause of the problem was in `scaffold.pl` as I was able to determine that `scaffold.pl` was running some code twice instead of once using an example that did not have anything to do with GroupTable.  So, I was intrigued to see that you were getting the error message without using `scaffold.pl`.

Do we think that it's a good idea to troubleshoot and bugfix this macro, or is the whole idea of using an editable html table with client-side javascript to assemble the answers from the table a bad idea? 
 If anyone has any ideas or suggestions, please feel free to comment here.  

My whole reason for writing this macro was to avoid having to use `parserMultiAnswer.pl` to grade a group table and to try to do a proof of concept (editable html table + javascript works).  There's no way to use Matrices instead, because I need non-commuting variables that satisfy at least (xy)^(-1) = y^(-1) x^(-1) and I haven't figured out how to make that work yet.

Thanks!
Paul

Alex-Jordan:
I think the aesthetic of the editable table is interesting. Just to understand that device better and learn how to leverage it in other places, I'm game to keep testing things.

I haven't looked through the code, but how does this handle nonabelian groups? Say you want the table for $D_6$? How will the `GroupTable()` synatx look?

paultpearson:
For D6, the group table syntax will look the same because all of the answer checking is handled by MathObjects.  In the MathObject Matrix context (and I would assume the Permutation context, too) you can add named constants to the context that are matrices (or permutations) and are thus named non-commutative objects.  I should probably update the documentation to use this example for D6 instead:

```
DOCUMENT();        
loadMacros(
"PGstandard.pl",
"PGML.pl",
"GroupTableAnswer.pl",
);
TEXT(beginproblem());

Context("Matrix");
$I = Matrix([[1,0],[0,1]]);
$A = Matrix([[0,1],[1,0]]);
$B = Matrix([[0,1],[-1,-1]]);
$C = Matrix([[-1,-1],[0,1]]);
$D = Matrix([[-1,-1],[1,0]]);
$K = Matrix([[1,0],[-1,-1]]);
Context()->constants->are( I=>$I, A=>$A, B=>$B, C=>$C, D=>$D, K=>$K );
Context()->operators->undefine("+","-","inverse");
Context()->flags->set(
    formatStudentAnswer=>'parsed'
);

$grouptable = Compute("I,A,B,C,D,K,  A,I,C,B,K,D,   B,K,D,A,I,C,   C,D,K,I,A,B,   D,C,I,K,B,A,   K,B,A,D,C,I");

BEGIN_PGML
# Group table input

Suppose [``I=[$I]``], [``A=[$A]``], [``B=[$B]``], [``C=[$C]``], [``D=[$D]``], [``K=[$K]``].  Complete the group table for the group defined by these six elements.

[@ GroupTable('grouptable_1', 'Group table', '\circ', ['I','A','B','C','D','K'], ['I','A','B','C','D','K'], $grouptable) @]*
END_PGML

ENDDOCUMENT();
```

paultpearson:
I have updated the documentation in the macro file to give the dihedral group of order 6 as the main example.  However, I have not yet updated the macro to address the issues that Alex Jordan pointed out, so the macro is not yet ready for merging into pg-devel.

Alex-Jordan:
OK. What I was getting at is that the problem author still has to do this part:
```
$grouptable = Compute("I,A,B,C,D,K,  A,I,C,B,K,D,   B,K,D,A,I,C,   C,D,K,I,A,B,   D,C,I,K,B,A,   K,B,A,D,C,I");
```
There is no expectation of WeBWorK working out the answers for you. For a moment, I thought that by specifying the row and column headers, something would multiply the nx1 vector with a 1xn vector and automatically end up with the (reduced) products in a nxn matrix for you.

Suppose you are a student answering in the cell for `A*A`. How do you prevent them just entering "A*A" instead of "I"?
* You could disallow multiplication. However I was thinking this might be used for a row and column like `[1,a,a^2,b,ab,a^2b]` where you expect multiplication in some answers in some cells. But you expect `ab`, not `ba^2`.
* You could disallow multiplication and make the constants be longer strings like "a^2b". But then you'd wrestle with students typing "a^2 b" instead of a^2b".

Alex-Jordan:
I wonder if the principles behind `bizarroArithmetic` could be used here. In your example, you leveraged 2x2 matrices to represent the group. But what if `I,A,B,C,D,K` (or `1,a,b`) are just regular variables, and we change the multiplication operator into something noncommutative? If the official answer in a cell is `ab`, then we could recognize that `ba` is not the same thing. If we also leveraged the matrix representations, we could additionally recognize that `ba^2` is the same, but not presented in the right form, and give students special feedback.

I'll think about if `bizarro` can help. We'd still want associativity. And want a bizarro exponentiation that agrees with repeated bizarro multiplication.

paultpearson:
Regarding "There is no expectation of WeBWorK working out the answers for you.":

When adding constants to the context, using prime multiples of the matrices is one easy way to ensure students must reduce their answers.  I have updated the documentation to show this.

Regarding automatic generation of the multiplication table: This can be done using hashes (multiply name1 * name2 to get a value, then look up the name for that value) and I looked into it earlier but was annoyed because hashes aren't ordered and importing ordered hashes was too much for me to handle at the time.  I can play around with it some more, though.

paultpearson:
Regarding using bizarro arithmetic to create a non-abelian context: I have tried to make that work, but my difficulty lies in creating something that is both "bizarre" and satisfies (xy)^(-1) = y^(-1)x^(-1), which is a fundamental relation in every group.  I would love to have a general non-abelian context to use.

Here's what I've tried for the `contextNonAbelian.pl` macro file:

```
=head1 NAME

contextNotAbelian.pl - Enables noncommutative multiplication. 

CURRENTLY THIS DOES NOT WORK BECAUSE (xy)^(-1) does equal y^(-1)x^(-1) but does not equal itself.  This is a fundamental relation among variables that cannot be left out.

=head1 DESCRIPTION

The point of this technique is to catch answers in unsimplified forms. A custom answer checker is generally used whereby the student answer is first checked against the correct via regular arithmetic. Then one or more of the bizarro arithmetic flags should be set, and the two answers should be compared again.

The bizarro arithmetic is basically defined as:

  a bizarroOp b = f(g(a) regularOp g(b))

where f and g are inverse functions defined on all of R, and are odd functions. Applications suggest that f(x) = x^3+3x is a good choice.
  
To enable the bizarro arithmetic operators, load this file with the macros, and then use any subset of:

Context()->operators->set(
'+' => {class => 'bizarro::BOP::add', isCommand => 1},
'-' => {class => 'bizarro::BOP::subtract', isCommand => 1},
'*' => {class => 'bizarro::BOP::multiply', isCommand => 1},
' *' => {class => 'bizarro::BOP::multiply', isCommand => 1},
'* ' => {class => 'bizarro::BOP::multiply', isCommand => 1},
'/' => {class => 'bizarro::BOP::divide', isCommand => 1},
' /' => {class => 'bizarro::BOP::divide', isCommand => 1},
'/ ' => {class => 'bizarro::BOP::divide', isCommand => 1},
'//' => {class => 'bizarro::BOP::divide', isCommand => 1},
'**' => {class => 'bizarro::BOP::power', isCommand => 1, perl=>undef},
'^' => {class => 'bizarro::BOP::power', isCommand => 1, perl=>undef},
);

and then turn on the bizrro arithmetic with:

Context()->flags->set(bizarroAdd=>1);
Context()->flags->set(bizarroSub=>1);
Context()->flags->set(bizarroMul=>1);
Context()->flags->set(bizarroDiv=>1);
Context()->flags->set(bizarroPow=>1);

Sample usage 

This will check if a student has simplified according to the rules of division, 
for example
 
         2 x^4/x^2

will be rejected, as will

       4/2 x^2

assuming that
 
      $ans = Formula("2x^2");

ANS($ans -> cmp(
   checker=>sub{
      my ( $correct, $student, $ansHash ) = @_;
      return 0 if $ansHash->{isPreview} || $correct != $student;
      $student = $ansHash->{student_formula};
      $correct = $correct->{original_formula} if defined $correct->{original_formula};
      $student = Formula("$student"); $correct = Formula("$correct");
      return 0 unless ($correct == $student);
      Context()->flags->set(bizarroDiv=>1);
      delete $correct->{test_values};
      delete $student->{test_values};
      Value->Error("Your answer is correct, but please simplify it further") unless (($correct == $student) or ($student == $correct));
      Context()->flags->set(bizarroDiv=>0);
      return 1;
})); 

=cut




###########################
#
#  functions used in defining non Abelian arithmetic
#

package NotAbelian;

sub f {
  my $x = shift;
  return ($x**3 + 3*$x);
}

sub g {
  my $x = shift;
  # return (cuberoot(($x+sqrt(($x)**2+4))/2) + cuberoot(($x-sqrt(($x)**2+4))/2)); # June 10, pretty good but (xy)^(-1) _____ works for y^(-1)x^(-1) but not for (xy)^(-1)
  return exp($x);
}

sub cuberoot {
  my $x = shift;
  return 0 if ($x == 0);
  return abs($x)/$x*(abs($x))**(1/3);
};

###########################
#
#  Subclass the addition
#
package NotAbelian::BOP::add;
our @ISA = ('Parser::BOP::add');

sub _eval {
  my $self = shift;
  my $context = $self->context;
  my ($a,$b) = @_;
  if ($context->flag("NotAbelianAdd")) {
    if (($a == 0) or ($b == 0)) {return 2*$a+2*$b+1;}
    else { return $a + 2*$b };
  } else {
    return $a + $b;
  }
}

sub call {(shift)->_eval(@_)}

###########################
#
#  Subclass the subtraction
#
package NotAbelian::BOP::subtract;
our @ISA = ('Parser::BOP::subtract');

sub _eval {
  my $self = shift;
  my $context = $self->context;
  my ($a,$b) = @_;
  if ($context->flag("NotAbelianSub")) {
    if (($a == 0) or ($b == 0)) {return 2*$a+2*(-$b)+1;}
    else {return $a - 2*$b };
  } else {
    return $a - $b;
  }
}

sub call {(shift)->_eval(@_)}

###########################
#
#  Subclass the multiplication
#
package NotAbelian::BOP::multiply;
our @ISA = ('Parser::BOP::multiply');

sub _eval {
  my $self = shift;
  my $context = $self->context;
  my ($a,$b) = @_;
  if ($context->flag("NotAbelianMul")) {
    return $a * ($b)**3;
    #return $a + 2 * $b;
  } else {
    return $a * $b;
  }
}

sub call {(shift)->_eval(@_)}


###########################
#
#  Subclass the division
#
package NotAbelian::BOP::divide;
our @ISA = ('Parser::BOP::divide');

sub _eval {
  my $self = shift;
  my $context = $self->context;
  my ($a,$b) = @_;
  if ($context->flag("NotAbelianDiv")) {

    # commented out June 10
    #if ($a == 1) {return $a - 2 * $b;}
    #else {return $a - 2 * $b;}

    return $a / (($b)**3);

    #return NotAbelian::cuberoot($a) * $b;
  } else {
    return $a / $b;
  }
}

sub call {(shift)->_eval(@_)}

###########################
#
#  Subclass the power
#
package NotAbelian::BOP::power;
our @ISA = ('Parser::BOP::power');

sub _eval {
  my $self = shift;
  my $context = $self->context;
  my ($a,$b) = @_;
  if ($context->flag("NotAbelianPow")) {
    if (($a == 1) or ($b == 0)) {return ($a+1)**($b+1);}
    else { 
      # return ($a)**($b+1) 
      return ($a) - 2*($b);
    }
  } else {
    return $a ** $b;
  }
}

sub call {(shift)->_eval(@_)}

1;
```

paultpearson:
Here's a pg file I've used with `contextNonAbelian.pl`

```
DOCUMENT();

loadMacros(
"PGstandard.pl",
"PGML.pl",
"contextNotAbelian.pl",
);
TEXT(beginproblem());

Context("Numeric")->variables->are(x=>'Real',y=>'Real');
Context()->flags->set(
    formatStudentAnswer=>'parsed'
);
#Context()->operators->undefine("+","-","inverse","**","^","/");
Context()->operators->undefine("+","-","inverse");

Context()->operators->set(
'*' => {class => 'NotAbelian::BOP::multiply', isCommand => 1},
' *' => {class => 'NotAbelian::BOP::multiply', isCommand => 1},
'* ' => {class => 'NotAbelian::BOP::multiply', isCommand => 1},
#'/' => {class => 'NotAbelian::BOP::divide', isCommand => 1},
#' /' => {class => 'NotAbelian::BOP::divide', isCommand => 1},
#'/ ' => {class => 'NotAbelian::BOP::divide', isCommand => 1},
#'//' => {class => 'NotAbelian::BOP::divide', isCommand => 1},
'**' => {class => 'NotAbelian::BOP::power', isCommand => 1, perl=>undef},
'^' => {class => 'NotAbelian::BOP::power', isCommand => 1, perl=>undef},
);
Context()->flags->set(NotAbelianMul=>1);
#Context()->flags->set(NotAbelianDiv=>1);
Context()->flags->set(NotAbelianPow=>1);


BEGIN_PGML
## Non Abelian Product

If a group is not abelian, then 
[`xy=`] [_____]{"xy"} [`\ne`] [______]{"yx"} [`=yx`] for some [`x`] and [`y`].

[`x^2y=`] [________]{"x^2y"}

[`xy^{-1}=`] [________]{"xy^(-1)"}

[`xy^{-2}=`] [________]{"xy^(-2)"}

[`(xy)^{-1}=`] [________]{"y^(-1) x^(-1)"} 

END_PGML

ENDDOCUMENT();

```

macros/GroupTableAnswer.pl
```diff
@@ -0,0 +1,292 @@
...
+    ######################################## 
+    my $cor = $ans_hash->{correct_ans};
+    my @cor_array = split(',' , $cor);
+
+    my $str = '';
+    foreach my $i (0..($nr-1)) {
+        $str .= join(','  , @cor_array[($i*$nc)..(($i+1)*$nc - 1)]) . '<br>';
+    }
+    $ans_hash->{correct_ans} = $str;
+
```
drgrice1, line 283: Lines 263 - 282 should be taken out of this method and moved into the GroupTable method above.  You will need to make a few minor modifications to make it work there.  This shouldn't be done when the student answers are processed, but should be done when the cmp object is created.

macros/GroupTableAnswer.pl
```diff
@@ -0,0 +1,292 @@
...
+    my $nr = $ans_hash->{num_rows};
+    my $nc = $ans_hash->{num_cols};
+
+    #########################################
+    my $stu = $ans_hash->{student_ans};
+    my @stu_array = split(',' , $stu);
+    #warn scalar(@stu_array);
+
+    my $str = '';
```
drgrice1, line 234: $stu, @stu_array, and $str are shadowed multiple times below.  This is extremely bad programming practice, and Perl will give warnings about this if "use warnings" is given.

drgrice1:
I managed to get your initial problem to accept the correct answer as correct by changing the Matrix answer into a flat list as in your D6 example.  You should update your initial problem that you posted to avoid confusion for others testing this.

taniwallach:
@paultpearson - Can you spare some time to address @drgrice1's comments so we can consider whether this can be considered for WW/PG 2.16?

paultpearson:
I do not have time for this right now. Please do not merge yet.

pstaabp:
@paultpearson Have you had any time to look into this?  We're looking at WW2.17 right now and we could probably get this in the release if we can iron out the bugs.

paultpearson:
I still have no time to work on this.

dlglin:
Closed due to inactivity.  Can be reopened if activity resumes.

PR #484, opened 2020-06-16
Graphtool macro for interactive graphing in problems via javascript

drgrice1:
This adds a macro that can be used in PG problems to inject an interactive javascript graph into the problem.  This is primarily intended for basic general studies type courses where graphing is introduced.  Admittedly the javascript graph tool (in https://github.com/openwebwork/webwork2/pull/1116) is largely designed to be similar to a graphing tool in a particular proprietary online homework system.

I have attached a few example pg problems demonstrating the usage of the macro:  [GraphToolExamples.zip](https://github.com/openwebwork/pg/files/4786200/GraphToolExamples.zip)

I have been using this macro for problems in my courses for several semesters, and it has been working well.  Students find its usage in problems rather straight forward.

### DIFF
--- a/macros/parserGraphTool.pl
+++ b/macros/parserGraphTool.pl
@@ -270,6 +270,7 @@ sub ans_rule {
 		($self->{bBox}[0],$self->{bBox}[3]) rectangle ($self->{bBox}[2],$self->{bBox}[1]) }
 	}
 }
+\\definecolor{borderblue}{HTML}{356AA0}
 \\pgfdeclarelayer{background layer}
 \\pgfdeclarelayer{foreground layer}
 \\pgfsetlayers{background layer,main,foreground layer}
@@ -323,7 +324,7 @@ END_TIKZ
 		if (@yTicks);
 
 		# Border box
-		$tikz .= "\\draw[blue,rounded corners=14pt,thick] " .
+		$tikz .= "\\draw[borderblue,rounded corners=14pt,thick] " .
 			"($self->{bBox}[0],$self->{bBox}[3]) rectangle ($self->{bBox}[2],$self->{bBox}[1]);\n";
 
 
--- a/macros/parserGraphTool.pl
+++ b/macros/parserGraphTool.pl
@@ -1,0 +1,520 @@
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2018 The WeBWorK Project, http://openwebwork.sf.net/
+# $CVSHeader: pg/macros/parserMultiAnswer.pl,v 1.11 2009/06/25 23:28:44 gage Exp $
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+# Artistic License for more details.
+################################################################################
+
+=head1 NAME
+
+parserGraphTool.pl - Allow students to enter basic graphical answers via interactive JavaScript.
+
+=head1 DESCRIPTION
+
+GraphTool objects let you provide an interactive graphing tool for students to enter graphical
+answers.
+
+To create a GraphTool object pass a list of graph objects (discussed below) for the students to
+graph to GraphTool().  For example:
+
+	$gt = GraphTool("{line,solid,(0,0),(1,1)}", "{circle,dashed,(2,2),(4,2)}");
+
+or
+
+	$gt = GraphTool("{line,solid,(0,0),(1,1)}")->with(bBox => [-20, 20, 20, -20]);
+
+Then, for standard PG use $gt->ans_rule() to insert the JavaScript graph into the problem (or a
+print graph when a hard copy is generated), and $gt->cmp to produce the answer checker.  For
+example:
+
+	BEGIN_TEXT
+	Graph the line \(y = x\).
+	$PAR
+	\{$gt->ans_rule()\}
+	END_TEXT
+
+	ANS($gt->cmp);
+
+For PGML you can just do
+
+	BEGIN_PGML
+	Graph the line [`y = x`].
+
+	[_]{$gt}
+	END_PGML
+
+=head1 GRAPH OBJECTS
+
+There are four types of graph objects that the students can graph.  Lines, circles, parabolas,
+and fills (or shading of a region).  The syntax for each of these objects to pass to the
+GraphTool constructor is summarized as follows.  Each object must be enclosed in braces.  The
+first element in the braces must be the name of the object.  The following elements in the
+braces depend on the type of element.
+
+For lines the name "line" must be followed by the word "solid" or "dashed" to indicate if the
+line is expected to be drawn solid or dashed.  That is followed by two distinct points on the
+line.  For example:
+
+	"{line,dashed,(1,5),(3,4)}"
+
+For circles the name "circle" must be followed by the word "solid" or "dashed" to indicate if
+the circle is expected to be drawn solid or dashed.  That is followed by the point that is to be
+the center of circle, and then by a point on the circle.  For example:
+
+	"{circle,solid,(1,1),(4,5)}"
+
+For parabolas the name "parabola" must be followed by the word "solid" or "dashed" to indicate
+if the parabola is expected to be drawn solid or dashed.  The next element in the braces must be
+the word "vertical" for a parabola that opens up or down, or "horizontal" for a parabola that
+opens to the left or right.  That is followed by the vertex and then another point on the
+parabola.  For example:
+
+	"{parabola,solid,vertical,(1,0),(3,3)}"
+
+For fills the name "fill" must be followed by a point in the region that is to be filled.  For
+example:
+
+	"{fill,(5,5)}"
+
+The student answers that are returned by the JavaScript will be a list of the list objects
+discussed above and will be parsed by WeBWorK and passed to the checker as such.  The default
+grader is the default list_checker.  Most of the time that will not work as desired, and you
+will need to provide your own list_checker.  This can either be passed as part of the cmpOptions
+hash discussed below, or directly to the GraphTool object's cmp() method.
+
+=head1 OPTIONS
+
+There are a number of options that you can supply to control the appearance and behavior of the
+JavaScript graph, listed below.  These are set as parameters to the with() method called on the
+GraphTool object.
+
+=over
+
+=item bBox (Default: bBox => [-10, 10, 10, -10])
+
+This is an array of four numbers that represent the bounding box of the graph.  The first
+two numbers in the array are the coordinates of the top left corner of the graph, and the last
+two numbers are the coordinates of the bottom right corner of the graph.
+
+=item gridX, gridY (Default: gridX => 1, gridY => 1)
+
+These are the distances between successive grid lines in the x and y directions, respectively.
+
+=item ticksDistanceX, ticksDistanceY (Default: ticksDistanceX => 2, ticksDistanceY => 2)
+
+These are the distances between successive major (labeled) ticks on the x and y axes,
+respectively.
+
+=item minorTicksX, minorTicksY (Default: minorTicksX => 1, minorTicksY => 2)
+
+These are the number of minor (unlabeled) ticks between major ticks on the x and y axes,
+respectively.
+
+=item snapSizeX, snapSizeY (Default: snapSizeX => 1, snapSizeY => 1)
+
+These restrict the x coordinate and y coordinate of points that can be graphed to being
+multiples of the respective parameter.  These values must be greater than zero.
+
+=item staticObjects (Default: staticObjects => [])
+
+This is an array of fixed objects that will be displayed on the graph.  These objects will not
+be able to be moved around.  The format for these objects is the same as those that are passed
+to the GraphTool constructor as the correct answers.
+
+=item graphOptions (Default: undefined)
+
+This is an advanced option that you usually do not want to use.  It is usually constructed by
+the macro internally using the above options.  If defined it should be a single string
+containing three comma separated arguments, and will override all of the above options.  It will
+be passed directly to the JavaScript graphTool method as its final three parameters.  The first
+argument in the string is JavaScript object notation that will be passed directly to the JSX
+graph board when it is initialized.  It may consist of any of the valid attributes documented
+for JXG.JSXGraph.initBoard at L<https://jsxgraph.org/docs/symbols/JXG.JSXGraph.html#.initBoard>.
+The second and third arguments are the snapSizeX and snapSizeY options discussed above.  For
+example the following value for graphOptions will give the same result for the JavaScript graph
+as the default values for the options above:
+
+	graphOptions => "{ boundingBox: [-10, 10, 10, -10]," .
+		"defaultAxes: {" .
+			"x: { ticks: { ticksDistance: 2, minorTicks: 1} }," .
+			"y: { ticks: { ticksDistance: 2, minorTicks: 1} }" .
+		"}," .
+		"grid: { gridX: 1, gridY: 1 }" .
+	"}, 1, 1"
+
+=item printGraph (Default: undefined)
+
+If the graphOptions option is set directly, then you will also need to provide a function that
+will generate the corresponding hard copy graph.  Otherwise the hard copy graph will still be
+generated using the above options, and will not look the same as the java script graph.
+
+=item cmpOptions (Default: cmpOptions => {})
+
+This is a hash of options that will be passed to the cmp() method.  These options can also be
+passed as parameters directly to the GraphTool object's cmp() method.
+
+=item texSize (Default: texSize => 400)
+
+This is the size of the graph that will be output when a hard copy of the problem is generated.
+
+=back
+
+=cut
+
+sub _parserGraphTool_init {
+	if ($main::displayMode ne 'TeX' && !$main::GraphToolHeaderSet) {
+		main::TEXT(
+			'<link rel="stylesheet" type="text/css" href="/webwork2_files/js/vendor/jsxgraph/jsxgraph.css">' .
+			'<link rel="stylesheet" type="text/css" href="/webwork2_files/js/apps/GraphTool/graphtool.css">' .
+			'<script type="text/javascript" src="/webwork2_files/js/vendor/jsxgraph/jsxgraphcore.js"></script>' .
+			'<script type="text/javascript" src="/webwork2_files/js/apps/GraphTool/graphtool.min.js"></script>'
+		);
+		$main::GraphToolHeaderSet = 1;
+	}
+	main::PG_restricted_eval('sub GraphTool { GraphTool->new(@_) }');
+}
+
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Thanks for this. I did some testing today, and it worked. I have some observations that might make it more accessible.

* Say you are making a circle. You click once to plant a center, then you click a second time to define a radius. If you mis-planted the center, it would be nice if you could hit ESC to abort. But it seems you must make a circle, then clear it. (I do like how you must click twice instead of clicking once and using drag to identify a radius. This seems more accessible to people with mobility disabilities.)

* The circle tool requires me to set a radius by choosing two grid points. So for example I would not be able to construct the circle `x^2+y^2=3`, since 3 is not the sum of two squares. Is it possible for there to be a different circle tool where you click a center and type a radius?

* In the problem I loaded, the curves are blue and the fill color is yellow. I assume that is the default. Do you happen to know what is the hex value of that yellow? I'd like to see if it satisfies color contrast threshold with the white. I think that WCAG would want the contrast ratio to be 4.5:1 for the yellow compared to the white.

* Is there any hope of keyboard-only usability? If I keyboard navigate to a tool, say the circle tool, then I can hit SPACE to activate it. But then I cannot use the keyboard to move the pointer around.

* Within one problem, is it configurable to omit some of the tools? For example, could the solid-line, dashed-line, and shading tools be left off?


* I have some feedback on the sample problems, given that they may end up being used as templates later. I'll leave that out of this thread for now.

drgrice1:
>     * Say you are making a circle. You click once to plant a center, then you click a second time to define a radius. If you mis-planted the center, it would be nice if you could hit ESC to abort. But it seems you must make a circle, then clear it. (I do like how you must click twice instead of clicking once and using drag to identify a radius. This seems more accessible to people with mobility disabilities.)

You can also click on another tool to cancel graphing an object.  For example if you are graphing a circle and have graphed the center but not the other point, then click on the "Object Selection Tool" and it will cancel graphing of the circle.

It may also be possible to make the escape button cancel graphing, but may take some work.  The underlying javascript library jsxgraph does not expose keyboard events.

>     * The circle tool requires me to set a radius by choosing two grid points. So for example I would not be able to construct the circle `x^2+y^2=3`, since 3 is not the sum of two squares. Is it possible for there to be a different circle tool where you click a center and type a radius?

See comment below on tool configurability.

>     * In the problem I loaded, the curves are blue and the fill color is yellow. I assume that is the default. Do you happen to know what is the hex value of that yellow? I'd like to see if it satisfies color contrast threshold with the white. I think that WCAG would want the contrast ratio to be 4.5:1 for the yellow compared to the white.

The RGBA colors are 255,255,150,250 which translate into a hex code of #FFFF96.  I doubt that this attains a contrast ratio of 4.5:1.  I am not convinces that the WCAG guidelines for text readability entirely apply to this situation though.  That is for a foreground/background ratio that is best for "text" readability.  This color should really be considered a background color, and it is probably more important that other text in the fill area is still readable (like axis labels).  It would be good to ensure that the fill color is easily distinguishable from the background color also.  Note that the colors chosen are the colors used by the comparable graphing tool in Pearson MyLab, and they are also concerned with accessibility issues.

>     * Is there any hope of keyboard-only usability? If I keyboard navigate to a tool, say the circle tool, then I can hit SPACE to activate it. But then I cannot use the keyboard to move the pointer around.

It may be possible, but will take considerable effort to implement.  As noted above the jsxgraph library does not expose keyboard events.  Furthermore, it would take considerable though as to how exactly this should work.

>     * Within one problem, is it configurable to omit some of the tools? For example, could the solid-line, dashed-line, and shading tools be left off?

I intend to make the tools that are present configurable in this way, but that is a future project.  I also would like to make it so that custom tools could be added by a particular problem.  This will take some revision to the current javascript.

drgrice1:
I made it so the escape key cancels graphing of an object.  See the last push to https://github.com/openwebwork/pg/pull/484.

drgrice1:
As to the circle graphing and the misplaced center, note that there is no need to cancel graphing due to a misplaced center.  You can complete the graph of the circle, and then move the center point to where you want it to go by clicking and dragging.

drgrice1:
I have added a mechanism that can be used to add custom graph objects and tools.  There is a lot of work ahead on documenting this.  I have attached some examples (also poorly documented) showing how to use this mechanism.

The CustomThreePointParabolaTool examples demonstrate how to create a new tool.  Note that the built in "parabola" graph object is replaced by the custom graph object in this example since it uses the same name.  Also, the only difference  between CustomThreePointParabolaTool.pg and CustomThreePointParabolaToolMacro.pg is that the latter file separates the new tool out into the macro GraphToolThreePointParabola.pl.

The GraphCircleCustomEg.pg also shows how to create a new graph object and tool (customLine and CustomLineTool) and how to create a new tool that uses an existing graph object (CustomCircleTool).  You would never use this example as the new objects and tools are identical to the built in objects and tools, but it shows how to do it.

The CustomFill.pg example shows how to create a new graph object that is derived from a built in object and modifies how it behaves.  The new "fill" graph object derives from the built in "fill" graph object and replaces the built in object.  It is then created by the built in FillTool.  Note that the built in fill graph object is an inequality fill.  All points are shaded that satisfy the same inequalities relative to the graphed objects as the point that defines the fill.  This custom fill object is a flood fill.  It starts at the defining point of the fill object and fills to the borders of the region it is in.

[GraphToolCustomExamples.zip](https://github.com/openwebwork/pg/files/4872798/GraphToolCustomExamples.zip)

Alex-Jordan:
I just tested things again using Firefox, Chrome, and Safari. All working as expected.

- I was confused by the CustomFill.pg example. It starts with two intersecting parabolas already drawn, and one region already filled. Then there are instructions to make a certain circle, which I did, submitted, and it counted correct. I felt like this is supposed to be demonstrating a new fill tool like you say here, but the problem text wasn't updated to tell the tester to use it.

- It would be nice if the delete key were bound to the delete button. So that if you have selected an object and hit delete, it deletes the object. On my Mac laptop, hitting delete is just defaulting to back-navigation of the browser.

- Suppose I use a tool like the line tool. After I make the line, the active tool turns to the object selection tool. That might be the right thing to happen, but I wonder if you thought about leaving it be whatever tool was just used. For example, in one problem I needed to plot three lines. So I needed to re-select the line tool for the second and third line.

- Should there be a line segment tool in the default tool set? In doing the example from SolveSysLinearInequals.pg, it seems like maybe the really "correct" answer would have the region you create bounded by segments, not with lines that extend infinitely.

- Is there somewhere appropriate to document that this is not keyboard accessible (yet)? If it can be made keyboard accessible in time for 2.16, great. If not, some warning might be a good idea. I know at my institution, using this without keyboard accessibility would be challenged pretty quickly. I wouldn't want anyone to invest in coding a bunch with it, only to be shut down by policy.

Since this is an entirely new feature that doesn't touch existing code, I see no reason not to merge at any time following a successful basic test. If now is a good time to merge, I can do that. If you have more planned anyway for the near future, I will wait. This is a great tool, thanks for doing this.

drgrice1:
The custom fill example is not an example of a valid problem. It is an example of how to use the newly added capabilities of the graph tool macro to modify the built in tools.  It is intended for developers, and not to be used as a problem for students. The static objects that are graphed are to demonstrate the difference from the built in full method, and that it works in print also as only static objects are printed.

It shouldn't be hard to bind the delete keyboard button to the delete button of the graph. However, general keyboard control of the graph tool is a long way out, and probably is going to take a lot of hackery. I don't intend to work on it at this point. If you want to document that somewhere, feel free.   As I have mentioned this graph tool is designed too be like the Pearson MyLab tool.  This graphing tool has everything that their graphing tool has. Their tool does not have keyboard controls. I am not sure why your institution would have a problem with a lack of keyboard controls. Most users don't use such things. Generally only more advanced users do.

When any tool completes it's action the select tool is activated. That really makes more sense than what you are suggesting if you think about it. It really doesn't take that much to click on the button each time you need to graph a line, and most students will expect that.

If you want a line segment tool, it shouldn't be hard to modify the line tool to create one. The way we teach that problem, it is best for them to be full lines.

I think the macro was good to go. I have been using it for a year, and it is functioning well. Students have no problem using it without additional instructions.

Alex-Jordan:
The keyboard navigation is for users with mobility disabilities, visual disabilities that are less severe than blindness, and is a prerequisite for future work to let blind users use this kind of tool. [This 11-minute video](https://www.youtube.com/watch?time_continue=250&v=eks3r-nE9lU&feature=emb_logo) profiles some students with disabilities and puts a human face on accessibility issues. The whole video is worth watching, and at 4:07 a profile starts of a student with a mobility disability that is compelling. So it's to support students with disabilities that PCC has policies about using accessible technology. It's also to support the faculty, who may have a similar disability, but also who can find they have a student with a disability in their class that makes an activity unusable, and they have no plan for a pedagogically equivalent activity. And it impact students. (I embedded a GeoGebra applet in our OER algebra book as an experiment, and within a month I had an email from disability services who was trying to help a student use it.)

And then there are the lawsuits against schools using inaccessible tools for violating the ADA. [Here](https://www.nfb.org/about-us/press-room/federal-court-rules-favor-blind-students?fbclid=IwAR3ybELDWsuJLjRFC_rJEHHZDQLdqp-_GrGgivcyq_bHaTklxXV9M5ROisQ) is one recent one (2019) where "the court specifically criticized the use of the Pearson MyMathLab product". IIRC there was another big one since then, but I couldn't find it in a quick search. These have been coming out about one per year in recent years, always in favor of the students, finding the school in violation. And Pearson and other companies get away without any repercussion because of the legal language they have in place with the school. It's always the school's fault for having a system where there is no alternative but to use MML (or WebAssign, or connect|Math, or whatever).

You say "they are also concerned with accessibility issues." However I would recommend against putting MyLab up as a standard to compare against. I know some individuals working there that care and try to do the right thing, but their systems are such juggernauts that did not have accessibility in mind from the ground up, and it's much much harder to retrofit with accessibility than it is to plan for it from the start. (I try to monitor accessibility in new WW features because among other reasons, it is ultimately easier to build it in from the start than to retrofit.) Meanwhile their legal department insulates them from these lawsuits, and their market share is so high they don't have to care about these lawsuits. And the big company is not concerned if it loses one client. Actually, it normally doesn't even lose the client; the school finds an accessible alternative (like WeBWorK) that can be used when needed, and the school keeps using the company's books and software in most cases. 

Back to the delete key thing. With that issue, even a user who is seemingly without a disability could naturally expect that behavior. And then when they hit delete, they will lose their work when the browser back-navigates. Far from the end of the world, but it is a possible annoyance I thought worth mentioning.

drgrice1:
It seems you have some challenging accessibility issues to work with at your institution.  So far we have not had such issues or restrictions.  In any case, keyboard controls will be difficult to implement.  I thought about the delete key thing, and that will be more challenging than I initially assessed as well.  The escape key implementation blindly checks for the escape key being pressed, and cancels graphing in that case.  It would be innocuous if the escape key were pressed for some other reason on the page.  However, that is not the case for the delete key.  So you would have to implement focus behavior for the graph, and check that.  That seems to be particularly challenging with the way that jsxgraph deals with things, and since it has no keyboard events.  Furthermore, I won't have time to work on this for a while.  My institution has decided to start two weeks early this semester, due to the coronavirus, and so I need to turn my attention to preparation for that.

Generally speaking this macro is in good shape for those that don't have the mobility accessibility issues to work with, and is a nice tool.

Alex-Jordan:
I'm testing this out again, just going through everything from before to see if any subsequent commits changed anything. Regarding the shading and color contrast, consider this [WCAG standard](https://webaim.org/articles/contrast/#sc1411). I believe that in the current situation, there should be a 3:1 contrast ratio. If it's possible to move the colors now, that will be one less issue to address at some point.

I'm red-green color blind so I do not always see color arrangements as pleasing that other people find pleasing. But with that caveat, I used [this contrast calculator.](https://webaim.org/resources/contrastchecker/) to see what I could find. If the fill color is this particular green (##00AD00) it has a 3:1 ratio with white. And if the main color is this blue variant (#0000EF) then that has ratio 3:1 with the green, and 9.34:1 with white. Again, I can't say if this is an ugly combination, but the blue is essentially the same as the current blue. This would move the yellow to a green. (Nothing that I see as "yellow" can reach a 3:1 ratio with white.)

As I have small bits of time I'll continue to review this and post any observations. Hoping to merge tomorrow or Monday.

Alex-Jordan:
These test problems that you provided earlier are no longer compiling: `CustomThreePointParabolaTool.pg`, `CustomThreePointParabolaToolMacro.pg`, `GraphCircleCustomEg.pg`, `CustomFill.pg`, `GraphToolThreePointParabola.pl`.

With `CustomThreePointParabolaTool.pg`, `GraphCircleCustomEg.pg`, and `CustomFill.pg`, changing instances of `GraphTool` to `parser::GraphTool` makes them compile. (I haven't looked more closely though until you confirm that they should otherwise function as expected.)

`GraphToolThreePointParabola.pl` was missing its `DOCUMENT();` call, but adding that and changing `GraphTool` to `parser::GraphTool` is not sufficient to get the problem to load.

I can't make `CustomThreePointParabolaToolMacro.pg` compile either.

(The original four test problems that don't have the custom tools are still working.)

drgrice1:
The following attached file is an update for the custom graph tool problems.  
[GraphToolCustomExamples.zip](https://github.com/openwebwork/pg/files/5287552/GraphToolCustomExamples.zip)
I added the parser namespace to the GraphTool package as is done for the parserMultiAnswer.pl macro in https://github.com/openwebwork/pg/pull/501.  That made it so the static methods used to add the custom graph objects need to use the namespace.  

Note that the .pl files are macros, and should not have the `DOCUMENT` call, and are not .pg problems.

I do think the green (#00AD00)  color is rather ugly with the rest of the colors used.  Furthermore, I find that that color makes it very difficult to distinguish various elements in the picture.  The fill tool icon is also green (#00FF00) when it is selected, and is difficult to distinguish from this green background.  The contrast tool you used shows these colors have a ratio of 2.19:1.  This breaks the part in the WCAG standard about user interface components.  I also think it is hard to distinguish the blue (#0000FF) of the graphed objects from this green.  According to the contrast tool you used the blue (#0000FF) and green (#00AD00) only have a ratio of 2.85:1.  I also find it difficult to see the black axes and labels if they are in the fill region with that color.   These WCAG standards bring to mind the Harrison Bergeron story.

drgrice1:
I also just noticed that the WCAG specification that you referenced says to have a 3:1 ratio to adjacent colors.  The fill color is not adjacent to the #FFFFFF white color.

drgrice1:
How about the color #FF0FFF.  I still don't find it particularly visually appealing.  But it does get a better contrast ratio with the other colors.  It is 3.1:1 versus white (#FFFFFF).  It is only 2.76:1 versus blue (#0000FF), but that probably is not as significant as the blue is only an outline.  Although it is only 2.26:1 versus the green of the fill icon, and that is probably still a problem.

If that green you chose works for you, I will just put it in as much as I don't like it.

drgrice1:
Maybe the contrast of the fill icon to the fill color is not important since the icon is outlined in black?

Alex-Jordan:
> Note that the .pl files are macros, and should not have the DOCUMENT call, and are not .pg problems.

Got it. I didn't notice it was `.pl` and was tossing it in alongside the others. Then copy-pasting the file name without looking closely.

> According to the contrast tool you used the blue (#0000FF) and green (#00AD00) only have a ratio of 2.85:1.

Yes, I suggested changing the blue to `#0000EF` to get 3:1.

>  These WCAG standards bring to mind the Harrison Bergeron story.

It is hard to grok elements of this for anyone who doesn't experience the disability. I'm a little on the spectrum of color-blindness, and it's easier for me to intuitively understand the significance of it all (even though I know we all understand the significance cerebrally).

> I also just noticed that the WCAG specification that you referenced says to have a 3:1 ratio to adjacent colors. The fill color is not adjacent to the #FFFFFF white color.

Well, in a technical sense, they are indeed adjacent if the border line is a dashed "open" variant. But I think that is besides the point. The contrast makes it clear that there is a difference, that the user perceives one side differently than the other, which is the point of the tool. An automated WCAG checker is not going to catch most/all of what we are discussing. But human staff in a disability services office will notice this, and point out that the motivation behind standard 1.4.11 applies here.

I think you know this, but I only proposed that green in order to put in some searching work up front to find a combo that hit the ratios. Since colors can be a minefield, how do you feel about a gray fill color? `#949494` is the lightest "pure gray" that has a ratio with white above 3:1. If the blue is toned down to `#0000F4`, that gives it a 3:1 ratio with the gray as well. I switched to that gray (RGB: 148, 148, 148) and it looks and works fine to me (but again, with the caveat that my vision works differently.) It's reminiscent of shading in with a graphite pencil.

I would like to merge this, get some more things merged as well in both repos to clear the docket some, and then maybe get 2.16 released this term. If you sign off on the gray idea, I'll just open a separate PR that does the color change. I don't mean to be giving you work. I'm on leave and I think you probably have classes to manage.

drgrice1:
Changing the colors isn't that hard.  The difference between the #0000F4 blue and the #0000FF blue is so negligible, I doubt anyone (even those with vision accessibility issues) can really perceive the difference even taking contrast to the other colors into account.  But okay.

The gray color is okay.  I will let you decide what colors you want.  I can implement them if you give them to me.  It isn't that much work.  Or if you prefer, you can do it.

I merged the corresponding pull request on the webwork side that contains the javascript for this macro.  The actual color changes will need to be made in the javascript there.

drgrice1:
One thing that I would like to add though.  Whatever colors are chosen, they shouldn't impair the ability of those that do not have vision accessibility issues to see things.  As I mentioned before, I find that the dark green that you chose makes it difficult to see what is behind it.  That is one of the objections I have to a dark fill color, and why I feel a lighter fill color is better.

Alex-Jordan:
> Whatever colors are chosen, they shouldn't impair the ability of those that do not have vision accessibility issues to see things.

Of course. I can't think of a scenario where some choice impairs a sight-abled person without also impairing a sight-disabled person. Building for universal design will not adversely impact anyone. You brought to my attention the issue with the fill icon. Right, the greens are too close. That is an issue for everyone, not only for color sight-abled people.

I think color is being used to do too much. One principle of universal design is that color alone should never indicate something important. It's often the first thing a designer (without accessibility experience) goes to, but there are more universal alternatives. 

Right now, orange is used to indicate focus on curves, and green on the fill icon. Is there any reason these should not be the same color? But wait, my larger point is that an alternative to coloring these exists. IIRC, canvas can do shadow blur around a shape. Could that be the focus indicator instead of (or in addition to) these colors? The color contrast concerns melt away if something else does the effective job of telling the user where focus is.

I'm always open to being wrong, but I think this could have a minimal color palette of three colors, not counting the faint grid lines:
* white for plain background
* something for filled background
* something for curves

All of these could get a black/gray shadow blur upon focus and no need for special coloration:
* curve
* fill icon
* point on a focused curve

Once there are only two non-white colors, it's easy to find them all such that they have good ratios and are also aesthetically pleasing.

I'm not suggesting that you make these changes. But I am mulling over opening a PR that does something like the above outline.

BTW, what about those grid lines, with their poor contrast to white? This is another great example. Nothing needs to change, because the purpose of those lines is to aid with pinpointing a location, and the explicit coordinates printed in the corner already do that.

drgrice1:
Shadowing and blurring would be a challenge to implement.  I don't think the jsxgraph library has such capabilities.  It also doesn't provide direct access to the internal HTML5 canvas elements.

At this point, focus is not indicated by color except for with the fill icon.  For curves focus is indicated by the visibility of the defining points.

Note that most of the colors used are defaults in the jsxgraph library.  The blue of the curves and the grid lines are among those.  The blue of the points isn't too hard to change.  The grid lines are more difficult.  In general, everything you are suggesting is going to take a lot of work to implement.

PR #485, opened 2020-06-30
add correct answer data to answer hash for Math Objects

Alex-Jordan:
Some Math Objects have a `data` property. The main example I have in mind is a `RadioButtons` object, in which case the `data` is the HTML `value` of the input (type `radio`) that is the correct answer. It's something like `B2`, by default. The answer hash doesn't currently give a way to identify that `value` of the correct option. This PR adds `correct_data` to the answer hash. Since it is an array reference, it's not helpful to stringify it. So in the `stringify` subroutine, we go one level deeper into its contents before stringifying.

@dpvc and others, does this raise any flags?

### DIFF
--- a/lib/AnswerHash.pm
+++ b/lib/AnswerHash.pm
@@ -257,9 +257,18 @@ sub score {
 sub stringify_hash {
   my $self = shift;
   Parser::Context->current(undef,$self->{correct_value}->context) if $self->{correct_value};
+  # special treatment for correct_data array
+  my $correct_data;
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
Not that anyone was on the cusp of merging this, but please stand by. I've discovered an issue with it.

Alex-Jordan:
This commit causes some unforeseen side effects, and now #494 accomplishes what I needed this for anyway.

PR #486, opened 2020-06-30
Correct aria label

Alex-Jordan:
Prior to this commit, if you use `parserRadioButtons.pl` to make a RadioButtons answer, the aria-labels on the buttons come out as:

- answer 1 option 1 
- answer 1 option 1 
- answer 1 option 2 
- answer 1 option 3

This commit corrects that to:

- answer 1 option 1 
- answer 1 option 2 
- answer 1 option 3 
- answer 1 option 4

### DIFF
--- a/macros/parserRadioButtons.pl
+++ b/macros/parserRadioButtons.pl
@@ -541,7 +541,7 @@ sub BUTTONS {
     $tag = $self->labelFormat($self->{labels}[$i]).$tag if $self->{displayLabels};
     if ($i > 0) {
       push(@radio,main::NAMED_ANS_RADIO_EXTENSION($name, $value, $tag,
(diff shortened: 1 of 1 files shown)


taniwallach (approved):
Bug from before the patch was seen in a sample problem in "view source", and is fixed by this patch.

It is worth noting that the first aria-label is added via the call to `main::NAMED_ANS_RADIO()` and the relevant code for that is in `macros/PGbasicmacros.pl` which is why the `else` case does not have an `aria_label` setting here.

It is also worth noting for the future that `generate_aria_label` seems to use `maketext` to internationalize much of the aria-label, but the string "option" used is both `macros/PGbasicmacros.pl` and in `macros/parserRadioButtons.pl` without `maketext` and at present "option" is not in the po files to provide translations.

PR #489, opened 2020-07-15
Fix applets in gateway quizzes

drgrice1:
Make it possible for applet problems to work in gateway quizzes.  Existing problems (written as described at https://webwork.maa.org/wiki/GeoGebraWeb1) will need to be modified to work in gateway quizzes, but will continue to work in homework problems without modification.  The changes that are needed are shown in the updated example in the POD documentation for macros/AppletObjects.pl.

To make this work an ADD_JS_FILE method was needed.  This was needed in general for macros in any case.  The method can be called to add a javascript file to the header from a pg problem file or macro, and the corresponding code added on the webwork side ensures that this is done without repetition.

I have also made it so that when NAMED_ANS or LABELED_ANS are called, the answer name provided is not preprended with the QUIZ_PREFIX.  That wasn't working in any case, as it ends up making the label for the answer rule and the label for the evaluator not match up in every case.  For a problem to work correctly in a gateway quiz the answer rule needs to have the QUIZ_PREFIX, and the evaluator the same, so the QUIZ_PREFIX needs to be put onto the label earlier.  The best way to achieve this is to use NEW_ANS_NAME to get a label, and then use that for both the answer rule and the LABELED_ANS/NAMED_ANS methods.  I also removed the completely unused SECTION_PREFIX.  That can be added back if something is actually implemented to use it (albeit not in the LABELED_ANS method for the same reason as for the QUIZ_PREFIX).

Note that this is broken into two commits.  The first is general code clean up, and the second is the primary content of this pull request.

See https://github.com/openwebwork/webwork2/pull/1121 for the corresponding webwork2 changes.

### DIFF
--- a/macros/AppletObjects.pl
+++ b/macros/AppletObjects.pl
@@ -1,13 +1,12 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright &copy; 2000-2018 The WeBWorK Project, http://openwebwork.sf.net/
-# $CVSHeader: pg/macros/AppletObjects.pl,v 1.24 2010/01/03 17:13:46 gage Exp $
-# 
+# Copyright &copy; 2000-2020 The WeBWorK Project, http://openwebwork.sf.net/
+#
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of either: (a) the GNU General Public License as published by the
 # Free Software Foundation; either version 2, or (at your option) any later
 # version, or (b) the "Artistic License" which comes with this package.
-# 
+#
 # This program is distributed in the hope that it will be useful, but WITHOUT
 # ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
 # FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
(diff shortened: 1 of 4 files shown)


macros/AppletObjects.pl
```diff
@@ -71,6 +70,7 @@ =head3  GeogebraWebApplet
 =cut
 
 sub GeogebraWebApplet {
+	ADD_JS_FILE("//web.geogebra.org/4.4/web/web.nocache.js", 0);
```
taniwallach, line 73: Should this be a full URL? Is it necessary to have a fixed version number in the path?

taniwallach (commented):
I took a quick look at the code, but would like some sample test problems and a test procedure to try this out.
---
Do we need to continue to support Flash content (until when)? Maybe we should add a warning about deprecation of Flash support?

Same goes for Java applets, which probably are no longer functional, as modern browsers do not allow a Java plugin.

macros/AppletObjects.pl
drgrice1: Yes, it is a full URL, and needs to be.  Or are you referring to the lack of the http or https protocol?

It would probably be nice to have the link be pulled in from a setting somewhere, but it was done this way before.  I didn't change that.  See line 1163 in pg/lib/Applet.pm.

drgrice1:
Here is one test file. This is a modified version of one of Drew's problems from Contrib. This problem will work in both homework and in gateway quizzes with this pull request. It won't work without this pull request though. I only made the minimum changes needed to make it work in gateway quizzes, so you can see the changes that problem editors need to make to existing problems to update them (although if those changes aren't made the problems will continue to work in homework).
[two-lines-parallel.zip](https://github.com/openwebwork/pg/files/5462447/two-lines-parallel.zip)

You should also test the original file Contrib/CUNY/CityTech/CollegeAlgebra_Trig/setLines/two-lines-parallel.pg. That file will work both with and without this pull request in homework, but will not work in gateway quizzes with or without this pull request.

By the way, don't forget that this pull request needs the corresponding pull request https://github.com/openwebwork/webwork2/pull/1121 to the webwork2 code to work.

macros/AppletObjects.pl
drgrice1: Actually, I remember when I was working on this pull request that I looked at the newer Geogebra javascript API, and I recall that it will not work with the applet code in webwork2.  I think that the version is more tightly bound to the way the macro utilizes the API, and so it probably does need to be hard coded in in this way.  At least for now, until someone wants to figure out a better way to do it.

taniwallach: I was referring to the missing http/https.I guess there may be reasons not to force https.

It would certainly be nice to get the URL from a setting, but it is harder to do that in PG, and with the intention to have the standalone render pick up more uses and possible to try in a future version of WW to split the renderer out of the Apache process - it is probably not a very good idea to try to pass it from webwork2 into pg.

For now - if it works - I guess we should leave it as you did it, and we can look into how to make it configurable in the future.

drgrice1:
I would like to see this merged into the code soon.  There are some things in this pull request that are important on their own.  Namely the things commented in my initial comment for this pull request.  The NAMED_ANS changes fix part of the problem with named answers in gateway quizzes.

taniwallach:
Tested - see the details from the testing at https://github.com/openwebwork/webwork2/pull/1121#issuecomment-782064140

taniwallach (approved):
The majority of the review comments are in https://github.com/openwebwork/webwork2/pull/1121

Alex-Jordan:
This branch has a small conflict now that something else was merged.

drgrice1:
Thanks for pointing that out.  It was @drdrew42's PGcore pg version.

taniwallach:
@Alex-Jordan have you tested this pair. If so - please add an approval to both.

drdrew42 (approved):

PR #490, opened 2020-07-17
AskSage should not match just any "100" in response header

drdrew42:
I have been running into sporadic issues with AskSage problems. They seemed to randomly fail when trying to connect to the Sage Cell Server.

It turns out the response header gives plenty of opportunity to match "100":

```http
checking for header:  HTTP/2 200 
date: Fri, 17 Jul 2020 16:51:30 GMT
content-type: application/json; charset=UTF-8
content-length: 437
set-cookie: __cfduid=dc7f19b456d5fb97ea5fbf50f3f71c4fa1595004689; expires=Sun, 16-Aug-20 16:51:29 GMT; path=/; domain=.sagemath.org; HttpOnly; SameSite=Lax
vary: Accept-Encoding
cf-cache-status: DYNAMIC
cf-request-id: 03ff493591000074018d33d200000001
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
server: cloudflare
cf-ray: 5b4577cf49197401-IAD at /Users/drdrew42/mojo/renderv2/lib/RenderApp/Controller/../../PG/lib/WeBWorK/PG/IO.pm line 327
```

So, I've tightened up the regex matching for the header to more accurately identify HTTP 100 and HTTP 200 responses.

I'm making a matching pull request for develop, but I think this should be merged now.

### DIFF
--- a/lib/WeBWorK/PG/IO.pm
+++ b/lib/WeBWorK/PG/IO.pm
@@ -326,8 +326,8 @@ sub query_sage_server {
 	 	my $header_block = shift(@lines);
 	 	warn "checking for header:  $header_block" if $debug;
 	 	next unless $header_block=~/\S/; #skip empty lines;
-	 	next if $header_block=~/HTTP/ and $header_block=~/100/; # skip continue line
-	 	if ($header_block=~/200/) { # 200 return is ok
+	 	next if ($header_block =~ m!HTTP[ 12/.]+100!); # skip continue line
+	 	if ($header_block=~ m!HTTP[ 12/.]+200!) { # 200 return is ok
 	 		$header_ok=1;
 	 		last;
 	 	}


taniwallach:
The motivation make sense, as does the code. Can you provide a very simple AskSage test problem with which the code can be tested.

drgrice1:
Some of Drew's problems in Contrib can be used for a test file, for example Contrib/CUNY/CityTech/Precalculus/setSequences_-_Geometric/geometric-seq-nth-term.pg.  Enter an answer for the nth term in a sequence in all parts, and this issue is rather likely to occur.  It is not guaranteed to occur though, so you might need to try a few times.

drgrice1:
By the way, you will need to check the very end of the debug output (Drew: why do you have debug enabled in this problem?).  If the error does not occur it will say "success is 1", and if the error does occur it will say "ERROR in contacting sage server. Did you accept the terms of service by setting {accepted_tos=>'true'} in the askSage options?"

taniwallach (approved):
I tested the version of this patch for develop: https://github.com/openwebwork/pg/pull/491 

I confirm that before the patch, the suggested sample problem would frequently fail and report:
```
ERROR in contacting sage server. Did you accept the terms of service by
setting {accepted_tos=>'true'} in the askSage options?
IO.pm: ERROR trapped during JSON call to sage:
Unable to make a sage call to https://sagecell.sagemath.org/service. at [PG]/lib/WeBWorK/PG/IO.pm line 442
at [PG]/lib/WeBWorK/PG/IO.pm line 502
```
and that after the patch those errors are not occurring. The code looks fine.

taniwallach:
Given that this is a fix for an issue with functionality/stability in the master branch - I think we should merge the hotfix in the near future.

drdrew42:
I am 100% in favor of merging this hotfix ASAP -- I would imagine that most AskSage problems are broken without it.

> Contrib/CUNY/CityTech/Precalculus/setSequences_-_Geometric/geometric-seq-nth-term.pg

I don't see debug enabled on this problem? It is quite possible that I accidentally committed a problem that still had the debug on - if you find which one, let me know? (a cursory grep didn't turn up anything)

drgrice1:
I don't see the debug flag now either.  I must have added it when I was testing things before.  I know after I did some testing yesterday I ran `git checkout -- .` on the git repository.  So that would have removed my changes.  My mistake.

drgrice1:
I think this has sat here long enough.  I am going to merge this later today, unless there are objections.

Issue #495, opened 2020-08-16
Feature request: cmp option to disable the code which triggers the "This answer is equivalent to the one you just submitted." message


taniwallach:
Certain types of problems use graders which do not "compare to" a fixed correct answer in the simple manner, and for such problems, the code which triggers the message "This answer is equivalent to the one you just submitted" often triggers it inappropriately. There is not a really simple mechanism to turn this off when it behaves inappropriately.

See: https://webwork.maa.org/moodle/mod/forum/discuss.php?d=3498 and https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4352 .

At present, there are three workaround to avoid this:

1. Clearing all the standard post-filters using `withPostFilter("erase")`.
2. "pop"ing off only this postfilter (assumes it is the last one installed) - see: https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4352
3. Having the custom grading code remove the value of the previous answer (something like` $ansHash->{prev_ans}='';`).
4. Using an additional post-filter just to remove the message.

I discovered that the first option seems to cause problems (even on initial problem load sometimes). The others are all somewhat inconvenient to use and less-than-obvious. The second also depends on the relevant post-filter remaining the last one added.

It would be nice to be able to pass a control setting to `cmp` to disable this warning from being triggered, as suggested in both threads.

There seem to be 2 files which can trigger this message:
  * `lib/Value/AnswerChecker.pm`
  * `macros/PGfunctionevaluators.pl` - reported by Davide as not being used an longer (pre-MathObjects).

PR #497, opened 2020-08-18
Add new bypass_equivalence_test switch

taniwallach:
Add bypass_equivalence_test switch to disable setting the message "This answer is equivalent to the one you just submitted" when it is not desired in a simple manner. This addresses the issue I reported in https://github.com/openwebwork/pg/issues/495 which I had originally not intended to work on right away.

It is primarily intended for use in problems whose graders do not "compare to" a fixed correct answer in the simple manner which the equivalence test code expects.

Can be tested by using the sample problem below by entering various equivalent answers one after the other with when the new setting is set to `1` to by bypass the message, to `0` to enable the message, or without being set (which leaves the message enabled).

```
DOCUMENT();
loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
);
Context("Numeric");
$aSoln = Compute("3x");
BEGIN_TEXT
\( x + 2x = \)
\{ $aSoln->ans_rule(8) \}
END_TEXT
ANS( $aSoln->cmp( bypass_equivalence_test => 1 ) );
ENDDOCUMENT();
```

### DIFF
--- a/lib/Value/AnswerChecker.pm
+++ b/lib/Value/AnswerChecker.pm
@@ -1721,6 +1721,7 @@ sub cmp_postfilter {
   $ans->{_filter_name} = "produce_equivalence_message";
   return $ans if $ans->{ans_message}; # don't overwrite other messages
   return $ans unless defined($ans->{prev_ans}); # if prefilters are erased, don't do this check
+  return $ans if ($ans->{bypass_equivalence_test});
   my $context = $self->context;
   Parser::Context->current(undef,$context);
   $ans->{prev_formula} = Parser::Formula($ans->{prev_ans});
--- a/macros/PGfunctionevaluators.pl
+++ b/macros/PGfunctionevaluators.pl
@@ -1037,10 +1037,11 @@ sub ORIGINAL_FUNCTION_CMP {
 			my $rh_ans = shift;	
 			#WARN_MESSAGE(pretty_print($inputs_ref));
 			my $isPreview = $inputs_ref->{previewAnswers}; # || ($inputs_ref->{action} =~ m/^Preview/);
+			return $rh_ans if ($rh_ans->{bypass_equivalence_test});
 			return $rh_ans unless !$isPreview # not preview mode
 				and $rh_ans->{ans_equals_prev_ans} # equivalent
 				and $rh_ans->{prev_ans} ne $rh_ans->{original_student_ans}; # not identical
-			
+
 			$rh_ans->{ans_message} = "This answer is equivalent to the one you just submitted.";
 			return $rh_ans;
 		}


lib/Value/AnswerChecker.pm
```diff
@@ -1721,6 +1721,7 @@ sub cmp_postfilter {
   $ans->{_filter_name} = "produce_equivalence_message";
   return $ans if $ans->{ans_message}; # don't overwrite other messages
   return $ans unless defined($ans->{prev_ans}); # if prefilters are erased, don't do this check
+  return $ans if ( defined( $ans->{bypass_equivalence_test} ) && $ans->{bypass_equivalence_test} );
```
drgrice1, line 1724: It is simpler here if you just use:
```perl
return $ans if ( $ans->{bypass_equivalence_test} );
```
If `$ans->{bypass_equivalence_test}` is not defined or is defined and is a false value, then the conditional will be false as desired here.
Perl doesn't emit an error or even a warning if the hash key is not defined in this situation.
The same is true where you check this condition in the file PGfunctionevaluators.pl.

I also don't like the extra spaces in the parentheses.  Of course this is just my coding style preference.  I generally am of the opinion that programming code should follow standard grammatical syntax rules to the extent that the programming language allows.

taniwallach: Thanks - cleaned up (with a forced push) to use the simpler test. Sorry, but I am too used to worrying about undefined values... but as you said and testing confirmed -  no warning is issued without the test if the hash has this key.

drgrice1: Looks good.

You even removed the spaces around the parenthetical.  There are many that prefer that coding style.

awmorp:
Great enhancement, thanks. Is there a way to set this flag globally rather than having to set it in each individual answer checker?

taniwallach:
> Great enhancement, thanks. Is there a way to set this flag globally rather than having to set it in each individual answer checker?

Thanks.

In principle, I think a new flag could be added to the list in `lib/Value.pm` in the definition of `$defaultContext`, say `alwaysBypassEquivalenceMessage` with the default value of `0`. Then the main `sub cmp_default` of the `Value` package (in 
`lib/Value/AnswerChecker.pm`) could be modified, and test the value of that flag and when necessary could add `bypass_equivalence_test => 1` to the return value of the base `cmp_default`.

I think such an approach would allow you to call
```
Context()->flags->set(
  alwaysBypassEquivalenceMessage => 1
);
```
to turn on the bypass for all `cmp` routines created in the context.

Go ahead and try to implement and test the idea. If it works, you can propose it as a PR, and then the maintainers and others will discuss the merit of merging in the new feature.

Alex-Jordan:
Looks good. Merging now.

PR #500, opened 2020-08-21
Fix malformed utf8 copyright (hot patch)

drgrice1:
This is already in develop. There has been several issues with this recently, and I think it may need to be re-considered as a candidate for a hot patch.


taniwallach:
This is the hotfix of the changes in https://github.com/openwebwork/pg/pull/500 and people have been reporting issues in the forum and here as issues - ex: https://github.com/openwebwork/pg/issues/493.

I'm merging it to get it into master.

PR #501, opened 2020-08-24
Fix parserMultiAnswer.pl issue when loaded twice

taniwallach:
Fix parserMultiAnswer.pl to resolve issue which caused memory/CPU issues when it was loaded twice, for example by problem parts loaded using includePGproblem(). 

The issue was first reported in the forums at https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4875 where a sample set of simple files to trigger the bug is included. A copy of the posts follow, and the test files are also [here in sample.zip](https://github.com/openwebwork/pg/files/5119286/sample.zip).

The fix in this PR was provided by @dpvc in the forum thread, and resolves the issue.

---

**Copied from the forums:**

by Nathan Wallach - Sunday, 23 August 2020, 5:57 AM

Some locally developed problems are coded as "problem pieces" which are then loaded via a control PG file using includePGproblem(). The typical usage case is a set of several possible options for each part of a question, which are easy to maintain as single files (rather than inside one large file which selects each case from a long list).

This approach works well in some settings, but fails when multiple "problem pieces" need to use parserMultiAnswer.pl.

When 2 "problem pieces" both use MultiAnswer, trying to load the control file leads to Apache using huge amounts of CPU and RAM and failing to respond.

From this it seems that there is some problem using MultiAnswer in problems loaded via includePGproblem() from different files.

A set of sample file is attached which is just a trivial problem which uses MultiAnswer to grade 2 input boxes together, and given twice to trigger the error. When test2.pg is loaded after test1.pg by load_pair.pg the issue hits, and the server thrashes. Replacing test2.pg by test2b.pg which does not use MultiAnswer bypasses the problem, but fails to provide the intended behavior.

Hopefully there is some method to avoid the issue.

---

by Davide Cervone - Monday, 24 August 2020, 7:09 AM

It turns out that the loading of parserMultiAnswer.pl twice in a single problem is the source of the issue, and the problem occurs during the _parserMultiAnswer_init. This is where the MultiAnswer() function is defined, and it seems that there is a conflict with making that definition if the MultiAnswer package is already in place. So one solution is to rename the MultiAnswer package to parser::MultiAnswer This requires two lines to be changed in parserMultiAnswer.pl: the MultiAnswer->new() call in the _parserMultiAnswer_init() function, and the package MultiAnswer a few lines later. Once these are changed, the two problems load successfully for me.

### DIFF
--- a/lib/TikZImage.pm
+++ b/lib/TikZImage.pm
@@ -2,7 +2,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
 # Copyright &copy; 2000-2018 The WeBWorK Project, http://openwebwork.sf.net/
-# $CVSHeader: pg/macros/parserMultiAnswer.pl,v 1.11 2009/06/25 23:28:44 gage Exp $
 # 
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of either: (a) the GNU General Public License as published by the
--- a/macros/PGML.pl
+++ b/macros/PGML.pl
@@ -1013,7 +1013,7 @@ sub Answer {
       }
       $rule = $ans->$method(@options);
       $rule = PGML::LaTeX($rule);
-      if (!(ref($ans) eq 'MultiAnswer' && $ans->{part} > 1)) {
+      if (!(ref($ans) eq 'parser::MultiAnswer' && $ans->{part} > 1)) {
         if (defined($item->{name})) {
           main::NAMED_ANS($item->{name} => $ans->cmp);
         } else {
--- a/macros/PGtikz.pl
+++ b/macros/PGtikz.pl
@@ -1,7 +1,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
 # Copyright &copy; 2000-2018 The WeBWorK Project, http://openwebwork.sf.net/
-# $CVSHeader: pg/macros/parserMultiAnswer.pl,v 1.11 2009/06/25 23:28:44 gage Exp $
 # 
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of either: (a) the GNU General Public License as published by the
--- a/macros/parserMultiAnswer.pl
+++ b/macros/parserMultiAnswer.pl
@@ -96,12 +96,12 @@ or
 loadMacros("MathObjects.pl");
 
 sub _parserMultiAnswer_init {
-  main::PG_restricted_eval('sub MultiAnswer {MultiAnswer->new(@_)}');
+  main::PG_restricted_eval('sub MultiAnswer {parser::MultiAnswer->new(@_)}');
 }
 
 ##################################################
 
-package MultiAnswer;
+package parser::MultiAnswer;
 our @ISA = qw(Value);
 
 our $answerPrefix = "MuLtIaNsWeR_";  # answer rule prefix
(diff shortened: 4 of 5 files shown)


drgrice1:
This breaks many of my problems.  I haven't figured out what exactly is causing it, but I will look into it.  It breaks the problems badly if MathQuill is enabled. In that case the problem won't even load.  If MathQuill is disabled, the problem loads but I get things like "Error in LABELED_ANS:|AnSwEr0001| -- inputs must be references to AnswerEvaluator objects or subroutines" and  "Error in Translator.pm::process_answers: Answer AnSwEr0001: Unrecognized evaluator type ||".

drgrice1:
I see now.  This breaks MultiAnswer for PGML.  See line 1016 of PGML.pl.  I knew that was there, and I should have known right away what the problem was.

drgrice1:
Note that the parserMultiPart.pl macro will probably also need to be fixed.  Although that is only for backwards compatibility, and I am not sure if any files still use that.

It would probably be a good idea to check for other cases like these in the code.

taniwallach:
> I see now. This breaks MultiAnswer for PGML. See line 1016 of PGML.pl. I knew that was there, and I should have known right away what the problem was.
> 
> If you add the parser:: prefix there, it fixes the problem.

Fixed as you suggested, and tested using a PGML file using MultiAnswer.

taniwallach:
> Note that the parserMultiPart.pl macro will probably also need to be fixed. Although that is only for backwards compatibility, and I am not sure if any files still use that.
> 
> It would probably be a good idea to check for other cases like these in the code.

Fixed and tested using a test file using `parserMultiPart.pl` (which intentionally triggers the warning that `MultiPart` is deprectated.

A search inside PG for other uses of `MultiAnswer` did not come up with anything else I think needs to be fixed.  It did discover two files which had CVS headers apparently copied from `macros/parserMultiAnswer.pl` which I removed as they only cause unnecessary confusion. (Maybe all the outdated `CVSHeader` lines should be removed one day.)

A quick search of `.pl` files in the OPL did not come up with anything I think needs to be fixed. 

Sample files (in addition to the prior ones) to test with `parserMultiPart.pl` and using `PGML` are in  [sample2.zip](https://github.com/openwebwork/pg/files/5122294/sample2.zip).

lib/TikZImage.pm
```diff
@@ -2,7 +2,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
 # Copyright &copy; 2000-2018 The WeBWorK Project, http://openwebwork.sf.net/
-# $CVSHeader: pg/macros/parserMultiAnswer.pl,v 1.11 2009/06/25 23:28:44 gage Exp $
```
drgrice1, line 5: Thanks.  These were cut and paste errors on my part.  You did the same thing I did, and searched the code for referenece to MultiAnswer, and found these!

taniwallach:
> I found that trying to use your load_pair.pg or load_tests.pg on my test server put apache2 into an infinite loop without the pull request. So this is clearly needed.

Yes, but only if someone tried to load MultiAnswer twice, so it seems not too many people ran into this problem to an extent which bothered them. As such, I doubt there is any need to make this a hotfix to WW 2.15, but it would be nice to get it merged and into develop.

drgrice1:
It occurred to me that the problem you are having with reloading the MultiAnswer via includePGproblem is not unique to this macro.  There are several other macros that use the same technique to define a convenience PG method for creating the object.  parserRadioButtons.pl does the same thing.  I noted that parserPopUp.pg uses the same technique that you are now implementing in the MultiAnswer macro.  The point is that someone should go through the macros and fix all of them with this same problem.

taniwallach:
Although there may be other macros which have similar issues, I think the fix for MultiAnswer is important enough to justify merging this and getting this into develop already.

drgrice1:
I agree.

drgrice1:
I will go ahead and merge it.

PR #502, opened 2020-09-05
Add methods to access radio button labels/choices from answer string

mikeshulman:
As discussed at https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4883, in order for a RadioButton included in a MultiAnswer group to have a correct answer that depends on other answers in the group, the user needs some way to convert between the "Bn" strings and the answer choices and/or labels.  This PR adds methods to RadioButton that do this, so that the format of "Bn" strings can remain a private implementation detail.

### DIFF
--- a/macros/parserRadioButtons.pl
+++ b/macros/parserRadioButtons.pl
@@ -199,7 +199,18 @@ to get the answer checker for the radion buttons.
 
 You can use the RadioButtons object in MultiAnswer objects.  This is
 the reason for the RadioButton's C<ans_rule()> method (since that is
-what MultiAnswer calls to get answer rules).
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I tested with:

```
$radio = RadioButtons(
  ["Red","Blue","Green","None of these"],
  "Blue", # correct answer
  last => ["None of these"], # can be a list
  labels => 'ABC',
);

for $n (0,1,2,3) {
TEXT(
'Choice:', $radio->answerChoice('B'.$n),$BR, 
'Label:', $radio->answerLabel('B'.$n),$PAR
);
}
```

and it works as expected.

PR #505, opened 2020-09-22
PG warnings cleanup and minor contextFraction issue

drdrew42:
These are a few minor warning cleanup items:
* Subroutine TEX redefined at... (in PGbasicmacros)
* Unquoted string "null" may clash with future reserved word at... (in PGML)
* "my" variable %bullet masks earlier declaration in same scope at... (in PGML)

and another minor change to contextFraction:
* method TeX called on a Fraction with denominator 1 will return a number, not a string
* this is an issue for homogenous responses from PG -- e.g. correct_answer_latex_string was not consistently a string

Remainder of changed files are all whitespace cleanup automatically made when resolving merge conflicts. Check off "ignore whitespace" when reviewing.


Alex-Jordan:
This appears to have more in it than described. For example, it adds this file `.github/FUNDING.yml`.

Alex-Jordan:
After looking through the rest of it, the addition of `.github/FUNDING.yml` appears to be the only thing different from the description.

drdrew42:
added fix to WeBWorK/PG/IO.pm that corrects the reference to the curlCommand given in site.conf

drgrice1:
> added fix to WeBWorK/PG/IO.pm that corrects the reference to the curlCommand given in site.conf

This is incorrect.  You are looking at your local copy of site.conf, and not at site.conf.dist.  In site.conf.dist it is $externalPrograms{curl}.

drdrew42:
> > added fix to WeBWorK/PG/IO.pm that corrects the reference to the curlCommand given in site.conf
> 
> This is incorrect. You are looking at your local copy of site.conf, and not at site.conf.dist. In site.conf.dist it is $externalPrograms{curl}.

This is the current version of site.conf.dist
https://github.com/openwebwork/webwork2/blob/03eca4cc1a6034d24cf28bd690f0f56ce14f284b/conf/site.conf.dist#L131

drgrice1:
That is the version of site.conf.dist in master.

drdrew42:
so is the point of changing this hash-key to ensure that your tikz feature is only tested against the develop branch of ww2?

drgrice1:
Mike actually wanted that changed.

drgrice1:
The point was that it was inconsistently named compared to the other $externalPrograms names.

drgrice1:
By the way, it isn't really my tikz feature.  I just built on the work that Mike and others had already done.

drgrice1:
Out of curiosity, what issues does the non-string return value of the TeX method in contextFraction.pl cause?  Perl doesn't exactly distinguish between the string "1" and the number 1.  They are both scalars.

macros/PGbasicmacros.pl
```diff
@@ -1416,10 +1416,10 @@ =head2 Display Macros
...
 =cut
 
-sub TEX {
-	my ($tex, $html ) = @_;
-	MODES(TeX => $tex, HTML => $html, HTML_tth => $html, HTML_dpng => $html);
-}
+# sub TEX {
+# 	my ($tex, $html ) = @_;
+# 	MODES(TeX => $tex, HTML => $html, HTML_tth => $html, HTML_dpng => $html);
+# }
```
drgrice1, line 1423: Rather than commenting this out, just delete it.

drdrew42:
In rendering through ww2, there is no issue - no. However, at least for consistency sake, the correct_answer_latex_string ought to always be a string, though in this case it was not. This arose as an issue for standardizing the PG output from a standalone problem renderer, where object typing should be consistent.

drgrice1:
> In rendering through ww2, there is no issue - no. However, at least for consistency sake, the correct_answer_latex_string ought to always be a string, though in this case it was not. This arose as an issue for standardizing the PG output from a standalone problem renderer, where object typing should be consistent.

I see.  I presume that the other problem renderer is not written in Perl then?  And the output when passed to this other language is then not quoted?  Is that what happens?

drdrew42:
Rendering has to happen in perl - but response types ought to be well-behaved for whomever would consume them.

drgrice1:
Hmm.  I am not sure that adding the quotes actually does anything here, since it is all in Perl, and Perl scalars don't actually have a "type".  It won't hurt though.

drdrew42:
it matters when spitting out JSON, which seems to be most useful... the fix addresses fraction objects that return a non-string correct_answer_latex_string when the answer is an integer. I have not yet encountered this issue with any other MathObject class.

drgrice1:
I thought that might be the case.  I wonder how the Perl json_encode method makes the distinction?  I mean "1" == 1 and "1" eq 1 both resolve to begin true.

drdrew42:
sure that's how perl handles equivalence... but json_encode will just use whatever is assigned to the key -- in this case, an int rather than a stringified int.

Alex-Jordan:
For example, there is this:
```
$n =  1  & '';  # $n is 0
$s = '1' & '';  # $s is ''
```
so at some level perl knows the difference and can write the json differently.

drgrice1:
> ```
> $n =  1  & '';  # $n is 0
> $s = '1' & '';  # $s is ''
> ```
> 
> so at some level perl knows the difference and can write the json differently.

Yeah, I just found this in a little Google searching.  The bitwise operator distinguishes.  That was introduced in version 5.22 of Perl.

Alex-Jordan:
This is a difficult pull request to review. There are 20 commits from 4 different authors, which makes me wonder what is the history of this branch, and if a rebase would help clear it up. And there are 15 affected files, with a lot of whitespace edits. At least one file has both whitespace edits and a more meaningful edit, so it's not as if I can trust that all the whitespace edits are harmless. I need to look through it all.

From what I can tell:
1. 11 files just have whitespace cleanup.
1. `contextFraction.pl` has a lot of whitespace edit, and one place where `$a` becomes `"$a"`. The reason for this is discussed in this thread.
1. A new file is added, `.github/FUNDING.yml`. What is this for? It's in two commits that Mike made, but it is not mentioned in the pull request description.
1. `PGbasicmacros.pl` loses the `TEX` subroutine. The PR description says it is redefined, but I'm missing where it gets redefined. I can only see that that subroutine is just removed.
1. The edits to `PGML.pl` that are described in the PR.

I don't understand the purpose of the last three items. I'd feel better if someone could comment on items 3, 4, and 5. Maybe @mgage and/or @dpvc.

drgrice1:
I tested merging this, and doing so does add duplicate entries in the git history of the extra commits that are listed.  A rebase onto the develop branch fixes this, although there are conflicts that need to be resolved.  The resolutions are straightforward though.

@Alex-Jordan The TEX subroutine was removed, and not replaced.  That is intended.  What he meant is that the warning about the TEX subroutine being redefined is fixed (because it no longer is).

dpvc:
I would say this does need to be rebased (or cherry-picked) since the drdrew42 branch is branched from the rederly master branch, not the develop branch, so technically, this PR is merging master into develop, which is not allowed.

Personally, I don't like having lots of white-space fixes mixed in with a PR with other changes, but I won't object if that's what you all want to do.  

I'm OK with the change to `contextFraction.pl` TeX output.  For PGML.pl, the

```
$self->End(null, shift);
```

should be

```
$self->End(undef, shift);
```

not 

```
$self->End("null", shift);
```

(This was my fault, having been working with Javascript and then going back to perl).

I don't understand the reported error message about %bullets, since the two instances are not in the same package, and I don't see any other occurrences anywhere else in PG, but I'm OK with changing the `my` to `our`.

I would remove the comment that describes the removed `TEX()` macro indicating that it was deprecated.  Luckily it seems that this macro is not used anywhere else in PG or in the OPL (which is good, since the second definition later in the file is completely different).

I can't tell you anything about the fundraising file.

PR #509, opened 2020-12-05
replace romberg approximation with faster function

doombot-exe:
using `normal_prob` from PGstatisticsmacros.pl requires excessive CPU resources and slows down problem rendering to the tune of ~2 sec per instance of `normal_prob`. Replacing the romberg approximation with the `uprob` function (from Statistics::Distributions, already loaded in PGstatisticsmacros.pl) greatly reduces the CPU load for problems that require this calculation (and does not have any impact on the degree of accuracy for the calculated result).


Alex-Jordan:
Thanks for this. Can you redirect the pull request into the `develop` branch of `pg`?

doombot-exe:
@Alex-Jordan ✅ 
Thank you.

PR #511, opened 2020-12-05
replace romberg approximation with faster function

doombot-exe:
(cherry picked from commit f5000573a8a0a1a8c1eef8e21814294c18d5f8cf)
Replacement PR for #509

using `normal_prob` from PGstatisticsmacros.pl requires excessive CPU resources and slows down problem rendering to the tune of ~2 sec per instance of `normal_prob`. Replacing the romberg approximation with the `uprob` function (from Statistics::Distributions, already loaded in PGstatisticsmacros.pl) greatly reduces the CPU load for problems that require this calculation (and does not have any impact on the degree of accuracy for the calculated result).

### DIFF
--- a/macros/PGstatisticsmacros.pl
+++ b/macros/PGstatisticsmacros.pl
@@ -33,41 +33,36 @@ sub normal_prob {
         warn 'You must also load PGnumericalmacros to use PGstatisticsmacros' unless defined(&_PGnumericalmacros_init);
 
 	my $a = shift;
-        my $b = shift;
+    my $b = shift;
  	my %options=@_;
 
-	my $mean = $options{'mean'} if defined ($options{'mean'});
-        $mean = 0 unless defined $mean;
-
-	my $deviation = $options{'deviation'} if defined ($options{'deviation'});
-        $deviation = 1 unless defined $deviation;
+	my $mean = $options{'mean'} // 0;
+	my $deviation = $options{'deviation'} // 1;
 
 	if ($deviation <= 0) {
 		warn 'Deviation must be a positive number.';
 		return;
-		}
+	}
 
-        my $z_score_of_a;
-	my $z_score_of_b;
-
+	my $prob;
 	if ( $a eq '-infty' ) {
-		$z_score_of_a = -6;
+		if ( $b eq 'infty' or '-infty') {
+			$prob = 1; # did you really need us to tell you that?
+		} else {
+			my $z_score_of_b = ($b - $mean)/$deviation;
+			$prob = 1 - uprob($z_score_of_b);
+		}
 	} else {
-		$z_score_of_a = ($a - $mean)/$deviation;
+		my $z_score_of_a = ($a - $mean)/$deviation;
+		if ( $b eq 'infty' or '-infty') {
+			$prob = uprob($z_score_of_a);
+		} else {
+            my $z_score_of_b = ( $b - $mean ) / $deviation;
+			$prob = uprob($z_score_of_a) - uprob($z_score_of_b);
+		}
 	}
 
-        if (($b eq 'infty') or ($b eq '+infty')) {
-		$z_score_of_b = 6;
-	} else {
-		$z_score_of_b = ($b - $mean)/$deviation;
-	}
-
-        my $function = sub { my $x=shift;
-                             $E**(-$x**2/2)/sqrt(2*$PI);
-                             };
-
-        my $prob = romberg($function, $z_score_of_a, $z_score_of_b, level => 8);
-        $prob;
+    return $prob;
 }
 
 =head3 "Inverse" of normal distribution
@@ -88,28 +83,26 @@ Load PGnumericalmacros.pl if you use this method.
 sub normal_distr {
 	warn 'You must also load PGnumericalmacros to use PGstatisticsmacros' unless defined(&_PGnumericalmacros_init);
 
-        my $prob = shift;
-        my %options=@_;
+	my $prob = shift;
+	my %options=@_;
 
-        my $mean = $options{'mean'} if defined ($options{'mean'});
-        $mean = 0 unless defined $mean;
+	my $mean      = $options{'mean'}      // 0;
+	my $deviation = $options{'deviation'} // 1;
 
-        my $deviation = $options{'deviation'} if defined ($options{'deviation'});
-        $deviation = 1 unless defined $deviation;
+	if ($deviation <= 0) {
+		warn 'Deviation must be a positive number.';
+		return;
+	}
+	if ($mean < 0 || $mean > 0.5) {
+		warn 'Probability must be between 0 and 0.5.';
+		return;
+	}
 
-        if ($deviation <= 0) {
-                warn 'Deviation must be a positive number.';
-                return;
-                }
+	$prob = 0.5 - $prob;
+	my $z_score_of_b = udistr($prob);
 
-        my $function = sub { my $x=shift;
-                             $E**(-$x**2/2)/sqrt(2*$PI);
-                             };
-
-        my $z_score_of_b = inv_romberg($function, 0, $prob);
-
-        my $b = $z_score_of_b * $deviation + $mean;
-        $b;
+	my $b = $z_score_of_b * $deviation + $mean;
+	$b;
 }
 
 
--- a/macros/PGstatisticsmacros.pl
+++ b/macros/PGstatisticsmacros.pl
@@ -39,22 +39,22 @@ sub normal_prob {
 	my $mean = $options{'mean'} // 0;
 	my $deviation = $options{'deviation'} // 1;
 
-	if ($deviation <= 0) {
+	if ( $deviation <= 0 ) {
 		warn 'Deviation must be a positive number.';
 		return;
 	}
 
 	my $prob;
 	if ( $a eq '-infty' ) {
-		if ( $b eq 'infty' or '-infty') {
-			$prob = 1; # did you really need us to tell you that?
+		if ( $b eq 'infty' or $b eq '-infty' ) {
+			$prob = ($b eq 'infty') ? 1 : 0; # did you really need us to tell you that?
 		} else {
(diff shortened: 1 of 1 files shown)


drgrice1:
Can you provide a test problem to test that this works as it should, and to compare the speeds of the old versus the new?

doombot-exe:
I should clarify that not all calls to normal_prob() take the same amount of time, but the problem: `Library/Rochester/setProbability10NormalDist/ur_pb_10_1.pg` was the catalyst for this change.

It seems that the use of MathObjects as the inputs for this function is what actually causes the slowdown. Using pure perl inputs does not incur nearly so much delay.

Here is an extreme example:
```perl
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "PGML.pl",
  "PGnumericalmacros.pl",
  "PGstatisticsmacros.pl",
);

TEXT(beginproblem());
Context("Numeric");

for my $i (0..25) {
    my $wasteTime = Real( normal_prob(Real(-1.3), Real(1.1)) );
}

BEGIN_PGML
This problem is here just to waste your time.
END_PGML
ENDDOCUMENT();
```

drgrice1:
I tested this with the problem `Library/Rochester/setProbability10NormalDist/ur_pb_10_1.pg` that you mentioned with seed 3782.

With this pull request the problem rendered considerably faster than without.

However, the answers with this pull request were incorrect.  The answers without this pull request were correct.  The correct answers with the mentioned seed are 0.431245, 0.555170, 0.913085, and 0.969258 (accurate to 6 decimal places).

So something isn't working right here.

doombot-exe:
How embarrassing - I confirmed that `uprob()` gave identical results, then restructured the if/else block and managed to break things in the process. I've triple-checked now that these two methods provide answers with identical accuracy across several randomizations (including seed 3782).

drgrice1 (approved):
That is much better.  I checked several variants of the mentioned problem from the library, and am now seeing the same correct answers both with and without this pull request.  I didn't compute all of them myself other than the mentioned seed from before, but it should be good.

This certainly provides significant speed improvement.  It takes less than 0.1 second with this pull request to process the pg code of the mentioned problem, and without this pull requests it takes more than 5 seconds.

Your "time waste" test case takes less than a second to process with this pull request, and more than 40 seconds without.

These times are on my test server, and of course will vary on different machines.

Alex-Jordan:
I haven't tested this, but looking at the code I see a few oversights. And also I have a few suggestions for improving the functionality beyond its original design that would be easy to fold in at the same time.

* One is that the code commentary says that `'+infty'` may be used, but the new code does not appear to allow for that leading plus sign. I'm unsure if there are problems out there that use a leading plus sign.
* The new code assumes that `$a` will never be `'infty'`. A silly but valid use would be to call `normal_prob('infty','infty')`.
* The new code allows `$a` to be a number, and `$b` to be `'-infty'`, and treats that situation the same as if `$b` is `'infty'`.
* The new code allows `$a` to be greater than `$b` and returns a negative number.

So I suggest:
* Do a validation check that `$a <= $b` at the beginning, modified appropriately if either is an infinity string. Return an error if `$a <  $b`.
* If `'+infty'` is used, turn it into `'infty'` before the real work begins.
* This file is older than MathObjects, where you might have an object whose string method is `'inf'`. It would be handy if authors could use such an object instead of needing to conform to `'infty'`. So I was thinking that it would be nice to allow more strings that would be recognized as `'infty'`. Specifically `'inf'` and `'infinity'`, and all capitalized variations.
* Since we are moving toward utf8 acceptance everywhere, could `'∞'` be another alternative? (I am hoping that one day, the text method on an infinity MathObject will return '∞'.) 


The above was all about `normal_prob`. With `normal_distr`, I'm just wondering if the right thing happens when the submitted probability is the edge case numbers of 0 or 0.5. In particular, if it's 0.5, then I think the "z-score" would be infinity.

doombot-exe:
Thank you both for your helpful feedback:

## normal_prob

* I can use regex rather than literal string equality (`/^[+-]?inf[inty]*$/`), and that approach will support `inf`, `infty`, `infinity` (but also `infininininininininity`), and it will be easy to recognize inputs with or without `+`.

* The existing function did not require `a < b`, so I did not enforce that - but P must be non-negative, so that restriction probably should have been included from the start.

* IDK about accepting utf8 `∞` in pg code, is that something that should be seriously considered?

## normal_distr
* It seems that the 'old' function capped out around ~5.20832909 (for probability = 0.5); and it seems that the (only slightly) more efficient `udistr` grows a little too quickly as probability nears 0.5 (the two begin to noticeably diverge at 0.4999). See the attached test code to compare the two (you do not need to be on this branch when testing it out).

* If it's preferable, I can revert the changes to `normal_distr` - it is not nearly as computationally intensive as `normal_prob`.

## code for comparison

This code mimics the behavior of `normal_distr` under the 'old' PGstatisticsmacros.pl implementation as well as the 'new' behavior under the proposed changes in this PR. This code is independent of the changes in this PR.

```perl
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "PGnumericalmacros.pl",
  "PGstatisticsmacros.pl",
);

TEXT(beginproblem());

$function = sub { 
    my $x=shift;
    return $E**(-$x**2/2)/sqrt(2*$PI);	
};

for my $i (1..10) {
    my $prob = 0.5 - 10**(-$i);
    TEXT("old($prob) = ", inv_romberg($function, 0, $prob), $BR);
    TEXT("new($prob) = ", udistr(0.5 - $prob), $BR, $BR);
}

$prob = 0.5;
TEXT"inv_romberg($prob) = ", (inv_romberg($function, 0, $prob), $BR);
# udistr does not return a value for 0 (prob = 0.5)
# TEXT("normal_distr($prob) = ", udistr(0.5 - $prob), $BR);

ENDDOCUMENT();
```

Alex-Jordan:
I'm not sure it would be desirable/necessary to allow for things like `infininininity`. I'd be curious what someone else thinks, maybe @drgrice1 since he's already on this thread. But regex that captured capital/lowercase versions comes to mind as a good idea. I think the `i` flag does that (but I get confused moving from one regex engine to another). And maybe allow whitespace. So maybe if it passes `/^\s*\+?\s*(inf|infty|infinity)\s*$/i` then it gets turned into `infty` and if it passes `/^\s*-\s*(inf|infty|infinity)\s*$/i` then it gets turned into `-infty`.

> The existing function did not require a < b, so I did not enforce that - but P must be non-negative, so that restriction probably should have been included from the start.

Right, the old code did not enforce that. Its description mentions the interval (a,b) which I read as requiring a <= b, so I'd look at this as a flaw with the old code, where its description does not agree with what it actually does.

> IDK about accepting utf8 ∞ in pg code, is that something that should be seriously considered?

It's not a serious thing for now. But students submit answers with characters like ∞ more often now. One common route is they copy-paste from where they see it on the screen already, knowing the answer but not knowing the WeBWorK convention to use `inf` or `infinity`. If a Context knew to parse that string into the infinity MathObject, students would be able to get away with this. And once you are doing that, the MathObject might as well have ∞ be its string method. And then if you are writing a stats problem and using Math Objects, you could drop in the infinity Math Object. But it's a ways off before all this will happen, so forget I mentioned it :) 

For `normal_distr`, my instinct would be to check right away if the input `== 0.5`, and then just skip to returning `inf` or `infinity`. Which could be input to a `Compute()` call if the problem is using MathObjects. The problem author would need to understand that if their randomization permits 0.5 to be fed here, that they need to program some cases. And I'm not worried about this end changing behavior from the old code, because the old code is sort of just wrong. If a student uses some other tool to do their work, the probably will not get 5.2 as a z-score and would be inappropriately marked incorrect. My guess is that problem writers using these tools have just avoided getting near 0.5 as an input up to now.

I'm happy to make edits as a commit on top of what is here, if you are busy. And of course I will test once we think it's good. I was hoping to just merge it yesterday after a quick test, but I saw the `+infty` thing and then I couldn't stop the flood of ideas/observations.

drgrice1:
I agree that it would generally not be desirable to accept infinininity.

doombot-exe:
`normal_prob` now accepts (only) the three forms of infinity string (inf, infty, & infinity), case-insensitive.
`normal_distr` does not accept 0.5 as an input, this case is left to the problem author - there's no reason to call this function in such cases (and as mentioned, this function pre-dates MathObjects, so infinity is not really a 'valid' response).

As for students inputting utf8 infinity in their responses, that doesn't seem relevant to these changes. This is a function that is enabled during problem creation, and not one that students can use in their answers. Can you imagine a use-case where student responses would be passed into this function?

Alex-Jordan:
Hi Victor,

Thanks for your work on this. I hope to be able to review it and merge soon.

You can forget about what I brought up regarding '∞'. But since you asked, yes, this could (in the future) come up during problem setup. The situation would be that `$x` is a MathObject. Some seeds might make `$x` be a `Real`, and other seeds might make it be an `Infinity`. And for convenience, an author uses `normal_prob(0,$x)`. Currently, `$x`'s string method will output `inf`, so your addition of that as a valid string covers this situation. But I envision a future where the string method of an Infinity object will either be '-∞' or '∞'. This kind of thing is a big project for a future release, and would be geared toward sweeping implementation details like the string "inf" out of sight of the student user. I shouldn't have brought it up here.

Alex-Jordan:
Thanks again and I'm sorry to be so slow. I tested and everything I tried worked as intended.

I did some file cleanup at https://github.com/rederly/pg/pull/4, which goes into this branch. So if you approve of that, it will be merged and then this PR would follow suit.

Worth noting: once I am about this close to 0.5, using 0.4999999, the results I get from Excel and WolframAlpha agree with each other, but are outside the default error tolerance from what this now produces. The old version was much farther off, so this is a huge improvement. This is just a note, that the either `udistr` routine is not perfect, or WolframAlpha and Excel aren't perfect, or both. Anyway, an author just needs to be savvy enough to know to avoid such extreme values. And we can expect them to do that.

PR #516, opened 2021-01-03
add pg/macros/draggableProof  which gets it's javascript for nestable…

mgage:
… from cdnjs instead of a file in webwork2/htdocs/js

This is a minimal change to make versions of draggableProof.pl work no matter what js files are available in 
webwork2/htdocs/js .  

This will intercept calls to draggableProof.pl that used to go to the library macros at OPL/macros/MC.  That file 
expected a copy of nestable in the course html directory for a problem that used draggableProof.pl-- a file that was often not there. 

This is not a final solution.  I suggest creating a file PGdraganddrop.pl with this code (the mechanism is used for more than just proofs) and then make additional changes as to the source of the javascript files -- possibly in a webwork2/htdocs/js/pg tree.  Changing the file names alerts people that this is no longer exactly the same as draggableProof.pl and they can change the file to load in the loadMacros()  header of their PGdraganddrop problem.

### DIFF
--- a/macros/draggableProof.pl
+++ b/macros/draggableProof.pl
@@ -1,0 +1,284 @@
+# Done: show possible choices in TeX mode
+# To do: display student answers and correct answers in TeX mode properly.
+# To do: put jquery.nestable.js in a universal spot on every webwork server.
+
+loadMacros("PGchoicemacros.pl");
+
+sub _draggableProof_init {
+  PG_restricted_eval("sub DraggableProof {new draggableProof(\@_)}");
+
+# post global javascript
+# using sources in cdnjs.cloudflare.com
+
+main::POST_HEADER_TEXT(main::MODES(TeX=>"", HTML=><<END_SCRIPTS));
+
+	<script src= "https://cdnjs.cloudflare.com/ajax/libs/nestable2/1.6.0/jquery.nestable.min.js">
+	<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/nestable2/1.6.0/jquery.nestable.min.css">
+	</script>
+
+    <script>
+      function _nestableUpdate (e) {
+          var dd = e.length ? e : \$(e.target);
+          var ans = dd["0"].getAttribute("data-ans");
+          var li_tags = dd["0"].getElementsByTagName("li");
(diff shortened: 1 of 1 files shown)


pstaabp:
@mgage I notice this is your develop branch.  Perhaps this is better to have as a separate branch since it is a fix.  

Also, what are you thoughts on making a PGdraganddrop.pl file instead?  Push to a later date? ]

This does fix the dragging problem in draggable proof with the updated PR in  openwebwork/webwork2#516

drgrice1:
@pstaabp:  Does it matter what branch he puts the fix on his fork?

I will try to test this pull request in the jQuery update branch.

drgrice1:
@pstaabp:  I tested this with the jQuery update branch (https://github.com/openwebwork/webwork2/pull/1167), and it works fine.  I also tested the version of draggableProof.pl in the OPL, i.e., the file OpenProblemLibrary/macros/MC/draggableProof.pl, and it also works with no problem with the jQuery update branch assuming that the jquery.nestable.js file is in the expected location for that macro.  So I am not sure what dragging problem you are talking about.  I suspect that you may have not ensured that your browser cache was cleared.  I did see an issue the first time I loaded the problem with the OPL version of draggableProof.pl that resulted from my browser using the cached version of some javascript.

taniwallach (approved):
Tested with the current develop branch and with the problem `Library/Rochester/Authoring/setdraggable/ABCabc.pg`.
Before the patch - expected draggable items were not draggable.
After the patch - they work.

Since @drgrice1 also tested with the jQuery update branch - I think we can merge this.

drgrice1 (approved):
Yeah, sorry I should have added my approval before.

taniwallach:
Maybe we should install `jquery.nestable.min.js` and `jquery.nestable.min.css` using `npm`:
  - https://www.npmjs.com/package/nestable2
  - https://github.com/RamonSmit/Nestable2
so we can avoid the dependence on the CDN.

drgrice1:
@taniwallach:  I agree.  There are many things that we should install via npm.  Lets get my jquery update branch in first though.  Once that is in, then the basic infrastructure is in place to start using npm.

PR #518, opened 2021-01-13
Add aliases to all context classes, and allow alternative tokens

dpvc:
This PR implements features in MathObjects to address the issues raised in #517.

The first is to support `alias` in more of the context classes, like operators, parens, variables, etc.  This makes handling aliases more uniform, and includes some error checking (e.g., that the object the alias points to actually exists).  This involves the most changes, especially in the parser.  In particular, one mush resolve the aliases in a number of locations in order to get the definition of the base object.  That involved adding `resolve()` and `resolveDef()` methods to the basic Context Data object, and using those methods in `Parser.pm` and other file whenever a token's data is needed.  I think I got all of the locations this needs to be done, but there might be some lurking out there.  These changes should not cause problems with existing usage, since there are no aliases in most of the cases where these are used (since it is a new feature), but since this does touch on fundamental areas of the parser, it needs some serious testing.  The existing aliases (for functions and strings) produce messages based on the resolved names rather than the originals (e.g., `arcsin` becomes `asin` automatically), so the resolved name is used in error messages.  That might be more confusing with aliased operators, etc.  (So if `**` where aliased to `^`, then `x**` would give a message about `^` needing a second argument.  This might want to ben revisited.)

I have made the changes for this feature in the first commit, so you may want to look at that in isolation in order to make the changes easier to understand.

The second feature is the addition of an `alternatives` property that provides alternative tokens for any given Context object.  For example

```
Context()->operators->set('-' => {alternatives => ["\x{2212}"]});
```

would add U+2212 (minus) to the list of tokens that act as the minus sign.  This is similar to aliasing U+2212 to the minus definition, except that (1) there is a configuration option to control whether alternatives are used or not (aliases are always used), and (2) if an object is removed, so are its alternatives (while aliases would have to be removed separately).  As part of this change, the name patterns for the various objects have been loosened up to allow Unicode characters in variable and function names, etc, and to make them less restrictive in other ways.  Also, the pattern for numbers has been moved so that strings and other objects can start with a number (an issue that has been raised in the past).  E.g., unicode Greek letters can be used for variables, as can `x_0`.

The second commit adds this functionality, and the third adds the configuration flag for it.

Third, there is now a feature where the full-width unicode characters at U+FF01 to U+FF5E can be converted to their ASCII equivalents automatically.  This should make using Chinese keyboards easier (and addresses awmorp's issue).

This is implemented in the fourth commit.

Finally, I have added a number of alternative Unicode versions to the default context definitions as an illustration of how this works.  If this PR is approved, then other of the contexts in PG/macros may want to be extended to include appropriate unicode values, or require other changes to avoid overriding these changes.  I have not looked at the problem library to see what problems might be adjusting the contexts in ways that might benefit from changes.

Finally, there is a corresponding pull request for webwork2 that needs to be used in order to set the `parseAlternatives` and `convertFullWidthCharacters` options on a course-wide basis.  Without that PR, you have to set these in the given context, and call the context's `update` method, as in

```
Context("Numeric")->flags->set(
  parseAlternatives => 1,
  convertFullWidthCharacters => 1
);
Context()->update;
```

in order to get the new features.

### DIFF
--- a/lib/Parser/Context/Default.pm
+++ b/lib/Parser/Context/Default.pm
@@ -44,17 +44,18 @@ $operators = {
            class => 'Parser::BOP::add'},
 
    '-' => {precedence => 1, associativity => 'left', type => 'both', string => '-',
-           class => 'Parser::BOP::subtract', rightparens => 'same'},
+           class => 'Parser::BOP::subtract', rightparens => 'same', alternatives => ["\x{2212}"]},
 
    'U' => {precedence => 1.5, associativity => 'left', type => 'bin', isUnion => 1,
-           string => ' U ', TeX => '\cup ', class => 'Parser::BOP::union'},
+           string => ' U ', TeX => '\cup ', class => 'Parser::BOP::union',
+           alternatives => ["\x{222A}","\x{22C3}"]},
 
    '><'=> {precedence => 2, associativity => 'left', type => 'bin',
            string => ' >< ', TeX => '\times ', perl => 'x', fullparens => 1,
-           class => 'Parser::BOP::cross'},
+           class => 'Parser::BOP::cross', alternatives => ["\x{00D7}"]},
 
    '.' => {precedence => 2, associativity => 'left', type => 'bin',
-           string => '.', TeX => '\cdot ', class => 'Parser::BOP::dot'},
+           string => '.', TeX => '\cdot ', class => 'Parser::BOP::dot', alternatives => ["\x{2219}", "\x{2022}"]},
 
    '*' => {precedence => 3, associativity => 'left', type => 'bin', space => ' *',
            string => '*', TeX => '', class => 'Parser::BOP::multiply'},
@@ -110,7 +111,9 @@ $parens = {
    '(' => {close => ')', type => 'Point', formMatrix => 1, formInterval => ']',
            formList => 1, removable => 1, emptyOK => 1, function => 1},
    '[' => {close => ']', type => 'Point', formMatrix => 1, formInterval => ')', removable => 1},
-   '<' => {close => '>', type => 'Vector'},
+   '<' => {close => '>', type => 'Vector',
+           alternatives => ["\x{2329}","\x{3008}","\x{27E8}"],
+           alternativeClose => ["\x{232A}","\x{3009}","\x{27E9}"]},
    '{' => {close => '}', type => 'Point', removable => 1},
    '|' => {close => '|', type => 'AbsoluteValue'},
    'start' => {close => 'start', type => 'List', formList => 1,
@@ -133,7 +136,7 @@ $lists = {
 
 $constants = {
    'e'  => exp(1),
-   'pi' => {value => 4*atan2(1,1), TeX => '\pi ', perl => "pi"},
+   'pi' => {value => 4*atan2(1,1), TeX => '\pi ', perl => "pi", alternatives => ["\x{03C0}","\x{1D70B}"]},
    'i'  => {value => Value::Complex->new(0,1), isConstant => 1,                        string => "i", perl => "i"},
    'j'  => {value => Value::Vector->new(0,1,0)->with(ijk=>1), TeX => '\boldsymbol{j}', string => "j", perl => "j"},
    'k'  => {value => Value::Vector->new(0,0,1)->with(ijk=>1), TeX => '\boldsymbol{k}', string => "k", perl => "k"},
@@ -183,7 +186,7 @@ $functions = {
    'log'   => {class => 'Parser::Function::numeric', TeX => '\log', simplePowers => 1},
    'log10' => {class => 'Parser::Function::numeric', nocomplex => 1, TeX => '\log_{10}'},
    'exp'   => {class => 'Parser::Function::numeric', inverse => 'log', TeX => '\exp'},
-   'sqrt'  => {class => 'Parser::Function::numeric', braceTeX => 1, TeX => '\sqrt'},
+   'sqrt'  => {class => 'Parser::Function::numeric', braceTeX => 1, TeX => '\sqrt', alternatives => ["\x{221A}"]},
    'abs'   => {class => 'Parser::Function::numeric'},
    'int'   => {class => 'Parser::Function::numeric'},
    'sgn'   => {class => 'Parser::Function::numeric', nocomplex => 1},
@@ -219,7 +222,7 @@ $functions = {
 };
 
 $strings = {
-   'infinity'  => {infinite => 1},
+   'infinity'  => {infinite => 1, alternatives => ["\x{221E}"]},
    'inf'  => {alias => 'infinity'},
    'NONE' => {},
    'DNE'  => {},
@@ -239,6 +242,8 @@ $flags = {
   allowBadOperands => 0,               # 1 is used by Typeset context (types need not match)
   allowBadFunctionInputs => 0,         # 1 is used by Typeset context (types need not match)
   allowWrongArgCount => 0,             # 1 = numbers need not be correct
+  parseAlternatives => 0,              # 1 = allow parsing of alternative tokens in the context
+  convertFullWidthCharacters => 0,     # 1 = convert Unicode full width characters to ASCII positions
 };
 
 ############################################################################
@@ -369,7 +374,7 @@ $context->{name} = "Complex";
 #
 $context = $context{"Complex-Vector"} = $context{Complex}->copy;
 $context->operators->redefine(['><','.'],from=>'Vector');
-$context->parens->add('<' => {close => '>', type => 'Vector'});
+$context->parens->redefine('<',from=>'Vector');
 $context->parens->set(
   '(' => {type => "Point", formMatrix => 0},
   '[' => {type => "Point", formMatrix => 0},
--- a/lib/Parser/Context/Functions.pm
+++ b/lib/Parser/Context/Functions.pm
@@ -11,7 +11,7 @@ sub init {
   $self->{dataName} = 'functions';
   $self->{name} = 'function';
   $self->{Name} = 'Function';
-  $self->{namePattern} = qr/[a-zA-Z][a-zA-Z0-9]*/;
+  $self->{namePattern} = qr/\w+|\x{221A}/;
   $self->{tokenType} = 'fn';
 }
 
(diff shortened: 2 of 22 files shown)


lib/Parser/Context/Functions.pm
```diff
@@ -11,7 +11,7 @@ sub init {
   $self->{dataName} = 'functions';
   $self->{name} = 'function';
   $self->{Name} = 'Function';
-  $self->{namePattern} = qr/[a-zA-Z][a-zA-Z0-9]*/;
+  $self->{namePattern} = qr/\w+|\x{221A}/;
```
taniwallach, line 14: Maybe add a comment that `x{221A}` is Unicode square root.

lib/Parser/Context/Default.pm
```diff
@@ -44,17 +44,18 @@ $operators = {
            class => 'Parser::BOP::add'},
 
    '-' => {precedence => 1, associativity => 'left', type => 'both', string => '-',
-           class => 'Parser::BOP::subtract', rightparens => 'same'},
+           class => 'Parser::BOP::subtract', rightparens => 'same', alternatives => ["\x{2212}"]},
 
    'U' => {precedence => 1.5, associativity => 'left', type => 'bin', isUnion => 1,
-           string => ' U ', TeX => '\cup ', class => 'Parser::BOP::union'},
+           string => ' U ', TeX => '\cup ', class => 'Parser::BOP::union',
+           alternatives => ["\x{222A}","\x{22C3}"]},
```
taniwallach, line 51: Should `\x{22C3}` which is N-Ary Union be here? I'm not sure it is meant to be used as a binary operator (although if not, abuse is likely)
https://www.compart.com/en/unicode/U+22C3

lib/Parser/Context/Variables.pm
```diff
@@ -26,15 +26,12 @@ sub init {
   $self->{dataName} = 'variables';
   $self->{name} = 'variable';
   $self->{Name} = 'Variable';
-  $self->{namePattern} = qr/[a-zA-Z][a-zA-Z0-9]*/;
+  $self->{namePattern} = qr/\w+/;
   $self->{tokenType} = 'var';
-  $self->{precedence} = 5;
+  $self->{precedence} = 10;
```
taniwallach, line 31: The precedence change is "black magic" for me. Maybe for the record put an explanation here or in the file.

taniwallach (approved):
@dvpc - Amazing.

Thanks for including all these various Unicode forms of these standard symbols/operations.
I tested Unicode minus, infinity, pi, and the full-width parenthesis as samples of what is now allowed and that the course settings turn the new features on and off.

I did my best to follow the logic of the changes, and did some testing, but cannot say I fully followed all the code changes or tested all the cases. Even the cursory reading makes me greatly appreciate the thought and effort which went into these changes, as well as that which you put into designing MathObjects in general. The WW community certainly owes you greatly for this, as does the greater Mathematical community for your work on MathJax. Thanks!

I think that full testing may require turning some students into test-subjects to see if some less-than-obvious bug/issue exists in the revised code.

drgrice1 (approved):
This looks good, and will be a feature that many will utilize.  Contrary to my original assessment of this concept, I will probably use it as well.

Everything looks good in my initial analysis of the code.  This is a big change though, and testing will be important.  I think that it would be a good idea to merge this soon, so that this can get better testing from the community than it will receive in a pull request.

lib/Parser/Context/Functions.pm
dpvc: Done.

lib/Parser/Context/Variables.pm
dpvc: I've added a comment to the file.

The change is because I moved the number pattern from being before all the specifically named items (functions, variables, operators, etc.) too after them, so that names can include leading numbers (e.g., a string `1st-degree` would be allowed now, but not before).  I set the number precedence to 5 (standard tokens are 0), to get a little room to insert others in between, if needed in the future.  Then I changed this precedence to 10 to put it after numbers (with a little room for more in-between).  This does mean that variable names can't start with a number.  One can set this to between 0 and 5 to allow variables that start with a leading number, if that is desired.

drdrew42:
I am still digesting, but my review of this code is in progress.

lib/Parser/Context/Default.pm
dpvc: I suppose it depends on the philosophy behind allowing unicode characters to be used.  Does one want to map everything that people might use, or only the "correct" ones.  For example, do you map en-dash to minus (I know parsers that do), or just the dash and minus?  Do you map U+00B5 (micron sign) to U+03BC (Greek small letter mu)?

I'm not sure people looking for the union symbol will make the distinction between the correct one (U+222A) and the n-ary one (U+22C3), and certainly the original "U" (U+0055) is not correct.  

I'm OK with taking it out, if people think that is best, but was using it as an example of how to map multiple Unicode points as alternatives (there are other examples of that as well, so it is not critical).

taniwallach: I also am not sure what is the right call here.
Abuse of the "wrong" one is likely and this is a "union" operator.

lib/Parser/Context/Variables.pm
taniwallach: Thanks for explaining/.

Alex-Jordan:
@dpvc What Tani said: Amazing.

I've tested all but the double-wide characters part. Maybe @awmorp can take a specific look into that?

I was looking for cracks, like what happens if you alias an alias. And are alias variables excluded from test points. And it all worked. (I'm not surprised.)

Is there any merit to making `**` an alias for `^` by default?

Alex-Jordan (approved):

dpvc:
@drgrice1, thanks for the review.

@Alex-Jordan, thank you as well.

> I was looking for cracks, like what happens if you alias an alias. And are alias variables excluded from test points.

Good man!  Those were certainly important considerations.  There are still some things that might be problematic, like making an alias to an item, then removing the item it is an alias to; that should probably cause an error.  (As it currently stands, you could remove the target of the alias and then create a new alias back to the original one, and get an alias loop that would never resolve.  But if you do that, you deserve it!)  But for the most part, I think it is pretty solid.

> Is there any merit to making `**` an alias for `^` by default?

Yes, in that it ties the two together, so any changes to `^` are automatically reflected in `**` (so you don't have to make duplicate changes).  But no, in that if you do, then error messages about `**` will refer to `^` (so `x ** ` with no power will complain about `^` needing a second argument).

I am thinking about a second PR for later that cleans up some of default context declarations, like removing the `' *'`, `'* '`, `' /'` and `'/ '` entries for the never-used non-standard precedence settings that were indented to allow `sin 2x` to be `sin(2x)` rather than `sin(2) x`, and `1/2x` to be `1/(2x)` rather than `(1/2)x`, which would be obtained by `1/2 x`.  That is, these were for a context in which spaces were significant.  It turned out to be impractical to make it work with only precedence settings, and the precedences listed in `Default.pm` are not the actual ones used (they are switched later on).  It has bothered me ever since that they are there, because if you want to change or disable multiplication or division, you have to change all three versions of each.  Since no one really knows about the ones with spaces, that is a problem.

Cleaning that up needs to be done somewhat carefully, as there are files in `pg/macros` that modify them, and since macro files get duplicated in local macro libraries, it is important to make their removal not cause errors on currently working macro files that aren't in the OPL.  I didn't want to clutter up this PR any further than it already is.

There should also be a third PR to add alternatives to other context files in `pg/macros`, like for inequalities.

I thought it best to get the infrastructure in place before making those kinds of changes.

awmorp:
> I've tested all but the double-wide characters part. Maybe @awmorp can take a specific look into that?

I don't have a Webwork development environment so I can't test the fullwidth substitution fully. But I did try to test the substitution code in a pg problem ([code here](https://gist.github.com/awmorp/0e85ca438c1f590eb8d9d3fd365e3785)). The transliteration operation 
`tr/\N{U+FF01}-\N{U+FF5E}/\x21-\x7E/`
from [here](https://github.com/openwebwork/pg/pull/518/commits/33bfe1c80a79f8b1ba027bf04d520b9dccc85706#diff-458d805ec4c1f9a287b4ae35f467e0b97cc764c6beafec6e9e2805e6c9308546R101) gives me an error message: 
> ERRORS from evaluating PG file: 
>  Invalid range "}-\" in transliteration operator at line 27 of (eval 979)

The only way I could get it to work was to use literal fullwidth characters in the range:
`tr/！-～/!-~/`
According to [this documentation](https://perldoc.perl.org/perlop#Quote-and-Quote-like-Operators) it should be ok to use \N{U+xxxx} in a tr so not sure if it is a quirk with pg ?

awmorp:
> not sure if it is a quirk with pg ?

It was a quirk with pg. Replacing the \ with ~~ did the trick. I've tested the fullwidth substitution using [this pg question](https://gist.github.com/awmorp/0e85ca438c1f590eb8d9d3fd365e3785) and it seems to be working as expected. Thanks!

Alex-Jordan:
I merged https://github.com/openwebwork/webwork2/pull/1174, and merging this one now since it has 3 approvals plus @awmorp's testing.

Alex-Jordan:
On a server that is at the current develop branch tips of pg (c43b1cd0ea521f8b895d5682893b252076f7644f) and webwork2 (02fadca640148e97a60527e4fc3b528f3f98c9ea) with a clean diff, the following PG problem breaks:

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
);

Context("Interval");
$domain = Interval("[1,inf)");

ENDDOCUMENT();
```

There is an error report that `Infinite endpoints must be open at line 9 of (eval 1216)`. 

I used git bisect and it traced the issue to 15ce46cb6d742c35fa1083fb8204b36ab6f82944, from this PR.

dpvc:
Thanks, I've pushed another commit and will open a new PR for it.

Alex-Jordan:
Possibly another issue from the work here. For example, there is this OPL problem: `Library/LoyolaChicago/Precalc/Chap9Tools/Connally3-9-Tools-35.pg`. At one point it has the line:

`$ans2 = Compute("-INF");`

This leads to the error:

```
Operand of '-' can't be a word; see position 1 of formula at line 100 of [PG]/macros/Parser.pl
```

PR #520, opened 2021-01-20
Add cdn draggable proof

doombot-exe:
draggableProof is a macro that is not 'official' PG (yet?) - so we add it to available macros with a reference to pull the necessary jquery.nestable javascript dependency from cdn. Version is hardcoded so that this is stable.


doombot-exe:
sorry, this was not intended to go to openwebwork - this was intended as a local PR. I will make another for openwebwork:dev if there is interest in adding draggableProof to the 'official' PG macros.

PR #521, opened 2021-01-30
Use the correct close string (resolves issues reported by Alex)

dpvc:
This fixes the problem reported by [Alex](https://github.com/openwebwork/pg/pull/518#issuecomment-770272968).  It was due to using a hale-open, half-close interval (where the close delimiter isn't the usual one for the open delimiter).  This puts back the original action on this line that was incorrectly changed in the original PR.


drgrice1 (approved):
This fixes the problem.  I will go ahead and merge it.  I don't think this needs any detailed reviews.

drgrice1:
Thanks for the fix, by the way.

dpvc:
No problem (it was my error, after all).

PR #523, opened 2021-02-02
Update AnswerChecker.pm

pschan-gh:
Minor nitpick with plural form of "fraction of integers" in Hints.


Alex-Jordan:
In the interest of modularizing the code, there could be a separate method for pluralization that defaults to appending an 's' but has code for, at minimum, recognizing "fraction of integers" and doing something different. It could also handle the awkward code in there now for dealing with "interval or union".

pschan-gh:
Indeed, that's probably a better solution in the long run. Thanks!

Alex-Jordan:
Also this has me thinking about localization issues. Sorry for the mission creep!

Issue #525, opened 2021-02-16
MathObject answers no longer accept numbers starting with a decimal


drgrice1:
This seems to be another regression introduced by https://github.com/openwebwork/pg/pull/518.

If a number starts with a decimal you get the error "Missing operand before '.'".  If you precede the decimal with a zero, it works as before.

dpvc:
Ah, yes, that would be because the for product is now before the number pattern in the token pattern (that was to allow other tokens to begin with numbers).  I'll look into how best to resolve that.

drgrice1:
Here is a minimal example in which this occurs (if it helps).

```perl
DOCUMENT();
loadMacros("PGstandard.pl", "MathObjects.pl", "PGML.pl");
TEXT(beginproblem());

$ans = Compute(".3333");

BEGIN_PGML
Compute [`\frac{1}{3}`].

[_______]{$ans}
END_PGML
ENDDOCUMENT();
```
This problem fails to load and gives the error "Missing operand before '.'"

dpvc:
I made a pull request to resolve the issue.

PR #526, opened 2021-02-16
Fix problem with decimals with no leading 0 (#525)

dpvc:
This PR adds a `patternPrecedence` property to context items that allows individual tokens to be prioritized in the pattern list.  It then uses that to put the `.` operator pattern after the number pattern, so that decimal numbers with no leading zero will be identified first, and then if `.` is not used as a decimal, it will be recognized as a dot product.

Without the patch, `Compute(".123")` produces an error about a missing operand for `.` but with the patch, it should produce .123 as expected.   Also

```
Context("Vector");
TEXT(Compute("<1,2> . <3,4>"));
```

should produce an output of 11.

Resolves issue #525.


drgrice1 (approved):
This fixes the problem
Thanks.

PR #529, opened 2021-02-21
Switch from opening a new window to opening a dialog when an image is clicked

drgrice1:
The current method of having a new browser window or tab open when an image in a PG problem is clicked is intrusive.  Furthermore the behavior that occurs is different depending on which browser is in use.

This changes to opening a bootstrap modal dialog  when an image is clicked instead.  That dialog can be moved around and the image in the dialog can be resized using the zoom in and zoom out buttons in the dialog.  The image can also be resized using the +/- keys or the mouse wheel.

This is the pg part that creates the DOM element.

The javascript for this is in https://github.com/openwebwork/webwork2/pull/1207

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -3034,12 +3034,7 @@ sub image {
 			|| $displayMode eq 'HTML_asciimath'
 			|| $displayMode eq 'HTML_LaTeXMathML'
 			|| $displayMode eq 'HTML_img') {
-			my $wid = ($envir->{onTheFlyImageSize} || 0) +30;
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):
This works for me as advertised.

taniwallach (approved):
Tested with the paired https://github.com/openwebwork/webwork2/pull/1207 which needs some minor merge conflicts fixed due to recent changes. Other than that, I think this is ready to be merged.

pstaabp (approved):

pstaabp:
Looks great.

PR #530, opened 2021-02-21
allow access to course environment PG_VERSION

drdrew42:
this PR provides access to the pg version as read in from `pg/VERSION` by the CourseEnvironment.pm module from ww2

requested in issue: openwebwork/webwork2#1204

also tagging openwebwork/webwork2#1121

### DIFF
--- a/lib/WeBWorK/PG/IO.pm
+++ b/lib/WeBWorK/PG/IO.pm
@@ -50,6 +50,7 @@ BEGIN {
                 fileFromPath
                 directoryFromPath
                 createDirectory
+				pgVersion
         );
 
 	our %SHARE = map { $_ => __PACKAGE__ } @SHARED_FUNCTIONS;
@@ -507,6 +508,10 @@ END
 		}
 	}
 	return $ret;
+}
+
+sub pgVersion {
+	return $CE->{PG_VERSION};
 }
 
 =back


drgrice1:
I was thinking about this, and I think it would make more sense to pass the PG version in as an environment variable, rather than creating a method in IO.pm.  This can be easily done by adding something like `$envir{pgVersion} = $ce->{PG_VERSION},` to `lib/WeBWorK/PG.pm`.  Of course this requires modifying the webwork2 code, rather than the pg code.

drdrew42:
On the ww2 side, my intuition is that using the problem environment hash (as produced in WeBWorK/PG.pm) would be less convenient than using the course environment (WeBWorK/CourseEnvironment.pm) for the purposes of retrieving the pg version. The course environment already contains the pg version and is readily available across ww2 because it is embedded in every apache request object.

With regards to the specific issue of problems that need branching logic to account for changes to PG, I would consider it undesirable to promote the direct use of the PGcore object inside pg code (which is what would be required if the pg version were to be stashed inside the problem environment hash).

drgrice1:
There are already course environment variables added to that hash.  This would still be using the course environment to retrieve the pg version in the same way that this pull request does.  This doesn't change that.

I don't understand why it is undesirable to access the information via `$PG->{envir}{pgVersion}`.  The PGcore object is there and accessible in pg code, regardless of whether you "promote" its usage or not.  There is really nothing wrong with using that for special case things like this.

drgrice1:
Also, dealing with a method that may or may not be defined is messier than dealing with a hash key that may or may not be defined.

If this is a hash key and you want the version with a fall back you can do
```perl
$var = $PG->{envir}{pgVersion} // 0;
```

On the other hand with the method approach you need to do something like
```perl
$var = pgVersion eq "pgVersion" ? 0 : pgVersion;
```
That is really not very intuitive.

mgage:
I kind of agree with this.   It means that the pgVersion is one of the things supplied to PG when it is initialized and 
has to be specified by what ever framework calls it. 

We could ask that PG knows it’s own version (e.g. hardwired variable) but then we would want a hook on github that automatically updated
this variable with each release. That was too complicated for me at the time I introduced the VERSION files. :-)

drdrew42:
> Also, dealing with a method that may or may not be defined is messier than dealing with a hash key that may or may not be defined.

This is certainly a fair point. However, I still do not like putting ww2 in charge of determining the pg version. In such case as a user updates pg without similarly updating ww2, we would be in a situation where `$PG->{envir}{pgVersion}` is undefined and it cannot then be accurately determined which pg version to account for in the pg code. 

I am not suggesting that we base our decision on worst-case-scenarios such as this, but the IO.pm approach does have the benefit of resolving this issue without requiring an update to ww2.

|| pro | con |
|--:|:-:|:-:|
|pg/IO.pm | doesn't require ww2 update <br /> readability in pg code | null checking subs is a pain  |
|webwork/PG.pm | easier null checking | pg depends on update to ww2 |

Now that I look at it, null checking for this subroutine is not terribly egregious: 
```perl
my $pgVersion = (defined &pgVersion) ? pgVersion() : 0;
```

drgrice1:
Using different versions of webwork2 and pg is something that should be discouraged.  Even more than discouraged, it should be something that really just shouldn't be done.  There are many incompatibilities between the older different versions already, and the number of these incompatibilities will only continue to increase if the WeBWorK project really wants to get anywhere.  The PG code is going to have to change to accomodate communication between whatever frontend is used and PG.  The use case scenarios that are coming up are making the need for this greater.

Alex-Jordan:
I agree with the principle that PG should know its own version, and not rely on WW2. I look at it from the perspective that WW2 is a front end for PG, and that other front ends are under development. I know that WW2 and PG are still intertwined. But isn't a long term goal to separate them?

drgrice1:
PG will be intertwined with any frontend.  This is really about frontend/PG communication, not specifically WW2/PG communication.

drgrice1:
If the argument is for PG to know its own version (which would be good), then this pull request still does not do that.  This pull request uses the course environment, which is a front end thing.

I think that when Mike suggested putting this into IO.pm, he meant to read the PG version from the pg/VERSION file using IO.  That can be done from IO.pm, and does not depend on any frontend.

drgrice1:
How about doing this for the pgVersion sub instead:
```perl
sub pgVersion {
	my $PG_VERSION = "unknown";

	my $PG_version_file = __FILE__ =~ s/lib\/WeBWorK\/PG\/IO\.pm/VERSION/r;
	if (-r $PG_version_file) {
		open(my $fh, "<", $PG_version_file) or warn "Can't open $PG_version_file.", return $PG_VERSION;
		local $/;
		my $data = <$fh>;
		close($fh);
		my $PG_COPYRIGHT_YEARS = "";
		eval($data);
		warn $@ if $@;
	}

	return $PG_VERSION;
}
```
This is an approach that is completely independent of any front end.  `$PG->{envir}` and `$CE` both come from the front end, and so using either of those will depend on what the front end decides to give you.  The webwork2 code defines `$CE->{PG_VERSION}` in essentially the same way, but a frontend would not have to do this and could set that to be whatever the front end wants it to be.

Just a possible thought.  I am sure there are problems here too.

dpvc:
How about this:  in `pg/lib/PGcore.pm`, between the `package PG;` and the `use strict` add

``` perl
use File::Spec;
sub getVersion {
  my $dir = (File::Spec->splitpath(File::Spec->rel2abs(__FILE__)))[1];
  do "${dir}../VERSION";
  warn "Error loading PG VERSION file: $!" if $!;
  warn "Error processing PG VERSION file: $@" if $@;
  return $PG_VERSION || 'unknown';
}
my $PG_VERSION = getVersion();
```

and then in the `new` subroutine, in the `my $self = {...}`, add

``` perl
                VERSION                   => $PG_VERSION,
```

That will load the version out of the `VERSION` file and include it in the `$PG` object, so you can use `$PG->{VERSION}` to get the version.  This is purely within PG.

This will show the error messages during startup if there is a problem loading or running the file, but it will not prevent WeBWorK from starting up (and the version will be listed as "unknown" in that case.

drdrew42:
This one's just a nit, but instead of File::Spec, I think it's more straightforward to use dirname from File::Basename.
I'm also curious what everyone's thoughts are on stashing the PG_VERSION inside of %ENV? This approach would ensure that all aspects of the WeBWorK codebase have access to PG's self-declared version. From there we can place it into the PGcore object in the new method as @dpvc suggested above...

```perl
BEGIN {
	use File::Basename qw(dirname);
	my $dir = dirname(__FILE__);
	do "${dir}/../VERSION";
	warn "Error loading PG VERSION file: $!"    if $!;
	warn "Error processing PG VERSION file: $@" if $@;
	$ENV{PG_VERSION} = $PG_VERSION || 'unknown';
}
```

dpvc:
I'm fine with the change to use `dirname`.  I haven't use the file-name management tools in perl in quite some time, so just pulled the first one I could find.  The `BEGIN` is probably also a good idea.  I had used a function, here, in order to not have extra variables (`$dir`) in the PGcore package, but it is a minor issue.

I'm OK with using `ENV` for this (and setting `$PG->{VERSION}` to it in the `new` method).

drgrice1:
I don't think that `$dir` is added to the PGcore package with @drdrew42's approach.  If I am not mistaken the `my` declaration keeps the scope local to the BEGIN block.  Testing in a Perl script seems to confirm this.  I can't find documentation on that though.

I think adding PG_VERSION to the %ENV hash is not a bad idea.

I don't like that this doesn't work with `use strict`, but that isn't really a big deal.  If you want to make this work with `use strict` you could use something like my original approach:
```perl
BEGIN {
	my $PG_VERSION = "unknown";

	use File::Basename qw(dirname);
	my $dir = dirname(__FILE__);
	my $PG_version_file = "${dir}/../VERSION";
	if (-r $PG_version_file && open(my $fh, "<", $PG_version_file)) {
		local $/;
		my $data = <$fh>;
		close($fh);
		my $PG_COPYRIGHT_YEARS = "";
		eval($data);
		warn "Error processing PG VERSION file: $@" if $@;
	} else {
		warn "Error loading PG VERSION file.";
	}

	$ENV{PG_VERSION} = $PG_VERSION;
}
```
That also works here (of course after the `use strict` though).

I will let you determine what is best here.

dpvc:
> I don't think that `$dir` is added to the `PGcore` package with @drdrew42's approach

You are correct, the braces make the `my` variable local to that scope.  My apologies.

dpvc:
It may be possible to make @drdrew42's code work in strict mode (or temporarily turn off strict for the `$PG_VERSION` access, or something like that).  I didn't try to hard to try to make that work, as the non-strict code is minimal, and as I don't have a lot of time to put toward this.  I just thought it was more direct than having to have an extra PG macro to get the version.

drgrice1:
Yeah, you can add `no strict "vars"` before `$ENV{PG_VERSION} = $PG_VERSION || 'unknown'` to temporarily disable strict.  I don't know.  The strict stuff doesn't really matter.  Just quibbling!

drgrice1:
You could do
```perl
BEGIN {
	use File::Basename qw(dirname);
	my $dir = dirname(__FILE__);
	my $PG_VERSION = "unknown";
	my $PG_COPYRIGHT_YEARS = "";
	eval `cat "${dir}/../VERSION"`;
	warn "Error loading PG VERSION file: $!"    if $!;
	warn "Error processing PG VERSION file: $@" if $@;
	$ENV{PG_VERSION} = $PG_VERSION;
}
```
This works in strict mode, and doesn't cheat by even temporarily disabling strict.

drdrew42:
That works, and so does referencing the variable with the package name.
```perl
BEGIN {
	use File::Basename qw(dirname);
	my $dir = dirname(__FILE__);
	do "${dir}/../VERSION";
	warn "Error loading PG VERSION file: $!"    if $!;
	warn "Error processing PG VERSION file: $@" if $@;
	$ENV{PG_VERSION} = $PGcore::PG_VERSION || 'unknown';
}
```

drgrice1:
This is adding package variables then.  This adds the variables $PGcore::PG_VERSION and $PGcore::PG_COPYRIGHT_YEARS.  You can even access the values of those variables in a pg problem.

The `do` call is probably safer than the `eval` call.

drdrew42:
Yes, definitely safer -- I'll make a new PR later today, after my block of afternoon meetings, then close this PR.

dpvc:
I think this latest is the winning version.  I thought I had tried `$PGcore::PG_VERSION`, but I may not have had the rest of it right at that point.  Kudos.

> The `do` call is probably safer than the `eval` call.

Agreed, and it also doesn't require spawning a subprocess, or making a new `$externalPrograms` variable for `cat`.

drdrew42:
I had a meeting end early, and another got pushed *yay*
closed in favor of #531

PR #531, opened 2021-02-24
set pg version info in %ENV and PGcore obj+package

drdrew42:
replacement/improvement on #530 (@dpvc @drgrice1 thanks!)

with this PR, pg will now declare its version when PGcore.pm is loaded -- making the version available through:
* the ENV hash: `$ENV{PG_VERSION}` (available everywhere _outside_ of a problem's pg code)
* the PGcore object: `$PG->{PG_VERSION}` (in a problem's pg code)
* the PGcore package: `$PGcore::PG_VERSION` (also available in pg code)

### DIFF
--- a/lib/PGcore.pm
+++ b/lib/PGcore.pm
@@ -14,8 +14,17 @@
 # Artistic License for more details.
 ################################################################################
 package PGcore;
-
 use strict;
+
+BEGIN {
+    use File::Basename qw(dirname);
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
I think I've already rather thoroughly tested and reviewed this!

dpvc (approved):

PR #532, opened 2021-02-24
Update WeBWorK to use MathJax version 3.

drgrice1:
See  https://github.com/openwebwork/webwork2/pull/1215

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -76,39 +76,7 @@ sub DOCUMENT {
 	#use strict;
 	#FIXME
 	# load java script needed for displayModes
-	if ($envir{displayMode} eq 'HTML_MathJax') {
-		TEXT(
-			qq?<script type="text/javascript">
-				if (!window.MathJax) 
-				{
-					window.MathJax = {
-						tex: {
-							autoload: {
-								color: [],
-								colorV2: ['color']
-							},
-							packages: {'[+]': ['noerrors']}
-						},
-						loader: {
-							load: ['input/asciimath', '[tex]/noerrors']
-						},
-						startup: {
-							ready: function() {
-								var AM = MathJax.InputJax.AsciiMath.AM;
-								for (var i = 0; i < AM.symbols.length; i++) {
-									if (AM.symbols[i].input == '**') {
-										AM.symbols[i] = {
-											input: "**", tag: "msup", output: "^", tex: null, ttype: AM.TOKEN.INFIX
-										};
-									}
-								}
-								return MathJax.startup.defaultReady()
-							}
-						}
-					};
-				}
-				</script>?."\n");
-	} elsif ($envir{displayMode} eq 'HTML_jsMath') {
+	if ($envir{displayMode} eq 'HTML_jsMath') {
 		my $prefix = "";
 		if (!$envir{jsMath}{reportMissingFonts}) {
 			$prefix .= '<script>noFontMessage = 1</script>'."\n";
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -77,27 +77,38 @@ sub DOCUMENT {
 	#FIXME
 	# load java script needed for displayModes
 	if ($envir{displayMode} eq 'HTML_MathJax') {
-		##
-		# The following is used to prevent MathJax from being loaded more than one on a page. (Happened in WW3 library browser)
-		#
-		# If there are any errors replace the line starting "window.MathJax" with the string in $loadScript.
-		#
-	    TEXT(
-		 qq?<script type="text/x-mathjax-config">
-                  MathJax.Hub.Config({
-                     MathMenu: {showContext: true}
-                  });
-                  </script>
-				  <script type="text/javascript"> 
-				  if(!window.MathJax) 
-				  (function () {
-  					var script = document.createElement("script");
-  					script.type = "text/javascript";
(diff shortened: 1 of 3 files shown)


macros/PG.pl
```diff
@@ -77,27 +77,38 @@ sub DOCUMENT {
...
+									if (AM.symbols[i].input == '**') {
+										AM.symbols[i] = {
+											input: "**", tag: "msup", output: "^", tex: null, ttype: AM.TOKEN.INFIX
+										};
+									}
+								}
+								return MathJax.startup.defaultReady()
+							}
+						}
+					};
```
dpvc, line 108: Since this configuration gets used in two places (here and in `Instructor/ShowAnswers.pm`), I'm wondering if this could be put into a separate .js file that is loaded here and there rather than repeating the same code.  That way, the code in both places won't get our of sync.

dpvc (approved):

macros/PG.pl
drgrice1: That sounds good.

dpvc: If you do make it a separate file, you may want to load it in the template along with the other scripts, as it needs to come before the script that loads `tex-chtml.js` (otherwise setting `MathJax` here will overwrite the `MathJax` from that file).  So if you use `defer` on loading this file followed by `defer` on loading `tex-chtml.js` (with loading polyfil with `defer` before both), you should be OK.

drgrice1: I didn't add the `defer` attribute to the polyfill javascript.  So that should have it to then?

drgrice1: Would the `async` attribute be better for the polyfill.io script?  That shouldn't need to wait for the DOM content to be loaded.

taniwallach:
@drdrew42 - Note that the change in  https://github.com/openwebwork/pg/pull/532/commits/851410c9adfd80e7d1dca511448eeaccb92ae0f8 will certainly require similar changes in the standalone renderer.

macros/PG.pl
dpvc: I think that might allow it to run out of order.  For example, if polyfill took a long time to get to the page, the since it is allowed to run whenever it arrives, it might be that the `defer` items are run first.  I don't know for sure about that, but I do know that `defer` requires the order to be preserved, while `async` does not.

macros/parserGraphTool.pl
```diff
@@ -524,14 +524,14 @@ sub cmp_preprocess {
...
+		# starts another MathJax math mode that is ended by the atttempts table.  Both of the MathJax elements are
+		# removed by the injected javascript.
 		$ans->{preview_latex_string} = <<"END_ANS";
-</script>
+\\]
 <div id='${ans_name}_student_ans_graphbox' class='graphtool-answer-container'></div>
 <script>
-jQuery("#${ans_name}_student_ans_graphbox").parent().find('span').remove();
-jQuery("#${ans_name}_student_ans_graphbox").parent().find('script[type^="math/tex"]').remove();
+jQuery("#${ans_name}_student_ans_graphbox").parent().find('mjx-container').remove();
```
dpvc, line 534: I don't think you want to remove this container, as some of the the MathJax CSS is tied to this.  It is also what sets the size, adds the MathJax menu, and so on.  Can you tell me what the purpose is?  I think the original removals were for the preview span and the script that contained the original TeX code.  MathJax 3 doesn't include either of those.

drgrice1: The original removal was to remove the preview span and script, as you said, although it wouldn't contain any TeX code.  The result before the javascript acted would have been
```html
<span class="MathJax_Preview">[math]</span><script type="math/tex; mode=display"></script>
<div id='${ans_name}_correct_ans_graphbox' class='graphtool-answer-container'></div>
<script>
jQuery("#${ans_name}_correct_ans_graphbox").parent().find('span').remove();
jQuery("#${ans_name}_correct_ans_graphbox").parent().find('script[type^="math/tex"]').remove();
...
</script>
```
Then the javascript removed the span and type="math/tex" script.  The reason for this was that MathJax created an empty element that added space above the graph.  Furthermore, if you click on the graph it shows the popover twice.

Now the reason for the javascript is much the same.  The result with this pull request before the javascript acts is
```html
\[\]
<div id='${ans_name}_correct_ans_graphbox' class='graphtool-answer-container'></div>
<script>
jQuery("#${ans_name}_correct_ans_graphbox").parent().find('mjx-container').remove();
...
</script>
\[\]
```
With this, MathJax again creates empty elements where both of the `\[\]` occur.  The javacript "attempts" to remove those elements.  The reason is those elements add space above and below the graph, as well as the double popover issue.

The point here is that there is no TeX.  The answers are the javascript graphs.  Yes, my original code and this replacement code are hacks.

Now, when I tested this originally this was working, and the mjx-container elements were removed, but now when I test this those elements are not being removed.  There is probably a race condition occurring between when the MathJax typesetting javascript is run and when this jQuery ready function injected here is run.  So I will need to do something different.  I have a fix that I will commit shortly.  I am going to add a flag to the answer hash that is checked in the attempts table that tells it this is not a TeX preview, so don't add the TeX delimiters.

macros/PG.pl
drgrice1: Okay.  I will leave it on `defer` then.

PR #537, opened 2021-02-26
Follow alias chain, not just first alias.  (#518)

dpvc:
[Alex's issue](https://github.com/openwebwork/pg/pull/518#issuecomment-785426797) turns out to be due to the `resolve()` method only following one alias, instead of following an alias chain.  This PR causes `resolve()` to follow aliases until the end.  MathObjects try to prevent there from being an alias loop (one alias pointing to an alias that points back to the first), but it is possible to make them if you try hard, so this could potentially loop infinitely if a page author worked hard to make that true.  One could introduce a counter and a maximum depth for alias links to avoid that.

### DIFF
--- a/lib/Parser/String.pm
+++ b/lib/Parser/String.pm
@@ -18,7 +18,7 @@ sub new {
   my ($value, $ref) = @_;
   my $def = $strings->{$value};
   my $VALUE = uc($value); my $DEF = $strings->{$VALUE};
-  ($value, $def) = ($VALUE, $DEF)  if !$def && $DEF && (!$DEF->{caseSensitive} || $value eq $VALUE);
+  ($value, $def) = ($VALUE, $DEF) if !$def && $DEF && !$DEF->{caseSensitive};
   ($value, $def) = $equation->{context}->strings->resolve($value);
   my $str = bless {
     value => $value, type => $Value::Type{string}, isConstant => 1,
--- a/lib/Parser/String.pm
+++ b/lib/Parser/String.pm
@@ -17,12 +17,9 @@ sub new {
   my $equation = shift; my $strings = $equation->{context}{strings};
   my ($value, $ref) = @_;
   my $def = $strings->{$value};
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I reported the issue with a particular PG file from the OPL, but as best as I can tell I copied and pasted the wrong problem file. That problem doesn't have the line with `Compute("-INF")`. I'm sorry. I can't see anything wrong with that problem file, even without this new commit.

This file: `Library/UCSB/Stewart5_2_2/Stewart5_2_2_9/Stewart5_2_2_9.pg` causes the issue. And I'm still getting the issue even with this commit.

Again, I'm sorry for the false lead.

lib/Parser/String.pm
```diff
@@ -17,11 +17,12 @@ sub new {
   my $equation = shift; my $strings = $equation->{context}{strings};
   my ($value, $ref) = @_;
   my $def = $strings->{$value};
+  my $VALUE = uc($value);
```
drgrice1, line 20: Cool, so `uc` not only changes the contents of the variable to upper case, it also changes the variables name to upper case!  I like your naming scheme.

dpvc: :-)

This commit still has one problem that I am trying to fix.  But have had to stop for the moment.  I'll let you know when it is correct.

dpvc:
OK, I've resolved the issue that was coming from that problem.  It had to do with how the situation for case-insenesitive strings was being handled now that aliases are allowed.  Please review again.

drgrice1 (approved):
This looks good to me.  Testing with the problem that @Alex-Jordan posted before this pull request, I see the error that he sees.  The problem works as expected with this pull request.

I also tested this with several other problems that use `-infinity` and didn't see any problems.

drgrice1:
@Alex-Jordan: Please take a look at this.  Go ahead and merge it if you approve.

Alex-Jordan:
Thanks @dpvc , works with my testing.

PR #538, opened 2021-03-01
Do a typecheck before comparing functions.  (bugzilla bug 4559)

dpvc:
This PR adjusts the function-equivalence check to call the correct-answer's `typeMatch()` method, which checks that the types of the return values are correct (the general comparison is less strict than that, as it allows promotion of types).

Resolve bugzilla [bug 4559](https://bugs.webwork.maa.org/show_bug.cgi?id=4559), where the bug is described, and a test-case is given.


Alex-Jordan (approved):

drgrice1 (approved):
This looks good.

With the problem attached in the bug I see the behavior described in the bug without this pull request.  That is if you first enter (x,x) it is marked incorrect, and then if you enter <x,x> it is marked correct but you get the message that this answer is equivalent to the previous answer.

With this pull request, you don't get the misleading equivalent answer message, and grading is still the same.

Issue #535, opened 2021-02-25
Context("Point") accepts [a,b] and {a,b} as correct.


somiaj:
When using Context("Point"), all answers (a,b), [a,b], and {a,b} are accepted as correct. In some cases (working with sets vs order pairs in my case), I found out students were entering in {a,b} as elements of a Cartesian product and being marked correct. I have written a custom grader that rejects answers that have [] or {} in their string, but think there should at least be an option (maybe there is but I didn't see it) to reject points written with [] or {}.

Here is an example problem, with my custom grader included (unsure if there is a better way to achieve what I did, and I had trouble writing a single regex line that worked, that is why I split it into multiple conditions).

```
DOCUMENT();      

loadMacros(
   "PGstandard.pl",
   "MathObjects.pl",
   "PGcourse.pl"
);

TEXT(beginproblem());

Context("Point");
$ans1 = Compute("(5,7)");
$ans2 = Compute("(3,-2)");
$cmp2 = sub {
  my ( $c, $s, $ansHash ) = @_;

  $sstr = $s->string;
  if ( ($sstr =~ /[\[]/)||($sstr =~ /[\]]/)||($sstr =~ /{/)||($sstr =~/}/) )
  { return 0; }
  if ( $c == $s ) { return 1; }
  else { return 0; }

};

Context()->texStrings;
BEGIN_TEXT

Enter in the point \( (5,7) \). Answer will accept
\( \lbrace 5,7 \rbrace \) and \( [5,7] \).
$BR
\{ ans_rule(5) \}

$PAR
Enter in the point \( (3,-2) \). Answer will not accept
\( \lbrace 3,-2 \rbrace \) or \( [3,-2] \).
$BR
\{ ans_rule(5) \}

END_TEXT
Context()->normalStrings;

ANS($ans1->cmp);
ANS($ans2->cmp(checker => $cmp2));

ENDDOCUMENT();
```

Alex-Jordan:
Hello @somiaj  This is a user support question, which will get good responses if you post it in the WeBWorK forums:
https://webwork.maa.org/moodle/mod/forum/index.php?id=3

You may need to sign up for that if you don't already have an account. 

Since I have a quick answer for you, I'll mention it here. You can change the way the parser views delimiters following this:
https://webwork.maa.org/wiki/Modifying_Contexts_(advanced)#Lists_and_Delimiters

drgrice1:
[...]

PR #539, opened 2021-03-01
Adjust the default class for [...] and {...}, and do type-matching in list comparisons.  (#535)

dpvc:
This PR addresses the issues raised in #535.

1.  The default class for `[...]` and `{...}` when delimiting a list of numbers used to be `Point`, which causes the Point context to allow `{1,2}` to match `(1,2)` unexpectedly.  This PR changes `[...]` to produce a List instead and `{...}` to produce sets.  These two are still removable, however, so `[1]` is still just `1` and `{2}` is just `2`.  This is because these brackets and braces are supposed to be allowed as alternative parentheses (to make it easier to match them in a complex expression), and because MathObject strings use both `(...)` and `[...]` (again, to make matching them easier to see).  So while `{1,2}` will produce a set, `{1}` will not, unless `removable` is set to `0`, as it is in the Interval context.

2.  The default list checker did not call `typeMatch()` before comparing a correct entry to a student entry, and so `(1,2)` would match `{1,2}` as points.  This PR adds the `typeMatch()` call so that the comparison is made only between objects that should actually be compared.  Some people may have taken advantage of this deficiency, so some problems may break due to this change.  I don't know of any, but after 15 years of working this way, no doubt some problems will rely on this mis-feature.

This probably wants extensive testing before merging.

Resolves issue #535.

### DIFF
--- a/lib/Parser/Context/Default.pm
+++ b/lib/Parser/Context/Default.pm
@@ -110,11 +110,11 @@ $operators = {
 $parens = {
    '(' => {close => ')', type => 'Point', formMatrix => 1, formInterval => ']',
            formList => 1, removable => 1, emptyOK => 1, function => 1},
-   '[' => {close => ']', type => 'Point', formMatrix => 1, formInterval => ')', removable => 1},
+   '[' => {close => ']', type => 'List', formMatrix => 1, formInterval => ')', removable => 1},
    '<' => {close => '>', type => 'Vector',
            alternatives => ["\x{2329}","\x{3008}","\x{27E8}"],
            alternativeClose => ["\x{232A}","\x{3009}","\x{27E9}"]},
-   '{' => {close => '}', type => 'Point', removable => 1},
+   '{' => {close => '}', type => 'Set', removable => 1},
    '|' => {close => '|', type => 'AbsoluteValue'},
    'start' => {close => 'start', type => 'List', formList => 1,
                removable => 1, emptyOK => 1, hidden => 1},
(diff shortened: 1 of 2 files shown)


drgrice1 (approved):
This looks good.

I tested my problems that give partial credit for the incorrect delimiters, and the graders still work with this change.  This is probably due to the fact that custom list checkers are used, and in some cases the answers are also in MultiAnswer objects and checkTypes => 0 is used, and in some other cases the grading is done in a post filter anyway.

drgrice1:
I should add that I also tested this with basic problems that just use the Point context with the answers being a single point or with a list of points, and the grading works as is expected (as per issue #535).  That is points delimited by brackets and braces are graded as incorrect, and points delimited by parentheses are graded correct.

Alex-Jordan:
I have at least one context (not yet contributed) that may be affected by this change. The context is downstream from Full, and has a complicated custom checker that examines the type of MathObject that was submitted, and does things depending on which one. If it's a "Point", do this. If it's a "Formula", do that. Etc. It doesn't have any contingency for "Set" since that wasn't a possible MathObject to create. 

I can easily address it by adding how to treat a "Set" answer. And I'm not objecting to the change here. But there could be consequences to this inside people's problem or macros code that we may want to prepare for in the forums and in release notes.

Do we want this in time for 2.16?

drgrice1:
I think this would be good for 2.16, but we probably want to get a consensus that this is how we want this to work.  I am for it.

Issue #536, opened 2021-02-26
Typo in options


kmill:
Reading through the source code, I noticed that "associativity" was written as "asscoiativity" here:

https://github.com/openwebwork/pg/blob/a284c3b31a26fb316c2fa6b34246df8ba2895a3e/macros/contextPiecewiseFunction.pl#L118

This seems to be a typo given the rest of the file.  (I'm not using WeBWorK right now and I'm not depending on this being fixed -- I just thought I'd let someone know.)

dpvc:
Thanks for pointing that out!  I've made a pull request to fix it.

PR #540, opened 2021-03-01
Correct typo in context property.  (#536)

dpvc:
This PR corrects a typo in the contextPiecewiseFunctions.pl file.  It would mean that the associativity of `in` is left rather than right, but since it is unlikely anyone would put two of these back to back anyway, it never really hurt anything.

Resolves issue #536.


drgrice1 (approved):
I will go ahead and merge this trivial fix.

PR #541, opened 2021-03-03
Add options to add TeX libraries and add to the TeX preamble in the PGtikz.pl macro

drgrice1:
This makes that method more versatile.  For example, with this you can use the pgfplots package.

This was also something that was requested when the PGtikz.pl macro was originally introduced (see https://github.com/openwebwork/pg/pull/432#issuecomment-638989408).

You can test this with the attached pg file.
[tikz-package-test.zip](https://github.com/openwebwork/pg/files/6077223/tikz-package-test.zip)

### DIFF
--- a/lib/TikZImage.pm
+++ b/lib/TikZImage.pm
@@ -33,6 +33,8 @@ sub new {
 		tex           => '',
 		tikzOptions   => '',
 		tikzLibraries => '',
+		texPackages   => {},
+		addToPreamble => '',
 		ext           => 'png',
         imageName     => ''
 	};
@@ -76,6 +78,19 @@ sub tikzLibraries {
 	return &$self('tikzLibraries', @_);
 }
 
+# Set additional TeX packages to load.  This accepts a single hash parameter.
+sub texPackages {
+	my $self = shift;
+	return &$self('texPackages', $_[0]) if ref($_[0]) eq "HASH";
+	return &$self('texPackages');
+}
+
+# Additional TeX commands to add to the TeX preamble
+sub addToPreamble {
+	my $self = shift;
+	return &$self('addToPreamble', @_);
+}
+
 # Set the image type.  The valid types are 'png', 'gif', 'svg', and 'pdf'.
 # The 'pdf' option should be set for print.
 sub ext {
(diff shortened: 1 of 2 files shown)


taniwallach:
Before patch:
```
Warning -- There may be something wrong with this question. Please inform your instructor including the warning messages below.

Error messages

Can't locate object method "texPackages" via package "TikZImage" at line 22 of (eval 9765) 
```

---

After patch - no error from PG, but the image did not render for me.
In `/var/log/apache2/error.log` (of Docker) I found:
```
Can't exec "/usr/bin/convert": No such file or directory at /opt/webwork/pg/lib/TikZImage.pm line 157.
Convert operation failed. at /opt/webwork/pg/lib/TikZImage.pm line 168.
apache2Can't print graph to /opt/webwork/webwork2/htdocs/tmp/Madar_devel/images/c59732e3-9ca4-3993-ac63-3f05bc82fb29___c615607d-28ab-306a-97bb-ee032fb14d02.png<BR> at /opt/webwork/pg/lib/PGcore.pm line 719.
```
so I installed: `apt-get update; apt-get install imagemagick` in the running Docker container.

Still not working...  but then I saw https://github.com/openwebwork/pg/pull/432#issuecomment-636221535 and
edited `/etc/ImageMagick-6/policy.xml` and to change the line
```
  <policy domain="coder" rights="none" pattern="PDF" />
```
to
```
  <policy domain="coder" rights="read" pattern="PDF" />
```
and deleted the empty temporary file. Then a problem reload generated the image.

---

We need to make instructions about editing `/etc/ImageMagick-6/policy.xml` and the potential security implications make it into the release notes and the installation instructions.

taniwallach (approved):
The additional features make sense and seem very useful.
Testing with the sample file was successful, once the prerequisite packages were installed and configuration settings made.

drgrice1:
The attached problem will not work without this pull request.

The issue that you are seeing with `imagemagick`'s `convert` program is documented in `site.conf.dist`.  Although we will need to add that to release documentation as well as you say.  This is true for the PGtikz.pl macro without this pull request.  This was added by #432 and https://github.com/openwebwork/webwork2/pull/1030.

pstaabp (approved):
I too needed to configure all the external programs in site.conf, but afterward, worked great. 

I'll definitely use this as I am a big tikz user.

PR #542, opened 2021-03-04
PreTeXt: remove sidebyside

Alex-Jordan:
This is catching up with PreTeXt developments. Formerly, PTX did not allow `tabular`, `image` or `video` elements without certain wrappers like `figure`, `table`, and `sidebyside`. We never allowed the first two as WW output, and used `sidebyside` as the wrapper in all cases. Now this practice is not needed. PTX allows "naked" `tabular`, `image` and `video` elements. Clearing away all the `sidebyside` stuff simplifies this WW code, and also prevents PTX from issuing warnings that it detects a pointless `sidebyside`.

I'd like to include this in 2.16. I suspect no one else who is active would want to test PreTeXt stuff. If you would like to review this, I suggest just eyeing the code edits to see that I didn't leave a `<` somewhere where it should have been cut, and typos like that.

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2323,29 +2323,23 @@ sub PTX_cleanup {
 	# Wrap <p> tags where necessary, and other cleanup
 	# Nothing else should be creating p tags, so assume all p tags created here
 	# The only supported top-level elements within a statement, hint, or solution in a problem
-	# are p, blockquote, pre, sidebyside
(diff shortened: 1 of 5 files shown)


drgrice1 (commented):
Since these changes only change the PTX display format, I can approve this without extensive testing.  I tested to make sure that problems using these macros still work.  Mainly just to ensure there aren't any syntax errors introduced, as that is the only thing that would cause problems for other display formats.

drgrice1 (approved):

PR #543, opened 2021-03-04
Unexciting code cleanup stemming from memory leak effort

drdrew42:
This is mainly uncontroversial code-cleanup that happened during the process of chasing down leaks.

**Overview:**
- PGresponsegroup does not use `Tie::IxHash`
- fix a lingering typo in Translator from the update to "new_" `rf_fun` and `temp_ans`
- stop declaring `$answergroup` as a global in PG.pl (line 465)
- remove detritus from Translator 
- similar cleanup to PG.pl 

The change in PG.pl is the only substantive edit in this PR.

Globally scoping the PGanswergroup objects prevented its children (PGresponsegroup, AnswerHash) from being GC'ed. I imagine that this wasn't done purposefully, but because PG.pl is `reval`ed into the safe without any restrictions (no opcode masking, no strict) this oversight was easily missed.

These changes are not dependent on the leak cleanup PR openwebwork/webwork2#1226 in the ww2 codebase, but they should be implemented together for effective memory-leak prevention.

### DIFF
--- a/lib/PGresponsegroup.pm
+++ b/lib/PGresponsegroup.pm
@@ -19,7 +19,7 @@ use strict;
 use Exporter;
 use PGUtil  qw(not_null) ;
 use PGanswergroup;
-use Tie::IxHash;
+# use Tie::IxHash;
 
 #############################################
 # An object which contains the student response(s)
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -51,9 +51,9 @@ sub DOCUMENT {
 	$solutionExists        		= $PG->{flags}->{solutionExists};
 	$hintExists            		= $PG->{flags}->{hintExists};
 	$pgComment                  = '';
-	%external_refs         		= %{ $PG->{external_refs}};
+	# %external_refs         		= %{ $PG->{external_refs}};
 	
-	@KEPT_EXTRA_ANSWERS =();   #temporary hack
+	$PG->{flags}{KEPT_EXTRA_ANSWERS} = [];
 	
 	my %envir              =   %$rh_envir;
 	# Save the file name for use in error messages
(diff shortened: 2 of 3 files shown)


lib/PGresponsegroup.pm
```diff
@@ -19,7 +19,7 @@ use strict;
 use Exporter;
 use PGUtil  qw(not_null) ;
 use PGanswergroup;
-use Tie::IxHash;
+# use Tie::IxHash;
```
Alex-Jordan, line 22: If `Tie::IxHash` is not needed, would it be better to remove this line than comment it out?

macros/PG.pl
```diff
@@ -51,9 +51,9 @@ sub DOCUMENT {
 	$solutionExists        		= $PG->{flags}->{solutionExists};
 	$hintExists            		= $PG->{flags}->{hintExists};
 	$pgComment                  = '';
-	%external_refs         		= %{ $PG->{external_refs}};
+	# %external_refs         		= %{ $PG->{external_refs}};
```
Alex-Jordan, line 54: And here as well. Perhaps we should have a policy about keeping old code in comments. Maybe the policy would be never do it. Or maybe, you can do it but you need to add commentary about that code. Or something.

drgrice1: I agree, except possibly in rare cases. Delete it!

drdrew42:
deletions are definitely coming - I figured commenting out for draft purpose was appropriate. If all is successful, then 💯  deleted!

drgrice1:
Yeah, this is a work in progress, so that makes sense.

drgrice1:
It is probably not ready for reviews yet.

drdrew42:
This PR is out here specifically for testing in combination with openwebwork/webwork2#1226
The release of certain objects from memory will not happen without the changes contained herein. Specifically, AttemptsTable and PGanswergroup objects.

All that being said, nothing in the changes here should break anything 🤞  (but again, we don't know without testing...)

PR #544, opened 2021-03-08
Make PGML work for hardcopy if used in a set header file

drgrice1:
Move the PGML `preamble` from the `problemPreamble` environment variable into the TeX `hardcopyPreamble.tex` template files and restructure things a bit so that PGML will work in a header file.

To make this work the following things needed to be done:

The PGML detect previous newline hack had to be removed.  Instead nicer spacing was implemented that applies to both PGML and PG problems.

The catcode changes for PGML problems that ignores carriage returns needed to be scoped (and was moved into the pgmlSetup definition). This was actually a problem in general when PG and PGML problems were used in the same homework set.  Any PG problem after the first PGML problem ended up not being displayed correctly as a result.

Now PGML no longer uses the problemPreamble environment variable (in fact nothing does).

One (perhaps controversial) change is that `parindent=0pt` is set globally.  Previously this was set only for PGML problems.  The reason for this is that even PG problems need this.  This makes the hardcopy output for PG problems more like the output that is shown on screen when the problem is worked online.

Note that this is paired with the webwork2 pull request https://github.com/openwebwork/webwork2/pull/1239 where the `hardcopyPreamble.tex` files are changed.

This fixes issue https://github.com/openwebwork/webwork2/issues/1066.

### DIFF
--- a/macros/PGML.pl
+++ b/macros/PGML.pl
@@ -1488,7 +1488,7 @@ sub Format {
   my $parser = PGML::Parse->new(shift);
   my $format;
   if ($main::displayMode eq 'TeX') {
-    $format = "{\\pgmlSetup\n".PGML::Format::tex->new($parser)->format."\\par}%\n";
+    $format = "\n{\\pgmlSetup\n" . PGML::Format::tex->new($parser)->format . "\\par}%\n";
   } elsif ($main::displayMode eq 'PTX') {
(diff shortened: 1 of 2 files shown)


taniwallach:
@Alex-Jordan @drdrew42 @pstaabp - This address something we had intended to defer, but if it works, we can merge it and close the issue. Are we interested in targeting the WW 2.16 release?

Issue #455, opened 2020-02-27
bug with MathObjects reduction


Alex-Jordan:
We've hit a bug with a reduction rule in MathObjects. Something like `-5/(-2x)` is incorrectly reducing to `5/2*x`. Here is a MWE.

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
  "PGML.pl",
);

TEXT(beginproblem());

$answer = Formula("-5/(-2 x)");
$reduced = $answer->reduce;

BEGIN_PGML

The answer is [$answer], which reduces to [$reduced].

END_PGML

ENDDOCUMENT();
```

(I'm having trouble logging in to the forum while the transition is underway, or else I might have posted this there.)

drgrice1:
You need to do a password reset on the wiki.  After you do that you will be able to sign in to the wiki and the Moodle forums.  Go to https://webwork.maa.org/wiki/Special:PasswordReset to do this.

mgage:
Glenn,
Do you know if the emails for password reset are being sent out by the wiki?

drgrice1:
I am not sure.  I was hoping so.  I will look into it.

drgrice1:
It seems that the password reset emails are not being sent (or at least received).

dlglin:
When I request a password reset email it is sent and caught in our spam filter, likely because it came from apache@🌻.invalid (see https://github.com/openwebwork/webwork.maa.org/issues/29).

Alex-Jordan:
This is still a bug (just tested on develop). The wiki login discussion may have distracted from the issue I reported.

dpvc:
I have made a PR (#546) that addresses this issue.  Sorry for the long wait for it.  It has scrolled off the top of my email screen, where I don;'t often look, and so was in limbo until I cleared out some more recent items.

PR #546, opened 2021-03-14
Use parens again if negative is factored out.  (#455)

dpvc:
This PR address issue #455, which causes the display for some reduced formulas be incorrect (parens are missing).  The expression evaluates properly, but doesn't print properly.  This is due to a setting that was trying to prevent `(-2)x` from showing as `-(2*x)` when it is reduced.  

The reduction works by pulling negatives out and cancelling when possible, so `-5/(-2*x)` becomes `-(5/(2*x))`, then `-(5/(-(2*x))), then `-(-(5/(2*x))` and finally `5/(2*x)`.  But at the `(-2*x)` to `-(2*x)` reduction, the parens are suppressed (so it would print as `-2*x` , and the further reduction didn't change that, so you got `5/2*x` instead of `5/(2*x)`.  This PR removes the no-paren marker when the negation is moved further out, allowing the parens to be included again.

### DIFF
--- a/lib/Parser/BOP.pm
+++ b/lib/Parser/BOP.pm
@@ -253,6 +253,8 @@ sub makeZero {
 sub makeNeg {
   my $self = shift;
   $self->{lop} = shift; $self->{rop} = shift;
+  delete $self->{lop}{noParens};
+  delete $self->{rop}{noParens};
   return Parser::UOP::Neg($self);
 }
 


Alex-Jordan (approved):
I tested with the original problem and a few more expressions I thought might be affected. All I can see is that it fixes the original issue without any side effects. So I recommend for merging.

drgrice1 (approved):
This checks out for me too.

PR #547, opened 2021-03-17
Remove a prefix from the answer rule insertion of  parserGraphTool.pl

drgrice1:
This was from my original implemenation, and was not supposed to make it here.
The setNumber and probNum may not be defined, and in that case this fails.

### DIFF
--- a/macros/parserGraphTool.pl
+++ b/macros/parserGraphTool.pl
@@ -496,9 +496,8 @@ END_TIKZ
 	} else {
 		$self->constructJSXGraphOptions;
 		my $ans_name = $self->ANS_NAME;
-		my $prefix = ($main::setNumber =~ tr/./_/r) . "_" . $main::probNum;
-		$out .= "<div id='${prefix}_${ans_name}_graphbox' class='graphtool-container'></div>" .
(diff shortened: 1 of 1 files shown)


pstaabp:
This looks minor.  It doesn't appear that the id tag is ever used,  is that right?  If not, it doesn't really matter what it is named (as long as it is unique).

drgrice1:
The id tag is used by the graphtool.js file.  That file is loaded by the macro and the method called on line 500 here with the id as the argument.  The file is located in the webwork2 code (in htdocs/js/apps/GraphTool).  That is the problem here.  If setNumber and probNum are undefined, then `$prefix = "not defined"`.  The space in that string doesn't work for an id.  The ANS_NAME should be unique without that.

Alex-Jordan:
Tested and it didn't break a problem set. I also tried a problem in an undefined set, before and after, and I find no difference. I'm not sure how to trigger the `not defined` issue before the PR. But I'm just going to merge this anyway.

drgrice1:
To trigger the "not defined" thing load a GraphTool problem via webwork2/html2xml

PR #549, opened 2021-03-18
Update the AppletObjects.pl macro for changes in the ADD_JS_FILE usage

drgrice1:
This adds the `js` prefix to the location of the javascript files added by AppletObjects.pl via ADD_JS_FILE.

This should be merged if and only if https://github.com/openwebwork/webwork2/pull/1258 is merged.

### DIFF
--- a/macros/AppletObjects.pl
+++ b/macros/AppletObjects.pl
@@ -28,8 +28,8 @@ See also L<http://webwork.maa.org/pod/pg_TRUNK/lib/Applet.html>.
 
 # Add basic functionality to the header of the question
 sub _AppletObjects_init{
-	ADD_JS_FILE("apps/Base64/Base64.js", 1);
-	ADD_JS_FILE("apps/AppletSupport/ww_applet_support.js", 1);
+	ADD_JS_FILE("js/apps/Base64/Base64.js", 1);
+	ADD_JS_FILE("js/apps/AppletSupport/ww_applet_support.js", 1);
 };
 
 =head3  FlashApplet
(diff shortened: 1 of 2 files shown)


drgrice1:
@taniwallach: I don't really think this one needs that much testing.  It is a pretty straightforward change.

taniwallach (approved):
With this PR and https://github.com/openwebwork/webwork2/pull/1258 active - the 3 files are all loaded properly when I tested `Library/CSUOhio/appletDemonstrationProblems/DerivativeDraw11.pg` a random OPL problem using `AppletObjects.pl`. That problem has a flash applet - so the applet itself does not work, but that was not what needed to be tested.

taniwallach:
Ready to merge once https://github.com/openwebwork/webwork2/pull/1258 is merged.

taniwallach:
I'm recommending adding minor changes to the documentation in `PG.pl` via https://github.com/drgrice1/pg/pull/1 to address what @pstaabp reported in https://github.com/openwebwork/webwork2/pull/1258#issuecomment-803570690 .

taniwallach (changes requested):
Recommend the patch to the `PG.pl` documentation (in the comments regarding `ADD_CSS_FILE` and `ADD_JS_FILE`)  be added, as they also relate to the changes made in https://github.com/openwebwork/webwork2/pull/1258 and I don't see any reason to make a separate PR for that.

taniwallach (approved):
> In my last commit I reversed the polarity of that second argument. So now if the argument is 0 (or not given) then it is assumed to be in the htdocs directory. Furthermore, I added that argument to the ADD_CSS_FILE method. This maintains the usage of the ADD_CSS_FILE method from before, and makes the ADD_JS_FILE usage consistent with that as you were expecting.

I just tested quickly with the revised version of this and the changes to webwork2 made in https://github.com/openwebwork/webwork2/pull/1258 (the quote above is taken from the discussion there). 

I think the revised default to assume additions come from `htdocs` is a good decision for both functions in PG. Thanks @drgrice1 and thanks for making the POD documentation.

I think this pair of PRs are ready to merge.

PR #552, opened 2021-03-23
stop random_coprime from producing error messages in apache2 log

Alex-Jordan:


Alex-Jordan:
@drgrice1 Does this kill those error messages?

If you ask for `random_coprime([2,4], [6,7], [2,5])`, then that will call `random_coprime([ [],  [[2],[4]]  ], [6,7], [2,5])`.  The empty `[]` represents 1-tuples with gcf 1. The `[[2],[4]]` is 1-tuples with gcf > 1.


Now the first argument is an array reference of array references, and the recursion begins. The next call will be `random_coprime([ [[2,7],[4,7]],  [[2,6],[4,6]] ], [2,5])`.

And then `random_coprime([ [[2,7,2],[4,7,2],[2,7,5],[4,7,5],[2,6,5],[4,6,5]],  [[2,6,2],[4,6,2]] ])`.

But you can see how in that first step (or later steps in other examples) the first argument could be empty and yet the recursion should continue. That's why it needs to look at `@{$newcandidates[0]} || @{$newcandidates[1]}`, not just `@{$newcandidates[0]}`.

drgrice1:
I thought it should be `||` but it was effectively what my other pull request was.

This takes care of the error messages.

PR #553, opened 2021-03-23
account for PTX output from GraphTool

Alex-Jordan:
GraphTool was assuming output is either TeX or some flavor of HTML. This accounts for PTX, in which case it outputs nothing, which is fine for now. Without this edit, there is an assumption of some flavor of HTML output, and HTML comes out mixed in with PreTeXt XML.

We really need some method that says "if the output is HTML" without directly using `$displayMode` in code like this. Something that captures the `$displayMode` variants like `MathJax`, `images`, and whatever else remains. But this small thing here is fine for now.

### DIFF
--- a/macros/parserGraphTool.pl
+++ b/macros/parserGraphTool.pl
@@ -176,7 +176,7 @@ This is the size of the graph that will be output when a hard copy of the proble
 =cut
 
 sub _parserGraphTool_init {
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Yeah, I am never quite sure what to do about PTX output.  I guess it should have been obvious not to output HTML stuff though.  I also tend not to think about it as I don't use it.  I count on you to fix that!

PR #555, opened 2021-03-25
Change mathQuillOpts a bit and disable MathQuill for parserWordCompletion.pl

drgrice1:
The stringify_answers method of Translator.pm now JSON encodes the mathQuillOpts key of the answer hash if defined and a hash reference.  This allows the value of that key to be a hash.  This is easier to work with than manually creating JSON strings.  Furthermore, it can be directly passed to the mqedit.js javascript as a data attribute with an object value.

So this changes the way the mathQuillOpts answer hash key works.

Previously to disable the spaceBehavesLikeTabs option (i.e. to enable typing of spaces) and the disable rootsAreExponents you would need to do
      `$ans->cmp(mathQuillOpts => "spaceBehavesLikeTab: false, rootsAreExponents: false")`
Now you will do
      `$ans->cmp(mathQuillOpts => { spaceBehavesLikeTab => 0, rootsAreExponents => 0 })`

This changes the usage from before, but I doubt many were using these options yet.
    
Note that you can also use `$ans->cmp(mathQuillOpts => "disabled")` to disable MathQuill answers for an answer blank.  (This is done in the webwork2 code).
    
This also automatically disables MathQuill answer boxes when the answer is a parserWordCompletion.pl answer.

This is paired with the webwork2 pull request https://github.com/openwebwork/webwork2/pull/1288.

### DIFF
--- a/lib/AnswerHash.pm
+++ b/lib/AnswerHash.pm
@@ -122,6 +122,7 @@ BEGIN {
 package AnswerHash;
 use Exporter;
 use PGUtil qw(not_null pretty_print);
+use JSON;
 
 # initialization fields
 my %fields = (		'score'					=>	undef,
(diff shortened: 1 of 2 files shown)


Alex-Jordan (approved):
Except for my small comment about `disabled` in https://github.com/openwebwork/webwork2/pull/1288, this works as described.

You got it to automatically disable for parserWordCompletion,pl answers. Do you have any thoughts on doing the same for parserNumberWithUnits.pl answers? That is the main reason local faculty are not using MathQuill. It turns answers like `2 m^2` into `2m^(2)` and `2 m/s` into `(2 m)/s` which are not understood by the parser. I tell faculty they could use text mode, but it's too much effort to train students about this issue.

drgrice1:
Sorry about the mistakes on the option names in the comments.  I fixed that above.  I don't care if the option is `disabled` or `disable`.  That is in the webwork code, and not in the PG code, but that matters little.

Disabling this for the parserWordCompletion.pl macro is easy.  Doing so for the parserNumberWithUnits.pl is not.  Furthermore, in my opinion disabling this for that macro is not the correct thing to do.  We discussed this (either last week or the week before) in our weekly meeting, and agree that there needs to be a better way of dealing with that.  It is more likely that the parser or the macro need to be modified to make answers with units work better.  So the correct thing to do for now is to use this `disabled` option for these problems until the better solution is figured out.

PR #556, opened 2021-04-01
Fix POD generation errors.

drgrice1:


taniwallach (approved):
I checked on https://webwork.maa.org/pod/pg
and see the documentation which was once just comments in the file on the site, and checked places where I see links were fixed and they work on the site.
I did not look into the other formatting related changes to the POD code.

pstaabp (approved):

PR #557, opened 2021-04-01
Implement a better method for generating svg images with TikZImage.pm.

drgrice1:
The current implementation generates a pdf using pdflatex, and then converts the pdf to svg using imagemagick's convert utility.

This changes it so that it uses latex to generate a dvi with the pgfsysdriver redefined to pgfsys-dvisvgm.def, and then converts the dvi to svg using dvisvgm.  This is the best way that I have found of creating svg images from tikz/tex.

This does require another external program (dvisvgm).  That is added in a corresponding pull request to webwork2.

I have attached a couple of examples to compare the png image generation to the svg image generation.
[examples.zip](https://github.com/openwebwork/pg/files/6254945/examples.zip)

To test this, make sure to set the dvisvgm external command in site.conf.  See https://github.com/openwebwork/webwork2/pull/1308.

### DIFF
--- a/lib/TikZImage.pm
+++ b/lib/TikZImage.pm
@@ -109,6 +109,7 @@ sub header {
 	my @output = ();
 	push(@output, "\\documentclass{standalone}\n");
 	push(@output, "\\usepackage[svgnames]{xcolor}\n");
+	push(@output, "\\def\\pgfsysdriver{pgfsys-dvisvgm.def}\n") if $self->ext eq 'svg';
 	push(@output, "\\usepackage{tikz}\n");
 	push(@output, map {
 			"\\usepackage" . ($self->texPackages->{$_} ne "" ? "[$self->texPackages->{$_}]" : "") . "{$_}\n"
@@ -145,17 +146,21 @@ sub draw {
 	print $fh $self->footer;
 	close $fh;
 
-	my $pdflatex_command = "cd " . $working_dir . " && " .
-		WeBWorK::PG::IO::pdflatexCommand() . " > pdflatex.stdout 2> pdflatex.stderr image.tex";
+	my $ext = $self->ext;
+	my $latex_binary = WeBWorK::PG::IO::externalCommand($ext eq 'svg' ? 'latex' : 'pdflatex');
 
 	# Generate the pdf file.
-	system "$pdflatex_command";
-	if (-r "$working_dir/image.pdf" ) {
-		my $ext = $self->ext;
+	system "cd " . $working_dir . " && $latex_binary > pdflatex.stdout 2> /dev/null image.tex";
 
+	if (-r "$working_dir/image." . ($ext eq 'svg' ? 'dvi' : 'pdf') ) {
 		# Convert the file to the appropriate type of image file
-		system WeBWorK::PG::IO::convertCommand() . " $working_dir/image.pdf $working_dir/image.$ext"
-		if $ext ne 'pdf';
+		if ($ext eq 'svg') {
+			system WeBWorK::PG::IO::externalCommand('dvisvgm') .
+				" $working_dir/image.dvi 2>&1 /dev/null --no-fonts --output=$working_dir/image.svg"
+		} elsif ($ext ne 'pdf') {
+			system WeBWorK::PG::IO::externalCommand('convert') .
+				" 2>&1 /dev/null $working_dir/image.pdf $working_dir/image.$ext"
+		}
 
 		if (-r "$working_dir/image.$ext") {
 			# Read the generated image file into memory
--- a/lib/TikZImage.pm
+++ b/lib/TikZImage.pm
@@ -142,7 +142,7 @@ sub draw {
 		or warn "Can't open $working_dir/image.tex for writing.";
 	chmod(0777, "$working_dir/image.tex");
 	print $fh $self->header;
-	print $fh $self->tex . "\n";
+	print $fh $self->tex =~ s/\\\\/\\/gr . "\n";
 	print $fh $self->footer;
 	close $fh;
 
--- a/lib/TikZImage.pm
+++ b/lib/TikZImage.pm
@@ -162,6 +162,7 @@ sub draw {
 	system "cd " . $working_dir . " && $latex_binary image.tex > pdflatex.stdout 2> /dev/null";
 
 	if (-r "$working_dir/image.$tex_ext") {
+		chmod(0777, "$working_dir/image.$tex_ext");
 		# Convert the file to the appropriate type of image file
 		if ($ext eq 'svg') {
 			if ($self->svgMethod eq 'dvisvgm') {
--- a/lib/TikZImage.pm
+++ b/lib/TikZImage.pm
@@ -35,7 +35,7 @@ sub new {
 		tikzLibraries => '',
 		texPackages   => {},
 		addToPreamble => '',
-		ext           => 'png',
+		ext           => 'svg',
         imageName     => ''
 	};
 	my $self = sub {
@@ -46,7 +46,7 @@ sub new {
 			if ($field eq 'ext') {
 				my $ext = shift;
 				$data->{ext} = $ext
-				if ($ext eq 'png' || $ext eq 'gif' || $ext eq 'svg' || $ext eq 'pdf');
+				if ($ext && ($ext eq 'png' || $ext eq 'gif' || $ext eq 'svg' || $ext eq 'pdf'));
 			}
 			else {
 				$data->{$field} = shift;
--- a/lib/TikZImage.pm
+++ b/lib/TikZImage.pm
@@ -33,7 +33,7 @@ sub new {
 		tex           => '',
 		tikzOptions   => '',
 		tikzLibraries => '',
-		texPackages   => {},
+		texPackages   => [],
 		addToPreamble => '',
 		ext           => 'svg',
 		svgMethod     => 'pdf2svg',
@@ -79,10 +79,13 @@ sub tikzLibraries {
 	return &$self('tikzLibraries', @_);
 }
 
-# Set additional TeX packages to load.  This accepts a single hash parameter.
+# Set additional TeX packages to load.  This accepts an array parameter.  Note
+# that each element of this array should either be a string or an array with one
+# or two elements (the first element the package name, and the optional second
+# element the package options).
 sub texPackages {
 	my $self = shift;
-	return &$self('texPackages', $_[0]) if ref($_[0]) eq "HASH";
+	return &$self('texPackages', $_[0]) if ref($_[0]) eq "ARRAY";
 	return &$self('texPackages');
 }
 
@@ -116,12 +119,13 @@ sub header {
 	my @output = ();
 	push(@output, "\\documentclass{standalone}\n");
 	push(@output, "\\def\\pgfsysdriver{pgfsys-dvisvgm.def}\n") if $self->ext eq 'svg' && $self->svgMethod eq 'dvisvgm';
-	my $xcolorOpts = $self->texPackages->{xcolor} ? $self->texPackages->{xcolor} : 'svgnames';
+	my @xcolorOpts = grep { ref $_ eq "ARRAY" && $_->[0] eq "xcolor" && defined $_->[1] } @{$self->texPackages};
+	my $xcolorOpts = @xcolorOpts ? $xcolorOpts[0][1] : 'svgnames';
 	push(@output, "\\usepackage[$xcolorOpts]{xcolor}\n");
 	push(@output, "\\usepackage{tikz}\n");
 	push(@output, map {
-			"\\usepackage" . ($self->texPackages->{$_} ne "" ? "[$self->texPackages->{$_}]" : "") . "{$_}\n"
-		} grep { $_ ne 'xcolor' } keys %{$self->texPackages});
+			"\\usepackage" . (ref $_ eq "ARRAY" && @$_ > 1 && $_->[1] ne "" ? "[$_->[1]]" : "") . "{" . (ref $_ eq "ARRAY" ? $_->[0] : $_) . "}\n"
+		} grep { (ref $_ eq "ARRAY" && $_->[0] ne 'xcolor') || $_ ne 'xcolor' } @{$self->texPackages});
 	push(@output, "\\usetikzlibrary{" . $self->tikzLibraries . "}") if ($self->tikzLibraries ne "");
 	push(@output, $self->addToPreamble);
 	push(@output, "\\begin{document}\n");
--- a/lib/TikZImage.pm
+++ b/lib/TikZImage.pm
@@ -115,12 +115,13 @@ sub header {
 	my $self = shift;
 	my @output = ();
 	push(@output, "\\documentclass{standalone}\n");
-	push(@output, "\\usepackage[svgnames]{xcolor}\n");
 	push(@output, "\\def\\pgfsysdriver{pgfsys-dvisvgm.def}\n") if $self->ext eq 'svg' && $self->svgMethod eq 'dvisvgm';
+	my $xcolorOpts = $self->texPackages->{xcolor} ? $self->texPackages->{xcolor} : 'svgnames';
+	push(@output, "\\usepackage[$xcolorOpts]{xcolor}\n");
 	push(@output, "\\usepackage{tikz}\n");
 	push(@output, map {
 			"\\usepackage" . ($self->texPackages->{$_} ne "" ? "[$self->texPackages->{$_}]" : "") . "{$_}\n"
-		} keys %{$self->texPackages});
+		} grep { $_ ne 'xcolor' } keys %{$self->texPackages});
 	push(@output, "\\usetikzlibrary{" . $self->tikzLibraries . "}") if ($self->tikzLibraries ne "");
 	push(@output, $self->addToPreamble);
 	push(@output, "\\begin{document}\n");
--- a/lib/TikZImage.pm
+++ b/lib/TikZImage.pm
@@ -36,6 +36,7 @@ sub new {
 		texPackages   => {},
 		addToPreamble => '',
 		ext           => 'svg',
+		svgMethod     => 'pdf2svg',
         imageName     => ''
 	};
 	my $self = sub {
@@ -98,6 +99,12 @@ sub ext {
 	return &$self('ext', @_);
 }
 
+# Set the method to use to generate svg images.  The valid methods are 'pdf2svg' and 'dvisvgm'.
+sub svgMethod {
+	my $self = shift;
+	return &$self('svgMethod', @_);
+}
+
 # Set the file name.
 sub imageName {
 	my $self = shift;
@@ -109,7 +116,7 @@ sub header {
 	my @output = ();
 	push(@output, "\\documentclass{standalone}\n");
 	push(@output, "\\usepackage[svgnames]{xcolor}\n");
-	push(@output, "\\def\\pgfsysdriver{pgfsys-dvisvgm.def}\n") if $self->ext eq 'svg';
+	push(@output, "\\def\\pgfsysdriver{pgfsys-dvisvgm.def}\n") if $self->ext eq 'svg' && $self->svgMethod eq 'dvisvgm';
 	push(@output, "\\usepackage{tikz}\n");
 	push(@output, map {
 			"\\usepackage" . ($self->texPackages->{$_} ne "" ? "[$self->texPackages->{$_}]" : "") . "{$_}\n"
@@ -147,19 +154,25 @@ sub draw {
 	close $fh;
 
 	my $ext = $self->ext;
-	my $latex_binary = WeBWorK::PG::IO::externalCommand($ext eq 'svg' ? 'latex' : 'pdflatex');
+	my $tex_ext = $ext eq 'svg' && $self->svgMethod eq 'dvisvgm' ? 'dvi' : 'pdf';
+	my $latex_binary = WeBWorK::PG::IO::externalCommand($tex_ext eq 'dvi' ? 'latex' : 'pdflatex');
 
 	# Generate the pdf file.
-	system "cd " . $working_dir . " && $latex_binary > pdflatex.stdout 2> /dev/null image.tex";
+	system "cd " . $working_dir . " && $latex_binary image.tex > pdflatex.stdout 2> /dev/null";
 
-	if (-r "$working_dir/image." . ($ext eq 'svg' ? 'dvi' : 'pdf') ) {
+	if (-r "$working_dir/image.$tex_ext") {
 		# Convert the file to the appropriate type of image file
 		if ($ext eq 'svg') {
-			system WeBWorK::PG::IO::externalCommand('dvisvgm') .
-				" $working_dir/image.dvi 2>&1 /dev/null --no-fonts --output=$working_dir/image.svg"
+			if ($self->svgMethod eq 'dvisvgm') {
+				system WeBWorK::PG::IO::externalCommand('dvisvgm') .
+					" $working_dir/image.dvi --no-fonts --output=$working_dir/image.svg > /dev/null 2>&1";
+			} else {
+				system WeBWorK::PG::IO::externalCommand($self->svgMethod) .
+					" $working_dir/image.pdf $working_dir/image.svg > /dev/null 2>&1";
+			}
 		} elsif ($ext ne 'pdf') {
 			system WeBWorK::PG::IO::externalCommand('convert') .
-				" 2>&1 /dev/null $working_dir/image.pdf $working_dir/image.$ext"
+				" $working_dir/image.pdf $working_dir/image.$ext > /dev/null 2>&1";
 		}
 
 		if (-r "$working_dir/image.$ext") {
@@ -173,9 +186,8 @@ sub draw {
 			warn "Convert operation failed.";
 		}
 	} else {
-		warn "File $working_dir/image.pdf was not created.";
-		if (open(my $err_fh, "<", "$working_dir/pdflatex.stdout"))
-		{
+		warn "File $working_dir/image.$tex_ext was not created.";
+		if (open(my $err_fh, "<", "$working_dir/pdflatex.stdout")) {
 			while (my $error = <$err_fh>) {
 				warn $error;
 			}
--- a/lib/WeBWorK/PG/IO.pm
+++ b/lib/WeBWorK/PG/IO.pm
@@ -287,26 +287,6 @@ sub curlCommand {
 	return $CE->{externalPrograms}{curl};
 }
 
-=item convertCommand
-
-	convert -- path to convert defined in site.conf
-
-=cut
-
-sub convertCommand {
-	return $CE->{externalPrograms}{convert};
-}
-
-=item pdflatexCommand
-
-	pdflatex -- path to pdflatex defined in site.conf
-
-=cut
-
-sub pdflatexCommand {
-	return $CE->{externalPrograms}{pdflatex};
-}
-
 =item copyCommand
 
 	copyCommand -- path to cp defined in site.conf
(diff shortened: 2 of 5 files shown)


Alex-Jordan:
Just looking over the example .pg files for now. I guess you can put things like `\(` in the heredoc because WeBWorK expands that to `\\(` before the perl interpolation happens, is that right?

How do you put actual dollar, percent, and at signs in the tikz code? Like if they were part of the text labeling an axis or something? Is it possible without using `${DOLLAR}` etc?

Can authors leave comments in the tikz code? Like `% this is a comment`?

Let's add the `BEGIN_TIKZ` substitution. If the early examples demonstrating this to new users avoid the heredoc syntax, it will be more comfortable.

drgrice1:
Yeah, the `\(...\)` construct works.  You can also do `~~$...~~$` and that works as well.

To add the dollar symbol you can do `\node at (3,-3){\(~~\$50,000\)};`.  `$DOLLAR` doesn't work though.

Using `% this is a comment` works.

I am not sure that the `BEGIN_TIKZ ... END_TIKZ` idea will work.  The TikZ code needs to be associated to a particular instance of a TikZImage object.  It is not the same as with `BEGIN_PGML ... END_PGML` and such.

drgrice1:
Yeah, so I looked into attempting to do the `BEGIN_TEXT ... END_TEXT` a bit.  That certainly would not be an easy thing to accomplish.  I don't think that is going to happen now.

drgrice1:
Okay.  So it wasn't that hard after all.  I have to update the documentation, but I will commit it soon.

drgrice1:
Note that using the `heredoc` still works (as in the attached examples in the initial comments on this pull request).  You could also use a `nowdoc` if desired (assuming you don't need interpolation).  The BEGIN_TIKZ/END_TIKZ construction is a `heredoc` though.  Generally that is what you want so that you can interpolate variables for dynamic images.

drgrice1:
Here is an example using the `BEGIN_TIKZ/END_TIKZ` construction.  Note it is a little different than the other `BEGIN/END` constructions, as it needs an object.

[BEGIN_TIKZ_eg.zip](https://github.com/openwebwork/pg/files/6253955/BEGIN_TIKZ_eg.zip)

pstaabp:
Overall, this looks good (with pun intended since svg is cleaner).

I wonder if instead of calling:
```
$graph_image->ext('svg') if $displayMode ne 'TeX';
```
inside a problem, that we have setting for this instead.  It seems like this should be a global or course-level setting rather than a problem setting.   We could have a 
```
$pg{specialPGEnvironmentVars}{tikz_image_output} = "svg";
```
inside `defaults.config` and override it in `site.conf` or `course.conf`.    Would the other setting then be 'png'?  

And then of course the conversion would need to handled somewhere instead of inside the problem itself.

drgrice1:
This is definitely not a case where it makes sense to be a course level or system level setting. This is a problem authoring decision. The point is that an author must decide if the images created by their tikz code work well with SVG or not. I am fine with setting the default to SVG, and then having the author override that to PNG if their images don't work well.  Generally SVG is the right way to go if that works.

I am not fond of having to set the override extension in the way that is done here, but there are limitations with the module that make this necessary.

drgrice1:
Attached are updated examples for the latest changes.
[examples.zip](https://github.com/openwebwork/pg/files/6254945/examples.zip)

Alex-Jordan:
Here is a screenshot from a modified version of the testing problem `BEGIN_TIKZ_eg.pg`. The modification is that it displays the svg image, then changes to png and displays it again. Then back to svg and change the image dimension from 200x200 to 100x100. Is something is wrong with the scaling for svg? Or do I need to do something to the tikz code, knowing the pixel dimensions of the ultimate output image?

Screen Shot 2021-04-13 at 4 13 49 PM

drgrice1:
I am not entirely certain of what you are doing.  I made a guess as follows, but it doesn't produce the output you show.  Here is my example attempting to reproduce your image.

```perl
DOCUMENT();

loadMacros("PGstandard.pl", "PGML.pl", "PGtikz.pl");

TEXT(beginproblem());

$a = random(-10, 9);
$b = random(-10, 9);

$graph_image = createTikZImage();
$graph_image->tikzLibraries("arrows.meta");
$graph_image->tikzOptions("x=0.6cm,y=0.6cm");
$graph_image->BEGIN_TIKZ
\tikzset{>={Stealth[scale=1.8]}}
\filldraw[draw=blue,fill=white,rounded corners=14pt,thick,use as bounding box] (-11,-11) rectangle (11,11);
\begin{scope}
	\clip[rounded corners=14pt] (-11,-11) rectangle (11,11);
	\draw[line width=0.2pt,color=gray,step=1] (-11,-11) grid (11,11);
\end{scope}
\huge
\draw[<->,thick] (-11,0) -- (11,0) node[above left,outer sep=2pt]{\(x\)};
\draw[<->,thick] (0,-11) -- (0,11) node[below right,outer sep=2pt]{\(y\)};
\foreach \x in {-10,-8,...,-2,2,4,...,10} \draw[thin] (\x,5pt) -- (\x,-5pt) node[below]{\(\x\)};
\foreach \y in {-10,-8,...,-2,2,4,...,10} \draw[thin] (5pt,\y) -- (-5pt,\y) node[left]{\(\y\)};
\filldraw[red] ($a,$b) circle[radius=5pt];
% this is a comment
\node at (3,-3){\(~~\$50,000\)};
END_TIKZ

BEGIN_PGML
[@ image(insertGraph($graph_image), width => 200, height => 200) @]*

[@ $graph_image->ext('png'); image(insertGraph($graph_image), width => 200, height => 200) @]*

[@ $graph_image->ext('svg'); image(insertGraph($graph_image), width => 100, height => 100) @]*
END_PGML

ENDDOCUMENT();
```

Alex-Jordan:
If I take `BEGIN_TIKZ_eg.pg` from your most recent examples.zip and just run it as is, I get:

Screen Shot 2021-04-13 at 9 09 11 PM

That's using the 300 pixel width written into the problem, and apparently I am seeing some default height. The image of course is a zoomed in section of the intended graph. 

If I change the `BEGIN_PGML...ENDPGML` part to the following, it is how I end up with the image I previously posted:

```
BEGIN_PGML

[@ image(insertGraph($graph_image), width => 200, height=>200, tex_size => 1000) @]*

END_PGML

$graph_image->ext('png');

BEGIN_PGML

[@ image(insertGraph($graph_image), width => 200, height=>200, tex_size => 1000) @]*

END_PGML

$graph_image->ext('svg');

BEGIN_PGML

[@ image(insertGraph($graph_image), width => 100, height=>100, tex_size => 1000) @]*

END_PGML
```

Alex-Jordan:
If there is a discrepancy between what I get and what you get, it could be some issue about the version of `latex` or `dvisvgm`. This server has `pdfTeX 3.14159265-2.6-1.40.18 (TeX Live 2017/Debian)` and `dvisvgm 2.1.3`.

drgrice1:
Argh!  I tested this on my server which has Ubuntu 18.04 and the versions of texlive/dvisvgm installed that you have.  I get the same crap.  So apparently newer versions of TeX and dvisvgm (and probably ghostscript which dvisvgm uses) are needed.

Looking closely I see that the pgfsys-dvisvgm.def file in the older texlive distribution is quite different than the one in the newer tex distribution on Ubuntu 20.04.  On Ubuntu 18.04 the dvisvgm binary is part of the texlive-binaries package, but on Ubuntu 20.04 it is in the separate dvisvgm package.  So apparently things have changed quite a bit with dvisvgm.

drgrice1:
I added a system configuration option as @pstaabp had suggested earlier to set the default image extension used by the PGtikz.pl macro.  This is not ideal, but unfortunately needed as the versions of latex and dvisvgm on Ubuntu 18.04 are not good enough to generate 'svg' images.  See https://github.com/openwebwork/webwork2/pull/1320 for the settings.

Alex-Jordan:
I never really tried making tikz images before this change. Would it make sense for the default to still be svg, but the configuration option is between the `convert` method or the `dvisvgm` method?

I'm not sure if the same or similar issues happen with the `convert` method. There must be something I do not understand about the svg production before this PR. With test files I try, even the images that are supposed to come out as svg still come out as png. Like with this test file from an old PR:

```
##DESCRIPTION
# TEST tikz from a pgml problem
##ENDDESCRIPTION
DOCUMENT();
loadMacros(
    "PGstandard.pl",
    "MathObjects.pl",
    "PGML.pl",
    "PGtikz.pl"
);
TEXT(beginproblem());
##############################################################
#  Setup
##############################################################
$drawing = createTikZImage();
$drawing->tikzOptions("main_node/.style={circle,fill=blue!20,draw,minimum size=1em,inner sep=3pt}");
$drawing->tex(<<END_TIKZ);
\node[main_node] (1) at (0,0) {1};
\node[main_node] (2) at (-1, -1.5)  {2};
\node[main_node] (3) at (1, -1.5) {3};
\draw (1) -- (2) -- (3) -- (1);
END_TIKZ
$path = insertGraph($drawing);
Context("Numeric");
##############################################################
#  Text
##############################################################

BEGIN_PGML
path = [$path]
path = [@ protect_underbar($path) @]
[@ $BR @]*
alias = [@ alias($path) @]*
alias = [@ protect_underbar(alias($path)) @]*

image = [@ image($path, width => 100, tex_size => 400) @]*

svg = [@ embedSVG($path) @]*
END_PGML
ENDDOCUMENT();
```

pstaabp:
Something a bit crazy is going on.  I'm looking at the `png_svg_cmp.pg` file that you have in the examples.  It appears to default to PNG in that both of them are clearly png files. 

If I add the line
```
$graph_svg->ext('svg');
```
then I do get it as a SVG file.   I was looking if there is any config that I set, but it doesn't appear to.  I'm guessing others are not seeing this.

pstaabp:
I just looked at openwebwork/webwork2#1320 and I have not pulled that in, so that's not causing this.

drgrice1:
@Alex-Jordan:  The convert method always really generates png images, even for the svg extension.  If you select the svg extension it just puts the png image into an svg container.

@pstaabp:  Even without openwebwork/webwork2#1320 the default is now 'png'.  I haven't updated the example file, so you will get 'png' from both.  You need to add `$image->ext('svg')` to get 'svg' output.

pstaabp:
@drgrice1 Thanks.  I was walking through the code, trying to figure out what's going on.  I got the impression that line 38 in `TikZImage.pm` set the default.  Evidently not.

Alex-Jordan:
OK, shooting through all of the options I know of, what about `pdf2svg` then? It's what PreTeXt uses for the PDF -> SVG conversion in a similar situation. Although there is movement afoot that may lead to using `dvipsvgm`. 

Using `pdf2svg` (assuming it even works for this) could be a separate project.

drgrice1:
@pstaabp:  Line 38 of TikZImage.pm sets the underlying default of TikZImage package.  However, that is overridden by line 99 of PGtikz.pl.  That sets the extension to the tikzImageExtension if that is defined, and falls back to 'png' if not.

@Alex-Jordan:  I would prefer to use dvisvgm as the current version does a better job in general.  I guess I will switch this pull request to using pdf2svg for now though.

Alex-Jordan:
> @Alex-Jordan: I would prefer to use dvisvgm as the current version does a better job in general. I guess I will switch this pull request to using pdf2svg for now though.

I completely agree. I am wondering if the configuration option that you just introduced could control whether this process uses `dvisvgm` versus `pdf2svg` instead of controlling whether the default extension is svg or png.

drgrice1:
> I completely agree. I am wondering if the configuration option that you just introduced could control whether this process uses `dvisvgm` versus `pdf2svg` instead of controlling whether the default extension is svg or png.

I was thinking the same thing, and will try to implement that.

pstaabp (approved):

Alex-Jordan (approved):
Things are mostly working for me now, with https://github.com/openwebwork/webwork2/pull/1320. 

One thing that is not is `tikz_tex_package_test.pg` from the most recent examples zip. It does not create the images (the web browser displays a broken image). If I comment out this line: `$drawing_png->addToPreamble("\pgfplotsset{compat=1.16}");` in each image construction, then they come through. So I guess that version is not present on my system. If these examples make it to the wiki or something like that, maybe these lines should be removed.

drgrice1:
Yeah, on Ubuntu 18.04 the newset compat level for pgfplots is 1.15.  So Change compat=1.16 to compat=1.15 and it will work.  That will still work for newer versions, but force the older compat level.  It isn't necessary to set the compat level, but pgfplots runs in backwards compatibility mode if you don't which can result in images that don't look as good.

taniwallach:
My impression is that this is almost ready, so the question is mainly when to merge into 2.16, and not if we will get it into the final version of 2.16. If correct - please add this and the webwork2 side to the WW 2.16 project.

pstaabp:
I'm getting some errors now with the newer code.  I pulled in  openwebwork/webwork2#1320 and set up  the path to `pdf2svg`.  I didn't change anything else. 

On the `png-svg-cmp.pg` example, I'm getting the PNG to work, but not the svg.  Drilling down into the `htdocs/tmp/XXX/images` directory, the svg is the following: 

image

Setting
```
$pg{specialPGEnvironmentVars}{tikzSVGMethod} = "dvisvgm";
```
in `localOverrides.conf`

generates the right SVG file.  

@drgrice1  I know you set this up for administrators to choose the better method to use.  I'm just wondering I found the case where `pdf2svg` doesn't work well.

pstaabp:
I also got the error:
```
Convert operation failed.
```
from line 186 of `TikZImage.pm`

drgrice1:
@pstaabp:  I tested both pdf2svg and dvisvgm on Ubuntu 20.04 with that problem, and don't get any errors.  I also tested the pdf2svg on Ubuntu 20.04, and don't have any problems, although do with dvisvgm as @Alex-Jordan did.

drgrice1:
@pstaabp:  The system administrator is the right one to choose the binary to use.  Generally speaking at this point the only determination will be if the latex/dvisvgm versions are new enough.  The author will then need to choose the output extension to use based on if the svg output works well or not.  Both programs work to generate an svg for all cases that I have tested on both Ubuntu 20.04 and Ubuntu 18.04 assuming the pdf/dvi generation by latex works.  As to if that svg is the desired output or not is another issue.  So the author will have to examine and determine if a png image will need to be used instead.  The png method will generally be more reliable, just not as pretty.

I am not sure how you are getting the issue you are seeing.

pstaabp:
@drgrice1 could be just my platform.  I'm running everything natively on a Mac, but am generally successful.   I'll try to debug a bit more.  If you're not seeing errors, I'm not so worried.

drgrice1:
@pstaabp:  You could try commenting out lines 199-201 of TikZImage.pm (those delete the temporary files used in /opt/webwork/webwork2/tmp) and then follow the image generation process in TikZImage.pm on the command line, and see if you get the same result, and where things go wrong.

pstaabp (approved):
I seem to have a `pdf2svg` that isn't working, so unless I can figure out that, I won't be able to test that.  I think the framework of different backends for the svg generation is sound and flexible.

drgrice1:
@pstaabp:  I think you have found a case where pdf2svg does not work, but dvisvgm does.  What versions of those programs do you have?  For dvisvgm you can run `dvisvgm -V`.  For pdf2svg you will have to look deeper.  On Ubuntu you can check the package version.  I am not sure how to check that with your setup.

drgrice1:
By the way, the working version of dvisvgm on Ubuntu 20.04 is 2.8.1, and the working version pdf2svg is 0.2.3.

pstaabp:
dvisvgm: 2.11.1
pdf2svg: 0.2.3_6

Running on a Mac, I'm using the dvisvgm that comes with texlive and pdf2svg is part of the home-brew project.

drgrice1:
I see.  There are actually quite a few other things that can go wrong with pdf2svg.  One possibility is that Ubuntu applies a patch do pdf2svg in their package that fixes something.  Another is that the versions of Poppler and Cairo that are installed are different (or missing).  The pdf2svg program is rather simple program that just calls on those libraries.  Also, its code has not changed since 2017.  The Poppler and Cairo libraries have changed quite a bit though.

drgrice1:
dvisvgm is quite actively maintained.  Your version is quite new, and certainly on a system that has that should be using it.  So on your system, you certainly want to change the settings to do so.

Perhaps it would be good to provide a test case for system administrators to determine which of the two to use.

pstaabp:
> I see. There are actually quite a few other things that can go wrong with pdf2svg. 

It does not seem to generate helpful error comments. I did comment out the lines to delete the temporary files and ran `pdf2svg` directly on the pdf that was generated and got the
```
Unable to open file
```
error.  I can open the file in many other programs, so not sure what is up.  

I also just tested `pdf2svg` on another pdf file (generated from latex) and it worked.  I'll keep troubleshooting.

drgrice1:
That sounds like a permissions issue.  Try a `chown` or `chmod` on the pdf file.

pstaabp:
I had thought about that.  I fixed permissions and ran as sudo and no luck.

drgrice1:
Maybe it is the pdf file that has an issue.  Did you regenerate that from the tex file?

pstaabp:
It's a weird permissions problem.  I had to set chmod 666 on the file to get it read.  I assume the web service runs as the same user that apache does.  correct?

drgrice1:
The web servcie does run with the same user that apache does.  Note the TikZImage module changes the permissions of the generated tex file to 0777.  Was it the permissions on that file that was the problem?  It doesn't change the permissions on the other generated files.

pstaabp:
The .tex file is 0777.  Everything else is 644 and although both the owner and group are correct, it doesn't seem to like that.  However if I'm in the temp directory with the image.tex and image.pdf files, then
```
 sudo -u _www pdf2svg image.pdf image.svg
```
works fine where `_www` is the user for running the server.   That's why I'm wondering if it is running as a different user.

drgrice1:
That would be odd.  Usually everything runs under the server user.  Although there are ways to make that different.  There are shared server setups that sandbox certain processes or things run from certain locations for instance.

drgrice1:
Although, even in those scenarios, usually it wouldn't change users mid process.  It is writing the tex file, so it is doubtful it could change users after doing that for the rest of the process.

pstaabp:
> Although, even in those scenarios, usually it wouldn't change users mid process. It is writing the tex file, so it is doubtful it could change users after doing that for the rest of the process.

Also, it works under the `dvisvgm` which is running under the same situation too.  I'm perplexed and since I'm testing on a non-standard server, I'm not worried about this and we should push ahead.

pstaabp:
I think this is ready to go.  @drdrew42 have you had a chance to look at this?

Alex-Jordan:
The ability to load a package with options is not working for me. It is also not working on my production server, before this PR, so it's not because of any code changes here. But I discovered it while testing this again with the example problems.

If I use the following on my Ubuntu 18.04.5 server, it works. It loads the `xcolor` package. But if I try to pass an option to `xcolor` like with `$graph_image->texPackages({ xcolor => "cmyk" });` then it fails to create an image. (It's not specific to `xcolor` and that option. I experienced this with other packages and options too.)

```
DOCUMENT();

loadMacros("PGstandard.pl", "PGML.pl", "PGtikz.pl");

TEXT(beginproblem());

$a = random(-10, 9);
$b = random(-10, 9);

$graph_image = createTikZImage();
$graph_image->tikzLibraries("arrows.meta");
$graph_image->tikzOptions("x=0.6cm,y=0.6cm");
$graph_image->texPackages({ xcolor => "" });
$graph_image->BEGIN_TIKZ
\color{Orchid}
\node {\(x\)};
END_TIKZ

BEGIN_PGML

[@ image(insertGraph($graph_image), width => 300, tex_size => 1000) @]*

END_PGML

ENDDOCUMENT();
```

Alex-Jordan:
I take it back. I swear I tried other packages with options that failed, but now I tried one that works. And the apache error log indicates that `[cmyk]{xcolor}` causes a failure because of a package option clash (with tikz I presume).

Alex-Jordan:
All is working for me.

It would be good if we could get one more test though, since we've discovered that OS-specific issues arise with this.

I have a few follow-up features, and I'm just waiting on this one to be merged first.

drgrice1:
@Alex-Jordan:  Okay, so I guess I need to add this to the documentation.  You can't load the `xcolor` package.  However, the `xcolor` package is already loaded with the `svgnames` option, so you don't need to.  This is done explicitly before loading the `tikz` pacakge.  Note that the `tikz` package automatically loads the `xcolor` package.  So usually with latex if you are using `tikz`, the only reason to explicitly load the `xcolor` package is to change the options to something other than the defaults, and to do this you must load the `xcolor` package before the `tikz` package.  Currently the `TikZImage` perl package loads the `xcolor` package with the `svgnames` option, then it loads the `tikz` package, and then the packages set in the options.  If we want to allow more versatility with the `xcolor` options then either a separate option for that would need to be added, or the code would need to filter the `xcolor` package and its options out of the `texPackages` option and add the options to the earlier loading of the `xcolor` package.  The latter method is more complicated to implement, but could be done.



  The loading of the other packages comes after that (and must).

drgrice1:
@Alex-Jordan:  I decided to implement the second version of what I described in my last comment.  That is now you can just do what you were trying to do and load the `xcolor` package with the `cmyk` option, and it will work.

drgrice1:
@pstaabp:  Could you test with the latest commit, and see if that resolves the permissions issue that you are experiencing?

pstaabp:
@drgrice1 This fixes that permissions problem now.

drgrice1:
@pstaabp:  So the images are properly generated on your local test server then?

pstaabp:
@drgrice1 before this patch, it wouldn't generate the svg image using `pdf2svg`. Now it works fine.

drgrice1:
Yeah, that is what I meant.  I forgot that it was working with dvisvgm.  Good deal.

Alex-Jordan:
>  I decided to implement the second version of what I described in my last comment. That is now you can just do what you were trying to do and load the xcolor package with the cmyk option, and it will work.

Thanks. I got to thinking what other package option clashes might arise? Since the packages and their options are a hash reference, we cannot require the packages to load in the order that the problem author writes them.  I thought to add an alternative where order would be enforced with an array reference. An extreme example:

```
$image->texPackages([
    {package1 => {option1a => value1a}},
    {
        package2 => {option2a => value2a}, 
        package3 => {option3a => value3a, option3b => value3b}
    }
])
```

but I think that is not easy to merge with the `xcolor` handling.

Probably it is incredibly rare that packages you might use with a tikz picture have the opportunity for a clash, so this observation is not all that important.

drgrice1:
> I got to thinking what other package option clashes might arise? Since the packages and their options are a hash reference, we cannot require the packages to load in the order that the problem author writes them. I thought to add an alternative where order would be enforced with an array reference. An extreme example:
>
> Probably it is incredibly rare that packages you might use with a tikz picture have the opportunity for a clash, so this observation is not all that important.

You are right.  This may not be that important, but could be an issue.  How about just an array or arrays, where the inner arrays may have 1 or 2 elements.  The first being the package name, and the second (if provided) the package options?  So it would be used like
```perl
$image->texPackages([
    [package1, options1],
    [package2]
]])
```
Here `options1` would be a string that is in the format that it would be passed to the package in latex already.  So the latex that is output would be
```latex
\usepackage[options1]{package1}
\usepackage{package2}
```
Of course the xcolor package would be filtered out.

I don't see the need to separate the options into yet another embedded hash.  Furthermore, latex package options do not always fit into the format that are assumed with that.

drgrice1:
The last commit changes from hash to an array for the texPackages option of the TikZImage package.
An example of the new usage is
```perl
$image = createTikZImage();
$image->texPackages([
    "pgfplots",
    ["hf-tikz", "customcolors"],
    ["xcolor", "cmyk,table"]
]);
```
Note that the `xcolor` package breaks the order rule.  It is always loaded, and is always loaded first.  Passing it is only needed if you want to change the options for the package.

drgrice1:
Here is the updated example that uses the pgfplots package.  It doesn't load any other packages though, so it is not a good example of how to use the texPackages option.
[tikz_tex_package_test.zip](https://github.com/openwebwork/pg/files/6343995/tikz_tex_package_test.zip)

Alex-Jordan:
I looked at the most recent commit that moves to an array reference for packages, and it looks good. Still approve for merge.

somiaj:
Now that this has been merged into PG-2.16, I have tested it out, using the .pg file in `tizk_tex_package_test.zip`. The .png file generates just fine but the .svg didn't with the default `pdf2svg`. Changing to `dvisvgm` worked. I also didn't notice any issues with `dvisvgm`, though my server is running Debian 10 with slightly older version at `2.6.1`. This is what I have in my apache2 error logs when trying to create an svg with `pdf2svg`.

```
Use of uninitialized value in concatenation (.) or string at /home/webwork/pg/lib/TikZImage.pm line 176.
Convert operation failed. at /home/webwork/pg/lib/TikZImage.pm line 192.
```

drgrice1:
@somiaj:  Yeah, if you are testing things with the release candidate branch, make sure that your configuration files are all updated.  Thanks for testing though.  We need as many people testing things as we can get.

somiaj:
@drgrice1 What is the preferred way to change convert options? I have been wanting to use Tikz to create some graphs, but the .png quality was always low. Adding `-density 300` to the convert option in `TikZImage.pm` on line 181 has made the .png images that are being created better quality. Here I think my issue is my images are actually fairly small, so the defaults don't work (your example the defaults worked fine). If there isn't a method, maybe it could be useful to add an option to be able to pass to convert? (Though the svg images work just fine, so I can just use those).

I also assume that using `$v_1$` for math cannot be supported (though `\(v_1\)` works, just means I have to update my graphs from my .tex files).

Anyways, thanks for the support, and all my tests appear to be working fine, these are just things I noticed.

Alex-Jordan:
There are two open pull requests to handle convert options. One for `pg`, one for `webwork2`.

https://github.com/openwebwork/pg/pull/566

https://github.com/openwebwork/webwork2/pull/1334

drgrice1:
Wait for #566.  That adds the capability you are looking for.  @Alex-Jordan just beat me to this post.

drgrice1:
You can use `$v_1$`, just use `~~$v_1~~$`.  Yeah, it is a bit ugly, so `\(v_1\)` is nicer.

somiaj:
@drgrice1 I was working with creating some images with this package and noticed the following behavior.

+ Write a problem to create an image and save it.
+ View the problem in a homework set.
+ Go edit the problem, and change the image.
+ When I "View" the edited images everything works fine, but when I "Update" the problem it shows the old image in the assignment.

It appears that the old image is used if it is attached to an assignment (something is being cached?) thus making it so changes don't appear in the assignment. I tried deleting the image from `htdocs/tmp`, but that didn't change things, the old image was still generated. But when I unassigned the assignment and reassigned it, the new image was generated.

drgrice1:
@somiaj:  Check things carefully to see that you are using the files you think you are using.  That is, make sure you are using the actual PG file for the homework problem that you think you are.  There was a bug in the PG problem editor that actually changed the file the instructor was using in homework sets when the instructor edited a PG file.  That was fixed in a recent pull request, but since you are mentioning restarting the server it is possible your server was on an earlier git commit before it was last restarted.

somiaj:
@drgrice1: That wasn't the issue, other changes in the file would be visible, it was only the image that would not change.

I tested this again this morning and got the same behavior again. I am able to see any changes to the problem except the image. But I think I was able to track down the issue, and it is due to my browser. Firefox was just caching the image, and since the random string that is generated is the same for the problem, the image name was the same and my browser was using the cached version. A hard reload got the correct image showing. Firefox doesn't seem to cache the image right away when I was actively testing it, but if I wait a bit and came back, it would use its cached version.

Thanks again for your feedback.

PR #558, opened 2021-04-03
reduce extra digit checking in digits toltype

Alex-Jordan:
I think the default on this should be more loose, as close as it can be to the current relative tolerance defaults. This change means you must have the first three digits right (not counting trailing zeros which may always be omitted) and then if you give a 4th digit, it must be right. But now there will be no checking beyond that 4th digit. Assuming truncation or rounding are both set to be permitted, for π, these would now be correct:

3.14
3.141\d*
3.142

and these would now be incorrect:

3.15
3.14[^12]\d*
3.142\d+

A change is that, for example, 3.1417 is not correct by default before this change, but now it would be.

### DIFF
--- a/lib/Value.pm
+++ b/lib/Value.pm
@@ -62,7 +62,7 @@ like equality are "fuzzy", meaning that two items are equal when they are "close
     zeroLevel    => 1E-14,
     zeroLevelTol => 1E-12,
     tolTruncation => 1,
-    tolExtraDigits => 3,
+    tolExtraDigits => 1,
     #
     #  For Formulas:
     #
@@ -117,7 +117,7 @@ $defaultContext = Value::Context->new(
     zeroLevel    => 1E-14,
     zeroLevelTol => 1E-12,
     tolTruncation => 1,
-    tolExtraDigits => 3,
+    tolExtraDigits => 1,
     #
     #  For Formulas:
     #


Alex-Jordan:
>  3.14[^12]\d*
>  3.142\d+

Almost. 3.14000 will still be brought to 3.14 and counted as correct. Similarly for 3.142000.

pstaabp (approved):
Seems reasonable and I will defer to @Alex-Jordan on this what the best default will be.

drgrice1 (approved):
Looks fine to me.  This just changes defaults, which can always be overridden if desired.

Alex-Jordan:
This is better for the default, in principle because it is closer to relative with tolerance 0.001.

No one will be taking it to the extreme that we take it here (making it the default instead of relative) but the way it is now is too intolerant. My own school's courses have not have issues, largely because we use our own problems that are generally coded well. But we let some outsiders use our server, including a high school in Wisconsin. They use OPL problems, and this caused issues. You'd have something like a correct answer that really should have been like `2^x` or `e^(ln(2) x)`, but was coded to be `e^(0.693147x)`. A student would enter `2^x` and then you'd have test values that were off in the 5th or 6th digit, so marked incorrect.

drgrice1:
This has two approvals.  I will go ahead and merge this.

PR #559, opened 2021-04-06
alt option for image()

Alex-Jordan:
Currently to get alt text in an HTML image, you must awkwardly use `extra_html_tags`. This adds `alt` as an option, which is not only less awkward, but also allows PreTeXt to do its natural thing with an alt text description.

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2929,6 +2929,8 @@ sub image {
 		width    => 100,
 		height   => '',
 		tex_size => 800,
+		# default value for alt is undef, since an empty string is the explicit indicator of a decorative image
+		alt => undef,
 		extra_html_tags => '',
 	);
 	# handle options
@@ -2944,6 +2946,7 @@ sub image {
 	my $width       = $out_options{width};
 	my $height      = $out_options{height};
 	my $tex_size    = $out_options{tex_size};
+	my $alt         = $out_options{alt};
 	my $width_ratio = $tex_size*(.001);
 	my @image_list  = ();
 
@@ -2994,10 +2997,11 @@ sub image {
 			|| $displayMode eq 'HTML_asciimath'
 			|| $displayMode eq 'HTML_LaTeXMathML'
 			|| $displayMode eq 'HTML_img') {
-			$out = qq!<IMG SRC="$imageURL" class="image-view-elt" tabindex="0" role="button" WIDTH="$width" $height_attrib $out_options{extra_html_tags}>!
+			my $altattrib = (defined $alt) ? ('alt="' . ($alt =~ s/"/&quot;/gr) . '"') : '';
+			$out = qq!<IMG SRC="$imageURL" class="image-view-elt" tabindex="0" role="button" WIDTH="$width" $height_attrib $out_options{extra_html_tags} $altattrib>!
 		} elsif ($displayMode eq 'PTX') {
 			my $ptxwidth = 100*$width/600;
-			$out = qq!<image width="$ptxwidth%" source="$imageURL" />!
+			$out = (defined $alt) ? qq!<image width="$ptxwidth%" source="$imageURL"><description>$alt</description></image>! : qq!<image width="$ptxwidth%" source="$imageURL" />!
 		} else {
 			$out = "Error: PGbasicmacros: image: Unknown displayMode: $displayMode.\n";
 		}
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2908,12 +2908,7 @@ sub row {
 =head2 Macros for displaying static images
 
     Usage:
-    $string = image($image, width => 100, height => 100, tex_size => 800)
-    $string = image($image, width => 100, height => 100, extra_html_tags => 'align="middle"', tex_size => 800)
-    $string = image([$image1, $image2], width => 100, height => 100, tex_size => 800)
-    $string = caption($string);
-    $string = imageRow([$image1, $image2 ], [$caption1, $caption2]);
-             # produces a complete table with rows of pictures.
+    $string = image($image, width => 100, height => 100, tex_size => 800, alt => 'alt text', extra_html_tags => 'style="border:solid black 1pt"');
 
 =cut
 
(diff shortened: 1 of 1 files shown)


macros/PGbasicmacros.pl
```diff
@@ -2994,10 +2997,11 @@ sub image {
 			|| $displayMode eq 'HTML_asciimath'
 			|| $displayMode eq 'HTML_LaTeXMathML'
 			|| $displayMode eq 'HTML_img') {
-			$out = qq!<IMG SRC="$imageURL" class="image-view-elt" tabindex="0" role="button" WIDTH="$width" $height_attrib $out_options{extra_html_tags}>!
+			my $altattrib = (defined $alt) ? ('alt="' . ($alt =~ s/"/&quot;/gr) . '"') : '';
+			$out = qq!<IMG SRC="$imageURL" class="image-view-elt" tabindex="0" role="button" WIDTH="$width" $height_attrib $out_options{extra_html_tags} $altattrib>!
```
drgrice1, line 3001: It would probably be better to use `encode_pg_and_html` here instead of this regular expression to deal with quotes.  That method will deal with the quotes and other html entities that might need to be encoded.

Alex-Jordan:
Thanks, I didn't know I had access to something like that without adding a package. Added that, and remobed the ternary operator I was using, which felt too cluttered here.

drgrice1 (approved):
Looks good. I tested it and it works well.

drgrice1:
There is one more thing that should probably be done here. That is add this to the usage above in the PHbasicmacros.pl file so this gets into the POD documentation.

Alex-Jordan:
So I looked at that just now.

* There is an example with an `align` attribute. That seems to be deprecated, so I will maybe change that example to put some inline style attribute, unless you have a better suggestion.
* I did not realize you can pass an array reference to `image` with more than one image. So maybe `alt` should be able to accept an array reference as well. I'll look into that.

drgrice1:
Yes, the align attribute has been deprecated for a long time. Change that.

Alex-Jordan:
It seems to me that there are legacy subroutines `images()`, `caption()`, `captions()`, `imageRow()`, and that syntax `image([$oneimage,$anotherimage],...)` only exists to support these legacy things. Also, these legacy subroutines do show up in the OPL.

I propose removing the last three example lines from the POD. Do you see any reason to handle it differently?

macros/PGbasicmacros.pl
```diff
@@ -2908,12 +2908,7 @@ sub row {
 =head2 Macros for displaying static images
 
     Usage:
-    $string = image($image, width => 100, height => 100, tex_size => 800)
-    $string = image($image, width => 100, height => 100, extra_html_tags => 'align="middle"', tex_size => 800)
-    $string = image([$image1, $image2], width => 100, height => 100, tex_size => 800)
```
drgrice1, line 2913: I don't think that you should remove the array ref version of this.  It is still supported.

drgrice1:
I noticed that the header on the part for the `image` sub is "Macros for displaying static images".  Perhaps at one point only static images would work, but now dynamically generated images also work.  That should probably be changed.

The POD in the PGbasicmacros.pl file really needs some help.  It seems that not a lot of time was spent in creating what is there, as well as things have changed and the documentation was not updated.  The quick usage and example statements could be greatly improved with a little explanation.

See https://webwork.maa.org/pod/pg/macros/PGbasicmacros.html#Macros-for-displaying-static-images for what this looks like with the WeBWorK 2.15 code.

I am not suggesting that you do this here.  Just pointing this out.  If POD documentation were improved, that would make the other "documentation" efforts easier.

macros/PGbasicmacros.pl
Alex-Jordan: The thing is that because of how `wantarray` is used in the return statement, it can't easily be used that way. It is wired so that if you use `\{ image([$gr1,$gr2]) \}` you will only see `$gr1`. It's being called in scalar context, and its written to only return the first image in that situation. Not sure why, maybe something to do with the legacy code.

So you'd have to use it in some way to account for this. It works to do something like:
`\{ join(' ',image([$gr1,$gr2])) \}`

Useful if you understand about arrays, about image()'s ability to return an array or a scalar, and if you understand join.

I'll leave it there, maybe with more explanation though.

drgrice1: So build the `join` into the image return.  In other words change
`return wantarray ? @output_list : $output_list[0];`
to
`return wantarray ? @output_list : join(' ', @output_list);`

Alex-Jordan: One idea: in scalar context, it could go ahead and join all the images with a default space character (or something) rather than only return the first image.

Alex-Jordan: Heh, posted to a page that had not refreshed with your comment.

drgrice1: 👍

Alex-Jordan:
Unless I forgot something, this is ready for another look. Summary:

* POD editing
* You can pass `alt => 'alt text'` or `alt => ['alt text 1', 'alt text 2', ...]`
* If the image is just an image (not an array ref) then either the alt string will be used, or the first entry if alt is an array ref
* If the image is an array ref and the alt is just a string, the alt string will be repeated so many times
* If the image is an array ref and the alt is also an array ref, the entries from the alt array ref will be used until they are used up. Any additional images get no alt text.
* Now in scalar context, if `image()` was passed an array ref of images, it returns the join of their strings with a space in between. Formerly it only returned the first one. It's possible somewhere, something intentionally relied on that behavior. But it seems unlikely. And the OPL has no instances of `image([` so I suspect that `image()` was never used that way. It gets passed an array ref by some deprecated legacy subroutines, and those ones use it in array context.

drgrice1 (approved):
This looks good.

drgrice1:
@Alex-Jordan:  I added a pull request to this pull request with some suggested changes to the POD documentation.  There is no change to the wording, just the formatting of the generated POD output.

drgrice1:
I think this is good to go.  I will go ahead and merge it so you can continue work on your related pull request.

PR #561, opened 2021-04-07
direct use of WWPlot or TikZImage objects in image()

Alex-Jordan:
This makes it so if `$gr` is either a WWPlot or TikZImage object, then you can just do:

``
image($gr,...)
``

instead of 

``
image(insertGraph($gr),...)
``

(This is related to #560.)

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2975,7 +2975,7 @@ sub image {
 	my @output_list = ();
 	while(@image_list) {
 		my $image_item = shift @image_list;
-		$image_item = main::insertGraph($image_item) if (ref($image_item) eq 'WWPlot' or ref($image_item) eq 'TikZImage');
+		$image_item = insertGraph($image_item) if (ref $image_item eq 'WWPlot' || ref $image_item eq 'TikZImage');
 		my $imageURL = alias($image_item)//'';
 		$imageURL = ($envir{use_site_prefix})? $envir{use_site_prefix}.$imageURL : $imageURL;
 		my $out = "";
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2911,9 +2911,9 @@ Usage:
 
     image($image, width => 100, height => 100, tex_size => 800, alt => 'alt text', extra_html_tags => 'style="border:solid black 1pt"');
 
-where C<$image> can be a local file path or URL, C<width> and C<height> are
-pixel counts for HTML display, while C<tex_size> is per 1000 applied to
-linewidth (for example 800 leads to 0.8\linewidth)
+where C<$image> can be a local file path, URL, WWPlot object, or TikZImage object,
+C<width> and C<height> are pixel counts for HTML display, while C<tex_size> is
+per 1000 applied to linewidth (for example 800 leads to 0.8\linewidth)
 
     image([$image1,$image2], width => 100, height => 100, tex_size => 800, alt => ['alt text 1','alt text 2'], extra_html_tags => 'style="border:solid black 1pt"');
     image([$image1,$image2], width => 100, height => 100, tex_size => 800, alt => 'common alt text', extra_html_tags => 'style="border:solid black 1pt"');
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2924,7 +2924,8 @@ this produces an array in array context and joins the elements with C<' '> in sc
 
 #   More advanced macros
 sub image {
-	my $image_ref  = shift;
+	my $image_ref = shift;
+	$image_ref = main::insertGraph($image_ref) if (ref($image_ref) eq 'WWPlot' or ref($image_ref) eq 'TikZImage');
 	my @opt = @_;
 	unless (scalar(@opt) % 2 == 0 ) {
 		warn "ERROR in image macro.  A list of macros must be inclosed in square brackets.";
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2925,7 +2925,6 @@ this produces an array in array context and joins the elements with C<' '> in sc
 #   More advanced macros
 sub image {
 	my $image_ref = shift;
-	$image_ref = main::insertGraph($image_ref) if (ref($image_ref) eq 'WWPlot' or ref($image_ref) eq 'TikZImage');
 	my @opt = @_;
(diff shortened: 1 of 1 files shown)


macros/PGbasicmacros.pl
```diff
@@ -2919,7 +2919,8 @@ =head2 Macros for displaying static images
 
 #   More advanced macros
 sub image {
-	my $image_ref  = shift;
+	my $image_ref = shift;
+	$image_ref = main::insertGraph($image_ref) if (ref($image_ref) eq 'WWPlot' or ref($image_ref) eq 'TikZImage');
```
drgrice1, line 2923: There is no need for the `main::` namespace.  We are already in the `main` namespace here, and not in another package namespace.

Also, it is a good idea to get into the habit of using `||`, `&&`, and `!` instead of `or`, `and`, and `not` in logic statements.  It is a good rule of thumb to only use the word versions in flow control.  The word versions can cause issues if you aren't careful due to their low precedence.  The operators `||` and `or` are not equivalent.

Of course, another change to the image subroutine means another change to the POD usage is needed!  It would probably be enough to add a comment below the `Usage:` examples about the types of objects that the `$image` parameter may be.

drgrice1 (commented):
What about the case of an array of `WWPlot` or `TikZImage` objects?  I think the check should be done later in the method where an `ARRAY` ref is checked for, and there is probably a need to take care to check for an immediate ref to a `WWPlot` or `TikZImage` object, as well as for `WWPlot` or `TikZImage` objects in an `ARRAY`.  Most likely the actual `insertGraph` call should be inside the `while(@image_list)` loop.

Edit:  After looking at your other related pull request adding the alt attribute, I see that this is related to your comment there about deprecating the array usage.

Alex-Jordan:
I'm in the middle of your suggestions. I think I will wait on POD until the other PR is resolved. There would be conflicts there. 

Perhaps this is not an appropriate place to bring this up, but tikz image production is not working here in two (possibly related) ways. I don't mean because of this change. But I was working with it to test this, and I can't complete the test because of the tikz production not working. And I also wonder if it's moot with the tikz PR that you have open.

(1) If I just do the documented example like:
```
DOCUMENT();
loadMacros(
  "PGstandard.pl",
  "PGtikz.pl",
);

$image = createTikZImage();
$image->tex(<<END_TIKZ);
\draw (0,0) circle[radius=1.5];
END_TIKZ

BEGIN_TEXT
\{image(insertGraph($image))\}
END_TEXT

ENDDOCUMENT();
```
There is no image. There is an image URL, but nothing is there. Do I need to look somewhere to make sure write permissions are what they ought to be for creating the image file?

(2) Perhaps related. If I just try to save the result from `insertImage($image)`, I get errors. Like this file:
```
DOCUMENT();
loadMacros(
  "PGstandard.pl",
  "PGtikz.pl",
);

$image = createTikZImage();
$image->tex(<<END_TIKZ);
\draw (0,0) circle[radius=1.5];
END_TIKZ

$test = insertGraph($image);

ENDDOCUMENT();
```
with these error messages:
```
 'require' trapped by operation mask at /usr/lib/x86_64-linux-gnu/perl/5.26/Encode.pm line 5
   Died within Encode::Alias::find_alias called at line 114 of /usr/lib/x86_64-linux-gnu/perl/5.26/Encode.pm
   from within Encode::getEncoding called at line 132 of /usr/lib/x86_64-linux-gnu/perl/5.26/Encode.pm
   from within Encode::find_encoding called at line 166 of /usr/lib/x86_64-linux-gnu/perl/5.26/Encode.pm
   from within Encode::encode called at line 93 of [WW]/lib/Apache/WeBWorK.pm
   from within TikZImage::draw called at line 717 of [PG]/lib/PGcore.pm
   from within PGcore::insertGraph called at line 568 of [PG]/macros/PG.pl
   from within main::insertGraph called at line 12 of (eval 5146)
Compilation failed in require at /usr/lib/x86_64-linux-gnu/perl/5.26/Encode/Alias.pm line 22
```

Does anything strike you as familiar or otherwise understood with these issues I am having?

drgrice1:
Did you make the changes to /etc/ImageMagick-6/policy.xml noted in site.conf.dist?

drgrice1:
The comment starts on line 117 of that file.  This is also on the release notes page.

drgrice1:
Note that if you are using my pull request openwebwork/pg#557 (and the defaults with svg) that isn't needed.

drgrice1:
Everything looks good. 

Yeah, resolving merge conflicts is an art of itself.

drgrice1:
One thing to note is that if my pull request #557 is merged, then this will need to check for the `ref eq PGtikz` instead (or also as the TikZImage package could still be directly used).

Alex-Jordan:
I figured it must be something like that. My hunch is I did everything I should have done for the production server moving to 2.16, but failed to review all of the site.conf changes on this development server.

drdrew42 (approved):
Neat! 👍

pstaabp (approved):
Nice simplifying code. 

I did notice that I didn't have to change anything in the policy on image-magick.  I seem to be running version 7.  Also on my develop machine, it may have a different default policy.  I'll have to check when upgrading my production machine.

Alex-Jordan:
When #559 is resolved, I will rebase this, handle any conflicts, and complete the suggestions here.

Alex-Jordan:
I updated POD to mention the argument can be a WWPlot or TikZImage. If #557 is resolved before this, I will change "TikZImage" to "PGtikz".

drgrice1:
I will take a look after dinner. 

If this is merged before #557 I will add the package name change there.

drgrice1 (approved):
I tested this with both a single image, and a list of images.  It works well.

drgrice1:
Also, the documentation looks good.

pstaabp:
> If this is merged before #557 I will add the package name change there.

@drgrice1  Should we go ahead with this one and then change #557  as needed?

pstaabp (approved):

taniwallach:
> > If this is merged before #557 I will add the package name change there.
> 
> @drgrice1 Should we go ahead with this one and then change #557 as needed?

Three approvals and positive feedback in the discussion thread. It seems this one is ready to merge first, unless @drgrice1 has some technical reason to sequence in the other order.

I think the release notes should mention the new syntax provide, and make the relevant Wiki page.

drgrice1:
The sequencing doesn't matter to me.

pstaabp:
I'm going to go ahead and merge this.  We have 3 positive reviews.

PR #563, opened 2021-04-13
HTML escape math content for MathJax

Alex-Jordan:


Alex-Jordan:
This addresses https://github.com/openwebwork/webwork2/issues/1315.

drgrice1 (approved):
This should be good.  Another option is to re-implement the script tag handling, so that this is more like the way it was before.  MathJax 3 has ways for this to be done.  To do this change this line to
```perl
$out = '<span><script type="math/tex">'.$in.'</script></span>';
```
and add
```javascript
options: {
	renderActions: {
		findScript: [10, function (doc) {
			for (const node of document.querySelectorAll('script[type^="math/tex"]')) {
				const display = !!node.type.match(/; *mode=display/);
				const math = new doc.options.MathItem(node.textContent, doc.inputJax[0], display);
				const text = document.createTextNode('');
				node.parentNode.replaceChild(text, node);
				math.start = {node: text, delim: '', n: 0};
				math.end = {node: text, delim: '', n: 0};
				doc.math.push(math);
			}
		}, '']
	}
}
```
See http://docs.mathjax.org/en/latest/upgrading/v2.html

macros/PGbasicmacros.pl
```diff
@@ -2112,7 +2112,7 @@ sub general_math_ev3 {
 
 	my $out;
 	if($displayMode eq "HTML_MathJax") {
-		$out = '<span>'.$in_delim.'</span>';
+		$out = '<span>'.encode_pg_and_html($in_delim).'</span>';
```
dpvc, line 2115: Although this should work, it is more aggressive than necessary, as it encodes lots of other things that don't need to be.  For example, it encodes the backslashes from the `\(`, `\)`, `\[`, and `\]`, which, while it works, does make for some strange-looking page source.

Also, there are other locations that also need similar treatment (in webwork2).  In particular, 

https://github.com/openwebwork/webwork2/blob/05c982497e68d105829e190475ad73979df89c38/lib/WeBWorK/Utils/AttemptsTable.pm#L364-L366

and

https://github.com/openwebwork/webwork2/blob/05c982497e68d105829e190475ad73979df89c38/lib/WeBWorK/Utils/AttemptsTable.pm#L364-L366

Actually, it looks like there are other display modes that need similar treatment (e.g., the plainText and essay ones just above the linked lines).

dpvc:
In terms of the `renderActions` you could go that way, but you would need to change line 2115 to be more like

``` perl
$out = '<span><script type="math/tex'.($mode eq 'display' ? '; mode=display' : '').'">'.$in.'</script></span>';
```

in order to handle both inline and display math.

Note that this still will process `\(...\)` and `\[...\]` delimited math (and AsciiMath delimiters), so other usages should still work.  In particular, `$BM...$EM` and `$BDM...$EDM` should still be processed OK, provided the intervening math uses `$LTS` and `$GTS` (there doesn't seem to be one for `$AMP`, perhaps there should be?), which it would have to be for the other modes anyway.

Alex-Jordan:
I'm going to close my attempt at this for now. It seems relatively easy to do it right, but I think it could be some time now (a week?) before I come back to WW development, and I don't want the presence of this PR to inhibit someone else maybe getting it done.

PR #564, opened 2021-04-14
Wrap math content in `<script type="math/tex"></script>` tags for MathJax

drgrice1:
This is a replacement for @Alex-Jordan's previous attempt #563.

This is paired with openwebwork/webwork2#1322 and won't work without it.

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2112,7 +2112,7 @@ sub general_math_ev3 {
 
 	my $out;
 	if($displayMode eq "HTML_MathJax") {
-		$out = '<span>'.$in_delim.'</span>';
+		$out = '<span><script type="math/tex' . ($mode eq 'display' ? '; mode=display' : '') . '">' . $in . '</script></span>';
 	} elsif ($displayMode eq "HTML_dpng" ) {
 		# for jj's version of ImageGenerator
 		#$out = $envir->{'imagegen'}->add($in_delim);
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):
These work for me.

There is a bonus improvement. Before when the page first loads, you see the unprocessed math content for a second before MathJax processes. With this, the script tag content is not visible until the math is processed. I think that is an improvement, although it's worth noting that the rest of the text will tend to move more now that there is not unprocessed math serving as a placeholder.

pstaabp (approved):

macros/PGbasicmacros.pl
```diff
@@ -2112,7 +2112,7 @@ sub general_math_ev3 {
 
 	my $out;
 	if($displayMode eq "HTML_MathJax") {
-		$out = '<span>'.$in_delim.'</span>';
+		$out = '<span><script type="math/tex' . ($mode eq 'display' ? '; mode=display' : '') . '">' . $in . '</script></span>';
```
dpvc, line 2115: Is there a need for the surrounding `<span>`?  I don't see one myself.

dpvc (approved):
One unnecessary `<span>`, but other than that, OK.

macros/PGbasicmacros.pl
drgrice1: It was there before, and I just left it.   It isn't needed as far as I can tell.  I removed it.

Issue #567, opened 2021-04-23
contextFraction.pl issue with previous answer being `ln|6/10|` and flag `reduceConstantFunctions=>0`


Alex-Jordan:
Below is a sample PG problem, minimalized to demonstrate an issue. Save this problem in a set, with the open and close dates such that you can submit an answer (not just check).

If you try to submit the answer `ln|6/10|` it will be a successful submission (although probably the wrong answer since I minimialized things).

Now try to submit it again, and there are errors. In 2.15, I get pink WeBWorK warnings:
```
    Argument "|" isn't numeric in numeric ge (>=) at line 877 of [PG]/macros/contextFraction.pl
    Argument "|" isn't numeric in numeric ge (>=) at (eval 1812) line 877.
```

With 2.16, I get a more severe error screen.

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
  "PGML.pl",
  "contextFraction.pl",
);

TEXT(beginproblem());

$b = 1;
($lb,$ub) = (2,3);
Context("Fraction");
Context()->flags->set(reduceConstantFunctions=>0);
$f = Formula("1/(x+$b)")->reduce;
$frac = Fraction($ub+$b,$lb+$b);
$A = Formula("ln($frac)");

BEGIN_PGML
Evaluate the definite integral.

[``\int_{[$lb]}^{[$ub]} [$f] \, dx``]

[_]{$A}{40}

END_PGML


ENDDOCUMENT(); 
```

The errors disappear if I either do not use Fraction, or if I comment out this line: `Context()->flags->set(reduceConstantFunctions=>0);`

drgrice1:
With webwork 2.16 I am only getting the same errors that you say you get with webwork 2.15.  Different line numbers, but the same error other than that.  Not a more severe error screen.  Still seems like it shouldn't be giving any error though.

drgrice1:
I also noted that if you add `reduceConstants => 0` to the flags the error goes away.

dpvc:
I have made a PR to resolve this issue.

pstaabp:
This has been addressed.

PR #569, opened 2021-04-25
Fix issue with contextFraction when checking for repeated formulas (#567)

dpvc:
This PR fixes several problems that cause the issue described in #567 with the Fraction context.  The main problem was that the `string()` and `TeX()` methods were handling the parameters incorrectly (there are two that were missing, and `$prec` was getting the value from the first of them, which is what was causing the error).  The same issue also occurs for the Real class (which is probably why it made it into the Fraction class).

There is also a potential issue in the function-equivalence checking post-filter, which is that the context flags used there are not necessarily the same as the ones used in the main `cmp()` method.  So I've added the calls that set/reset the flags, which are duplicated from the main function.

Finally, I adjust the Union `string()` and `TeX()` to handle undefined `$prec` in the same way as in other MathObject classes (the current fix seems to assume that 0 is the lowest precedence, but there is no guarantee of that).

### DIFF
--- a/lib/Value/Real.pm
+++ b/lib/Value/Real.pm
@@ -195,7 +195,7 @@ sub atan2 {
 ##################################################
 
 sub string {
-  my $self = shift; my $equation = shift; my $prec = shift;
+  my $self = shift; my $equation = shift; shift; shift; my $prec = shift;
   my $n = $self->{data}[0];
   my $format = $self->getFlag("format",$equation->{format} ||
 			        ($equation->{context} || $self->context)->{format}{number});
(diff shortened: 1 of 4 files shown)


lib/Value/Real.pm
```diff
@@ -195,7 +195,7 @@ sub atan2 {
 ##################################################
 
 sub string {
-  my $self = shift; my $equation = shift; my $prec = shift;
+  my $self = shift; my $equation = shift; shift; shift; my $prec = shift;
```
taniwallach, line 198: The extra shifts seem to be for "open" and "close" which some such routines need.

pstaabp (approved):
looks good.

drdrew42 (approved):

PR #292, opened 2016-12-31
Tikz -- proof of concept only -- don't pull into develop (yet) on openwebwork

mgage:
This is pull request illustrates proof of concept for rendering tikz code within a WeBWorK project. 

This branch is NOT ready for prime time and should not be pulled into develop(yet). We've place it here so you can take a look at it as it is developed. You can pull it into your local instances of WeBWorK ( split off from the develop branch) and see what you think of it. 

There are two example problems in pg/t/tikz_test that show how it is used (at present).  The user front end will probably change significantly as we get feedback. 

```
$path = tikz_graph(<<END_TIKZ);
\begin{tikzpicture}[main_node/.style={circle,fill=blue!20,draw,minimum size=1em,inner sep=3pt]}] 
\draw (-4,0) -- (4,0);
\draw (0,-2) -- (0,2);
\draw (0,0) circle (1.5);
\draw (0, 1.5) node[anchor=south]{N} -- (2.5,0)node [above]{y};
\draw (1.2,0.9) node[right]{\((\vec x, x_{n})\)};
\end{tikzpicture}
END_TIKZ
```
This is work of Mike Gage and Peter Peluso at U. of Rochester, based on the initial perl script developed by Mathew Vassair working under the direction of Dr. Sternin and Dr. D'Agostino at Brock University.

PR #572, opened 2021-04-28
tikz image package for PreTeXt output

Alex-Jordan:
This makes it such that for PTX display mode (used when producing PreTeXt output) the tikz "image" is actually a .tgz file with .tex, .pdf, .svg, and .png.

To test, check that earlier tikz examples still work. There are sample files in #557.

Next, check PreTeXt output is as described. To do that, you could use something like the following URL. You need to set:
* the server
* courseID should be a course ID 
* sourceFilePath should be a file path to some tikz example file accessible from that course
* userID and course_password should be a user in the course (login proctor or higher), and their password 

https://webwork.yoursite.edu/webwork2/html2xml?&amp;problemSeed=1&amp;answersSubmitted=0&amp;sourceFilePath=local/BEGINTIKZ.pg&amp;displayMode=PTX&amp;courseID=testing&amp;userID=anonymous&amp;course_password=anonymous&amp;outputformat=ptx

Ignore whatever your web browser displays and instead look at the page source. You should see something like:
```
<image source="/wwtmp/testing//images/4c09146e-8bc3-3ae3-aac3-ad011135bebb___ca219939-7303-3697-8505-574f0e733dc4.tgz" />
```
with that .tgz extension. Download that file (so, like from https://webwork.yoursite.edu/wwtmp/testing//images/4c09146e-8bc3-3ae3-aac3-ad011135bebb___ca219939-7303-3697-8505-574f0e733dc4.tgz) and see that it contains what it should contain.

### DIFF
--- a/lib/TikZImage.pm
+++ b/lib/TikZImage.pm
@@ -48,7 +48,7 @@ sub new {
 			if ($field eq 'ext') {
 				my $ext = shift;
 				$data->{ext} = $ext
-				if ($ext && ($ext eq 'png' || $ext eq 'gif' || $ext eq 'svg' || $ext eq 'pdf'));
+				if ($ext && ($ext =~ '^(png|gif|svg|pdf|tgz)$'));
 			}
 			else {
 				$data->{$field} = shift;
@@ -96,8 +96,10 @@ sub addToPreamble {
 	return &$self('addToPreamble', @_);
 }
 
-# Set the image type.  The valid types are 'png', 'gif', 'svg', and 'pdf'.
+# Set the image type.  The valid types are 'png', 'gif', 'svg', 'pdf', and 'tgz'.
 # The 'pdf' option should be set for print.
+# The 'tgz' option should be set when 'PTX' is the display mode.
+# It creates a .tgz file containing .tex, .pdf, .png, and .svg versions of the image
 sub ext {
 	my $self = shift;
 	return &$self('ext', @_);
@@ -166,49 +168,57 @@ sub draw {
 	close $fh;
 
 	my $ext = $self->ext;
-	my $tex_ext = $ext eq 'svg' && $self->svgMethod eq 'dvisvgm' ? 'dvi' : 'pdf';
-	my $latex_binary = WeBWorK::PG::IO::externalCommand($tex_ext eq 'dvi' ? 'latex' : 'pdflatex');
+	my $svgMethod = $self->svgMethod;
+	my $latex     = WeBWorK::PG::IO::externalCommand('latex');
+	my $pdflatex  = WeBWorK::PG::IO::externalCommand('pdflatex');
 
-	# Generate the pdf file.
-	system "cd " . $working_dir . " && $latex_binary image.tex > pdflatex.stdout 2> /dev/null";
+	# Make only the dvi, only the pdf, or both in case we are making tgz with svg via dvisvgm
+	if (($ext eq 'svg' || $ext eq 'tgz') && $svgMethod eq 'dvisvgm') {
+		system "cd $working_dir && $latex image.tex > latex.stdout 2> /dev/null";
+	}
+	if ($ext ne 'svg' || ($ext eq 'svg' && $svgMethod ne 'dvisvgm')) {
+		system "cd $working_dir && $pdflatex image.tex > pdflatex.stdout 2> /dev/null";
+	}
 
-	if (-r "$working_dir/image.$tex_ext") {
-		chmod(0777, "$working_dir/image.$tex_ext");
-		# Convert the file to the appropriate type of image file
-		if ($ext eq 'svg') {
-			if ($self->svgMethod eq 'dvisvgm') {
-				system WeBWorK::PG::IO::externalCommand('dvisvgm') .
-					" $working_dir/image.dvi --no-fonts --output=$working_dir/image.svg > /dev/null 2>&1";
-			} else {
-				system WeBWorK::PG::IO::externalCommand($self->svgMethod) .
-					" $working_dir/image.pdf $working_dir/image.svg > /dev/null 2>&1";
-			}
-		} elsif ($ext ne 'pdf') {
-			system WeBWorK::PG::IO::externalCommand('convert') .
-				join('',map {" -$_ " . $self->convertOptions->{input}->{$_}} (keys %{$self->convertOptions->{input}})) .
-				" $working_dir/image.pdf" .
-				join('',map {" -$_ " . $self->convertOptions->{output}->{$_}} (keys %{$self->convertOptions->{output}})) .
-				" $working_dir/image.$ext > /dev/null 2>&1";
-		}
-
-		if (-r "$working_dir/image.$ext") {
-			# Read the generated image file into memory
-			open(my $in_fh,  "<", "$working_dir/image.$ext")
-				or warn "Failed to open $working_dir/image.$ext for reading.", return;
-			local $/;
-			$data = <$in_fh>;
-            close($in_fh);
-		} else {
-			warn "Convert operation failed.";
+	# Make derivatives of the dvi
+	if (-r "$working_dir/image.dvi") {
+		if (($ext eq 'svg' || $ext eq 'tgz') && $svgMethod eq 'dvisvgm') {
+			$self->use_svgMethod($working_dir, "image.dvi");
 		}
 	} else {
-		warn "File $working_dir/image.$tex_ext was not created.";
-		if (open(my $err_fh, "<", "$working_dir/pdflatex.stdout")) {
-			while (my $error = <$err_fh>) {
-				warn $error;
-			}
-			close($err_fh);
+		warnLaTeXOutput($working_dir, "image.dvi");
+	}
+
+	# Make derivatives of the pdf
+	if (-r "$working_dir/image.pdf") {
+		if (($ext eq 'svg' || $ext eq 'tgz') && $svgMethod ne 'dvisvgm') {
+			$self->use_svgMethod($working_dir, "image.pdf");
 		}
+		if ($ext eq 'tgz') {
+			$self->use_convert($working_dir, "image.pdf", "png");
+		}
+		elsif ($ext ne 'svg') {
+			$self->use_convert($working_dir, "image.pdf", $ext);
+		}
+	} else {
+		warnLaTeXOutput($working_dir, "image.pdf");
+	}
+
+	# Make the tgz
+	if ($ext eq 'tgz') {
+		my $tar = WeBWorK::PG::IO::externalCommand('tar');
+			system "cd $working_dir && $tar -czf image.tgz image.tex image.pdf image.svg image.png > /dev/null 2>&1";
+	}
+
+	# Read the generated image file into memory
+	if (-r "$working_dir/image.$ext") {
+		open(my $in_fh,  "<", "$working_dir/image.$ext")
+			or warn "Failed to open $working_dir/image.$ext for reading.", return;
+		local $/;
+		$data = <$in_fh>;
+		close($in_fh);
+	} else {
+		warn "Image file production failed.";
 	}
 
 	# Delete the files used to generate the image.
@@ -217,6 +227,50 @@ sub draw {
 	}
 
 	return $data;
+}
+
+sub warnLaTeXOutput {
+	my $working_dir = shift;
+	my $file = shift;
+	my $stdout = ($file =~ /dvi$/) ? 'latex.stdout' : 'pdflatex.stdout';
+	warn "File $file was not created.";
+	if (open(my $err_fh, "<", "$working_dir/$stdout")) {
+		while (my $error = <$err_fh>) {
+			warn $error;
+		}
+		close($err_fh);
+	}
+}
+
+sub use_svgMethod {
+	my $self = shift;
+	my $working_dir = shift;
+	my $file = shift;
+	my $svgfile = $file;
+        $svgfile =~ s/\.(dvi|pdf)$/\.svg/;
+	if ($file =~ /\.dvi^/) {
+		system WeBWorK::PG::IO::externalCommand('dvisvgm') .
+			" $working_dir/$file --no-fonts --output=$working_dir/$svgfile > /dev/null 2>&1";
+	} else {
+		system WeBWorK::PG::IO::externalCommand($self->svgMethod) .
+			" $working_dir/$file $working_dir/$svgfile > /dev/null 2>&1";
+	}
+	warn "File $working_dir/$svgfile was not created" unless (-r "$working_dir/$svgfile");
+}
+
+sub use_convert {
+	my $self = shift;
+	my $working_dir = shift;
+	my $file = shift;
+	my $ext = shift;
+	my $outfile = $file;
+	$outfile =~ s/\.pdf$/\.$ext/;
+	system WeBWorK::PG::IO::externalCommand('convert') .
+		join('',map {" -$_ " . $self->convertOptions->{input}->{$_}} (keys %{$self->convertOptions->{input}})) .
+		" $working_dir/$file" .
+		join('',map {" -$_ " . $self->convertOptions->{output}->{$_}} (keys %{$self->convertOptions->{output}})) .
+		" $working_dir/$outfile > /dev/null 2>&1";
+	warn "File $working_dir/$outfile was not created" unless (-r "$working_dir/$outfile");
 }
 
 1;
--- a/lib/PGalias.pm
+++ b/lib/PGalias.pm
@@ -254,19 +254,7 @@ sub make_alias {
 	
 	if ($ext eq 'html' 		    ) {
 	   $adr_output = $self->alias_for_html($aux_file_id,$ext)
-	} elsif (   $ext eq 'gif'  
-		     or $ext eq 'jpg' 
-		     or $ext eq 'png'
-		     or $ext eq 'svg'
-		    or $ext eq 'pdf'
-		    or $ext eq 'mp4'
-		    or $ext eq 'mpg'
-		    or $ext eq 'ogg'
-		    or $ext eq 'webm'
-		    or $ext eq 'css'
-		    or $ext eq 'js'
-		    or $ext eq 'nb'
-		    ) {
+	} elsif ($ext =~ /^(gif|jpg|png|svg|pdf|mp4|mpg|ogg|webm|css|js|nb|tgz)$/) {
 		if ($displayMode =~ /^HTML/ or $displayMode eq 'PTX') {
 			 $adr_output=$self->alias_for_html($aux_file_id, $ext);
 		} elsif ($displayMode eq 'TeX') {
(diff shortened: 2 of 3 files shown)


lib/TikZImage.pm
```diff
@@ -48,7 +48,7 @@ sub new {
 			if ($field eq 'ext') {
 				my $ext = shift;
 				$data->{ext} = $ext
-				if ($ext && ($ext eq 'png' || $ext eq 'gif' || $ext eq 'svg' || $ext eq 'pdf'));
+				if ($ext && ($ext =~ '^(png|gif|svg|pdf|tgz)$'));
```
drgrice1, line 51:
```suggestion
				if ($ext && ($ext =~ /^(png|gif|svg|pdf|tgz)$/));
```
The right hand side of the match operator needs to be a regular expression.

lib/TikZImage.pm
```diff
@@ -166,49 +168,57 @@ sub draw {
...
 		}
 	} else {
-		warn "File $working_dir/image.$tex_ext was not created.";
-		if (open(my $err_fh, "<", "$working_dir/pdflatex.stdout")) {
-			while (my $error = <$err_fh>) {
-				warn $error;
-			}
-			close($err_fh);
+		warnLaTeXOutput($working_dir, "image.dvi");
+	}
```
drgrice1, line 190: The logic is not correct here.  This will add warnings to the output anytime that the dvi file does not exist, even when that file should not exist.  For example, if a png image is requested or if an svg image is requested via pdf2svg.  The inner `if` here should be the outer `if`, and the `else` should belong to the inner `if`.  I.e.,
```suggestion
if (($ext eq 'svg' || $ext eq 'tgz') && $svgMethod eq 'dvisvgm') {
	if (-r "$working_dir/image.dvi") {
			$self->use_svgMethod($working_dir, "image.dvi");
	} else {
		warnLaTeXOutput($working_dir, "image.dvi");
	}
}
```

drgrice1, line 190: The logic is not correct here.  This will add warnings to the output anytime that the dvi file does not exist, even when that file should not exist.  For example, if a png image is requested or if an svg image is requested via pdf2svg.  The inner `if` here should be the outer `if`, and the `else` should belong to the inner `if`.  I.e.,

```suggestion
if (($ext eq 'svg' || $ext eq 'tgz') && $svgMethod eq 'dvisvgm') {
	if (-r "$working_dir/image.dvi") {
			$self->use_svgMethod($working_dir, "image.dvi");
	} else {
		warnLaTeXOutput($working_dir, "image.dvi");
	}
}
```

lib/TikZImage.pm
```diff
@@ -166,49 +168,57 @@ sub draw {
...
 
 	my $ext = $self->ext;
-	my $tex_ext = $ext eq 'svg' && $self->svgMethod eq 'dvisvgm' ? 'dvi' : 'pdf';
-	my $latex_binary = WeBWorK::PG::IO::externalCommand($tex_ext eq 'dvi' ? 'latex' : 'pdflatex');
-
-	# Generate the pdf file.
-	system "cd " . $working_dir . " && $latex_binary image.tex > pdflatex.stdout 2> /dev/null";
-
-	if (-r "$working_dir/image.$tex_ext") {
-		chmod(0777, "$working_dir/image.$tex_ext");
```
drgrice1, line 176: You dropped this in your new code.  This was the fix for the permissions error that @pstaabp was getting on OSX with the pdf2svg method.

drgrice1 (changes requested):
There are some changes needed here.  This breaks `svg` image generation as it currently stands.  There are some other issues as well.  See the comments.

There is another problem when generating the `svg` image with the `ptx` extension.  Note that the `tex` file used with `dvisvgm` is different than the `tex` file used in all other cases.  `dvisvgm` needs `\def\pgfsysdriver{pgfsys-dvisvgm.def}` to work properly.  You might get lucky and it will still work without that depending on what is in the `TikZ` code.  Even if it does work, the generated `svg` may not look as good.  With your current code the `pdfsysdriver` definition is not added with the `ptx` extension, and the same `tex` file is used in all cases.

lib/TikZImage.pm
Alex-Jordan: Elsewhere you mention regular expressions are slow. Should I switch this back to a bunch of `eq` and `||`?

drgrice1: No, this is probably better. Regular expressions are allowed than direct string comparisons, and so if there is a direct string comparison that should be preferred. In this case however, we are comparing a single regular expressions to several string comparisons. This is at least more concise coding wise.

drgrice1: I did a quick timing test on this.  So running
```
my $ext = 'png';
for (0 .. 10000000) {
	my $test = $ext eq 'png' || $ext eq 'gif' || $ext eq 'svg' || $ext eq 'pdf' || $ext eq 'tgz';
}
```
took about 1.2 seconds, while runing
```
my $ext = 'png';
for (0 .. 10000000) {
	my $test = $ext =~ /^(png|gif|svg|pdf|tgz)$/;
}
```
took about 4.1 seconds.

drgrice1: This code is not going to be executed ten million times though.  It will probably occur once or twice per problem.  So these timing considerations aren't that significant.

Alex-Jordan: That's interesting. You made `$ext` be "png" and I wonder if it would affect the timing since "png" is first. In the string comparison, does perl even bother with the subsequent comparisons if the first one works out to true?

drgrice1: Good point.  So if I set `my $ext = 'tgz'` then the string comparison executed ten million times takes about 2.5 seconds, and the regular expression takes about 4.1 seconds.  If I set it to `svg` (the middle value) then the string comparison takes about 1.9 seconds ant the regular expression about 3.9 seconds.  So a value later in the list is slower, but the never quite as slow as the regular expression.

Alex-Jordan:
> With your current code the pdfsysdriver definition is not added with the ptx extension, and the same tex file is used in all cases.

Just to be clear, you mean the `tgz` extension. I decided it was inappropriate to get "ptx" down at this level in `TikzImage.pl`. In `PGtikz.pl`, if the display mode is "PTX" then that makes the extension override to `tgz` just like how tex display mode overrides it to "pdf". And "tgz" actually is a file extension.

Thanks for being on top of this though. I was not aware of this detail with the pdfsysdriver. I'm thinking what to do about it. I see a few obvious ways, but don't like them for various reasons.

lib/TikZImage.pm
drgrice1: Here is a more comprehensive test (full script):
```perl
#!/usr/bin/env perl

use strict;
use warnings;

use Time::HiRes qw/time/;

my $start = time;
for (0 .. 1000000) {
	for my $ext ('png', 'gif', 'svg', 'pdf', 'tgz') {
		my $test = $ext eq 'png' || $ext eq 'gif' || $ext eq 'svg' || $ext eq 'pdf' || $ext eq 'tgz';
	}
}
print "string compare: " . (time - $start) . "\n";

$start = time;
for (0 .. 1000000) {
	for my $ext ('png', 'gif', 'svg', 'pdf', 'tgz') {
		my $test = $ext =~ /^(png|gif|svg|pdf|tgz)$/;
	}
}
print "regex compare: " . (time - $start) . "\n";
```
This outputs something like
```
string compare: 1.15723896026611
regex compare: 2.274906873703
```
Several executions gave similar results.

drgrice1:
Yeah, the `tgz` extension then.  Was it `ptx` before?  Not sure where I got that.  In any case, it is probably better, as you said, to not have `ptx` at this level.  Note it is `pgfsysdriver` (not `pdfsysdriver`).  It is a TikZ and PGF thing.  `dvisvgm` uses a modified backend driver when latex is run so that it can inject postscript specials into the dvi output.  Then when `dvisvgm` uses those specials to improve its images.  I am now rather well versed on this after messing with the TikZJax stuff.  It uses the same technique.  I actually improved on the pgfsysdriver Jim Fowler uses for TikZJax by looking at the dvisvgm version in my PGtikzjax.pl macro.

lib/TikZImage.pm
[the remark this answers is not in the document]
Alex-Jordan: Yeah, I had it the other way around and while I was trying to clean up the code and streamline it, I switched them to use `warnLaTeXOutput` fewer times. But I see now that won't be good. It's reversed in my branch now, soon to be pushed.

lib/TikZImage.pm
```diff
@@ -219,4 +229,48 @@ sub draw {
...
+	}
+}
+
+sub use_svgMethod {
+	my $self = shift;
+	my $working_dir = shift;
+	my $file = shift;
+	my $svgfile = $file;
+        $svgfile =~ s/\.(dvi|pdf)$/\.svg/;
+	if ($file =~ /\.dvi^/) {
```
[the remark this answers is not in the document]
Alex-Jordan: Yikes. It's not even the right regex, `/\.dvi$/`.

lib/TikZImage.pm
```diff
@@ -219,4 +229,48 @@ sub draw {
...
+		close($err_fh);
+	}
+}
+
+sub use_svgMethod {
+	my $self = shift;
+	my $working_dir = shift;
+	my $file = shift;
+	my $svgfile = $file;
+        $svgfile =~ s/\.(dvi|pdf)$/\.svg/;
```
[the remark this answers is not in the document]
Alex-Jordan: Done.

lib/TikZImage.pm
Alex-Jordan: Do you think this is needed for both methods? It seems harmless to do it for both. Just an extra line of code though now that I'm splitting it up.

Alex-Jordan:
> Was it ptx before?

Probably, in the discussion I started several days ago.

Alex-Jordan:
Standy by. I can't do complete `dvisvgm` testing but I realized I could partly do some. And am doing now.

Alex-Jordan:
OK, I think it is ready for another look. On my system, it "works" with dvisvgm as well as it ever did, with the messed up scaling.

lib/TikZImage.pm
```diff
@@ -219,4 +247,47 @@ sub draw {
...
+	my $working_dir = shift;
+	my $file = shift;
+	my $ext = shift;
+	my $outfile = $file =~ s/\.pdf$/\.$ext/r;
+	system WeBWorK::PG::IO::externalCommand('convert') .
+		join('',map {" -$_ " . $self->convertOptions->{input}->{$_}} (keys %{$self->convertOptions->{input}})) .
+		" $working_dir/$file" .
+		join('',map {" -$_ " . $self->convertOptions->{output}->{$_}} (keys %{$self->convertOptions->{output}})) .
+		" $working_dir/$outfile > /dev/null 2>&1";
+	warn "File $working_dir/$outfile was not created" unless (-r "$working_dir/$outfile");
```
drgrice1, line 290: In the `use_svgMethod` and `use_convert` subroutines, there is no need for the `$file` parameters.  The value of `$file` is always going to be `image.$ext`.  Just use "image", which is again just an internal name.  I suggest changing to the following that does the job, conveys the pertinent information if errors occur, and is cleaner.  Of course you also need to adjust the calls to these above.
```suggestion
sub use_svgMethod {
	my $self = shift;
	my $working_dir = shift;
	if ($self->svgMethod eq 'dvisvgm') {
		system WeBWorK::PG::IO::externalCommand('dvisvgm') .
			" $working_dir/imag.dvi --no-fonts --output=$working_dir/image.svg > /dev/null 2>&1";
	} else {
		system WeBWorK::PG::IO::externalCommand($self->svgMethod) .
			" $working_dir/image.pdf $working_dir/image.svg > /dev/null 2>&1";
	}
	warn "Failed to generate svg file." unless -r "$working_dir/image.svg";
}

sub use_convert {
	my $self = shift;
	my $working_dir = shift;
	my $ext = shift;
	system WeBWorK::PG::IO::externalCommand('convert') .
		join('',map {" -$_ " . $self->convertOptions->{input}->{$_}} (keys %{$self->convertOptions->{input}})) .
		" $working_dir/image.pdf" .
		join('',map {" -$_ " . $self->convertOptions->{output}->{$_}} (keys %{$self->convertOptions->{output}})) .
		" $working_dir/image.$ext > /dev/null 2>&1";
	warn "Failed to generate $ext file." unless -r "$working_dir/image.$ext";
}
```

drgrice1 (changes requested):
This is certainly not looking pretty.  I offered some suggestions to clean this up some, but I almost think it would be better to just call the draw method multiple times from special code for handling the `PTX` format in PGtikz.pl.

lib/TikZImage.pm
```diff
@@ -156,60 +157,87 @@ sub draw {
...
+	if (($ext eq 'svg' || $ext eq 'tgz') && $svgMethod eq 'dvisvgm') {
+		open($fh, ">", "$working_dir/image-dvisvgm.tex")
+			or warn "Can't open $working_dir/image-dvisvgm.tex for writing.";
+		my @header = $self->header;
+		splice @header, 1, 0, "\\def\\pgfsysdriver{pgfsys-dvisvgm.def}\n";
+		print $fh @header;
+		print $fh $self->tex =~ s/\\\\/\\/gr . "\n";
+		print $fh $self->footer;
+		close $fh;
+		system "cd $working_dir && $latex image-dvisvgm.tex > latex.stdout 2> /dev/null && mv image-dvisvgm.dvi image.dvi";
```
drgrice1, line 179: This should probably use the `mv` external program via `WeBWorK::PG::IO::externalCommand('mv')`.

drgrice1 (commented):
I just noticed that this now breaks hardcopy generation as the 'pdf' extension is not properly accounted for.

lib/TikZImage.pm
```diff
@@ -156,60 +157,87 @@ sub draw {
...
-			close($err_fh);
+	}
+	if ($ext eq 'tgz') {
+		if (-r "$working_dir/image.pdf") {
+			$self->use_convert($working_dir, "image.pdf", "png");
+		} else {
+			$self->warnLaTeXOutput($working_dir, "image.pdf");
 		}
 	}
+	elsif ($ext ne 'svg') {
```
drgrice1, line 217: This should be
```suggestion
	elsif ($ext ne 'svg' && $ext ne 'pdf') {
```
No need to convert "image.pdf" to "image.pdf".   Furthermore this actually reduces the quality of the pdf image.

Alex-Jordan:
@drgrice1 I appreciate your time reviewing this and I regret that there are so many issues. I wanted to ask at this point, if you'd prefer (a) I continue combing it out or (b) I could revert to how things were before and basically wrap it within the draw subroutine like:

```
if ($ext ne 'tgz')
    {what was there before}
else
    {new block of code that builds everything and puts it into a tgz}
```

There would be code duplication in that new block, which is what motivated me to try this way. But putting it in a new block would insulate what you already worked on so much that was already working.

drgrice1:
Continue on the path we are on.  This will work.

lib/TikZImage.pm
Alex-Jordan: I would commit this directly but it should be paired with the changes in the calls above. So I committed to my local branch.

lib/TikZImage.pm
Alex-Jordan: Oops, I accidentally staged and committed this a few items back.

lib/TikZImage.pm
drgrice1: Yeah, I should have also made the suggestions in the calls above.  Then you could have added them as a batch and committed them.  I was just playing with this suggestion feature that @drdrew42 pointed out.

Alex-Jordan:
@drgrice I believe I addressed your comments with the last four commits.

drgrice1 (approved):
This looks good.  You have addressed my primary concerns and everything works.

The only thing I don't like is the `if (-r "$working_dir/image.pdf") convert else warn` repetition.  Note that when generating the `tgz` output you will get the warnings for the same latex run twice when an error occurs if you are using pdf2svg.

I noticed there is another `chmod` that you removed.  Before this pull request the tex file permissions were also set to "0777".  That doesn't seem to be needed for me on linux.  I checked and that was something that @mgage had in his original TikZ pull request (#292).  That may be another OSX permissions thing.

@pstaab:  Could you take a look at this?  See if it still works on OSX.

lib/TikZImage.pm
```diff
@@ -219,4 +248,42 @@ sub draw {
 	return $data;
```
drgrice1, line 248: There is a minor thing here also.  This is my mistake that I am going to ask you to fix.  This should probably use the external rm program (also defined in site.conf).

drgrice1: I think all of these external programs (mv, cp, rm, mkdir, tar, and gzip) are here for non-linux operating system usage.  All of these are always available on linux and always in the path.  So there is no need to use the full path.

Alex-Jordan:
> The only thing I don't like is the if (-r "$working_dir/image.pdf") convert else warn repetition.

I think I took care of that in an upcoming commit.

> I noticed there is another chmod that you removed.

I'll put it back, because it seems harmless. I did not intentionally remove it. My eyes blurred and I did not see the difference, and I convinced myself an extra copy had snuck in through copy/paste or shuffling lines of code around.

I changed the `rm`.

Alex-Jordan:
I just pushed those things I mentioned in my last post. I noticed I had a typo in the commit message right before these and I couldn't help but correct it with `git commit --amend` before I did these newer commits. So I had to force push, and you (or anyone may need to trash the branch and get it fresh.

Alex-Jordan:
@pstaabp I'm seconding the call if you wouldn't mind testing this again. I made even more edits after @drgrice1 's last post, including bringing back the `chmod` line. But it would still be good to hear that nothing is broken on your system.

drgrice1 (approved):
Yeah, this is better.

I am not sure what is actually needed of the `chmod` calls, but we will see what @pstaabp's tests show.

drgrice1:
> I just pushed those things I mentioned in my last post. I noticed I had a typo in the commit message right before these and I couldn't help but correct it with `git commit --amend` before I did these newer commits. So I had to force push, and you (or anyone may need to trash the branch and get it fresh.

Yeah, I do the same thing for quick minor tweaks.  I don't mind this on pull requests.  I usually delete branches between testing in any case.

pstaabp (approved):
It seems to be working on my system.

pstaabp:
Hold on.  When I switch back to pdf2svg method, then I get the following:
```
94eabc5b___73d89ae3-bf01-3da5-94cc-7557c35d815d.svg<BR> at XXX/pg/lib/PGcore.pm line 718.
Failed to generate svg file. at XXX/pg/lib/TikZImage.pm line 269.
Failed to generate svg file. at XXX/pg/lib/TikZImage.pm line 269.
Image file production failed. at XXX/pg/lib/TikZImage.pm line 248.
```

pstaabp:
Finally got this working with `pdf2svg`.  Settings on my end.   I now doubly approve this and will go ahead and pull it in.

PR #573, opened 2021-05-13
Switch from using a jquery-ui accordion to using a bootstrap accordion for scaffolds.

drgrice1:
Also the javascript and css are moved out of the macro and into separate files (in webwork2) loaded via ADD_JS_FILE and ADD_CSS_FILE.

See the corresponding webwork2 pull request openwebwork/webwork2#1361

Note that there were two motivating factors for this pull request.  The first is that since the update to the latest version of jquery-ui, the arrows on the scaffold headers have problems.  When the arrow points down it turns white and is very hard to see.  The other is part of the goal to remove all jquery-ui usage in webwork.

### DIFF
--- a/macros/scaffold.pl
+++ b/macros/scaffold.pl
@@ -1,7 +1,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright &copy; 20014 The WeBWorK Project, http://openwebwork.sf.net/
-# $$
+# Copyright &copy; 2000-2020 The WeBWorK Project, http://openwebwork.sf.net/
 # 
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of either: (a) the GNU General Public License as published by the
@@ -263,7 +262,11 @@ within that section.
 
 =cut
 
-sub _scaffold_init {};   # don't reload this file
+sub _scaffold_init {
+	# Load style and javascript for opening and closing the scaffolds.
+	ADD_CSS_FILE("js/apps/Scaffold/scaffold.css");
+	ADD_JS_FILE("js/apps/Scaffold/scaffold.js");
+};
 
 #
 #  The Scaffoling package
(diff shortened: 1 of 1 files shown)


drgrice1:
Also note that this can be done with less code and more cleanly with bootstrap version 5 (which is now officially released).  In fact, I have borrowed some of the bootstrap 5 styling to make this look better than bootstrap 2.3.2 does by default.

macros/scaffold.pl
```diff
@@ -321,7 +324,7 @@ sub End {
   Scaffold->Error("Scaffold ended with section was still open") if $self->{current_section};
   my $self = $scaffold;
   push(@{$self->{output}},splice(@$PG_OUTPUT,0));                      # collect any final non-section output
-  $self->open_sections(@{$self->{open}});                              # make the open sections be displayed
+  $self->hide_other_results(@{$self->{open}});                         # hide results of unnopened sections in the results table
```
Alex-Jordan, line 327:
```suggestion
  $self->hide_other_results(@{$self->{open}});                         # hide results of unopened sections in the results table
```

Alex-Jordan: Trying Andrew's GitHub tip.

Alex-Jordan: I have a two-part exercise (three answer blanks, then two). If I enter some answers in the first three blanks, even one of them incorrect, then when I click submit, the results table still has five rows. The bottom two rows are grayed out. This is all while logged in as a student. Is this what I should be seeing? The wording here made me think I should only see three rows.

Screen Shot 2021-05-13 at 2 42 27 PM

Alex-Jordan:
This is working fine for me. I just had the observation and question about rows in the results table.

drgrice1:
That is the same behavior as before and that is the wording that was in there before.  It doesn't really hide the rows though, it dims them (sets the opacity to 0.5).

drgrice1:
There is a comment about hiding the rows instead by using display "none".  Se the comments on lines 439 - 443 of the current file (lines 431-435 in this pull request).

Alex-Jordan (approved):

pstaabp (approved):
looks good.

PR #576, opened 2021-05-21
Add a third optional argument to ADD_JS_FILE to add script tag attributes (defer, async, etc)

drgrice1:
Add a third optional argument to ADD_JS_FILE.  This argument is a hash reference of attribute value pairs.  These attributes will be added to the script tag that is added to the page.  The code for adding these attributes to the html that is output is added in a corresponding pull request to webwork2 (see openwebwork/webwork2#1374).

The usage of this argument is documented in the POD for the method.  Its usage is also demonstrated in the paserGraphTool.pl macro.

Also, fix a bug that prevented the optional second argument of ADD_CSS_FILE from working.

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -190,7 +190,7 @@ For example:
 =cut
 
 sub ADD_CSS_FILE {
-  my ($file, $external) = shift;
+  my ($file, $external) = @_;
   push(@{$PG->{flags}{extra_css_files}}, { file => $file, external => $external });
 }
 
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -203,16 +203,21 @@ from the webwork2/htdocs/ directory or from an external location.
 
 If external is 1, it is assumed the full url is provided.  If external is 0 or
 not given, then file name will be prefixed with the webwork2/htdocs/ directory.
+
+Addtional attributes can be passed as a hash reference in the optional third
+argument.  These attributes will be added as attributes to the script tag.
+
 For example:
 
 	ADD_JS_FILE("js/apps/Base64/Base64.js");
 	ADD_JS_FILE("//web.geogebra.org/4.4/web/web.nocache.js", 1);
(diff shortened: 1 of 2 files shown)


pstaabp:
This looks like this is mainly for `parserGraphTool`.  I can test some problems with this, but what should I look for?

drgrice1:
Actually the graph tool macro doesn't really need it. It can use it though. I just used that as an example.

The point of this pull request is really about future proofing by adding versatility.

pstaabp:
Reading through the docs and code, it seems that `JSXGraphOptions` field of the GraphTool uses it.  I got the bounding box and ticks changed via this field.  



> Actually the graph tool macro doesn't really need it. It can use it though. I just used that as an example.
> 
> The point of this pull request is really about future proofing by adding versatility.

Is there anything using the 3rd argument.  If not and just future-proofing, I feel safe to pull in.

drgrice1:
The only thing to test is that problems using the macros that use the ADD_JS_FILE method still work.  Those are graph tool problems and Geogebra applet problems mainly at this point.

pstaabp:
The graphTool problems work fine.  Looks like the geogebra problems that I found are java applets.  Are we still supporting those? 

I found some CUNY geogebra problems and those look good.

pstaabp:
I think we can merge unless we want another set of eyes.

macros/PG.pl
```diff
@@ -203,16 +203,21 @@ =head4 ADD_JS_FILE
 
 If external is 1, it is assumed the full url is provided.  If external is 0 or
 not given, then file name will be prefixed with the webwork2/htdocs/ directory.
+
+Addtional attributes can be passed as a hash reference in the optional third
```
taniwallach, line 207: typo "Addtional" => "Additional"

taniwallach (approved):
I did not have time to test, but it looks good and @pstaab reported testing.

macros/PG.pl
drgrice1: Fixed.

PR #579, opened 2021-05-28
scaffold section numbering

Alex-Jordan:
This adds an option for each `Scaffold()` to have numbering. There is a companion PR for styling at https://github.com/openwebwork/webwork2/pull/1383.

In addition to the `numbered => 1` option, I tweaked styling for nested scaffolds. In LaTeX, they now indent 1em for each level of depth. And in HTML, they no longer contract in width from the right, only from the left.

Here is a simple test file demonstrating sections with depth, and the various numbering and title combinations.

```
DOCUMENT();

loadMacros("PGstandard.pl","scaffold.pl");

Scaffold::Begin(numbered=>1);
  Section::Begin("The outer scaffold is numbered");
    Scaffold::Begin(numbered=>1);
        Section::Begin("The next scaffold is numbered");
        Section::End();
        Section::Begin("Its first and second sections have titles, but not the third");
        Section::End();
        Section::Begin();
            Scaffold::Begin();
                Section::Begin("This is an unnumbered scaffold");
                Section::End();
                Section::Begin("Its first and second sections have titles, but not the third");
                Section::End();
                Section::Begin();
                Section::End();
            Scaffold::End();
        Section::End();
    Scaffold::End();
  Section::End();
  Section::Begin("Here is the last section of the outer scaffold");
  Section::End();
Scaffold::End();

ENDDOCUMENT();
```

### DIFF
--- a/macros/scaffold.pl
+++ b/macros/scaffold.pl
@@ -186,6 +186,15 @@ the first section, you might want the first section to be closed, and
 have the student open it by hand before anwering the questions.  In
 this case, set this value to 0 (it is 1 by default).
 
+=item C<S<< numbered => 0 or 1 >>>
+
+This determines whether each section is automatically numbered before
+its title. If true, each section title will be preceded by a number
+and a period. The section's nesting level determines the style of
+numbering: a, i, A. Any deeper and numbering is just arabic.
+If there is no title, a default title like "Part 1:" is used, and
+in that case no extra numbering is added regardless of this option.
+
 =back
 
 Some useful configurations are:
@@ -284,6 +293,7 @@ our $afterAnswerDate = (time() > $main::envir{answerDate});
 our $scaffold;           # the active scaffold (set by Begin() below)
 my  @scaffolds = ();     # array of nested scaffolds
 my  $scaffold_no = 0;    # each scaffold gets a unique number
+my  $scaffold_depth = 1; # each scaffold has a nesting depth
 
 our $PG_ANSWERS_HASH = $main::PG->{PG_ANSWERS_HASH};  # where PG stores answer evaluators
 our $PG_OUTPUT = $main::PG->{OUTPUT_ARRAY};           # where PG stores the TEXT() output
@@ -305,7 +315,7 @@ our $PREFIX = "$main::envir{QUIZ_PREFIX}Prob-$main::envir{questionNumber}";
 #
 sub Begin {
   my $self = Scaffold->new(@_);
-  unshift(@scaffolds,$self); $scaffold = $self;
+  unshift(@scaffolds,$self); $scaffold = $self; $scaffold_depth++;
   $self->{previous_output} = [splice(@{$PG_OUTPUT},0)];  # get output and clear it without changing the array pointer
   $self->{output} = [];                                  # the contents of the scaffold
   return $self;
@@ -328,6 +338,7 @@ sub End {
   push(@$PG_OUTPUT,@{$self->{previous_output}},@{$self->{output}});    # put back original output and scaffold output
   delete $self->{previous_output}; delete $self->{output};             # don't need these any more
   shift(@scaffolds); $scaffold = $scaffolds[0];
+  $scaffold_depth--;
   return $scaffold;
 }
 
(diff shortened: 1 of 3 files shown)


Alex-Jordan:
I'm not sure what I was thinking, but obviously adding to and subtracting from `\parindent` does not really indent the levels in hardcopy output. I'll think about doing that better.

dpvc:
> adding to and subtracting from `\parindent` does not really indent the levels in hardcopy output

Do mean to adjust `\leftskip` instead?

Alex-Jordan:
@dpvc That's where I landed when I went back to look at alternatives last night. I'm still looking at if that could cause unforeseen consequences depending on what is inside each section. Probably not, at least not that I can think of so far. The other thing I was consider is using the changepage package.

Alex-Jordan:
With the 2.16 release candidate announced, is it still OK to direct this PR (and others) to the 2.16 branch? I'm sorry I haven't been able to attend the release team meetings. This coming week is my last week of class for the term, and I'll be up for most meetings after that.

drgrice1:
It is still okay to target small changes to the 2.16 branch.  Bigger changes will probably be deferred.  It doesn't hurt to target the branch though.  We can always change it and say it needs to wait.

Alex-Jordan:
I made it add/subtract to/from `\leftskip`, as @dpvc suggested. I chose 15pt since that is the default `\parindent` for 10pt documents, and maybe it would be nice for it to indent that way to align with outer paragraphs that have opening indentation.

Alex-Jordan:
I added a commit here that is PreTeXt-specific. We will be retiring the "stage" element in favor of the "task" element for subdividing an exercise, and that might as well happen now on the WeBWorK side.

Alex-Jordan:
This is ready for a review now alongside https://github.com/openwebwork/webwork2/pull/1383.

macros/scaffold.pl
```diff
@@ -495,7 +508,28 @@ sub Begin {
...
+	my @romanatom = (['i','v','x'],['x','l','c'],['c','d','m']);
+	my @romanmolecule = map{['',$_->[0],$_->[0]x2,$_->[0]x3,$_->[0].$_->[1],$_->[1],$_->[1].$_->[0],$_->[1].$_->[0]x2,$_->[1].$_->[0]x3,$_->[0].$_->[2]]}(@romanatom);
+	my @roman;
+	for my $i (@{$romanmolecule[2]}) {
+		for my $j (@{$romanmolecule[1]}) {
+			for my $k (@{$romanmolecule[0]}) {
+				push(@roman,$i.$j.$k);
+			}
+		}
+	}
```
drgrice1, line 525: Do you really anticipate having thousands of scaffolds nested at depth 2?  This code seems excessive.

drgrice1 (approved):
This looks okay.  I don't think that nesting accordions is that great of an idea in general.  I also am not fond of the numbering system implemented (a.i.A, a.i.B, a.ii.A, etc).

Since the defaults are maintained without the new option, I approve.

macros/scaffold.pl
Alex-Jordan: Since I went `a..zz` at level 1, this was for consistency to use about the same order of magnitude. 26^2 compared to 999. No, no one will ever need that many.

Alex-Jordan:
> I also am not fond of the numbering system implemented (a.i.A, a.i.B, a.ii.A, etc).

This is the default enumeration scheme with LaTeX. Well, it's really (a), then i., then A. Starting at depth 2 since the problem number itself is depth 1.

Alex-Jordan:
>  I don't think that nesting accordions is that great of an idea in general.

I agree and I have no plans personally to use this, but I need to make it work in a nested manner with enumeration for PreTeXt's purposes. There, an `exercise` can have `task` children, which can nest 3 deep within the exercise. It feels like every time there is something silly and extreme that I leave unimplemented, it's only a matter of time before a PreTeXt user comes asking about it in the forum there.

drgrice1:
I see.  In the context of a problem I prefer to use upper case alphabetic characters for parts, and never nest past the first level.  I try to avoid roman numerals because students can't count past three in that system (if that).

In any case, I think this pull request (and the paired webwork2 PR) is good.

It would be good if someone else took a look though.

macros/scaffold.pl
Alex-Jordan: Now I do wonder if it's worth producing arrays that are 26^2, 999,and 26^2 elements long, just to pick one entry from them.  I think I will add a conditional to check if the number is "small" first (which it will almost always be) and do something direct in that case.

Alex-Jordan:
I changed this to be less excessive. It now only supports numbers up to 99 at each depth (up to cu, xc, CU). And unlike before, if you exceed the maximum that it supports, you still get good numbers but they are left as arabic.

Alex-Jordan:
This pull request (and the accompanying https://github.com/openwebwork/webwork2/pull/1383) is holding up some PreTeXt development. Assuming @drgrice1 and I approve it, could I ask for one more review to meet our three-approval standard?

pstaabp (approved):
Looks good.

PR #580, opened 2021-06-06
Properly handle negative constants and variables.

dpvc:
This PR resolves the issue raised in the [WeBWorK forum](https://webwork.maa.org/moodle/mod/forum/discuss.php?d=5581) where the LimitedPolynomial context didn't rule out certain expressions that should not be allowed.  These turned out to be cause by not handling negation properly.

To test, use

```
loadMacros("contextLimitedPolynomial.pl");
Context("LimitedPolynimal-Strict");
Compute("-2 + x + 4");
Compute("-x + x");
```

Both of the `Compute()` statements should produce error messages (without the patch, they will produce formula objects.

### DIFF
--- a/macros/contextLimitedPolynomial.pl
+++ b/macros/contextLimitedPolynomial.pl
@@ -68,7 +68,7 @@ sub markPowers {
     $self->{index} = $vIndex->{$self->{name}};
     $self->{exponents} = [(0) x scalar(keys %{$vIndex})];
     $self->{exponents}[$self->{index}] = 1;
-  } elsif ($self->class eq 'Number') {
+  } elsif ($self->class eq 'Number' || ($self->isNeg && isConstant($self))) {
     my $vIndex = LimitedPolynomial::getVarIndex($self);
     $self->{exponents} = [(0) x scalar(keys %{$vIndex})];
   }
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
@drgrice1 Should this be retargeted to 2.16?

Alex-Jordan (approved):
Tested with Davide's suggestion, and I see the bug before this commit. After this commit, the bug is gone.

drgrice1:
I am getting `Unknown context 'LimitedPolynimal-Strict'` with your test code!

Other than that everything works as claimed.

I think this should be targeted for 2.16.

dpvc:
Sorry, should be `LimitedPolynomial-strict`.  (Lower case "strict")

drgrice1:
@dpvc:  Yeah, I know.  I guess the exclamation mark wasn't enough to indicate that I was kidding.

Thanks for the fix.

I am going to re-target this for PG 2.16 and merge it if that is okay.  It will also be added to develop of course.

dpvc:
OK, sounds good.

drgrice1:
... or nevermind on the re-target.  I will merge it, and then add it to PG 2.16.  It seems this branch doesn't want to re-target cleanly.

PR #582, opened 2021-06-15
Update the url for the geogebra javascript used by the AppletObjects.pl macro

drgrice1:
The web.geogebra.org/4.4/web/web.nocache.js link is no longer valid.

www.geogebra.org/apps/latest/web/web.nocache.js is what will need to be used.


drdrew42 (approved):

drdrew42:
No reason to wait, merging.

drgrice1:
It looks like we are going to have to tell users of previous webwork versions how to fix this for those versions.

PR #584, opened 2021-06-16
Feature/update geogebra api

drdrew42:
Really not that bad after all, it's just become necessary to pull the id, height, and width into the params object for GGBs.

Testing on existing GGB problems seems to have no issues, otherwise, I'd have suggested this for develop/next-pg-release.

### DIFF
--- a/lib/Applet.pm
+++ b/lib/Applet.pm
@@ -71,6 +71,7 @@ The module Applet stores common code for the different types of applets.
 package Applet;
 use URI::Escape;
 use MIME::Base64 qw(encode_base64 decode_base64);
+use JSON qw(encode_json);
 use PGcore;
 @ISA = qw(PGcore);
 
@@ -729,7 +730,7 @@ sub insertObject {
 
 	my $javaParameters = '';
 	my $flashParameters = '';
-	my $webgeogebraParameters = '';
+        my $webgeogebraParameters = qq!<div id = "$appletName"></div>\n<script>\nvar params = !;
 
 	if (PGUtil::not_null($self->{parameter_string})) {
 		$javaParameters = $self->{parameter_string};
@@ -741,14 +742,19 @@ sub insertObject {
 		foreach my $key (keys %param_hash) {
 			$javaParameters .= qq!<param name="$key" value="$param_hash{$key}">\n!;
 			$flashParameters .= uri_escape($key) . '=' . uri_escape($param_hash{$key}) . '&';
-			$webgeogebraParameters .= qq!data-param-$key = "$param_hash{$key}"\n!;
 		}
 		$flashParameters =~ s/\&$//; # trim last &
-		$webgeogebraParameters = qq!<article class="geogebraweb"
-			data-param-id     = "$appletName"
-			data-param-width  = "$width"
-			data-param-height = "$height"
-			!. $webgeogebraParameters . qq!\n></article>!;
+                $param_hash{id}     = $appletName;
+                $param_hash{height} = $height;
+                $param_hash{width}  = $width;
+                $webgeogebraParameters .= encode_json(\%param_hash) . ';';
+                $webgeogebraParameters .= qq!
+                    var $appletName = new GGBApplet(params, true);
+                    window.addEventListener('load', function() {
+                        $appletName.setHTML5Codebase('https://geogebra.org/apps/latest/web3d/');
+                        $appletName.inject('$appletName');
+                    });
+                </script>!;
 	}
 
 	my $objectText = $self->{objectText};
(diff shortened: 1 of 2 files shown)


drgrice1:
That is all?  And all of the current javascript that we have works with it?  I was expecting to need to change that more.

drdrew42:
I know! Once I got it working and then stripped out all the work-arounds, I ended up with very minimal changes after all.

drdrew42:
Oh, I didn't check this with gateways and multiple GGB per page... BRB

drdrew42:
okay, so this seems to work but the shift in applet naming convention now requires the applet `id` to be a valid javascript variable name.

Before these changes, did gateway quizzes support multiple ggb problems if any pair happened to collide with respect to applet IDs?

drgrice1:
No. Applets with the same ID won't work on the same page. I don't think there is any way around that.

drdrew42:
I'm going to see if I can get the problemUUID incorporated into this structure... IDK if this limitation is a problem for anyone, but it'd be nice to have it on automatic, just in case.

lib/Applet.pm
```diff
@@ -741,14 +742,19 @@ sub insertObject {
...
+		$param_hash{height} = $height;
+		$param_hash{width}  = $width;
+		$webgeogebraParameters .= encode_json(\%param_hash) . ';';
+		$webgeogebraParameters .= qq!
+			var $appletName = new GGBApplet(params, true);
+			window.addEventListener('load', function() {
+				$appletName.setHTML5Codebase('https://geogebra.org/apps/latest/web3d/');
+				$appletName.inject('$appletName');
+			});
+		</script>!;
```
drgrice1, line 757: I don't like the global variable that is created.  Nor do I like the inline javascript.  This could certainly be done in the ww_applet_support.js script instead, and only pass data attributes like before.

To do this, change the `article` to a `div`, add the `id` (as you have above), and remove `param-` from the added data attributes.   Then add the following to line 627 of `webwork2/htdocs/js/apps/AppletSupport/ww_applet_support.js`:
```javascript
if (ww_applet_list[appletName].type == 'geogebraweb') {
	var container = document.getElementById(appletName);
	var ggbApplet = new GGBApplet(Object.assign({}, container.dataset), true);
	ggbApplet.setHTML5Codebase('https://geogebra.org/apps/latest/web3d/');
	ggbApplet.inject(appletName);
}
```

The attached zip file contains three patches that demonstrate all of this.  The cleaner-pg-2.16 patch applies to the PG-2.16 branch (instead of this pull request), and is included to demonstrate directly what is changed from there.  The cleaner-update-geogebra-api.patch applies to this pull request branch and does the same thing.  The webwork2-update-geogebra-applet-js.patch applies to the WeBWorK-2.16 branch of the webwork2 code and makes the suggested javascript changes.

[patches.zip](https://github.com/openwebwork/pg/files/6670440/patches.zip)

Really, what needs to be done is for the ww_applet_support.js script to be completely rewritten and remove the flash and java support, and clean it up to the barebones javascript needed for Geogebra applets with no globals.

drgrice1 (commented):
This works, but I think there are better ways to do this.  See my code comments.

drgrice1:
Another thing that we need to do is to set up a way to have pg javascript (and css) in the pg code instead of the webwork2 (or whatever frontend) code.  That removes the need for pull requests to both pg and webwork2.  Furthermore this javascript will have to be duplicated to every other front end that is developed.  That is a major pain.

lib/Applet.pm
drdrew42: That's a really neat trick with `dataset`!

mgage:
we should add a directory  js
so the top level has pg/lib, pg/macros and pg/js  (perhaps pg/html as well?)
pg/js (and pg/html) will be publicly accessible from the web.  pg/lib and pg/macros will not be.

drdrew42:
> Another thing that we need to do is to set up a way to have pg javascript (and css) in the pg code instead of the webwork2 (or whatever frontend) code. That removes the need for pull requests to both pg and webwork2. Furthermore this javascript will have to be duplicated to every other front end that is developed. That is a major pain.

Yeah, I was trying to avoid this... but also, agreed, inline javascript isn't desirable either.

lib/Applet.pm
drgrice1: I guess it is a bit silly to complain about inline javascript and global variables here, when a few lines below in Applets.pm is another instance of inline javascript and a global variable.

drdrew42: it seems that even with this strategy, the applet object is still available as a global. Not sure how that's happening...

PR #586, opened 2021-06-25
adjust newlines in PTX output

Alex-Jordan:
This is a trivial adjustment of newlines in PTX output.

PTX output is XML. Someday I will change the `ptx` format to handle all formatting of the XML. For now, the XML is (sort of) pretty-printed using these manual newlines. It's not great. But this change is an improvement.

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -1320,7 +1320,7 @@ sub statement {
 }
 
 sub STATEMENT {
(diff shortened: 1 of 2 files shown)


taniwallach (approved):
This modifies 2 newlines for PTX only.
Approving and merging based on @Alex-Jordan's need.

drdrew42 (approved):

PR #587, opened 2021-07-01
Make PTX multiple choice answers show their name.

Alex-Jordan:
This makes PreTeXt static versions of popup and radio buttons answers have the answer name as an attribute of the element surrounding the choices. This is needed to associate the correct answer from the answer hash with the right element.


drgrice1:
This looks good. @Alex-Jordan  knows better than anyone what is needed here. I think this can be merged as it only affects pretext output.

drgrice1 (approved):

PR #595, opened 2021-07-20
Many updates to docker files and tests

pstaabp:


pstaabp:
This is a squash and merge for a number of commits from the day.

PR #597, opened 2021-07-27
Splitting webwork2 code from PG

pstaabp:
This PR is designed to remove all webwork2 code from the PG side.   You need openwebwork/webwork2#1463 as well. 

There are two main parts to this

1. WWSafe.pm (from webwork2) is moved to PG_ROOT/lib. 
2. A PGEnvironment module is created to mimic some of the functionality of CourseEnvironment.  

This will also better testing of PG code without needing any webwork2 modules. 

To Test if works with current webwork2:

1. check out this PR and openwebwork/webwork2#1463
2. The top of `WEBWORK_ROOT/conf/webwork.apache2.4.config` needs to be updated.  `pg_dir` is needed to be defined here and in `use lib ...`  See the .dist file for the changes. 
3. restart apache.

Test some problems including those with  tikz and other plotting that create images.  These use the `IO.pm` file, which has been rewritten.  

I'll post some other documentation here about testing some pg code that demonstrates the split from webwork2.


conf/pg_defaults.yml.dist
```diff
@@ -0,0 +1,18 @@
+# This file is loaded if PG is loaded without a webwork2 lib
+#
+# The following are configuration options that are needed by PG
+# and are a slimmed down list from those of WEBWORK_ROOT/conf/defaults.config
+
+options:
+  webworkDirs:
+    tmp: /tmp
+  externalPrograms:
```
taniwallach, line 9: `externalPrograms` is originally from `site.conf` (`site.conf.dist`) if I am not mistaken.

Also - at present some command line options were being provided as part of the settings there, at least `--no-shell-escape` for various latex varieties. As those switched are important, they should be included here in some manner.

pstaabp: @taniwallach the purpose of this is to allow PG to run without any of the webwork configuration files.  The external programs are needed, so they are put in this file.  

Note: if there is the presence of ww2, then the config files are loaded as usually.  

I'll look at the `--no-shell-escape` option.  For what I was trying to do, I do think they were needed.

drgrice1: I think that @taniwallach is pointing out that your comment says this is a slimmed down list from `WEBWORK_ROOT/conf/defaults.config`, and yet the only thing here from that file is webworkDirs{tmp} and everything else is actually from `site.conf`.

There is a lot more from `defaults.conf` that will probably be needed.  Most likely many (if not all) of the `$pg{specialEnvironmentVars} will be needed.  One in particular is the `$pg{specialEnvironmentVars}{entryAssist}`.  For that see the discussion at https://github.com/drdrew42/renderer/pull/83#issuecomment-970584673.  This is where the server default setting should go.

drgrice1 (commented):
This looks like a start.  I put a pull request to this that has some suggested clean up.

We will need to integrate this into the renderer and we will see what changes are needed as we go.

As to working for webwork2, that should be good to go.

conf/pg_defaults.yml.dist
pstaabp: I agree that other fields will need to go into this PG environment.  I initially built this to get some testing running that was independent of webwork2.  I didn't test everything, but more importantly set up a structure to move forward.

pstaabp:
@drgrice1 I don't see a PR for this branch.  Did you submit it?

drgrice1:
It doesn't look like I did.  I added it now.  I left `lib/chromatic/color.c` the way you have it, but why did you change that file?  It may have been wise not to do code clean up there in this pull request so as not to detract from the real point of this pull request.

pstaabp:
Whenever `color.c` compiles on my system I have tons of warnings.  I think all I did was rearrange the subroutines so they were defined before called, which took care of many of the warnings.  It didn't have anything to do with this PR, but I thought it was a quick fix.  Fine if you want to switch it back.

pstaabp:
@drgrice1 since we will start using `perltidy` with this repository, should we add the `.perltidyrc` to this PR so we are using the same rules?

drgrice1:
On my system I also have several warnings when `color.c` compiles.  Six of them to be precise.  Your changes remove one warning.  The warning "implicit declaration of function ‘print_colors’".  The other five are all about implicit int return types.  I don't think that it matters much.  These are only used for the CAPA stuff, and I don't think those are used much anymore.

I would also like to add the `.perltidyrc` file.  However, actually running that on the PG code base would rather dramatically change everything.  I guess we could just add the config file, but not run it on the code base.  Then we could start using it on new files, and perhaps a few files at a time on the rest of the code base.

pstaabp:
We should run perltidy on the `t` directory, most of which is new.  If you're okay with me adding it and running on `t` before merging this, I can do that.

drgrice1:
Sure. Sounds good.

pstaabp:
I realize all the tests are in another branch (`unit-tests-no-ww`) that depends on this.  I'll hold these additions for that branch.

Issue #598, opened 2021-08-05
PG-2.16:  Error Trig function missing input if in the form cos^5(x)


drgrice1:
See https://webwork.maa.org/moodle/mod/forum/discuss.php?d=5641.

drgrice1:
After a little investigation I see that this is not the trig function exponent form as originally reported.  It is the space between the exponent and the function input that is the issue.

dlglin:
Is the issue with the space new in 2.16?

drgrice1:
I think so.

dpvc:
I've make a hot fix branch that resolves the issue.

taniwallach:
Fix was merged.

PR #599, opened 2021-08-10
Resolve problem with space after function to a power. (#598)

dpvc:
This PR fixes an error introduced in 15ce46c that causes spaces after functions with powers to be misinterpreted.

You can test is via

```
TEXT(Compute("sin^2 (x)"))
```

which will throw an error in version 2.16 without the patch, and will produce `[sin(x)]^2` with the patch.

This resolves issue #598.

### DIFF
--- a/lib/Parser.pm
+++ b/lib/Parser.pm
@@ -266,7 +266,7 @@ sub pushBlankOperand {
 #    Otherwise, report the missing operand for this operator
 #
 sub Op {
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This fixes the problem for me.  Thanks for the fix.

I will merge this soon unless someone objects.

Alex-Jordan (approved):

pstaabp (approved):
good.

PR #602, opened 2021-08-31
Updated Implementation of Drag-And-Drop Type Macros

pschan-gh:
Paired with openwebwork/webwork2#1499

Resubmitting a previous pull request:

This pull request hopes to make the draggableProof.pl macro work better with quizzes by making use of NEW_HIDDEN_ANS_NAME and ADD_JS_FILE, and to implement a backend Drag-N-Drop module which could serve other types of drag-and-drop macros.

A draggableSubset.pl is introduced to facilitate writing problems which ask students to partition a given set into subsets.

### DIFF
--- a/lib/DragNDrop.pm
+++ b/lib/DragNDrop.pm
@@ -1,6 +1,6 @@
 ################################################################################
 # WeBWorK Online Homework Delivery System
-# Copyright &copy; 2000-2021 The WeBWorK Project, http://openwebwork.sf.net/
+# Copyright &copy; 2000-2021 The WeBWorK Project, https://github.com/openwebwork
 #
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of either: (a) the GNU General Public License as published by the
@@ -51,13 +51,11 @@ Example:
 
 It is imperative that square brackets be used.
 
-##############################################################################
 OPTIONAL:
 
  DragNDrop($answerInputId, $aggregateList, AllowNewBuckets => 1);
 
 allows student to create new buckets by clicking on a button.
-##############################################################################
 
 To add a bucket to an existing pool $bucket_pool, do:
 
@@ -80,7 +78,6 @@ then:
 
 An empty array reference, e.g. $bucket_pool->addBucket([]), gives an empty bucket.
 
-###############################################################################
 OPTIONAL:
 
  $bucket_pool->addBucket($indices, label => 'Barrel', removable => 1)
@@ -89,7 +86,6 @@ puts the label 'Barrel' at the top of the bucket.
 With the removable option set to 1, the bucket may be removed by the student via the click of a "Remove" button
 at the bottom of the bucket.
 (The first created bucket may never be removed.)
-###############################################################################
 
 To output the bucket pool to HTML, call:
 
@@ -105,15 +101,13 @@ See draggableProof.pl and draggableSubsets.pl
 
 =cut
 
-###############################################################################
-
 use strict;
 use warnings;
 
 package DragNDrop;
 
 sub new {
-	my $self = shift;
+	my $self  = shift;
 	my $class = ref($self) || $self;
 
 	# 'id' of html <input> tag corresponding to the answer blank. Must be unique to each pool of DragNDrop buckets
@@ -127,14 +121,14 @@ sub new {
 	my $defaultBuckets = shift;
 
 	my %options = (
-	AllowNewBuckets => 0,
-	@_
+		AllowNewBuckets => 0,
+		@_
 	);
 
 	$self = bless {
-		answerInputId => $answerInputId,
-		bucketList => [],
-		aggregateList => $aggregateList,
+		answerInputId  => $answerInputId,
+		bucketList     => [],
+		aggregateList  => $aggregateList,
 		defaultBuckets => $defaultBuckets,
 		%options,
 	}, $class;
@@ -148,19 +142,19 @@ sub addBucket {
 	my $indices = shift;
 
 	my %options = (
-	label => "",
-	removable => 0,
-	@_
+		label     => "",
+		removable => 0,
+		@_
 	);
 
 	my $bucket = {
-		indices => $indices,
-		list => [ map { $self->{aggregateList}->[$_] } @$indices ],
+		indices   => $indices,
+		list      => [ map { $self->{aggregateList}->[$_] } @$indices ],
 		bucket_id => scalar @{ $self->{bucketList} },
-		label => $options{label},
+		label     => $options{label},
 		removable => $options{removable},
 	};
-	push(@{$self->{bucketList}}, $bucket);
+	push(@{ $self->{bucketList} }, $bucket);
 
 }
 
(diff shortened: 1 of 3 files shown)


macros/draggableProof.pl
```diff
@@ -1,284 +1,404 @@
-# Done: show possible choices in TeX mode
-# To do: display student answers and correct answers in TeX mode properly.
-# To do: put jquery.nestable.js in a universal spot on every webwork server.
+################################################################
+=head1 NAME
+draggableProof.pl
 
-loadMacros("PGchoicemacros.pl");
+=head1 DESCRIPTION
```
drgrice1, line 5: Delete the DESCRIPTION header.  It has no content.

Either that or add a description.

drgrice1 (changes requested):
I have not done extensive testing on this yet, but there are a few things that need to be done.

First, our current practice is to indent with un-expanded tabs instead of spaces.

Second, please clean up the white space at the ends of the lines.

Third, try not to make lines excessively long.  Generally speaking there should be no lines longer than 120 characters.  Lines longer than 120 characters will wrap in many situations and hinder readability.

Fourth, there are some POD issues.  Generally any POD command should have an empty line before and after it.  In some cases the lack of these empty lines can cause POD generation errors, although I don't see any of those in this case.  However, the lack of those empty lines does cause the generated POD to be rather unreadable as it causes a lack of spacing where it should be.

Finally, added files should have the WeBWorK license added to the beginning.  You can cut and paste that from another file (for example macros/PGtikz.pl).  Make sure to update the copyright range to go up to 2021.  I know that many of the existing files do not have this, but that is a shortcoming that needs to be fixed.  New files should be added with this already.

drgrice1:
Note that when I say to indent with tabs, I mean in the code portion.  In the POD it is actually better to use spaces.

macros/draggableProof.pl
pschan-gh: A description has been filled in. Thank you!

drgrice1:
This looks quite a bit better.  You have resolved most of the issues.  The generated POD looks much better.

The whitespace at the end of lines issue though is still not resolved.  I recommend that you use a regex substitution to take care of this.  In vim you could run `:\s/\+$//g` and immediately take care of it.  Many text editors have regex substitutions, although the regex syntax might be different.  Some editors automatically remove whitespace at the end of lines.  Vim actually does this if the .editorconfig for the project specifies to do so.  We need to add an .editorconfig file!

pschan-gh:
OK, trying once more. I use the atom editor, and I guess I need to play around with the settings more. Thanks!

drgrice1:
Much better.  Thanks.

I will try to do some testing of this later today.

pschan-gh:
Inclusion of Context()->texStrings for problems which use PGML apparently causes the answer checker to fail. "main::Context()->normalStrings;" has been added to the prefilter of the answer checker to force the context to normal strings.

pstaabp (approved):
I was also looking at https://webwork.maa.org/wiki/DraggableProofs and it mentions that `draggableProof.pl` is not a standard macro.  I assume that this one will replace the one in `OpenProblemLibrary/macros/MC/draggableProof.pl`. 

Even though the wiki documentation isn't part of this PR, we should update that to reflect this (or we may need to wait until this is fully released).  I'd argue that the example on the wiki should be updated.  I chuckle at the inside joke between JT and Jason, but we could be a bit more professional.

drgrice1 (approved):
I have done rather extensive testing with this now, and this looks really good.  This seems to work with the existing problems in the OPL.  I tested almost all of them (there are 75 of them by my count).  I also tested with a few other problems that I have found and it works with them as well.

Very nice work!

drgrice1:
I forgot to test in gateway quizzes.  I have done that now.  The problems also work there!

By the way, I have attached a set definition file with all of the problems in the OPL that use this macro.
[DraggableProofOPL.zip](https://github.com/openwebwork/pg/files/7108530/DraggableProofOPL.zip)

macros/draggableProof.pl
```diff
@@ -1,284 +1,455 @@
...
+loadMacros(
+"PGchoicemacros.pl",
+"MathObjects.pl",
+);
 
+sub _draggableProof_init {
+	ADD_CSS_FILE("https://cdnjs.cloudflare.com/ajax/libs/nestable2/1.6.0/jquery.nestable.min.css", 1);
+	ADD_JS_FILE("https://cdnjs.cloudflare.com/ajax/libs/nestable2/1.6.0/jquery.nestable.min.js", 1);
+	ADD_CSS_FILE("js/apps/DragNDrop/dragndrop.css", 0);
+	ADD_JS_FILE("js/apps/DragNDrop/dragndrop.js", 0, { defer => undef });
```
drgrice1, line 152: I just noticed that you defer loading of the dragndrop.js file, but you do not defer loading of the jquery.nestable.js file.  You should also defer loading for that.

lib/DragNDrop.pm
```diff
@@ -0,0 +1,205 @@
...
+ADD_JS_FILE("https://cdnjs.cloudflare.com/ajax/libs/nestable2/1.6.0/jquery.nestable.min.js", 1);
+ADD_CSS_FILE("js/apps/DragNDrop/dragndrop.css", 0);
+ADD_JS_FILE("js/apps/DragNDrop/dragndrop.js", 0, { defer => undef });
+
+To initialize a bucket pool, do:
+
+my $bucket_pool = new DragNDrop($answerInputId, $aggregateList);
+
+$answerInputId is a unique identifier for the bucket_pool, it is recommended that
+it be generated with NEW_HIDDEN_ANS_NAME.
```
drgrice1, line 44: Another minor issue.  There is no such subroutine NEW_HIDDEN_ANS_NAME.  This should be NEW_ANS_NAME.

pschan-gh: I have updated the files in the latest PR based on your recommendations, thanks!

pschan-gh:
Thanks everyone for testing and reading through the code!

drgrice1:
Good.  Thanks.

lib/DragNDrop.pm
```diff
@@ -0,0 +1,228 @@
...
+		for my $index ( @{$bucket->{indices}} ) {
+			$out .= "<li data-shuffled-index='$index'>$self->{aggregateList}->[$index]</li>";
+		}
+		$out .= "</ol>";
+		$out .= "</div>";
+	}
+	$out .= '</div>';
+	$out .= "<br clear='all'><div><a class='btn reset_buckets'>reset</a>";
+	if ($self->{AllowNewBuckets} == 1) {
+		$out .= "<a class='btn add_bucket' data-ans='$self->{answerInputId}'>add bucket</a>";
```
drgrice1, line 201: It would be good to add the css `btn-secondary` class to both of these buttons.  This class does not exist in bootstrap version 2, and so at this point it will do nothing.  However, soon we will be upgrading to bootstrap version 5, and with version 5 the `btn` class is not intended to be used by itself.  The `btn-secondary` class is the flavor that will work best here.  If you don't do this now, I will add it later.  I have a pull request coming soon that implements this upgrade.

drgrice1 (commented):
There is a little clean up that should be done, but otherwise this looks good.

It would be a good idea to run perltidy on the files that are change or added in this pull request using the .perltidyrc that will be added by pull request #618.  Get a copy of that file, put it in the pg root directory, and then run `perltidy -pro=.perltidyrc -b -bext='/' lib/DragNDrop.pm macros/draggableProof.pl macros/draggableSubsets.pl` from the pg root directory.  Then commit the changes.

lib/DragNDrop.pm
pschan-gh: Thank you for the comments! I have updated the files per your suggestions. I will update the linked PR on openwebwork/webwork2 next.

drgrice1 (approved):
Thanks for making my suggested changes.  I think this is ready to go.

PR #512, opened 2020-12-09
Properly add parentheses to Fraction::Value objects based on precedence.

dpvc:
Currently, Fraction MathObjects may lose necessary parentheses when stringified or texified.  For example `Compute("x^(2/7)")->string;` will produce `x^2/7`.  This means that

```
$f = Compute("x^(2/7)");
$g = Compute("$f");
```
would produce `$g` equivalent to `Compute("(x^2)/7")` rather than equivalent to `$f`.  (See this [Forum post](https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4981) for more details.)

This PR resolves the issue by overriding the `string()` and `TeX()` methods to properly insert parentheses when needed.  These are added when the operator precedences require it, or (in `string()`) when the original had them (since the TeX version is a vertical fraction, they are are only added when precedence requires it).

To test, use

```
loadMacros("PGML.pl", "contextFraction.pl");

sub LaTeX {
  my $f = shift;
  return PGML::LaTeX('\(' . $f->TeX . '\)');
};

Context("Fraction");

$f1 = Compute("x^(2/7)");
$f2 = Compute("(2/7)*x");
$f3 = Compute("2/7 x");

TEXT($f1, " has parens", $BR);
TEXT(Compute("$f1"), " has parens", $BR);
TEXT($f2, " has parens", $BR);
TEXT(Compute("$f2"), " has parens", $BR);
TEXT($f3, " has no parens", $BR);
TEXT(Compute("$f3"), " has no parens", $BR);
TEXT($HR);

$x = Formula("x");
$f = Fraction(2,7);

$f4 = $x ** $f;
$f5 = $f * $x;

TEXT($f4, " has parens", $BR);
TEXT(Compute("$f4"), " has parens", $BR);
TEXT($f5, " has no parens", $BR);
TEXT(Compute("$f5"), " has no parens", $BR);

TEXT($HR, "No parens:", $BR);
TEXT(LaTeX($f1), $BR);
TEXT(LaTeX(Compute("$f1")), $BR);
TEXT(LaTeX($f2), $BR);
TEXT(LaTeX(Compute("$f2")), $BR);
TEXT(LaTeX($f3), $BR);
TEXT(LaTeX(Compute("$f3")), $BR);

TEXT($HR, "No parens:", $BR);
TEXT(LaTeX($f4), $BR);
TEXT(LaTeX(Compute("$f4")), $BR);
TEXT(LaTeX($f5), $BR);
[...]

PR #608, opened 2021-10-15
Fix calls to Parser::Value::TeX & too many (cooks)

drdrew42:
In PR #512, contextFraction.pl inserted new methods into the `string` and `TeX` method call chains (originally routed directly to Parser::Value::TeX, now passing through context::Fraction::Value). These methods had the impact of adding necessary parens to the string (as precedence would require -- [see reference](https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4981)).

However, strictly following precedence at the Fraction::Value layer causes [issues with unary negation](https://github.com/openwebwork/pg/pull/512#commitcomment-57951720) (as reported by @Alex-Jordan) -- because the precedence of `u-` is 6 (and precedence of `/` is 3). 

This issue is avoided in Fraction::BOP::divide::TeX via argument `$showparens eq 'nofractions'` via UOP::TeX (derived from `Parser::Context::Default::operations->{'u-'}{nofractionparens}`). This solution is not available at the Fraction::Value layer (and is already being handled on the BOP layer), so these additional parens will _now_ only be added when either:
- the parens are explicitly there in the original string definition, or
- precedence dictates AND we've explicitly set `showExtraParens => 2` in the context

Meanwhile the precedence parens remain a necessary addition for strings at the Fraction::Value layer because we cannot be assured that the original string contained them (see attached pg example of constructing a MathObject with precedence derived from perl operations [modified from code provided in #512]).

Furthermore, subsequent changes in #569 (stemming from issue #567) changed `context::Fraction::Fraction::TeX` to properly handle the incoming precedence. This had the effect of yet another layer trying to insert string and TeX parens based on precedence. This code was not previously being acted upon because the incoming precedence was not being handled properly (Fraction objects should never have ->{open} or ->{closed} defined as [noted](https://github.com/openwebwork/pg/pull/569/files#r626851711) by @taniwallach). This is definitely not the layer to be handling precedence issues and so that code has been removed.

See attached code: [check-parens.pg.txt](https://github.com/openwebwork/pg/files/7350814/check-parens.pg.txt)

### DIFF
--- a/macros/contextFraction.pl
+++ b/macros/contextFraction.pl
@@ -636,8 +636,9 @@ sub TeX {
   return $string unless $self->{value}->classMatch('Fraction');
   my $precedence = shift;
   my $frac = $self->context->operators->get('/')->{precedence};
+  my $noparens = shift;
   $string = '\left(' . $string . '\right)' if $self->{hadParens} ||
-    (defined $precedence && $precedence > $frac && $self->context->flag('showExtraParens') > 1);
+    (defined $precedence && $precedence > $frac && !$noparens);
   return $string;
 }
 
--- a/macros/contextFraction.pl
+++ b/macros/contextFraction.pl
@@ -618,7 +618,7 @@ sub reduce {
 #
 sub string {
   my $self = shift;
-  my $string = $self->SUPER::string($self, @_);
+  my $string = $self->SUPER::string(@_);
   return $string unless $self->{value}->classMatch('Fraction');
   my $precedence = shift;
   my $frac = $self->context->operators->get('/')->{precedence};
@@ -627,15 +627,17 @@ sub string {
 }
 
 #
-#  Add parentheses if they are needed by precedence
+#  Add parentheses if they were there originally, or
+#  are needed by precedence and we asked for exxxtra parens
 #
 sub TeX {
   my $self = shift;
-  my $string = $self->SUPER::TeX($self, @_);
+  my $string = $self->SUPER::TeX(@_);
   return $string unless $self->{value}->classMatch('Fraction');
   my $precedence = shift;
   my $frac = $self->context->operators->get('/')->{precedence};
-  $string = '\left(' . $string . '\right)' if defined $precedence && $precedence > $frac;
+  $string = '\left(' . $string . '\right)' 
+    if defined $precedence && $precedence > $frac && $self->context->flag('showExtraParens') > 1;
   return $string;
 }
 
@@ -898,7 +900,6 @@ sub string {
   if ($self->getFlagWithAlias("showMixedNumbers","showProperFractions") && CORE::abs($a) > $b)
     {$n = int($a/$b); $a = CORE::abs($a) % $b; $n .= " " unless $a == 0}
   $n .= "$a/$b" unless $a == 0 && $n ne '';
-  $n = "($n)" if defined $prec && $prec >= 1;
   return "$n";
 }
 
@@ -911,7 +912,6 @@ sub TeX {
   my $s = ""; ($a,$s) = (-$a,"-") if $a < 0;
   $n .= ($self->{isHorizontal} ? "$s$a/$b" : "${s}{\\textstyle\\frac{$a}{$b}}")
     unless $a == 0 && $n ne '';
-  $n = "\\left($n\\right)" if defined $prec && $prec >= 1;
   return "$n";
 }
 
--- a/macros/contextFraction.pl
+++ b/macros/contextFraction.pl
@@ -636,8 +636,8 @@ sub TeX {
   return $string unless $self->{value}->classMatch('Fraction');
   my $precedence = shift;
   my $frac = $self->context->operators->get('/')->{precedence};
-  $string = '\left(' . $string . '\right)' 
-    if defined $precedence && $precedence > $frac && $self->context->flag('showExtraParens') > 1;
+  $string = '\left(' . $string . '\right)' if $self->{hadParens} ||
+    (defined $precedence && $precedence > $frac && $self->context->flag('showExtraParens') > 1);
   return $string;
 }
 
--- a/lib/Parser.pm
+++ b/lib/Parser.pm
@@ -401,8 +401,8 @@ sub Close {
                     ($top->type eq 'Comma') ? $top->entryType : $top->typeRef,
                     ($type ne 'start') ? ($open,$paren->{close}) : () )};
         } else {
-	  $top->{value}{hadParens} = 1;
-	}
+          $top->{value}{hadParens} = 1 unless $open eq 'start';
+        }
         $self->pop; $self->push($top);
         $self->CloseFn() if ($paren->{function} && $self->prev->{type} eq 'fn');
       } elsif ($paren->{formInterval} eq $type && $self->top->{value}->length == 2) {


pstaabp:
This does seem to fix the problem with the unary minus and fractions.  We don't have to do this here, but it seems logical to not have parens at all around either 1/2 or -1/2.  For example:

Formula("-((x+3)/(x-4))");

returns $$-\frac{x+3}{x-4}$$

which is the way I would write it.   Throw these cases into the unit-test that I will write one day. :)

drgrice1 (approved):
This looks good.  Although, as it is only a display issue, I am not sure that this should be a hot fix to main.

drdrew42:
## Additional change:

### Before
Parser sets the `hadParens` flag at the very top level of the parsed formula. This is why we see `$pos = Formula("1/2")` receiving parens, and then they are inherited by `Formula("-$pos")` which parses `$pos->string` ([generated by the MWE provided here](https://github.com/openwebwork/pg/pull/512#commitcomment-57951720)). Notice that the same is not true for `$pos = Fraction(1,2)` and `Formula("-$pos")` because we're not parsing a string for `$pos` (and when parsing `"-$pos"`, the top-level `hadParens` flag is being set on the UOP, rather than on the Fraction).

### After

After this additional commit, we should see the same TeX from `Formula("1/2")` and `Fraction(1,2)` regardless of `reduceConstants`. The same should hold for `Formula("-$pos")`. Furthermore, the string for all of these versions of 1/2 is also consistent (with or without reduceConstants).

This additional change should have no other effect, since the `hadParens` flag is only used in contextFraction.pm and contextAlternativeIntervals.pm. In contextAlternativeIntervals, the presence of the `hadParens` flag is contained in a duplication of the code block from Parser that we're editing now, so this change should have no effect there. 

@Alex-Jordan I think this latest change should address the negated fractions.

macros/contextFraction.pl
```diff
@@ -618,7 +618,7 @@ sub reduce {
 #
 sub string {
   my $self = shift;
-  my $string = $self->SUPER::string($self, @_);
+  my $string = $self->SUPER::string(@_);
```
dpvc, line 621: Good catch, here and below.  Don't know what I was thinking.

macros/contextFraction.pl
```diff
@@ -627,15 +627,17 @@ sub string {
...
 sub TeX {
   my $self = shift;
-  my $string = $self->SUPER::TeX($self, @_);
+  my $string = $self->SUPER::TeX(@_);
   return $string unless $self->{value}->classMatch('Fraction');
   my $precedence = shift;
   my $frac = $self->context->operators->get('/')->{precedence};
-  $string = '\left(' . $string . '\right)' if defined $precedence && $precedence > $frac;
+  $string = '\left(' . $string . '\right)' if $self->{hadParens} ||
+    (defined $precedence && $precedence > $frac && $self->context->flag('showExtraParens') > 1);
```
dpvc, line 640: Adding the test for `showExtraParens` is not the right solution to the unary minus issue, as these parens are not optional for other situations.  For example, if `showExtraParens => 0` is set, then something like

```
(Formula(Fraction(1,2))**2)->TeX
```

will produce a fraction squared, but without the needed parentheses.  That is not the correct display for that situation.

Fortunately, the unary minus already passes a parameter to the `TeX()` method to indicate that there should be no parens around a fraction, and you can tap into that.  It is the next parameter in the argument list, so using

```suggestion
  my $noparens = shift;
  $string = '\left(' . $string . '\right)' if $self->{hadParens} ||
    (defined $precedence && $precedence > $frac && !$noparens);
```

should make it work both for unary minus and powers (or other higher-precedence operations).

dpvc (changes requested):
I have made a suggested edit, but otherwise looks fine.

macros/contextFraction.pl
dpvc: Yes, that is right on both counts.

dpvc: The `string` and `TeX` methods are a mess, and probably deserve an overhaul.  Lots of hacks in there to make it work that should probably be cleaned up.

drdrew42: also, you have a mixture of my thanks and apologies for belaboring all of this -- I've benefited greatly from hacking away on this issue, but I'm also super grateful for anything you're willing to throw my way in terms of making sense of the overall structure. 🙏

dpvc: Well, this takes a bit of history and overview to understand.

First, you need to understand the difference between the `Parser` namespace and the `Value` namespace.  The `Parser` objects are used to maintain an abstract syntax tree for a parsed formula, so the just represent the structure of the formula.  Thus things like addition and powers and function calls and how they fit together are coded here. In general, this is not about computation, but about the structure.

The `Value` objects are the ones that model the actual mathematical values of things like real and complex numbers, vectors and such, and do the computations involving them.  There is no structure here, just computation.  So if a real number cam from addition of two numbers or multiplication is not known here.  We just have the number, not how it was computed.

Originally, I wanted these two libraries to be able to operate separately, in particular the Value objects were supposed to be able to be used without the Parser, just for computational purposes.  As it turns out, over time they became more and more intertwined, and the separation is no longer possible.

The `Value::Formula` is also a complication, since it is really a `Parser` object under the hood, but it has all the computational methods for doing things like evaluating the formula, generating random points, comparing formulas and so on.  So it is a cross-over object.

A complication from the other direction is that the Parser can reduce constant operations and function calls, and so when it parses `1 + 2`, rather than making the abstract tree with a `Parser::BOP::add` node whose `lop` and `rop` fields are `Parser::Number` objects whose `value` fields are the Perl reals 1 and 2, the parser converts the perl reals into `Value::Real` objects and asks one to add itself to the other, and gets back the `Value::Real` result.  But this doesn't fit into a `Parser::Number` since its value is a perl real.  In the case of a `Value:Real()`, the parser knows how to convert the result back into a perl number for the `Parser::Number`, but in general, the parser doesn't really know much about `Value` objects or how they relate to its own objects.  The `Value` object that the parser gets back from reducing constants or function calls could be virtually anything, like something defined in an external macro file, for example a Fraction object, which the parser has never seen before.  That means the parser needs another mechanism of dealing with such results since it can't turn them back into its own parser nodes.  That is handled by the `Parser::Value` object, which is just a shell around a `Value` object as a holding place for it (whether it is a vector, list, or some subclass that is unknown to it, like a fraction).

These `Parser::Value` nodes are only created if the parser has to deal with a `Value` object (other than a Real or Complex object).  That can happen due to reducing constants, as in the example above, or by doing something like passing a `Value` object to the `Formula()` constructor (which will create a formula containing just a `Parser::Value` node holding the `Value` object.  So `Formula(Vector(1,2,3))` is different from doing `Formula("<1,2,3>")` (when `reduceConstants => 0` is set) since the former creates a `Parser::Value` node containing a `Value::Vector` object, while the later produces a `Parser::list::vector` node with three `Parser::Number` sub-nodes.

Now the context can override the objects that the Parser creates in several ways.  For example, you can set the class created by an operator in the context's operators hash.  For example, the `contextFraction.pl` file sets up the `Fraction` context using 

``` perl
$context->operators->set("/"  => {class => "context::Fraction::BOP::divide"});
```

(along with some other operator definitions).  This overrides the `/` operator to produce a `Fraction::BOP::divide` object rather than the usual `Parser::BOP::divide` that is the default for contexts based on the Numeric context.  When a `parser::BOP::divide` node is evaluated, it generates a `Fraction` object rather than a `Real`.  (Technically, it is a `context::Fraction::Fraction` object, which is the `Value` object that handles computation with fractions.)  This is how `Compute("1/2")` is made to produce a Fraction object in the end.

The context can also override the class used to create the various `Parser` nodes or `Value` objects that it needs.  For example, the Fraction context also sets

``` perl
$context->{value}{Real} = "context::Fraction::Real";
$context->{parser}{Value} = "context::Fraction::Value";
```

and some other definitions.  This tells the parser that when it wants to create a `Value::Real` it should instead create a `context::Fraction::Real` and when it wants to create a `Parser::Value` node, it should create `context::Fraction::Value` instead.  This means that if you do `Formula(Fraction(1,2))` (or `Formula("1/2")` when `reduceConstants => 1` is set), you should get a `Fraction::Value` node rather than a `Parser::Value` node.  The `Fraction::Value` node is what allows constant Fraction values to be handled specially (in particular for its `string` and `TeX` methods).

So we are finally in a position to answer your question about `Parser::UOP->{op}`.  When the `UOP` is acting on a fraction, it could be either a `Fraction::BOP::divide` or a `Fraction::Value` containing a `Fraction::Fraction`  object, depending on how the the operator was created.  If the fraction was a constant fraction and `reduceConstants` is true, the `Parser::UOP->{op}` will have a `Fraction::Value` object, and when its `TeX()` method is called, it will be the one you have been modifying in the `Fraction::Value` class.  That will call the `Fraction::Value` superclass (which is `Parser::Value`) `TeX()` method, and that in turn will call the `Value` object's `TeX()` method.  In this case, the `Value` object is a `Fraction::Fraction`, and so that is the `TeX()` method that will be called.  These are the ones that used to contain the `($n)` lines that you removed.

On the other hand, if the UOP is acting on a fraction that is not constant (e.g., `1/x`) or `reduceConstants` is false, then the `Parser::UOP->{op}` will be a `Fraction::BOP::divide`, and it is that object's `TeX()` method that is called.  That is a subclass of `Parser::BOP::divide`, and it overrides `TeX()` but not `string()`.  When its `TeX()` method is called, it calls `SUPER::TeX()`, so *that* is when `Parser::BOP::divide->TeX()` is finally called.  That will only happen if the fraction is not constant, or `reduceConstants` is false, and the parser tree holds the `Fraction::BOP::divide` and its two `Number` nodes, rather than the `Fraction::Value` node.

You can investigate the parse tree using `TEXT(Formula("...")->{tree})` to print the reference to the top node in the tree, and you could `pretty_print()` the tree to see the values of its fields, I would think.

Anyway, that is a long-winded response, but I hope it helps clarify what is happening, and why you have been seeing different values in the output for the table in the original issue thread.



Similarly, when the Fraction context is used (with `reduceConstants => 0`), `Formula(Fraction(1,2))` produces a `Parser::Value`

Issue #607, opened 2021-10-13
Issue with ParserAssignment.pl


dlglin:
There is an issue with ParserAssignment.pl when a student enters an answer for which PG can't generate enough points for comparison.  This was discovered with Library/FortLewis/DiffEq/1-First-order/05-Substitution-Bernoulli/Lebl-1-5-06.pg.  A MWE is attached.

Steps to reproduce:
In one of the above problems, enter an answer that is not defined on the domain (e.g. "y=sqrt(x-5)"), and click either Preview or Submit.  Now enter a correct answer and click Submit.  You will receive the error "The evaluated answer is not an answer hash : ||.", and the answer will be marked wrong.

It looks like this has something to do with the check for equivalence to the previously submitted answer.

[ParserAssignmentMWE.pg.txt](https://github.com/openwebwork/pg/files/7340572/ParserAssignmentMWE.pg.txt)

dlglin:
Did you try submitting a correct answer after that answer?

drdrew42:
*sigh* -- apparently I can only read a couple sentences before my brain shuts off. 

Confirmed, will investigate.

pstaabp:
Following up on this one.  This does fix the initial error from the OPL problem and the MWE.  

There is still the warning about not generating enough points.  I understand the issue, but what do students do with this information--except ask the instructor.   And most instructor's don't understand this.  

I think this should be closed but perhaps generate a more-helpful warning.

pstaabp:
This has been fixed by #609 and #645

PR #609, opened 2021-10-15
parserAssignment chokes when previous answer is undefined on the domain

drdrew42:
See #607 for details (incl. a MWE)

Submitting an answer whose domain is incompatible with the limits for the available variables causes an appropriate error message: `The domain of your function doesn't match that of the correct answer`

However, when followed up with an answer whose domain **is** compatible, you will get an error: `The evaluated answer is not an answer hash : ||. ` (This happens only based on domain, and is independent of whether or not the submitted answer is correct.)

The failure to complete the processing of the answer hash is due to the cmp_postfilter which compares the current and previous answers. 
_Note:_ So long as the submitted answer generates a feedback message, the comparison between the current and previous submissions is skipped.
Specifically, the `typeMatch` method call (which ensures that the current and previous responses are comparable objects _before_ actually comparing them) dies because the previous submission fails to generate valid points.

The solution is to simply remove `typeMatch` from `parser::Assignment::Formula`. It's not doing anything more than what (the superclass) `Value::Formula` is already doing. And `Value::Formula::typeMatch` is structured to handle situations where the generation of valid points fails. (Interesting to note: this looks like a bug that was fixed not long after parserAssignment was introduced -- but it was fixed for Value::Formula and the changes never made their way into parserAssignment... see 2e1ed6b41f8375d15b27c64490ff2e324f2826e9)

TL;DR -- parser::Assignment::Formula doesn't need to override `typeMatch`

postfilter: 🍻  to @dlglin for finding a teenaged bug!

### DIFF
--- a/macros/parserAssignment.pl
+++ b/macros/parserAssignment.pl
@@ -348,7 +348,18 @@ sub new {
   my $self = shift; $class = ref($self) || $self;
   my $f = $self->SUPER::new(@_);
   bless $f, $class if $f->type eq 'Assignment';
+  my $rhs = $f->getTypicalValue($f)->{data}[1];
+  Value->Error('Assignment of strings is not allowed.') if $rhs && $rhs->type eq 'String';
   return $f;
+}
+
+sub typeMatch {
+  my $self = shift; my $other = shift; my $ans = shift;
+  return 0 unless $self->type eq $other->type;
+  my $typeMatch = $self->getTypicalValue($self)->{data}[1];
+  $other = $self->getTypicalValue($other,1)->{data}[1];
+  return 1 unless defined($other); # can't really tell, so don't report type mismatch
+  $typeMatch->typeMatch($other,$ans);
 }
 
 sub cmp_class {
--- a/macros/parserAssignment.pl
+++ b/macros/parserAssignment.pl
@@ -351,18 +351,6 @@ sub new {
   return $f;
 }
 
-sub typeMatch {
-  my $self = shift; my $other = shift; my $ans = shift;
-  return 0 unless $self->type eq $other->type;
-  $other = $other->Package("Formula")->new($self->context,$other) unless $other->isFormula;
-  my $typeMatch = ($self->createRandomPoints(1))[1]->[0]{data}[1];
-  $main::__other__ = sub {($other->createRandomPoints(1))[1]->[0]{data}[1]};
-  $other = main::PG_restricted_eval('&$__other__()');
-  delete $main::{__other__};
-  return 1 unless defined($other); # can't really tell, so don't report type mismatch
-  $typeMatch->typeMatch($other,$ans);
-}
-
 sub cmp_class {
   my $self = shift; my $value;
   if ($self->{tree}{rop}{isConstant}) {


dpvc:
> It's not doing anything more than what (the superclass)

That's actually not true.  The super class (which is `Value::Formula`) `typeMatch()` function is

https://github.com/openwebwork/pg/blob/37637e2f880bccbbb8781b5d669421849ac81267/lib/Value/AnswerChecker.pm#L1653-L1661

and it returns `1` immediately if `$self->type` equals `$other->type`.  For a normal Formula, the `type` will be the type of the result of the formula (e.g., `Number` or `Vector`, etc.), and only goes on to try to analyze the return type if the named types don't match.  For an assignment, however, the `type` is always `Assignment`.  So using the super-class `typeMatch()` would always return `1` for any two assignments.

The current `typeMatch()` for assignments returns `0` if the types are not equal (i.e., if they aren't both assignments), and if they are both assignments, then it goes on to do additional type checking of the type of object that is being assigned to the variable.  That additional checking would not be performed if you just use the `Value::Formula->typeMatch()`, and that checking is important for feedback messages.  For example, in `Vector` context, if the correct answer is an assignment `v = <1,2,3>` and the student enters `v = 1`, the current `typeMatch()` would fail, and the message

```
Your answer isn't a variable equal to a vector
(it looks like a variable equal to a real number)
```

is issued.  With the changes in this PR, however, the answer will be marked incorrect silently (no helpful message). 

There are also other differences.  The current `typeMatch()` forces other to be converted to a Formula object, but that should be able to be dropped, because it is already known to be an Assignment object, which is a Formula.  But because the Assignment is actually a list-valued formula (the first item being the variable name and the second its value), the check for its assignment type is done on the second element from the list.  That is not the case in `Value::Formula->typeMatch()`, which operates on the formula's value as a whole.

So this PR changes the functionality of the Assignment object from its current form in a way that I would not support.

It's true that the Assignment `typeMatch()` should be updated to use `getTypicalValue()` (which postdates this code), 
so a possible version for the Assignment object might be

``` perl
sub typeMatch {
  my $self = shift; my $other = shift; my $ans = shift;
  return 0 unless $self->type eq $other->type;
  my $typeMatch = $self->getTypicalValue($self)->{data}[1];
  $other = $self->getTypicalValue($other,1)->{data}[1];
  return 1 unless defined($other); # can't really tell, so don't report type mismatch
  return 1 if $typeMatch->classMatch('String') && Value::isFormula($ans->{typeMatch});  # avoid infinite loop
  $typeMatch->typeMatch($other,$ans);
}
```

though I haven't actually tried it out.

drgrice1:
There seems to be a problem created by this pull request.  I have problems that use parserAssignment.pl and PGML's parsing to display equations.  For example, one problem has `[::[$e]/[$a] x + 1/[$b] = 1/[$c] x - 1/[$d]::]`.  This now gives the error
```
Errors parsing PGML:
Error parsing mathematics: No perl form for ' = '; see position 13 of formula
```
If I revert this pull request, the error does not occur.

drgrice1:
Here is a minimal example for which this issue occurs:
```perl
DOCUMENT();

loadMacros(
	"PGstandard.pl",
	"MathObjects.pl",
	"PGML.pl",
	"parserAssignment.pl",
	"PGcourse.pl"
);

TEXT(beginproblem());

parser::Assignment->Allow;

BEGIN_PGML
>>[::2/5 x + 1/7 = 1/3 x - 1/11::]<<
END_PGML

ENDDOCUMENT();
```

PR #613, opened 2021-11-11
Don't allow entryType change to modify original Value::Type entries

dpvc:
This PR fixes a problem with the Inequalities context reported in the [forum](https://webwork.maa.org/moodle/mod/forum/discuss.php?d=6562) where the use of a list of inequalities would cause an error when an answer is resubmitted.

The problem is due to the fact that the Inequalities context was changing a value that is part of `$Value::Types{interval}`, but `%Value::Type` is static code that is shared among all problems processed by a given https child process, so that change was propagating to problems processed subsequently by the same child.

The fix is to use `Value::Type()` to create a new `typeRef` object rather than modify the existing (shared) one.

This can be checked using the code in the original forum post.  With this patch, it should not cause an error when submitting the problem more than once.

This PR also fixes a second (unreported issue) in the Inequalities context, where direct calls to `Interval()` could produce incorrect results.  This can be tested with 

```
loadMacros("contextInequalitieis.pl");
Context("Inequalities");
TEXT(Interval('(', 1, 2, ')'));
```

which should produce the output `(1,2)`.  Without this patch, it produces just `)`.

### DIFF
--- a/macros/contextInequalities.pl
+++ b/macros/contextInequalities.pl
@@ -380,7 +380,7 @@ sub _check {
     unless $self->{lop}{isInequality} && $self->{rop}{isInequality};
   $self->Error("Inequalities combined by '%s' must both use the same variable",$self->{bop})
     unless $self->{lop}{varName} eq $self->{rop}{varName};
-  $self->{type} = Value::Type("Interval",2);
+  $self->{type} = $Value::Type{interval};
   $self->{varName} = $self->{lop}{varName};
   $self->{isInequality} = 1;
 }
@@ -400,7 +400,7 @@ sub _check {
     unless $self->{lop}{isInequality} && $self->{rop}{isInequality};
   $self->Error("Inequalities combined by '%s' must both use the same variable",$self->{bop})
     unless $self->{lop}{varName} eq $self->{rop}{varName};
-  $self->{type} = Value::Type("Interval",2);
+  $self->{type} = $Value::Type{interval};
   $self->{varName} = $self->{lop}{varName};
   $self->{isInequality} = 1;
 }
@@ -840,7 +840,7 @@ our @ISA = ("Value::Interval");
 
 sub new {
   my $self = shift;
-  $self = $self->SUPER::new(@_);
+  $self = Value::Interval->new(@_);
   $self = $self->demote if $self->classMatch("Inequality");
   return $self;
 }
@@ -878,7 +878,7 @@ sub _check {
     my $entryType = $self->typeRef->{entryType};
     return unless $entryType->{name} =~ m/^(unknown|Interval|Set|Union)$/;
     foreach my $x (@{$self->{coords}}) {return unless $x->{isInequality}};
-    $entryType->{name} = "Inequality";
+    $self->typeRef->{entryType} = Value::Type("Inequality", $entryType->{length}, $entryType->{entryType});
   }
 }
 


macros/contextInequalities.pl
```diff
@@ -840,7 +840,7 @@ package Inequalities::MakeInterval;
 
 sub new {
   my $self = shift;
-  $self = $self->SUPER::new(@_);
+  $self = Value::Interval->new(@_);
```
drgrice1, line 843: Would it be better to change this to
```suggestion
  $self = (shift)::Interval->new(@_);
```
and delete line 842 (as is done on line 856 for `Inequalities::List`)?  It seems that this makes the `$self` usage in this situation more clear.

drgrice1 (approved):
This looks good.  It seems to fix both of the mentioned issues in my tests.

Thanks for this.

macros/contextInequalities.pl
drdrew42: I think you meant `(shift)->SUPER::new(@_);`

drgrice1: That is literally what is on line 856, but for this I meant `(shift)::Interval->new(@_)`.  I was referring to the `$self` usage and adapted that to this change.  But looking closely now, I see that both are wrong.  It looks like line 842 should just be deleted, and the `my` added to line 843.

drdrew42: calling `Value::Interval->new` would be the same as `shift->SUPER::new` since this package @ISA `Value::Interval` so `my $self = (shift)->SUPER::new(@_);` would work the same as the two lines in the PR do now.

dpvc: No, the line is correct as is.  This is downgrading the `Inequalities::Interval` to a standard `Value::Interval`.  Neither of the suggestions would do that.  Also, removing the `my $self = shift` would leave the original object in the `@_` array, and that would be passed to `new()` method, which isn't correct.

dpvc: Sorry, my remark about shift is wrong.  D'oh.

dpvc: > calling `Value::Interval->new` would be the same as `shift->SUPER::new` since this package ISA `Value::Interval`

Just to be clear, that's not how it works.  The `new()` methods base their return values on the class of the caller, not on the `@ISA` value.  The `new()` methods typically start with

``` perl
  my $self = shift; my $class = ref($self) || $self;
``` 

and end with returning an object that is blessed using

``` perl
bless {...}, $class
```

So if you use `(shift)->SUPER::new(@_)`, that will produce an object with the same class as `(shift)` regardless of whether the `new()` is `SUPER::new()` or not.  Without that, every subclass would have to override `new()` and re-bless the return value to be its class (with it, subclasses are created as instances of the subclasses automatically).
It also allows `Value::Interval->new()` to produce an instance of `Value::Interval` since that is what `$class` will be in that case.  The `@ISA` value is used to method inheritance, not instance class creation.

drdrew42: Aha, yes, I see. (I also see that I was effectively suggesting a return to the prior code...) This ensures that object created has the Value::Interval class rather than the Inequalities::MakeInterval class.

What I'm still unclear on regards the original forum post -- and I think belongs to the other code changes. I'll make a separate comment for that.

drdrew42:
The original forum post suggests that the issue was caused by our Interval object being populated with a reference to a global Interval object, and then modifying that global to have entryType "Inequalities". *BUT* in two lines (`_check` in `Inequalities::BOP::and/or`) you have swapped out a function `Value::Type(...)` (which seems to return a new, unblessed, hash) for a global hash reference`Value::Type{...}`. 

This seems to be the opposite of what is described in the forum post -- we are instead exchanging a fresh hash for a reference sourced from a global?

And in the final change, `_check` in `Inequalities::List::List`, `$entryType` is a key-value in the hash returned by the typeRef  method (but again, that seems to be a method that returns a fresh hash every time). It doesn't make sense to me to set any key-values in that hash, since it doesn't seem to be stored anywhere?

Sorry in advance for any required pedantry. I must be overlooking something here.

dpvc:
> This seems to be the opposite of what is described in the forum post

The issue in the post is due to the original line 881 where the `$entryType->{name}` was modified, potentially modifying a global object.  The change to that line resolves that issue.

The other changes at 383 and 403 are because there is no need for a new instance of the type in those cases, and the common `$Value::type{Interval}` is sufficient and more efficient.  They were there originally to overcome the problem in line 881, and since that was fixed, it makes sense to use the common values on those two lines now.

> `$entryType` is a key-value in the hash returned by the typeRef method (but again, that seems to be a method that returns a fresh hash every time)

Not in this case.  The `Inequalities::List::List` inherits from `Parser::List::List`, which inherits from `Parser::List`, which inherits from `Parser::Item`, and the `typeRef` method there returns the value of `$self->{type}`, not a new copy of it.  Parser item have actual `type` properties, but Value objects don't, so Value objects have to make a new one when needed.  That is what the `typeRef` method in `Inequalities::common` does (and the one in `Value.pm`, which it could have called directly, I guess).  So `$entryType` is the value stored in `$self->{type}` and setting it does make sense (and is needed for proper operation of the list).

drdrew42 (approved):
I get it -- again we're seeing the difference between the Value side and the Parser side of MathObjects...

PR #614, opened 2021-11-11
Attempt to fix MathQuill issues with mixed numbers, units, and scientific notation.

drgrice1:
For this, switch to inserting mathquill responses on the PG side of things, and add a special precedence that is used only for student answers and only when MathQuill is enabled for the answer.

The mathquill latex answers are now stored in KEPT_EXTRA_ANSWERS instead of as an additional response appended to the answer.

This requires changes to MathQuill and for webwork2 (or the front end) to pass in an environment variable via the translation options that tells PG that MathQuill is enabled.  These changes are in a corresponding pull request to webwork2.

Special handling of MultiAnswer's is implemented.  It is now possible to disable a single part of a multi answer, instead of only having the option to disable all parts.

One of the upsides of adding mathquill responses here is that we have access to an answer value's context.  One use is implemented here.  That is to make MathQuill automatically disable the rootsAreExponents option if the root function is available in the context for an answer.

This is paired with https://github.com/openwebwork/webwork2/pull/1522


drgrice1:
@dpvc:  I am hoping that you can take a look a the mq_precedence change in this pull request, and advise as to if this is a good way to deal with this, or might there be unintended side effects.

The objective is to deal with things like the following:

If a student types an answer that is displayed in the MathQuill answer box as `\frac{x-2}{3}\frac{x+5}{7}`, then the text that will be submitted to webwork for the answer is `(x-2)/3 (x+5)/7`.  With the mq_precedence change this is correctly parsed by webwork to be the same as what the student sees for the answer they typed.  With the default precedence it is parsed to be `\frac{\frac{x-2}{3}(x+5)}{7}`, which is not desired.

Previously MathQuill just wrapped all fractions in parentheses to resolve this issue, but that creates problems when mixed numbers are desired.

sean-fitzpatrick:
I am not qualified to inspect the actual code here. But I did try checking out this PR (and the related one for webwork2) and then testing some of the problems where we had run into issues with using MathQuill when there are units.

One particular issue is not resolved by the changes so far: for a problem where the answer is an area, a student cannot simply enter an answer like `420 m^2` with MathQuill turned on. They will get the message "This does not look like a number with units". The answer that is logged by WeBWorK is `420 m^(2)`.
To get the question marked correct, students have to know to click the 'Text Mode' button on the MathQuill toolbar before entering their answer.

drgrice1:
@sean-fitzpatrick:  Check that your browser is not using a cached version of the javascript (clear your cache if needed).  This pull request does fix that.

drgrice1:
@sean-fitzpatrick: Also make sure that you run `npm install` from the webwork2/htdocs directory to install MathQuill.  This of course will not work without that.

pstaabp:
I just found `Library/FortLewis/Authoring/Templates/DiffCalc/AnswerWithUnits1.pg` for an example with units.  The second one has units of ft/s^2 and this looks like it accepts the units correctly.

drgrice1:
I have reverted the part that fixes units in this.  That wasn't working.  Although not for the reason that @sean-fitzpatrick mentions here.  In fact those `m^-1` will still work.  However, units like `ft/s^2` will not.  The fraction parentheses are now back for that. They are needed for something like `(xy)/(ab)` and MathQuill can't distinguish between something like `xy` and `ft`.  They are both just letters, and `f` and `t` could be variables so we can not assume `ft` means feet in MathQuill (not to mention other units that use more than one letter).

See the discussion at https://github.com/openwebwork/mathquill/pull/1.

drgrice1:
With the last commit to this pull request units should be working again.  When MathQuill is enabled for an answer, parentheses are allowed around the numerator and denominator of the unit expression.  This only applies to student answers.

sean-fitzpatrick:
Sorry for being slow returning to this. Thursday was a holiday (Remembrance Day) and Friday there was no school. So my kid-free work window closed just as I got your reply to clear my cache.

These pull requests only affect parsing of numbers with units, is that correct? If so, I'm willing to try them on my production server, and wait to see if I get any complaints from students.
I've got a development server but it just runs as a backend for PreTeXt books. I can test there if needed though.

drgrice1:
@sean-fitzpatrick:  Are you also checking out the corresponding webwork2 pull request?  This won't work without that.  (Just checking).   If you are willing to test it on your production server, that would be great.  This pull request does do more than just change the unit handling though (and the corresponding pull requests to webwork2 and openwebwork/mathquill).  The javascript is a complete rewrite of MathQuill using modern javascript features and typescript, and built with webpack.

sean-fitzpatrick:
I'll check to see whether colleagues are comfortable with the change. Although I suppose I'll hear it pretty quickly if something breaks.

And yes, I checked out both the webwork2 PR and the pg PR.
Do I understand correctly that the PR to mathquill is just changing the javascript that gets loaded, so there's nothing to change there? Or do I also need to reinstall mathquill?

drgrice1:
When you run `npm install` from the webwork2 htdocs directory it installs the javascript from the openwebwork/mathquill pull request.  So there is nothing more you need to do for that.

sean-fitzpatrick:
OK. But I *do* need to run the `npm install` to update the version of MathQuill, if I already have it installed.

drgrice1:
You only need to run `npm install` again if the last time that you ran it was before the last commit to the webwork2 pull request.  The last two commits to that pull request change which commit from the openwebwork/mathquill pull request is installed when you run `npm install`.  In general it won't hurt to run `npm install` again.  If your installation is up to date, then it won't do anything.  But if not, then it will update everything.

drgrice1:
@sean-fitzpatrick:  When you ran `npm install` after checking out https://github.com/openwebwork/webwork2/pull/1522, did you have any issues?  What versions of node and npm do you have installed?  I am curious because I downgraded node and npm to the versions in the Ubuntu repositories, and I did have problems with this.  I made changes to both [the openwebwork/mathquill pull request](https://github.com/openwebwork/mathquill/pull/1) and the https://github.com/openwebwork/webwork2/pull/1522 to fix this.

sean-fitzpatrick:
I haven't tried this yet. Monday I'm either teaching or prepping from 9 am until 7:30 pm, except for a break for dinner before my evening class.

But now that I think about this, I might run into trouble here: my production server is still on Ubuntu 18.04, and I'm not sure how much of the npm stuff needs 20.04.

(I upgraded my PreTeXt WeBWorK server this summer, and ran into an issue: the Perl module needed to run R was not installable. Someone showed me a page with instructions to do a manual force installation when cpan fails but I haven't had time to try it. We have stats courses running on the production server so I can't risk upgrading Ubuntu until I know I can install the R module.)

Wednesday is the first day in my week that I'm not spending the day running from classroom to office hours and back again. (It's one of two days in the week that my schedule allows me enough time to eat lunch, bit I digress.)

If I'm not too buried in marking and the like, I'll give things a try then.

drgrice1:
It sounds like you never ran `npm install` after you checked out the corresponding webwork2 pull request.  If you didn't do that, then you never had the new javascript.  All of the actual fixes for mixed numbers, units, and scientific notation are in the changes to the javascript.

I tested things on Ubuntu 18.04, and discovered that `npm install` does not execute the `prepare` script for version 3.5.2 of npm (the version on Ubuntu 18.04).  So I made further changes.  The mathquill build is now included in the git repository, and running `npm install` just downloads that, and does not build mathquill as it did before.

Alex-Jordan:
I'm thinking about the precedence issues. Forgive me for not testing, but do you happen to know if the right thing happens when the thing that the student sees is typeset like `\frac{a}{\frac{b}{c}}`? Does WeBWorK still receive `a/(b/c)`?

drgrice1:
In that case WeBWorK will receive ` a/( b/c ) ` to be precise.

pstaabp (approved):
checking a number of OPL problems with units.  Looks good.

drdrew42 (approved):

PR #615, opened 2021-11-18
Modify the translator to work for both webwork2 and the standalone renderer

drgrice1:
This keys in on the environment variable $ENV{MOJO_MODE} which will always exist when run via the standalone renderer as it is a Mojolicious app.  In that case the safe compartment is cached at compile time.  Otherwise it is loaded with the creation of each PG object as needed by webwork2.

Edit:  Credit should be given to @drdrew42 for actually writing the code that loads the safe compartment at compile time for the standalone renderer.


drdrew42 (approved):
Thanks for this @drgrice1, now we can point the Renderer back at PG/main 🥳

PR #618, opened 2021-12-10
Add unit tests

pstaabp:
This sets up some unit tests for some PG macros and contexts.  It also provides a framework and examples for how to do further unit testings.

Also, there is a GitHub action to run the tests on pushes to GitHub.


README.md
```diff
@@ -1,4 +1,8 @@
-#Welcome to WeBWorK
+![main workflow](https://github.com/pstaabp/pg/actions/workflows/coverage.yml/badge.svg)
+[![codecov](https://codecov.io/gh/pstaabp/pg/branch/unit-test/graph/badge.svg?token=H7WYHBDB9S)](https://codecov.io/gh/pstaabp/pg)
+![GitHub last commit](https://img.shields.io/github/last-commit/pstaabp/pg/unit-test)
+
+# Welcome to WeBWorK
```
drgrice1, line 5: The `#Welcome to WeBWorK` line should be the first line in this file.  The badges should come after that.  The first line should be a top-level heading.  markdownlint complains about this.  Of course markdownlint complains about a lot of other things in this file too, but this is the only one added by this pull request.  You do fix one issue below!

docker/README.md
```diff
@@ -0,0 +1,63 @@
+# Docker Instructions
+
+These are instructions to get the docker image running for running the unit tests in `/t`.
```
drgrice1, line 3: There is a trailing space on this line.

docker/README.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+Here, `CODECOV_TOKEN=xxxx-xxxx-xxxx` should be adapted to your actual [token](https://docs.codecov.com/docs/quick-start). It is passed to the docker container as argument to the `-e` option.
+
+HTTP Code `200` means that the data was sent successfully to codecov and is availablle at https://app.codecov.io/gh/pstaabp/pg
+
+### Using the Shell
+
+You can also just open up the `bash` of the Docker container via
+```bash
```
drgrice1, line 46: There should be an empty line before this one.

docker/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+#### Publish Results to [`codecov.io`](https://about.codecov.io/)
+
+```bash
+docker run -it --rm --name pg-unit-test -v `pwd`:/opt/webwork/pg -w /opt/webwork/pg -e CODECOV_TOKEN=xxxx-xxxx-xxxx eltenedor/pg-no-ww cover -report codecov
+```
+
+Here, `CODECOV_TOKEN=xxxx-xxxx-xxxx` should be adapted to your actual [token](https://docs.codecov.com/docs/quick-start). It is passed to the docker container as argument to the `-e` option.
+
+HTTP Code `200` means that the data was sent successfully to codecov and is availablle at https://app.codecov.io/gh/pstaabp/pg
```
drgrice1, line 41: The URL here should be properly marked up.

docker/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Building the Docker Image Locally
+
+Execute the following command from your `pg/docker` folder
+
+```bash
+docker build -t pg-no-ww -f pg-no-dww.Dockerfile
+```
+
+### Running the Test Suite
```
drgrice1, line 60: This is a duplicate header.  Find a different header name that is not the same as the one above (or change the one above).

docker/pg-no-ww.Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM perl:5.32
+# set metadata
+LABEL maintainer="fabian.gabel@tuhh.de"
+LABEL hub.docker.com="eltenedor/pg-no-ww"
```
drgrice1, line 4: These labels should be changed to something that belongs to openwebwork or should be deleted.

t/README.md
```diff
@@ -0,0 +1,75 @@
+# Unit Tests for PG
+
+The individual unit tests are located in each of the directories.
+
+Formal unit tests are located in the the `macros` and `contexts` directories that are designed to test the pg macros and contexts respectively.
+
+## Running the tests
+
+```
```
drgrice1, line 9: The language should be specified here.  `bash` in this case.

t/README.md
```diff
@@ -0,0 +1,75 @@
...
+cd $PG_ROOT
+prove -r .
+```
+
+will run all of the tests in `.t` files within subdirectories of `t`.
+
+### Running an individual test
+
+If instead, you want to run an individual test, for example the `pgaux.t` test suite,
+```
```
drgrice1, line 19: The language (`bash`) should be specified, and this line should be preceded with an empty line.

t/README.md
```diff
@@ -0,0 +1,75 @@
...
+## the following needs to include at the top of any testing  down to TOP_MATERIAL
+
+BEGIN {
+	die "PG_ROOT not found in environment.\n" unless $ENV{PG_ROOT};
+	die "WEBWORK_ROOT not found in environment.\n" unless $ENV{WEBWORK_ROOT};
+
+	$main::pg_dir = $ENV{PG_ROOT};
+	$main::webwork_dir = $ENV{WEBWORK_ROOT};
+
+}
```
drgrice1, line 47: Markdown files don't like the hard tab usage here.  They should probably be changed to four spaces.

t/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+use lib "$main::webwork_dir/lib";
+use lib "$main::pg_dir/lib";
+
+require("$main::pg_dir/t/build_PG_envir.pl");
+
+## END OF TOP_MATERIAL
+```
```
drgrice1, line 55: The next line should be an empty line.

drgrice1 (commented):
This generally looks good.  There are several issues with the README.md files revealed by markdownlint.

It would be a good idea to mention the addition of the perl dependency `HTML::TagParser` needed by the basicmacros.t test file.

All of the implemented tests pass, but many more tests will be needed.  Lots of work to do here.

docker/pg-no-ww.Dockerfile
pstaabp: Agreed.  We should have a openwebwork organization in docker as well.  Should I do now and switch things over?

drgrice1:
If you have what is needed to create an openwebwork docker organization, then go for it.

pstaabp:
Looks like that's a pay service for organizations.  I don't see anything about non-profits, but can look into it.

pstaabp:
Cleaned up your above comments and installed a local markdown linter. :) 

Also, switched the loading of the default config file similar to what you suggested for webwork3 in `PGenvirnoment.pm`

drgrice1:
For now you can just leave the docker LABEL's until we can figure out a way to set up an openwebwork docker account.

drgrice1:
I think that you accidentally deleted the `conf/pg_defaults.yml.dist` file.  Looking at your code I think you meant to rename it to `conf/pg_defaults.dist.yml`.  Did you forget to add the renamed file?  Although there is some inconsistency in the code regarding this.  In `PGEnvironment.pm` it uses `conf/pg_defaults.dist.yml`.  However, in `.github/workflows/coverage.yml` it uses `conf/pg_defaults.yml.dist`.

pstaabp:
The `pg_defaults.dist.yml` was ignore from `.gitignore`.  That's updated now.  I also updated `coverage.yml` which doesn't need to copy that file over now.

macros/PGbasicmacros.pl
```diff
@@ -103,6 +105,8 @@ sub _PGbasicmacros_init {
 	# initialize the my variable version of $displayMode from the "runtime" version
 	# of main::displayMode
 
+	# dd "in _PGbasicmacros_init";
+
```
drgrice1, line 109: This looks like a remnant of some local debugging.  This should be deleted.

drgrice1 (approved):
There are a couple of minor clean up details.  You added the .perltidyrc file, so I suggest that you run that on all of the added files in the`t` directory.  Those new files might as well come in clean.

Once these minor clean up details are taken care of, then I will merge this.

macros/PGbasicmacros.pl
drgrice1: It looks like the only file that needs perltidying is t/build_PG_envir.pl.

pstaabp: fix.

pstaabp:
Fixed the minor error and ran perltidy.

Issue #619, opened 2021-12-11
Preview of a matrix with a variable used triggers "The evaluated answer is not an answer hash : ||." warning


taniwallach:
When a question uses a MathObject matrix to generate an `ans_array` in which answers are typed, then "Preview My Answers" does not handle answers which contain variables nicely. Instead the PG warning
```
The evaluated answer is not an answer hash : ||.
```
gets issued.

For example, in the attached question [matrixQ.pg.txt](https://github.com/openwebwork/pg/files/7697506/matrixQ.pg.txt) enter `x` into one of the input boxes and 0 in the rest. 

"Check answers" gives a more helpful feedback message like:
```
In entry (1,1): 	Your answer isn't a number
(it looks like a formula that returns a number)
```
(when the variable was in the 1,1 location in the matrix).

When an input box is left empty, preview does not trigger the warning message.

dpvc:
This turns out to be due to the fact that preview mode doesn't produce the warning message about an entry being a formula not a number, and that another part of the code uses the presence of an error message to skip further checking.  Without the error message, code was run that didn't work when the student answer is a formula.

I have made a PR (#620) to resolve the issue.

drgrice1:
I added the ArrayLayout back to htdocs/js/apps/Problem/problem.scss.

taniwallach:
Resolved in PG-2.17 by https://github.com/openwebwork/pg/pull/620

PR #620, opened 2021-12-14
Fix problem with matrix type-checking in answer preview (#619)

dpvc:
This PR resolves the problem from #619 where a constant matrix where the student enters a formula produces bad messages when the student asks for an answer preview.  It adds a test for whether a typeMatch error is produced on one of the matrix entries, and the answer is marked incorrect silently in that case rather than going on with the check and crashing while trying to determine the matrix dimensions.

This also adds some CSS on the warning messages produced during answer checking so that the background is transparent (so you get the table cell background rather than that gray that is caused by changes in the theme CSS that occurred after the MathObjects code was written).

The test case is in #619.  Without the patch, you will get the warning message indicated there.  With the patch, you should get no pink messages, just the preview of the student answer.

### DIFF
--- a/lib/Value/AnswerChecker.pm
+++ b/lib/Value/AnswerChecker.pm
@@ -201,6 +201,7 @@ sub cmp_collect {
   $ans->{student_value} = $ans->{student_formula};
   $ans->{preview_text_string} = $ans->{student_ans};
   $ans->{preview_latex_string} = $ans->{student_formula}->TeX;
+  return 0 if $ans->{typeError};
   if (Value::isFormula($ans->{student_formula}) && $ans->{student_formula}->isConstant) {
     $ans->{student_value} = Parser::Evaluate($ans->{student_formula});
     return 0 unless $ans->{student_value};
@@ -593,6 +594,7 @@ sub ans_collect {
       $ans->{typeError} = 0;
       my $result = $data->[$i][$j]->cmp(@ans_cmp_defaults)->evaluate($entry);
       $OK &= entryCheck($result,$blank);
+      $ans->{typeError} = 1 if $result->{typeError};
       push(@row,$result->{student_formula});
       entryMessage($result->{ans_message},$errors,$i,$j,$rows,$cols);
     }
@@ -624,8 +626,7 @@ sub entryMessage {
 
 sub entryCheck {
   my $ans = shift; my $blank = shift;
-  return 0 if $ans->{typeError};
-  return 1 if defined($ans->{student_value});
+  return 1 if defined($ans->{student_value}) || $ans->{typeError};
   if (!defined($ans->{student_formula})) {
     $ans->{student_formula} = $ans->{student_ans};
     $ans->{student_formula} = $blank unless $ans->{student_formula};
--- a/lib/Value/AnswerChecker.pm
+++ b/lib/Value/AnswerChecker.pm
@@ -590,6 +590,7 @@ sub ans_collect {
 	$entry = $ans->{original_student_ans};
 	$ans->{student_formula} = $ans->{student_value} = undef unless $entry =~ m/\S/;
       }
+      $ans->{typeError} = 0;
       my $result = $data->[$i][$j]->cmp(@ans_cmp_defaults)->evaluate($entry);
       $OK &= entryCheck($result,$blank);
       push(@row,$result->{student_formula});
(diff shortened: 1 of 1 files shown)


lib/Value/AnswerChecker.pm
```diff
@@ -590,6 +590,7 @@ sub ans_collect {
 	$entry = $ans->{original_student_ans};
 	$ans->{student_formula} = $ans->{student_value} = undef unless $entry =~ m/\S/;
       }
+      $ans->{typeError} = 0;
```
taniwallach, line 593: `entryCheck` seems to use the value of `typeError` from `$result`, so it is not clear to me whether the line  `$ans->{typeError} = 0;` is necessary.
Could this have adverse consequences for something else which depends on it?

taniwallach (changes requested):
The primary change (the one in `entryCheck`) certainly avoids the warning message mentioned in https://github.com/openwebwork/pg/issues/619. 

I am not certain whether the addition of line inside `ans_collect` is needed. See the comment about that line.

The CSS change makes sense and gives the warning with the same background color as the rest of the region of the table.

--- 
For the sample problem in the issue, the preview for an answer which used to trigger the warning message will now be blank. This is certainly an improvement, and is the same sort of behavior as occurs (now and before) if an entry of the matrix is left blank.

I suspect that it would be quite difficult to get "preview" to report somehow about why the answer is invalid without using a simple `warn` statement which seems unwise (as that would include the "pink" header warning message "There may be something wrong with this question. Please inform your instructor including the warning messages below.").

lib/Value/AnswerChecker.pm
dpvc: I have removed the line that sets it (I was mistakenly thinking they used the same answer hash, but they aren't the same, as you point out).

dpvc:
> I suspect that it would be quite difficult to get "preview" to report somehow about why the answer is invalid 

No, that could be done but, preview is *not* supposed to give that type of information away (that is, it should not say that the answer is a formula but should be a number, for example).  So in that case, preview should be silent about the answer being wrong, as it is for other type-match problems during preview.

I did, however, make changes so that the preview will show the typeset version when there is a type error.  That should at least make the preview do something useful in that case.  (Only answers with blank entries or ones containing syntax errors will not be typeset.)

I hope that resolves your issues.

taniwallach:
After the second commit, some tests worked and some failed. It took me some time to understand what exactly was happening.

When test the first 2 commits of this PR (by patching PG 2.16 with the changes) - the behavior depended on whether a use of a variable is made in the **lower right matrix entry** or not. If the use of a variable is made elsewhere but not in the lower right entry - once again I am seeing:
```
The evaluated answer is not an answer hash : ||.
```
when previewing the answer. and no typeset matrix is shown.

However, if I use the variable in the lower right entry, then the error message is not shown and the typeset matrix is shown.

I was able to overcome this and get the fixes to work as intended (I think) by modifying `ans_collect` so that the final value of `$ans->{typeError}` is `1` if **any** `typeError` was seen in the double loop. My approach is in https://github.com/dpvc/pg/pull/11.

taniwallach (changes requested):
@dpvc - Thanks for making the change to get the typesetting output to work more broadly in preview.
However, after the second commit, an additional change seem to be needed, as I just wrote in the discussion. I think https://github.com/dpvc/pg/pull/11 will get it to work.
Sorry, I should just have made that post a review.

dpvc:
@taniwallach: OK, I've updated the branch.  See if that is OK.

taniwallach (approved):
Works now. Sorry that the patch was convoluted. Your final version certainly works, and I don't think we really need the history of all the changes on at a time.
Maybe force push to a branch with just the final changes as a single commit?
Or maybe a squash-merge is good enough.

dpvc:
Personally, I prefer the full history in my own repositories, but whatever your convention is of PG is fine.  I do believe that the squash-merge will do as you request, though I've never used one myself.

pstaabp:
I'm not seeing this error anymore on develop.  Was this fixed in a different PR?  I'm using the example posted in #619

drgrice1:
@pstaabp:  Note that there are two issues fixed by this pull request.  The primary issue that is still seen is the issue with a preview giving the warning "The evaluated answer is not an answer hash."  The secondary issue is the display issue in the results table.

I am still seeing the primary issue with the develop branch.  The secondary issue is fixed by the css styles that have been introduced in develop.

pstaabp (approved):
This looks good now.  Was getting errors related to LaTeXImage and DragNDrop not found, but merged in develop and all is good. 

Seems like since this is only one file, a squash merge is fine.

drgrice1:
Since there are multiple authors on this one, I agree with @dpvc that this should not be squashed.

taniwallach:
I'm not sure that there is any real need to keep the details - but whatever you decide is fine.

The changes I made were totally removed by @dvpc, as the later commit no longer reset `$ans->{typeError}` back to `0` in each iteration of the loop (line 593 in the first commit https://github.com/openwebwork/pg/pull/620/commits/b35a55508d8e891bca1e022e2b0b792684348439) so do not need the approach I had found to get things to work.
As @dvpc had left it in that line and I forgotten some of the details from prior discussions in the "resolved" conversation, I mistakenly thought the "reset" was still needed inside the loop, so was collecting the desired final value in the loop and setting it after the loop ended. Since the hash being changed is not something which affects the calls made during the loop, there is no need to do that.

PR #621, opened 2021-12-26
Allow paragaph breaks within comments and code blocks

dpvc:
This PR adjusts PGML to allow paragraph breaks within comments (and within code blocks), in response to a [forum post](https://webwork.maa.org/moodle/mod/forum/discuss.php?d=6602).  Code blocks used to allow breaks, but it produced an internal error that was not being reported, so this allows it to be handled properly.

You can test using

```
loadMacros("PGML.pl");

BEGIN_PGML
Text
[% comment with

multiple paragraphs%]
more text
END_PGML
```

which should produce `Text more text` instead of 

```
Text

multiple paragraphs%]
more text
```

which is produced without the patch.

### DIFF
--- a/macros/PGML.pl
+++ b/macros/PGML.pl
@@ -274,10 +274,14 @@ sub ForceBreak {
 
 sub Par {
   my $self = shift; my $token = shift;
-  $self->End(undef, shift);
-  $self->Item("par",$token,{noIndent => 1});
-  $self->{atLineStart} = $self->{ignoreNL} = 1;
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
I have observed this behavior before as well, and thought it would be nice to be able to comment out chunks in PGML.  Good change.

pstaabp (approved):
Works as advertised and welcomed change.

drgrice1:
Thanks @dpvc.  Merging.

PR #622, opened 2021-12-28
Fix an issue with parserAssignment.pl introduced in #609

drgrice1:
See the discussion in #609 for a test case.

### DIFF
--- a/macros/parserAssignment.pl
+++ b/macros/parserAssignment.pl
@@ -347,7 +347,8 @@ our @ISA = ("Value::Formula");
 sub new {
   my $self = shift; $class = ref($self) || $self;
   my $f = $self->SUPER::new(@_);
-  bless $f, $class if $f->type eq 'Assignment';
+  return $f unless $f->type eq 'Assignment';
+  bless $f, $class;
   my $rhs = $f->getTypicalValue($f)->{data}[1];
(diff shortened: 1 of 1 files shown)


pstaabp:
It does fix this error.  

Since I haven't used this macro before, reading the documentation, if there is a single variable, there seems to supposed to be an error, but Is the example about supposed to be skipped as an assignment since there isn't a single variable on the left?

drgrice1:
The parser::Assignment::Formula new method is called for the entire string in the example, and the type of that string is not an `Assignment`.  It is resolved to a `Number` by the `Value::Formula` package.  As such it does not have the `data` hash key which an `Assignment` would have.  That causes an error on the later lines of the new method.  Basically, not everything that is passed through the new method should end up being an `Assignment`.

@dpvc can hopefully give a better explanation, and I am hoping he can take a look at this when he gets a chance.

pstaabp:
Just a heads up, should this also be added to develop?  Or is the plan to pull in any hot fixes to main back to develop at some point?

drgrice1:
Yes, this will also be added to develop once it is approved.  At this point all hotfix pull requests that have been merged have also had a duplicate pull request merged into develop.

dpvc (approved):
This looks reasonable.

PR #623, opened 2021-12-30
Update the scaffold html markup for bootstrap 5.

drgrice1:
This is paired with openwebwork/webwork2#1532 and will not work without that one.

### DIFF
--- a/macros/parserGraphTool.pl
+++ b/macros/parserGraphTool.pl
@@ -490,18 +490,27 @@ END_TIKZ
 	} elsif ($main::displayMode ne 'PTX') {
 		$self->constructJSXGraphOptions;
 		my $ans_name = $self->ANS_NAME;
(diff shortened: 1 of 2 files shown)


pstaabp:
@drgrice1 This looks nice too.  Looks like the accordion buttons are now bootstrap.  Anything else to look at for scaffolding?

drgrice1:
That is pretty much it.  This pull request doesn't have that much too it.  PG doesn't directly use bootstrap that much.  I think there are a few other things that may be needed though.  I believe there are a few macros that use the `btn` class without one of the particular flavors (like `btn-primary`, `btn-secondary`, etc).  That doesn't work with bootstrap 5.

drgrice1:
Actually, it seems that the only place that the `btn` class is directly used is in AppletObjects.pl, and I added the `btn-primary` class to that for WeBWorK-2.16.  So that is good.

pstaabp (approved):
Looks good.

PR #624, opened 2022-01-01
Add the files from webwork2 htdocs that are used by pg problems to a new htdocs directory here.

drgrice1:
All of those files have been deleted from the webwork2 repository by a corresponding pull request (openwebwork/webwork2#1536).

Note that a package.json file has been added that installs jsxgraph. So make sure to run `npm install` from the htdocs directory so that the problems that use the parserGraphTool.pl macro will work.

Also the knowl.js and knowlstyle.css files are moved to htdocs/js/apps/Knowls in the pg htdocs directory.


drdrew42 (approved):

pstaabp:
It seems like all browsers support base64 encoding/decoding with atob() and btoa() https://developer.mozilla.org/en-US/docs/Web/API/atob

Should we take advantage of this and drop the Base64.js file?

drdrew42:
`atob` and `btoa` do not play nicely with UTF-8, I believe that is why we're using Base64.js instead...

pstaabp:
> `atob` and `btoa` do not play nicely with UTF-8, I believe that is why we're using Base64.js instead...

Yep.  https://developer.mozilla.org/en-US/docs/Web/API/btoa#unicode_strings

Aww.  Why support unicode? nobody uses that on the web. 😬

drgrice1:
We have discussed that.  That won't work for knowls.  The atob and btoa methods do not support multibyte utf8 characters, and knowls need to do that.  The Base64.js methods do support encoding those things.  Although I have a pull request that rewrites knowls as we have also discussed.  That also stops base64 encoding hints and solutions.  I stopped short of removing the base64 encoding capability of knowls, but could take that the rest of the way.  The other place that base64 encoding is used is for applets (Geogebra and such), and those could use the atob and btoa methods.  In fact another pull request I have does that (and more).

Alex-Jordan:
Which PR stops the base64 encoding of hints and solutions? That could affect PreTeXt handling of live embedded problems, so I'd like to put a note in there to remind myself to investigate that before it is merged.

drgrice1:
It is a pull request that I have not submitted yet.  I am waiting on openwebwork/webwork2#1536 to be merged so that I can do it as one pull request to the pg repository, instead of two pull requests, one to each of the pg and webwork2 repositories.

drgrice1:
I will add a comment mentioning you when I put that pull request in.

drgrice1:
On the atob/btoa versus Base64.encode/Base64.decode issue:

I have discovered that atob/btoa are not even sufficient for Geogebra applet problems.  I had switched the ww_applet_support.js javascript to use the atob/btoa methods instead, and was testing Contrib/CUNY/CityTech/Calculus/setLimits_-_Introduction/interactive-align-limits.pg.  This throws a `DOMException: String contains an invalid character` when you submit an answer.

So it looks like we are stuck with the Base64.js methods.

Issue #585, opened 2021-06-21
scaffold.pl: Issue with empty answers in MultiAnswer inside scaffold


taniwallach:
Due to a student complaint, I discovered that when the first input box of a MultiAnswer group (which has `allowBlankAnswers => 1`) is left empty, then the behavior inside a scaffold section differs from that without the use of scaffolding.

**Update:** seen also with `allowBlankAnswers => 0`, and for all but the last box of the MultiAnswer group.

A sample pair of minimal problems follow below. When the left box is left empty in the scaffolded version, the warning
```
Use of uninitialized value in string ne at line 380 of [PG]/macros/parserMultiAnswer.pl
```
is issued. No such warning is issued on the non-scaffolded version.

**Update: As pointed out by @drgrice1 below, the error was actually in line 378, so lines 379-380 copied below are not relevant to the issue.**

The error message was from testing on a server running a relatively recent copy of `PG-2.16`, where lines 379-380 of `macros/parserMultiAnswer.pl` have:
```
    return if $self->{checkTypes} && $ans->{student_value}->type ne $ans->{correct_value}->type &&
              !($self->{allowBlankAnswers} && $ans->{student_ans} !~ m/\S/) ;
```

---

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
  "parserMultiAnswer.pl",
  "scaffold.pl",
  "PGcourse.pl",
);

# Print problem number and point value (weight) for the problem
TEXT(beginproblem());

# Show which answers are correct and which ones are incorrect
$showPartialCorrectAnswers = 1;

Context("Numeric");

$a = 2;
$b = 3;

$ans1 = Real($a);
$ans2 = Real($b);

# ===============================================================


$mans1 = MultiAnswer( $ans1, $ans2 )->with(
  singleResult => 0,
  allowBlankAnswers => 1,
  checker => sub {
      my ( $correct, $student, $self, $ansHash ) = @_;
      my @stud_ans = @{$student};
      my @results = ( 0,0 );

      for ( $ii = 0; $ii < 2 ; $ii++ ) {
        if ( $$correct[ $ii ] == $$student[ $ii ] ) {
          $results[ $ii ] = 1;
        }
      }
      return( @results );
  }
);



# ===============================================================


###########################################
#  The scaffold
Scaffold::Begin(
      can_open => "when_previous_correct",
      is_open  => "correct_or_first_incorrect",
      instructor_can_open => "always",
[...]

PR #625, opened 2022-01-03
Fix an unintialized value in string ne warning for parserMultiAnswer.pl.

drgrice1:
Fix an issue when a MultiAnswer is used inside a scaffold that causes a warning to be displayed about the use of an uninitialized value in a string ne.

This fixes issue #585.

### DIFF
--- a/macros/parserMultiAnswer.pl
+++ b/macros/parserMultiAnswer.pl
@@ -375,7 +375,7 @@ sub perform_check {
   foreach my $ans (@{$self->{ans}}) {
     push(@correct,$ans->{correct_value});
     push(@student,$ans->{student_value});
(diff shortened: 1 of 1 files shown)


dpvc (approved):
This looks fine.

taniwallach (commented):
@drgrice1 Thanks for fixing this.

I suppose that if someone wanted to set `$ans->{ans_message}` to be `"0"` this could make trouble for them, but such an special case is less of a problem in the big picture than students getting the pink warning box with an error message if they left some boxes empty.

I'm happy his has been merged.

drgrice1:
If we want to allow such a message, then we would need to check that `$ans->{ans_message}` is defined.  However, problems are supposed to provide meaningful feedback with these messages, so I can't see any scenario where the message "0" would be meaningful.

Issue #627, opened 2022-01-04
Answer Preview display bug when changing open/close symbols


somiaj:
Changing the open/close symbols for a matrix is not honored in the Answer Preview. Consider the following example problem

```perl
DOCUMENT();
loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
  "PGML.pl",
  "PGcourse.pl"
);
TEXT(beginproblem());
$showPartialCorrectAnswers = 1;

Context('Matrix');
$M = Matrix([ [1,2,3], [4,5,6], [7,8,9] ]);
$M->{open}="|";
$M->{close}="|";


BEGIN_PGML
[```[$M]```]

Enter in the above determinant.

        [_]*{$M}

END_PGML
ENDDOCUMENT();
```

The change to the open/closed symbols for the determinant are correctly shown everywhere except the answer preview. See the following screenshot.

![display-bug](https://user-images.githubusercontent.com/18082412/147999179-30fb865b-e21c-45b7-8759-bc8419a58e9e.png)

somiaj:
Fixed with #628.

PR #628, opened 2022-01-04
Make sure matrix parentheses are transferred to student input for matrix ans_array (#627)

dpvc:
This PR resolves the issue from #627 where the open/close delimiters that were set manually on a matrix (to show as determinant bars) are transferred to the student input generated from an `ans_array`.  See that issue for the test case.


drgrice1 (approved):
This fixes the problem.  Thanks for looking into it.

PR #631, opened 2022-01-12
Fix some warnings caused by moving WWSafe.pm to PG

drgrice1:
There are warnings about the use of an uninitialized string caused by webwork_courses_dir hash key not being set.  This key is used in lib/WeBWorK/PG/IO.pm.

### DIFF
--- a/lib/PGEnvironment.pm
+++ b/lib/PGEnvironment.pm
@@ -40,11 +40,8 @@ use YAML::XS qw/LoadFile/;
 BEGIN {
 	eval {
 		require WeBWorK::CourseEnvironment;
-		# WeBWorK::CourseEnvironment->import();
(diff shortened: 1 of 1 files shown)


drgrice1:
By the way, to see the warnings look at the apache2 error log when you render a problem via the html2xml end point.  This will show `Use of uninitialized value $dir in pattern match (m//) at /opt/webwork/pg/lib/WeBWorK/PG/IO.pm line 285.`

pstaabp:
I'm seeing that error in the apache log.  Just to clarify, does `html2xml` get called when rendering in `ProblemSetDetail`?

pstaabp:
I just found it rendering in the LibraryBrowser and this fixes it.

drgrice1:
Yes, html2xml is used when rendering in `ProblemSetDetail.pm` or in `SetMaker.pm`.

Issue #632, opened 2022-01-14
PopUp menus block Value::Error from MultiAnswer problems.


somiaj:
I noticed that `Value->Error("message")` are not displayed from `parserMultiAnswer.pl` when using a PopUp menu.

```perl
DOCUMENT();
loadMacros(
    "PGstandard.pl",
    "PGML.pl",
    "parserPopUp.pl",
    "parserMultiAnswer.pl",
);
$showPartialCorrectAnswers = 1;
TEXT(beginproblem());

Context("Numeric");
$ans1 = Real("pi");
$ans2 = Real("e");
$ans3 = PopUp(["Choose","yes","no"],"yes");
$ans4 = PopUp(["Choose","yes","no"],"no");

$ma1 = MultiAnswer($ans1, $ans2)->with(
    singleResult => 1,
    allowBlankAnswer => 1,
    checkTypes => 1,
    format => "%s,%s",
    tex_format => "%s,%s",
    checker => sub {
        my ( $correct, $student, $self ) = @_;
        Value->Error("testing");
        return 0;
    }
);
$ma2 = MultiAnswer($ans1, $ans3)->with(
    singleResult => 1,
    allowBlankAnswer => 1,
    checkTypes => 1,
    format => "%s,%s",
    tex_format => "%s,%s",
    checker => sub {
        my ( $correct, $student, $self ) = @_;
        Value->Error("testing");
        return 0;
    }
);
$ma3 = MultiAnswer($ans3, $ans4)->with(
    singleResult => 1,
    allowBlankAnswer => 1,
    checkTypes => 1,
    format => "%s,%s",
    tex_format => "%s,%s",
    checker => sub {
        my ( $correct, $student, $self ) = @_;
        Value->Error("testing");
        return 0;
    }
);

BEGIN_PGML

+ MultiAnswer with two numbers: [_]{$ma1} [_]{$ma1}

+ MultiAnswer with one popup and one number: [_]{$ma2} [_]{$ma2}

+ MultiAnswer with two popups: [_]{$ma3} [_]{$ma3}

END_PGML

ENDDOCUMENT();
```

The second two MultiAnswer boxes don't display the Value->Error message.

somiaj:
Fixed with #633.

PR #633, opened 2022-01-14
Make sure MultiAnswer uses the correct context (#632)

dpvc:
This PR resolves the issue raised in #632, where error messages sometimes aren't reported.  The problem was that the wrong context was being checked.  The errors are stored in the context object, and with MultiAnswer, where the answers can use different contexts, the error message needs to be set in the proper one.  The test case from #632 uses a PopUp, and that has its own context (to store the string values of the pop-up), and that context was being left as the active one.  The solution is to set the context to that of the MultiAnswer (just as is usually done in the standard MathObject `cmp_parse` method, which is not used by the MultiAnswer object, which has its own `single_cmp` and `entry_cmp` methods).

### DIFF
--- a/macros/parserMultiAnswer.pl
+++ b/macros/parserMultiAnswer.pl
@@ -386,7 +386,7 @@ sub perform_check {
 
   Parser::Context->current(undef,$context);    # change to multi-answser's context
   my $flags = Value::contextSet($context,$self->cmp_contextFlags($ans)); # save old context flags
-  $context->{answerHash} = $rf_ans;            # attach the answerHash
+  $context->{answerHash} = $rh_ans;            # attach the answerHash
   my @result = Value::cmp_compare([@correct],[@student],$self,$rh_ans);
   Value::contextSet($context,%{$flags});       # restore context values
   $context->{answerHash} = undef;              # remove answerHash
--- a/macros/parserMultiAnswer.pl
+++ b/macros/parserMultiAnswer.pl
@@ -370,7 +370,8 @@ sub entry_check {
 #
 sub perform_check {
   my $self = shift; my $rh_ans = shift;
-  $self->context->clearError;
+  my $context = $self->context;
+  $context->clearError;
   my @correct; my @student;
   foreach my $ans (@{$self->{ans}}) {
     push(@correct,$ans->{correct_value});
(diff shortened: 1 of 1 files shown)


somiaj:
This patch fixes the issue for me. Thanks.

pstaabp:
Also verified that this fixes this issue.

macros/parserMultiAnswer.pl
```diff
@@ -382,8 +383,15 @@ sub perform_check {
   my $inputs = $main::inputs_ref;
   $rh_ans->{isPreview} = $inputs->{previewAnswers} ||
                          ($inputs_{action} && $inputs->{action} =~ m/^Preview/);
+
+  Parser::Context->current(undef,$context);    # change to multi-answser's context
+  my $flags = Value::contextSet($context,$self->cmp_contextFlags($ans)); # save old context flags
+  $context->{answerHash} = $rf_ans;            # attach the answerHash
```
drdrew42, line 389: shouldn't this be `$rh_ans`?

drdrew42, line 389:
```suggestion
  $context->{answerHash} = $rh_ans;            # attach the answerHash
```

taniwallach: The patch as is (with the typo on line 389) fixed the issue for me and the prior people who tested it.
The next line which calls `Value::cmp_compare` provides the answerHase `$rh_ans` explictly.
As such, maybe there is no reason to provide the `answerHash` in the context one line 389 and then to remove it on line 392?
At least in the test case provided in https://github.com/openwebwork/pg/issues/632 dropping both those lines from the proposed change seems to still fix the problem.

taniwallach:
Note also: if an input box is left empty then the error message will not be triggered even though the test problem from https://github.com/openwebwork/pg/issues/632 has `allowBlankAnswer => 1` set.

(I was testing with my revised version of https://github.com/openwebwork/pg/pull/620 also installed, but I do not think that matters.)

taniwallach:
There is a side-effect of this change (at least the version without adding/removing the answerHash to the context), and it can make problems that worked before the change fail to work now. Those problems were not really correct in that they were not changing context properly, but the misbehavior of contexts with MultiAnswer let them work.

The point of this post is mainly to provide a record of the problem in case someone else runs into something similar.

For example, the attached PG problem [t_R3_to_R4_seq01_eval_general_v.pg.txt](https://github.com/openwebwork/pg/files/7920872/t_R3_to_R4_seq01_eval_general_v.pg.txt) was working before the change, but after the change was reporting:
```
Formulas from different contexts can't be compared
```
The code was fixed by moving the line where the context is change to `Numeric` to before the answers are defined.

drgrice1:
Why does that problem even change contexts?  As far as I can see, it isn't needed at all.

taniwallach:
> Why does that problem even change contexts? As far as I can see, it isn't needed at all.

Quote probably not. It was coded in the first semester I was working with WeBWorK. The original intention was to have the answer be a matrix of formula's and use `ans_array` on that matrix. After starting to code the question I discovered that a MathObject matrix with functions as elements is treated as a Formula whose value is a Matrix and not as a matrix of formulas. That lead to reworking the code so the answers are single formulas and laid out visually with `display_matrix`. It then seemed that then the proper context at the end was the Numeric context. Since the code of the problem worked - it hung around as it was. When it broke today after putting the patch from this PR onto a production server, I realized that the issue was almost certainly caused by a bad interaction of the problem with the change made by the code in this PR.

I posted, as I suspect that there may be other people who wrote sloppy code which happened to work due to the bug which this PR fixes and which might break in some manner or another due to the change in the context used during the grading.

drdrew42 (approved):
> There is a side-effect of this change (at least the version without adding/removing the answerHash to the context)

I can confirm that replacing `$rf_ans` with `$rh_ans` still encounters this side-effect: "formulas from different contexts can't be compared".

I'm inclined to merge this PR and let it 'break' problems that maybe should not have worked before...

drgrice1:
The `rf_ans` typo at least needs to be fixed first.

pstaabp:
> I'm inclined to merge this PR and let it 'break' problems that maybe should not have worked before...

The change was made. I'll pull in and we'll go from here.

Issue #634, opened 2022-01-18
reduce incorrectly removes parenthesis in mathobject string.


somiaj:
When using reduce (with a double negative) parenthesis are removed from the `->string`. Here is my test case

```
DOCUMENT();
loadMacros(
  "PGstandard.pl",
  "PGML.pl",
);
TEXT(beginproblem());

$a = Formula("-4x^2")->reduce;
$test1 = Formula("e^(-$a)")->reduce;
$test2 = Formula("e^(-$a)");

BEGIN_PGML
**Test 1:**
+ String: [$test1->string]
+ TeX: [``[$test1]``]

**Test 2:**
+ String: [$test2->string]
+ TeX: [``[$test2]``]
END_PGML

ENDDOCUMENT();
```
This causes the `->string` to loose the parenthesis , but `->TeX` is fine.

![Screenshot if issue.](https://user-images.githubusercontent.com/18082412/150003499-3af420e6-0002-44db-82ab-3dc5c638a40c.png)

I think this is related to a similar bug, but it wasn't fully fixed in issue #455 and the pull request #456.

dpvc:
This turns out to be a bad interaction between two different reduction rules.  I have made a pull request to fix it (linked above).

somiaj:
Thanks, I can confirm this fixes the issue.

I notice you only have a pull request for `develop`, this bugfix seems reasonable to fix in `main` as well (imo).

dpvc:
I will let others make that call.  I'm not up on the release process any more.  If so, then the PR should be rebased to `main` and targeted there, and then targeted to `develop` as well.  But that may not be how things are done any longer.

drgrice1:
@dpvc:  That is pretty much it.  Rebase and retarget.

I am not sure if this should be a hotfix or not.  I am not certain that is so pressing.  It is really a display issue, and those are usually things that we defer to the next release.

What do others think?

somiaj:
@drgrice1 The reason I think it should be put in main is because although the original mathobject is correct, the string is used to build future mathobjects, so any mathobject built using this mathobject, such as `$test3 = Formula("5 + $test1")` will be incorrect (which is how I ran across this issue in the first place). I can patch locally, I just thought this might be pressing enough due to this.

dpvc:
[...]

PR #635, opened 2022-01-18
Remove noParens marker from operand (if any) when reducing double negatives. #634

dpvc:
This PR resolves a problem with missing parentheses after reducing a double negative involving a multiplication.  This was due to an interaction between the rules for `(-x)*y` and `-(-x)`.  The first rule technically produces `-(x*y)`, but marks the multiplication as not needing parentheses, and the second just returns `x`, so when used to reduce `-(-4*x^2)`, the inner `-4*x^2` (which is `(-4)*x^2)`) becomes `-(4*x^2)` with the multiplication marked for no parentheses, and then `-(-(4*x^2))` becomes just `4*x^2` market as needing no parentheses.  This is why `e^(-(-4*x^2))` produced `e^4*x^2` as its string output (though it was still structurally correct internally).

The easiest fix is to remove the `noParens` marker in the `-(-x)` reduction rule, as is done here.

Resolves issue #634.

### DIFF
--- a/lib/Parser/UOP/minus.pm
+++ b/lib/Parser/UOP/minus.pm
@@ -29,7 +29,10 @@ sub _eval {-($_[1])}
 sub _reduce {
   my $self = shift; my $op = $self->{op};
   my $reduce = $self->{equation}{context}{reduction};
-  return $op->{op} if $op->isNeg && $reduce->{'-(-x)'};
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
This fixes the issue.

drdrew42 (approved):

PR #639, opened 2022-01-20
Fix the drag and drop macro for bootstrap 5 and some other improvements and issue fixes

drgrice1:
Add a prefix to the css classes used by the javascript and css used for the DragNDrop macros.  The existing classes were very likely to clash with other classes and cause problems.  Note that the bootstrap 2 `hidden` class was used, and that class no longer exists for bootstrap 5.  I could have used the bootstrap 5 equivalent `d-none`, but decided to go with something that doesn't depend on an external library (although `btn` and `bnt-secondary` classes are used -- so the dependency is really still there).

Switch the `a` elements used for buttons to `button` elements as they should be in modern html markup.

The style is improved to be responsive to the screen width.

Finally, fix some warnings that were appearing in the error log.  See lines 210 and 299 of draggableSubsets.pl, and lines 210, 216, 248, 250, 253, and 312 of draggableProof.pl.

### DIFF
--- a/htdocs/js/apps/DragNDrop/dragndrop.css
+++ b/htdocs/js/apps/DragNDrop/dragndrop.css
@@ -1,8 +1,17 @@
+.dd-bucket-pool {
+	display: flex;
+	flex-direction: row;
+	flex-wrap: wrap;
+	justify-content: space-evenly;
+	gap: 1rem;
+	margin-bottom: 1rem;
+}
+
 .dd-container {
+	display: flex;
+	flex-direction: column;
 	width: 350px;
-	float: left;
-	margin: 10px !important;
-	padding: 0;
+	padding: 0.5rem;
 	color: #000000;
 	border: 1px solid #388E8E !important;
 	border-radius: 5px;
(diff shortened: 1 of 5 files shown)


pstaabp (approved):
Looks good.  I don't see the warnings anymore and this is nicely responsive.

pschan-gh:
Nice css improvements! Before, it always bothered me that the drag-n-drop "buckets" could move around as items are taken in and out of them. They don't do that anymore, so that's great.
I have begun writing a preliminary update of dragndrop.js which makes use of sortablejs (and also weaned off jquery). My question is: If a drag-n-drop problem is to use the sortablejs library bundled with webwork2, how should a pg macro access it with ADD_JS? Thank you!

drgrice1:
Thanks.  Flex is a very useful css tool.

Now, when I told you that the jquery.nestable.js library needed to be replaced, I thought at the time that it depended on jquery-ui.  I realize now that it does not.  It only depends on jquery.  That isn't as big of a deal.  Furthermore, I realized that this library is touch friendly.  If it were jquery-ui it would not be.  So there is less urgency to replace the jquery.nestable.js library.  Perhaps it still should be done though as that library is unmaintained.  The sortablejs library is maintained better.

In any case, the way to add the javascript would be
```
ADD_JS_FILE('node_modules/sortablejs/Sortable.min.js', 0, { defer => undef });
```

pschan-gh:
I see, thanks! 
It appears migrating to sortablejs requires only changing two lines in dragndrop.js. Further testing is required, but so far it seems to work quite well.
Incidentally, a colleague suggested that it might be useful to be able to use a statement in a draggableProof problem more than once. While I have no plan to incorporate this feature in the new future, the "cloning" functionality of sortablejs seems to readily accommodate this. So, I think migrating to sortablejs might be a good move to set the stage for further customizations.

PR #640, opened 2022-01-20
Change the way that coloring of inputs is done for correct and incorrect answers.

drgrice1:
Instead of creating a global javascript variable, use data attributes.  This is done in the paired pull request openwebwork/webwork2#1551 in the AttemptsTable.  That pull request also defers loading of the javascript as is needed.

Also, instead of using inline javascript to make clicking on the correct/incorrect link set focus on the input, add an event listener.
This is done from htdocs/js/apps/InputColor/color.js, so perhaps that file is not properly named anymore.

Note that if MathQuill is enabled this is handled for mathquill inputs in htdocs/js/apps/MathQuill/mqeditor.js.  It did so before, but in a hackish way.  Now it is much cleaner.

As with all of these changes this is also designed to work with webwork3 and the standalone renderer.


pstaabp:
Definitely checked out both branches, ran npm install and cleared the caches.  I can even verify directly looking the code in the browser that it has the new js code.

drgrice1:
@pstaabp:  For the problems that are not begin colored, do those problems happen to have `$showPartialCorrectAnswers = 0` in them?  If that is the case, then there will be no coloring of the inputs for that problem, regardless of if the answers are correct or not.  That is true both with and without this pull request.

pstaabp:
I've attached a quiz that generates the problem.  Now the first and 4th problems are highlighted correctly, but the 2nd and 3rd have no coloring.

[setquiz1.def.txt](https://github.com/openwebwork/pg/files/7951330/setquiz1.def.txt)

drgrice1:
@pstaabp: So this is exactly what I said above.  Both of problems 2 and 3 in this quiz set the variable `$showPartialCorrectAnswers = 0`,  and both of problems 1 and 4 set `$showPartialCorrectAnswers = 1`.  So both with this pull request and without this pull request there will be no coloring of inputs for problems 2 and 3, while there will be for problems 1 and 4.

Note that the default value of `$showPartialCorrectAnswers` is 1, so any problem that does not set that variable at all will still get coloring of inputs.

In any case, what you are seeing is not a change caused by this pull request, and that is intended behavior for these problems.

pstaabp:
> @pstaabp: So this is exactly what I said above.

Sorry, I didn't read that carefully.  Then I think this looks good.

pstaabp (approved):

drdrew42 (approved):
Looks good -- only minor feedback is that radio button highlighting looks a little strange (highlighting all options).

drgrice1:
@taniwallach's commented to me in a private Slack conversation exactly the opposite on regards to the behavior before this pull request, and I agree with him.  Before this pull request only the first radio button is highlighted.  Highlighting only the first radio button is misleading to students as it may be taken to be an indication of that particular choice being the correct or incorrect answer (also depending on which one the student has chosen).

drdrew42:
Eww, no. I wouldn't want only the first one highlighted either. Highlighting all is preferable to highlighting the first option only -- but IDK if there's any easy way to ID which option should get the highlight? e.g. student picks B, submits answers,  B is incorrect, radio button for B glows red.

drgrice1:
I agree that that would be best.  That should be achievable now that I think about it.  I will try to work that in.

There is another issue with radio buttons that I couldn't figure out a way to resolve.  It only applies to Google Chrome, but the highlighting does not look the best there for radio buttons.  It is workable, but there is a white square around the circular radio button and inside the highlighting.  I found something about this in a Google search, but didn't find a workaround.

drgrice1:
In the case that the student picks no answer, I think the only way will be to highlight all though.

drgrice1:
@drdrew42:  Now it colors radio buttons the way we discussed.  If one is checked, then that is the only one that is highlighted.  If none are checked all are highlighted.

drdrew42 (approved):
Awesome! 🥳

PR #642, opened 2022-01-24
Update Geogebra applet support for the new API as well as for future webwork3 usage.

drgrice1:
This removes flash, java, and canvas applet support.  Flash and java are dead anyway.

Sorry @mgage, but canvas applet support is not working.  When I said it was the other day, I clearly hadn't looked closely.  The problem you gave me would grade everything as correct regardless of the answer given.  If someone wants to make this work again they are welcome to do so, but for now this is removed.

I believe this is the last of what will be needed to make the pg javascript work for webwork3.  Now the standalone renderer needs to be updated to use the javascript here.

Note that debugMode is changed.  The visible state inputs are removed.  Browser developer tools provide better mechanisms for monitoring that if desired.  Setting debugMode to 1 or 2 enables console logging, 2 with more verbosity.


drgrice1:
@pstaabp: I added that.

pstaabp:
This looks great when viewing as a problem set.  I looked at this in the library--thinking this where instructors might be looking for problems and I guess we've suppressed errors in the library browser.  For example, I went to WeBWorK->demos->Flash Applets in the OPL and there are 4 problems there and clearly the applet is missing, but not errors.  

Eventually we either want to 1) deprecate these problems and show in the browser that is true  and 2) eventually remove it.  I'm wondering if there is a simple way to show this in the Library Browser currently.

drgrice1:
@pstaabp:  I added the warnings to the output in the Library Browser in openwebwork/webwork2#1567.  If you test that pull request with this one, you will see the warnings added here.  If you don't use this pull request, you will still see warnings but they are not as nice.

drgrice1:
> Eventually we either want to 1) deprecate these problems and show in the browser that is true and 2) eventually remove it. I'm wondering if there is a simple way to show this in the Library Browser currently.

1 is not an option.  Deprecating completely broken problems that can not be fixed is just inane.  2 is the only thing that works for flash applet problems.

drgrice1:
By the way, we were not suppressing errors in the library browser.  They just have never been shown there.  Before I switched to using the html2xml method, only the problem text was shown, but no errors.  Note that these flash problems did not work in the library browser then, even when flash problems did work.  When I switched to the html2xml method, the errors were still not shown because the formatRenderedProblem did not give those errors in the output with the simple format.  I remedied that with openwebwork/webwork2#1567 of course.

pstaabp (approved):
This looks great now.

drdrew42:
Warnings in the Library Browser appear as described... but for some reason, I'm getting `'require' trapped by operation mask` errors when I attempt to view flash problems in a problem set. The stack points back to lib/WeBWorK/Apache.pm L#93 where we UTF-8 encode our captured perl warnings. What am I missing here? I've got Encode loaded into the safe through defaults.config (which imports Encode twice, actually).

drgrice1:
Are you using a problem from the OPL?  If so which one?  I tested with Library/CSUOhio/appletDemonstrationProblems/graphLimits.pg (and several others), but perhaps there is something different with the one you are testing with.  With this problem I get the warning
```
Flash applets are no longer supported at line 49 of [PG]/macros/AppletObjects.pl
```

drgrice1:
I believe I have tested all of the applet problems in the OPL now rendered by Problem.pm, as well as on the ProblemSetDetail.pm page.  I don't get your error with any of them.

drdrew42:
I added multiple files from the same directory, and double checked with the specific one you mentioned -- I'm still getting the `'require' trapped by operation mask` error.

I'm sure it's got to be some misconfiguration, but I cannot identify where. There are plenty of other places where Encode is used on warnings and I'm not seeing this error elsewhere. 😠

drdrew42:
Well, I guess that's what I get for using an old WW install for a develop machine. I updated Encode from CPAN and things work as expected.

drdrew42 (approved):

PR #643, opened 2022-01-26
Fixed Gauss unit numerical factor

nick-chura:
I noticed that WeBWorK's definition of the unit Gauss has a numerical factor of 1E-5, but it looks like that should be 1E-4 according to [this](https://en.wikipedia.org/wiki/Gauss_(unit)) and [this](https://link.springer.com/content/pdf/bbm%3A978-3-540-74296-8%2F1.pdf).

### DIFF
--- a/lib/Units.pm
+++ b/lib/Units.pm
@@ -744,7 +744,7 @@ our %known_units = ('m'  => {
                            's'         => -2,
                          },
                 'G' => {
-                           'factor'    => 1E-5,
+                           'factor'    => 1E-4,
                            'kg'        => 1,
                            'amp'       => -1,
                            's'         => -2,


Alex-Jordan:
Looks good to me. (What are you working on where you saw this?!)

One thing though. Can you do this again from the `develop` branch instead of `master`? I tried a simple switching in GitHub but that did not do it nicely.

drgrice1:
It will need to be rebased.

nick-chura:
I worked on something similar for IMathAS, and someone pointed out that I had an incorrect definition for Gauss. Having consulted the WeBWorK units file for many of the unit definitions I used, I checked back here and saw the issue.

It's a single character change, so maybe it makes sense to just have you deny this PR and add that change to your develop branch?

drgrice1:
If you run `git rebase origin/develop` and then `git push --force`, then this pull request can be re-targeted into the develop branch.

Or we can put in a pull request that makes this change for you.

nick-chura:
Done -- thanks for the Git commands. Let me know if anything else needs to be adjusted on my end.

drgrice1 (approved):
Looks good.  Thanks for finding this.

Issue #644, opened 2022-01-28
Parser Assignment Issues after PR #609


drgrice1:
I think the change in pull request #609 may still need more work.  Prior to that pull request problems like the following would work:
```perl
DOCUMENT();

loadMacros("PGstandard.pl", "MathObjects.pl", "PGML.pl", "parserAssignment.pl");

parser::Assignment->Allow;

$ans = Formula("x = 5");

BEGIN_PGML
Enter [`[$ans]`].

[_]{$ans}{10}
END_PGML

ENDDOCUMENT();
```
Prior to #609 you could enter `x=5` and it would be graded as correct.  Now it fails to parse and gives the warning `The evaluated answer is not an answer hash : ||.`

You can change `Formula` to `Compute` and this works, but it seems that it should work with both.  At least it did before.

Note that if you add what was line 357 before this pull request back before the added code then this works again.  That is insert
`$other = $other->Package("Formula")->new($self->context,$other) unless $other->isFormula;`
before the current line 360.

@drdrew42 and @dpvc:  What do you think is the correct thing here?

pstaabp:
Fixed again.  Hopefully!

PR #645, opened 2022-01-29
Fix another regression caused by #609.

drgrice1:
If a problem sets the answer to be a formula with an assigment, (something like `Formula("x = 5")` then all answers are counted
incorrect and a warning is displayed that the evaluated answer is not an answer hash.  This worked prior to #609.  The cause of this was the removal of a line that should not have been removed that converts the other answer being compared to to a Formula if it is not already.

Note that the original issue that was attempted to be fixed by #609 still is fixed with this change.

This fixes issue #644 that I just submitted about this.

Attached are minimal working examples for the original issue as well as for this one.
[FormulaAssignment.pg.txt](https://github.com/openwebwork/pg/files/7962769/FormulaAssignment.pg.txt)
[ParserAssignmentMWE.pg.txt](https://github.com/openwebwork/pg/files/7962770/ParserAssignmentMWE.pg.txt)
See issue #607 for details on how to test ParserAssignmentMWE.pg, and issue #644 for details on testing FormulaAssignment.pg.

Note that this is definitely a hotfix.  This certainly brings in to question that #609 should have been a hotfix, considering that this is something that has been this way since 2008.

### DIFF
--- a/macros/parserAssignment.pl
+++ b/macros/parserAssignment.pl
@@ -357,6 +357,7 @@ sub new {
 sub typeMatch {
   my $self = shift; my $other = shift; my $ans = shift;
   return 0 unless $self->type eq $other->type;
+  $other = $other->Package("Formula")->new($self->context,$other) unless $other->isFormula;
   my $typeMatch = $self->getTypicalValue($self)->{data}[1];
   $other = $self->getTypicalValue($other,1)->{data}[1];
   return 1 unless defined($other); # can't really tell, so don't report type mismatch


somiaj:
I can confirm this fix works. I just ran across a very similar issue where if I the answer was `Compute("y=5+2x")` and the student entered in just `y=5` as their answer it produced the error described in #644. The grader would would still accept the correct answer (and also not error out if any variable was used, like `y=3t`, but be marked incorrect).

drgrice1:
@somiaj:  I can confirm the behavior you describe with the ParserAssignmentMWE.pg problem that I attached above.  If you enter `y=5` it gives the evaluated answer is not an answer hash.  This pull request fixes that also.

pstaabp (approved):
Having trouble getting my main branch running well, but the develop branch looks good and noticed that this does restore the line that was deleted.

drdrew42 (approved):

PR #646, opened 2022-01-29
Fix another regression caused by #609 (for develop).

drgrice1:
If a problem sets the answer to be a formula with an assigment, (something like `Formula("x = 5")` then all answers are counted
incorrect and a warning is displayed that the evaluated answer is not an answer hash.  This worked prior to #609.  The cause of this was the removal of a line that should not have been removed that converts the other answer being compared to to a Formula if it is not already.

Note that the original issue that was attempted to be fixed by #609 still is fixed with this change.

This fixes issue #644 that I just submitted about this.

This is #645 for develop.

### DIFF
--- a/macros/parserAssignment.pl
+++ b/macros/parserAssignment.pl
@@ -357,6 +357,7 @@ sub new {
 sub typeMatch {
   my $self = shift; my $other = shift; my $ans = shift;
   return 0 unless $self->type eq $other->type;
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
This does seem to fix this error and I did look back in the history and see that the line you added is restored.

drdrew42 (approved):

PR #651, opened 2022-02-05
Moving configuration and files to PG

pstaabp:
This is a draft PR to get a discussion going plus some help.  Here's some big picture ideas about this

- expansion of `PGEnvironment` module to 1) read the `conf/pg_defaults.yml` file 2) process much of it (say to build up nested directories)
- move pg-related configuration from `defaults.config` (on ww2 side) into `conf/pg_defaults.yml`
- addition of a standaloneRenderer code (in `bin`) from https://github.com/mgage/standaloneProblemRenderer and adaption of it to only use modules from pg-side.  
- To get standaloneRenderer working, need to move many modules from webwork side to pg side. 
- All of the modules on pg side should be independent of webwork.  The main thing needed is to replace `CourseEnvironment` with `PGEnvironment` and update the path to get configurations. 

A few things I'm thinking about/having trouble with probably because of my lack of knowledge of the way pg works in detail: 
- I think I have got Localization compiling on pg-side (though not tested).  I copied all of the .po files from ww side.  I think only the pg side things need to be there. 
- I commented out some things about the `Delayed Mailer`.  It seems like that shouldn't be on the pg-side.  Can we get this back to webwork2. 
- In `FormatRenderedProblem` (now on pg side), there are a number of calls to `$ce->{...}` which don't seem to exist in the code.  Is this automatically added by LTI ?  
- In `FormatRenderedProblem`, there is information about the theme?  Is this really needed?  
- Perhaps we need two different `FormatRenderedProblem` code (one on each side) or a wrapper on the webwork side?

To play with this so far, running `bin/standaloneRenderer.pl` on a problem with various flags.  It doesn't finish, but runs a lot of code.  

This clearly isn't done and once we can get the standalone renderer working, need to remove modules from webwork2 side.

Please let me know if there is something I have clearly not done well.  Also, I think there might be other modules needed to be moved to PG.


pstaabp:
A new draft.  First, in an attempt to clean up and clarify packages, there is now a `lib/Renderer` directory where I'm hoping to keep all packages related to rendering a problem.   This touches a lot, but I think helps separate in our minds what is PG. 

Although not currently working (hence a draft PR), the meat of this is in `Renderer::Problem`.  You create a problem passing in a problem source (not a file path), and then render the problem.  The idea will be to get the results as a hash of everything we typically expect back. Currently, this isn't working.  There's some issues with the safe compartment, but am pushing this up to get some feedback. 

I have a very simple script `bin/test_render_problem.pl` that loads in a problem from the OPL and attempts to set it up to render.  Currently I'm getting an error:
```
Can't locate object method "current" via package "Parser::Context" (perhaps you forgot to load "Parser::Context"?) at /opt/webwork/pg/macros/PG.pl line 16.
```

htdocs/package.json
```diff
@@ -12,8 +12,9 @@
         "prepare": "npm run generate-css"
     },
     "dependencies": {
+        "bootstrap": "^5.1.3",
```
drgrice1, line 15: No.

pstaabp: I think I added that before my latest round of changes.  I'm trying to strip out as much html/css/js as possible.

lib/Renderer/FormatRenderedProblem.pm
```diff
@@ -0,0 +1,591 @@
...
+
+	say "\$session_key";
+	print Dumper $session_key;
+
+	say "\$STRING_Preview";
+	print Dumper $STRING_Preview;
+	my $STRING_Preview     = '';
+	my $STRING_Submit      = '';
+	my $STRING_ShowCorrect = '';
+
```
drdrew42, line 466:
```suggestion
```
This looks like debug code, and why set these values to empty string?

lib/Renderer/FormatRenderedProblem.pm
```diff
@@ -0,0 +1,591 @@
...
+	# Interpolate values into the template
+	$template =~ s/(\$\w+)/$1/gee;
+
+	return $template unless $self->{inputs_ref}{send_pg_flags};
+	return JSON->new->utf8(0)->encode({ html => $template, pg_flags => $rh_result->{flags} });
+}
+
+sub saveGradeToLTI {
```
drdrew42, line 474: Is this expected to be PG's responsibility?

lib/Renderer/PG.pm
```diff
@@ -0,0 +1,570 @@
...
+
+use strict;
+use warnings;
+# use WeBWorK::Debug;
+use Renderer::Localize;
+use Renderer::ImageGenerator;
+use Renderer::Utils qw(runtime_use formatDateTime makeTempDirectory);
+use Renderer::Utils::RestrictedClosureClass;
+
+use Data::Dumper;
```
drdrew42:
```suggestion
```

lib/Renderer/PG.pm
```diff
@@ -0,0 +1,570 @@
...
+			'refresh');
+	}
+
+	if (defined $extras->{mailer}) {
+		#my $rmailer = new Renderer::Utils::RestrictedClosureClass($extras->{mailer},
+		#	qw/Open SendEnc Close Cancel skipped_recipients error error_msg/);
+		#my $safe_hole = new Safe::Hole {};
+		#$envir{mailer} = $safe_hole->wrap($rmailer);
+		# $envir{mailer} = new Renderer::Utils::RestrictedClosureClass($extras->{mailer}, "add_message");
+	}
```
drdrew42:
```suggestion
```

lib/Renderer/Problem.pm
```diff
@@ -0,0 +1,480 @@
...
+=cut
+
+# This is derived from the openwebwork/renderer branch.
+
+package Renderer::Problem;
+
+use PGEnvironment;
+use Renderer::Translator;
+use Renderer::Localize;
+use Data::Dumper;
```
drdrew42:
```suggestion
```

lib/Renderer/Problem.pm
```diff
@@ -0,0 +1,480 @@
...
+	}
+	$errorno = 0;
+	return ($answer, $errorno);
+}
+
+sub nullSafetyFilter {
+	return shift, 0;    # no errors
+}
+
+sub defineProblemEnvir {
```
drdrew42: isn't all this duplicating PG_env?

lib/Renderer/IO.pm
```diff
@@ -40,30 +40,30 @@ BEGIN {
...
 	our %SHARE = map { $_ => __PACKAGE__ } @SHARED_FUNCTIONS;
-	my $ww_version = "2.x";  # hack -- only WW2 versions are supported.
+	my $ww_version = "2.x";    # hack -- only WW2 versions are supported.
 	if (defined $ww_version) {
 		my $mod;
 		for ($ww_version) {
-			/^1\./          and $mod = "WeBWorK::PG::IO::WW1";
-			/^2\./          and $mod = "WeBWorK::PG::IO::WW2";
-			/^Daemon\s*2\./ and $mod = "WeBWorK::PG::IO::Daemon2";
+			/^1\./          and $mod = "Renderer::IO::WW1";
```
drdrew42, line 61: Isn't WW1 being deleted in this PR?

lib/Renderer/Localize.pm
```diff
@@ -0,0 +1,68 @@
+package Renderer::Localize;
+
+use base qw/Locale::Maketext/;
+
+use strict;
+use warnings;
+
+use feature "say";
```
drdrew42, line 8:
```suggestion
```

lib/Renderer/Localize.pm
```diff
@@ -0,0 +1,68 @@
...
+use Locale::Maketext::Lexicon {
+	# '*' => [Gettext => "/Users/pstaab/Downloads/test_localization/loc/*.po"],
+	'en'      => [ Gettext => "/opt/webwork/pg/lib/Renderer/Localize/en.po" ],
+	_auto     => 1,
+	_decode   => 1,
+	_encoding => undef
+};
+
+use Data::Dumper;
+
```
drdrew42, line 21:
```suggestion
```

drdrew42 (changes requested):

pstaabp:
Closed in favor of #709 and #726

PR #653, opened 2022-02-09
Point tool

Alex-Jordan:
I believe I did as @drgrice1  described at https://github.com/openwebwork/webwork2/pull/1515#issuecomment-1032595502. I made branch locally that merged your graphTool branch, then pushed it to GitHub to open this PR. I haven't tested this out. For anyone who would like to test, these attached problems are in use at my school under the previous implementation of pointTool.

[pointToolExamples.tar.gz](https://github.com/openwebwork/pg/files/8029311/pointToolExamples.tar.gz)


drgrice1:
It looks good.  Now I can put in my pull requests to your branch, and we can pick up where we left off.

drgrice1:
I added my pull requests to this branch.  Both pull requests move the point tool into the parserGraphToolPointTool.pl macro (I am not sold on that file name).  That macro loads the parserGraphTool.pl macro, so a problem does not need to load both.

The version 1 pull request loads the javascript from pg/htdocs/js/apps/GraphTool/pointtool.js.

The version 2 pull request injects the javascript via `TEXT(MODES(...))` in the macro itself.

I prefer version 1.

One of the reasons I want to do this is to use this as an example of how others can extend the graph tool macro with new tools.  Doing it with a separate macro like this makes this more clear than if it is added to the parserGraphTool.pl macro itself.

drgrice1:
By the way, to test the pull requests you should be able to just change the `parserGraphTool.pl` macro in the files to `parserGraphToolPointTool.pl`.

Alex-Jordan:
> One of the reasons I want to do this...

Crossed messages. I was just asking why in another place, as you no doubt have read by now.

Alex-Jordan:
I wonder if we could accomplish that with something less fundamental than point making. Like, an absolute value function tool, or a 5-point conic section tool, etc.

drgrice1:
Another reason for having the point tool in another macro is for code modularity.  If every new tool is just dumped into the main macro, the macro and javascript files will get to big and become unmanageable.

Furthermore, the point tool is inherently of a different nature than the other tools.  It is not loaded by default, at the least, and it is of a nature that it wouldn't generally be used in combination with the other tools.

I also don't want to set the precedent of someone wanting to add a new tool, and just dumping it into the main macro.

I am not sure why you feel that it is wrong for a macro to depend on another macro and add functionality.  There are many other macros that do this.  For example, the contextLimitedRadical.pl macro depends on the contextLimitedPowers.pl macro (and bizarroArithmetic.pl).   Also, the contextPolynomialFactors.pl depends on the contextLimitedPolynomial.pl macro.

As to concerns that authors would find the parserGraphTool.pl and not find the parserGraphToolPointTool.pl macro, what about adding a comment to the documentation in the parserGraphTool.pl macro (where the tools are described) that mentions its availability?

Alex-Jordan:
> I am not sure why you feel that it is wrong for a macro to depend on another macro and add functionality.

Well, that's not what I said, or at least not what I meant. The aspect that gives me pause is an author needing to know to load macro file B for a tool that is mostly the same as some other tool from macro file A. I believe a real situation is an instructor coming across something that uses parserGraphTool.pl and wondering if they could make their own graphing exercise that includes point plotting. And then concluding something like "well, I guess it can't be done" because it doesn't occur to them that there might be another tool. I can't think of a similar situation, perhaps because it is not a good practice. But it would be a bit like if there were `contextFraction.pl` and also `contextLimitedFraction.pl`. 

> As to concerns that authors would find the parserGraphTool.pl and not find the parserGraphToolPointTool.pl macro, what about adding a comment to the documentation in the parserGraphTool.pl macro (where the tools are described) that mentions its availability?

That will help. Not 100%, since there will be much of the target audience who will miss that.

It's really your tool, and it's a great tool. And I'm happy to evolve with it however you see best. I hope to tackle some WW development stuff this coming weekend, and I'll include wrapping up these two PRs to my repo.

drgrice1:
I completely disagree that this is not a good practice.  From a coding standpoint and modular code, this definitely is a good practice.

I realize that you are thinking more in terms of the authoring perspective than the coding perspective.  Although I feel that any author that finds the parserGraphTool.pl macro should also easily find the parserGraphToolPointTool.pl macro.  Any author that misses this is also likely to miss the existence of the point tool even if it is build in to the parserGraphTool.pl macro.

If you feel that strongly about this, then I am fine with leaving it in the parserGraphTool.pl macro.  Note that if it is decided to separate via one of the pull requests to this branch, make sure that only one is merged.

If it is decided that it should not be separate, then I will probably add some other tools as separate macros.  I have two nice tools already that would work for this.  One is a three point parabola tool.  That one replaces the built in parabola tools, and so can not be used in combination with them.  That would probably make more sense to you to be in a separate macro.  The other tool is a four point cubic graphing tool that is closely related in how it works.

Alex-Jordan:
Let's just take a poll at Friday's meeting (where I think I will be present this week). Tie breaks in your direction. And I fully expect some "I have no opinion"s.

Alex-Jordan:
I had another thought about this. In a future where there could be a half dozen or so additional tools (Point, Vee, Conic, Cubic, Exponential, Logarithmic, Polynomial, ...) consider a problem that will use all/most of the extensions. Like some sort of open-ended question, or something that does not want to give away the nature of what is meant to be plotted. In that situation, the author will need to load a half dozen macro libraries?

This got me wondering if `parserGraphToolPointTool.pl` should be `parserGraphToolExtensions.pl` and new things go there, all off by default as they are added. Would that help with the code separation you are interested in?

drgrice1:
Yeah, that would be another way to do it.

Alternately, we could just have methods in parserGraphTool.pl directly that can be called to add the additional tools.

Or yet another alternative is to reverse the macro inclusion that I was suggesting.  Put the point tool (and other future tools that are added) in other macro files, and have the parserGraphTool.pl just load all of them.

I don't know on those.  Just throwing out ideals.

Now, more important to me than the separation of code in the macro, is the separation of the javascript.  The file is already 1423 lines long.  In fact I was already thinking about splitting that up.  With the addition of the point tool it increases by 92 lines to 1515 lines.  The point tool even has a relatively small amount of code added.  This is probably the bare minimum of code that a new graph object and tool will add (it is probably the simplest).  The other two tools that I mentioned have more than 500 lines combined.  Throw in a an absolute value tool, a conic tool, exponential, logarithmic, etc, and the single file approach will become a nightmare to manage.

pstaabp:
This looks good.  Working for all of the problems although one of the problems generated the function $(-1)^x*4-2$.  Typically you want bases of exponential to be positive, but that's not part of this PR.

I think some modularity is good. I haven't looked deep into the code, but if this isn't done, a base class for any graph tool could be made with new tools inheriting from that class.

drgrice1:
This is still pending a pull request to this pull request.

Alex-Jordan:
I merged @drgrice1 's PR into this branch. So this is ready to either test again, or merge.

drgrice1:
I have tested it of course.  I will merge it unless anyone objects.

Issue #658, opened 2022-02-28
substitute incorrectly handles negative numbers when reduceConstants is set to 0


dlglin:
See the attached MWE.  If reduceConstants is set to 0, when a negative number is substituted into a formula, the parentheses are removed in the LaTeX string.

e.g.
Context() -> flags -> set(reduceConstants => 0);
$f = Formula("x^2");
$g = $f->substitute(x=>-3);

The value of $g is 9, but the LaTeX string for $g is "-3^2", which causes two issues:

1. $g is displayed incorrectly (without the parentheses) when shown on screen.
2. If $g is used in a subsequent calculation, then the LaTeX string is used, which leads to incorrect values.  e.g. $g-1 evaluates to -10

[reduceConstantsIssue.pg.txt](https://github.com/openwebwork/pg/files/8155227/reduceConstantsIssue.pg.txt)

pstaabp:
#659 now fixes this.

PR #659, opened 2022-03-01
Make sure negative numbers are handled as a negated number in Value items (#658)

dpvc:
This PR fixes the problem where substituting a negative constant into a formula might not get the needed parentheses.  It does so by providing a proper tree structure (negation together with a number) so that the parentheses are inserted when needed.

You can test using

```
Context()->variables->add(y => 'Real');
TEXT(Formula('x^y')->substitute(x => -3));
```

which should produce `(-3)^y` rather than `-3^y`, while

```
TEXT(Formula('x+y')->substitue(x => -3));
```

should produce `-3+y` (not `(-3)+y`).

Similarly, 

```
TEXT(Formula('x^y')->substitue(x => 3));
```

should produce `3^y`, as expected.

Resolves issue #658.

### DIFF
--- a/lib/Parser/Value.pm
+++ b/lib/Parser/Value.pm
@@ -25,7 +25,10 @@ sub new {
   return $value->{tree}->copy($equation) if ($type eq 'Formula');
   return $self->Item("String",$context)->new($equation,$value,$ref) if ($type eq 'String');
   return $self->Item("String",$context)->newInfinity($equation,$value,$ref) if ($type eq 'Infinity');
-  return $self->Item("Number",$context)->new($equation,$value,$ref) if ($type eq 'Number');
+  if ($type eq 'Number') {
+    my $number = $self->Item("Number",$context)->new($equation,CORE::abs($value),$ref);
+    return ($value < 0 ? $self->Item("UOP",$context)->new($equation,'u-',$number) : $number);
+  }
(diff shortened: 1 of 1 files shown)


pstaabp:
This looks good.  I also checked:
```
TEXT(Formula('x+y')->substitute(x => -3));
```
which returns `x+(-3)`.  I assume that is intended and not `x-3`.

dpvc:
@pstaabp, yes, this should produce `x+(-3)` just as `Formula('x + -3')` would.  To get `x-3` you need to use `reduce`.

drdrew42 (approved):

drgrice1:
This pull request seems to be causing a problem with the LimitedNumeric context.  Opening a problem using this context causes the system to hang.  A minimal working example is attached.
[MWE.pg.txt](https://github.com/openwebwork/pg/files/8367772/MWE.pg.txt)

drgrice1:
@dpvc:  The problem seems to be that the `class` method for a Parser::Legacy::LimitedNumeric::UOP::minus object returns 'MINUS' instead of 'UOP'.  As a result when the Parser::UOP new method is called, the isNeg method of lib/Parser/Item.pm returns false when it is called at line 33 of lib/Parser/UOP.pm.  Thus the Parser::Item::Value new method is called, which calls the Parser::UOP new method (in the change introduce here) and starts it all over again, thus causing an infinite recursion.

PR #661, opened 2022-03-11
Add keyboard controls for the GraphTool javascript and two new tools

drgrice1:
If you can now use tab and shift-tab to navigate to the tool buttons of the GraphTool, and then hit enter or space, then the tool will activate.  That isn't new, but before nothing would happen until you move the mouse onto the board.  Now for the graphing tools, a highlight point will immediately show up at the origin.  Then you can use the arrow keys to move that point around.  Of course you can still use the mouse and the point will follow that mouse as soon as it enters the board.  If you hit enter or space, then it will move to the next stage of the graphing process for the tool.  Of course for the point tool that will be it.  For the other tools another highlight point will appear that can be moved with either keyboard or mouse.

There are several keyboard shortcut keys added as well.  Those are:
Escape:  Cancel graphing for the current tool (not new)
Delete:  Delete the currently selected object on the graph (shortcut for the delete button)
Ctrl-Delete:  Remove all objects that have been graphed (shortcut for the clear button)
s:  Change the currently selected or currently being graphed object to solid (solid button shortcut)
d:  Change the currently selected or currently being graphed object to dashed (dash button shortcut)
Shift-n:  Move keyboard focus from the currently selected object to the first enabled button on the toolbar

I am not sold on shift-n.  I wanted Ctrl-Tab but that conflicts with the browser keyboard shortcut for opening a new tab.  It is hard to find keyboard combinations that don't conflict with browser keyboard shortcuts.  Any better suggestions?

Once objects are graphed you can successively hit Tab to progress through the points of the graphed objects in the order that they were added to the board.  Shift-tab will go backward of course.

Also added are two new tools.  A three point quadratic tool and a four point cubic tool.  The following attached problem is a demo that has all available tools active.  Note that I don't recommend that authors use both the new quadratic tool and the existing parabola tools at the same time as that would be confusing for students.
[Demo.pg.txt](https://github.com/openwebwork/pg/files/8229066/Demo.pg.txt)

There are a few new options added to the parserGraphTool.pl macro as well (passed into the javascript).

The first two are `xAxisLabel` and `yAxisLabel`.  These will be the labels displayed on the horizontal and vertical axes in the graph.  They can also be set to the empty string for no label. 

The other option is `ariaDescription`.  The contents of this string will be added to a hidden div that jsxgraph references in an `aria-describedby` attribute.  This is a new feature of the latest release of jsxgraph.

Note that this pull request depends on openwebwork/webwork2#1599 and #657.


drgrice1:
I should point out that there is now a concept of focus for the graphtool as a whole.  If you click outside the board or toolbar, then the GraphTool will lose focus and go inactive.  The keyboard shortcuts won't work.  Of course you can still use tab or shift-tab if you are before or after the board to re-focus the tool or just click/touch the board and make things active again.

Alex-Jordan:
I am starting some testing on this. Before I get to actually testing this, I am seeing an issue. Here is a problem:
```
###################################################################################
# Setup and Macros
###################################################################################
DOCUMENT();

loadMacros(
	"PGstandard.pl",
	"MathObjects.pl",
	"PGML.pl",
	"parserGraphTool.pl"
);

###################################################################################
# Variables and Formulas
###################################################################################
TEXT(beginproblem());

$x = non_zero_random(-5, 5);
$y = non_zero_random(-5, 5);

$gt = GraphTool("{point, ($x, $y)}")->with(
    availableTools => ["PointTool"],
	bBox => [-11, 11, 11, -11],
	staticObjects => [ "{point,(5,1)}", "{point,(0,4)}" ],
	cmpOptions => {
		list_checker => sub {
			my ($correct, $student, $ans, $value) = @_;

			my $score = 0;
			my @errors;
			my $count = 1;

			# Get the correct point.
			my ($cx, $cy) = $correct->[0]->extract(1)->value;

			for (@$student) {
				my $nth = Value::List->NameForNumber($count++);

				$score += 1, next
				if ($_->extract(1) eq $correct->[0]->extract(1) &&
					$_->extract(2) eq $correct->[0]->extract(2));

				push(@errors, "The $nth object graphed is not a point."),
				next if ($_->extract(1) ne $correct->[0]->extract(1));

				push(@errors, "The $nth object graphed is incorrect.");
			}

			return ($score, @errors);
		}
	}
);

###################################################################################
# Main Text
###################################################################################

BEGIN_PGML
Graph the point [`([$x], [$y])`].

[_]{$gt}
END_PGML

ENDDOCUMENT();

```

It looks like this:
Screen Shot 2022-03-31 at 10 07 46 PM

So far, so good. Now I click to edit, and "Take Action" to View without changing anything, and the axes are missing:


Screen Shot 2022-03-31 at 10 08 28 PM

In case it matters, before starting all of this, I ran `npm install` in both of the htdocs folders.

drgrice1:
This seems to only be happening in the problem editor.  I will see if I can figure out what is happening there.  I don't see this happening anytime it is rendered on its own page or when rendered in the homework sets editor.

drgrice1:
By the way, there is no need to clear your browser cache anymore.  With the new js/css build system that will not be an issue.

drgrice1:
@Alex-Jordan:  The issue inside the modal dialog of the PGProblemEditor is now fixed.  Thanks for pointing that out.

drgrice1:
Make sure to run `npm install` in the pg htdocs directory again to update the javascript.

pstaabp:
Overall, this looks great.  I would say stay with a SHIFT something for the shortcut to put focus on the first button.  That will definitely not conflict with anything.  SHIFT-N is fine.  

 Would it be helpful (and not too difficult) to also have a keyboard shortcut that selects the next object that is placed on the graph?

drgrice1:
Actually, `shift-n` does conflict with an extension that I have installed in Google Chrome.  Although, most users won't have that installed, and it is probably impossible to avoid all keyboard shortcuts that extensions may use.

I am not sure what you mean by a keyboard shortcut that selects the next object that is placed on the graph.  Currently when you graph an object, that object is selected as soon as you finish graphing it.  Also, if an object is selected, then you can use `tab` to go to the next object that was graphed after it (if there is one), or `shift-tab` to go to the previous object.  So as I see it there already is such a keyboard shortcut.

pstaabp:
> Also, if an object is selected, then you can use tab to go to the next object that was graphed after it (if there is one)

Yes, I see that now.  I guess the SHIFT-N is just to jump ahead to the buttons, instead of needing to TAB through all the objects on the graph, since that is another way to access the buttons.

drgrice1:
Yeah, `shift-n` is just a quick jump to the buttons.  The point is also that by using `shift-n` you can choose which object is selected when you jump to the buttons.  This affects things like changing the object from being solid to dashed or vice versa.

PR #667, opened 2022-03-29
Fix an infinite recursion caused by #659 for the Parser::Legacy::LimitedNumeric::UOP::minus package.

drgrice1:
See https://github.com/openwebwork/pg/pull/659#issuecomment-1081832326 in that pull request for an explanation and minimal working example.

### DIFF
--- a/lib/Parser/Legacy/LimitedNumeric.pm
+++ b/lib/Parser/Legacy/LimitedNumeric.pm
@@ -40,7 +40,7 @@ sub class {'MINUS'};
 # instead of 'UOP' as it is for the Parser::UOP::minus parent package.
 sub isNeg {
   my $self = shift;
-  return $self->class eq 'MINUS' && $self->{uop} eq 'u-' && !$self->{op}{isInfinite};
+  return !$self->{op}{isInfinite};
 }
 
 ##################################################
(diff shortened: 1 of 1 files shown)


lib/Parser/Legacy/LimitedNumeric.pm
```diff
@@ -36,6 +36,12 @@ sub _check {
 
 sub class {'MINUS'};
 
+# Override the parent method because the class for this package is 'MINUS'
+# instead of 'UOP' as it is for the Parser::UOP::minus parent package.
+sub isNeg {
+  my $self = shift;
+  return $self->class eq 'MINUS' && $self->{uop} eq 'u-' && !$self->{op}{isInfinite};
```
dpvc, line 43: You can probably just `return !$self->{op}{isInfinite}`, here, since this object's class is always `MINUS` (unless someone makes a subclass of it that changes that, in which case, they will have to override this method anyway), and from `u-` (unless the context is remapped).

dpvc (approved):
One potential simplification.

lib/Parser/Legacy/LimitedNumeric.pm
drgrice1: Yeah, that is true.  Good point.

pstaabp:
This fixes the issue and minor enough.  Will merge in.

PR #668, opened 2022-04-10
Clean up the html help files and fix html validation and accessibility issues in them.

drgrice1:
This was brought to my attention by a communication with someone working on accessibility issues for WeBWorK problems embedded in LibreTexts.

One issue is that of headings not being at the correct level.  This is actually a somewhat complicated issue for these help files.  For now I started the headings with `h2`s.  Generally that will work for in webwork and most other places.  It assumes there is an `h1` somewhere before.  However, the complication is if the help file is already inside a deeper level, or if the the problem itself is already in a deeper level.  Then the headings in the help file will not be at the correct level, and could mess up the heading structure of the page containing it.  So another option that may need to be considered is not having headings at all in the help files.

### DIFF
--- a/htdocs/helpFiles/Entering-Syntax.html
+++ b/htdocs/helpFiles/Entering-Syntax.html
@@ -1,4 +1,4 @@
-<h2>Syntax for entering answers to WeBWorK</h2>
+<h2 style="text-align: center">Syntax for entering answers to WeBWorK</h2>
 
 <h3 style="font-size: 1.25rem">Mathematical Symbols Available In WeBWorK</h3>
 <ul>
--- a/htdocs/helpFiles/Entering-Syntax.html
+++ b/htdocs/helpFiles/Entering-Syntax.html
@@ -1,137 +1,249 @@
-<h1 id="firstHeading">Syntax for entering answers to WeBWorK</h1>
+<h2>Syntax for entering answers to WeBWorK</h2>
 
+<h3 style="font-size: 1.25rem">Mathematical Symbols Available In WeBWorK</h3>
+<ul>
+	<li><code style="color: black">+</code> Addition.</li>
+	<li><code style="color: black">-</code> Subtraction.</li>
+	<li>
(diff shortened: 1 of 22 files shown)


drgrice1:
Attached is a problem that utilizes all help files for testing.
[helpFiles.pg.txt](https://github.com/openwebwork/pg/files/8459725/helpFiles.pg.txt)

pstaabp:
Since all of the help `.html` files seemed to be changed, is this just linting/spacing or any content changes?

drgrice1:
Yeah, the rest is just formatting. There are no content changes other than a few typos fixed.

htdocs/helpFiles/Entering-Syntax.html
```diff
@@ -1,137 +1,249 @@
...
-</li></ul>
-
-<a name="Scientific_Notation_Available_In_WeBWorK"></a><h4> <span class="mw-headline">Scientific Notation Available In WeBWorK</span></h4>
-<ul><li> 2.1E2 is the same as  210
-</li><li> 2.1E-2 is the same as .021
-</li></ul>
-<a name="Mathematical_Functions_Available_In_WeBWorK"></a><h4> <span class="mw-headline">Mathematical Functions Available In WeBWorK</span></h4>
-<p>Note that sometimes one or more of these functions is disabled for a WeBWorK problem because the 
-instructor wants you to calculate the answer by some means other than just using the function.
+<h2>Syntax for entering answers to WeBWorK</h2>
```
pstaabp, line 1: For consistency, should this be:
```
<h2 style="text-align: center">Syntax for entering answers to WeBWorK</h2>
```

drgrice1: I missed that.  Thanks.  I added that now.

However, should I also center the other headings in that file?  It looks a bit off with only the top heading centered.  That file (and its duplicate) are the only help files that have additional headings.  Wait, I see that the PDE-notation.html file does also.

PR #669, opened 2022-04-12
Move cssnano from dependencies to devDependencies in the htdocs/package.json file.

drgrice1:
This was a minor mistake made when I added this dependency.  It certainly doesn't matter much, it is just where it should be.


pstaabp:
This is simple, so will merge.

PR #673, opened 2022-04-19
Fix possibly undefined input in mqeditor.js.

drgrice1:
This can happen if a problem is in a scaffold that can not be opened.  The hidden MathAuill latex input is still output on the page, and the mqeditor.js code then acts on it, but can't find the corresponding text input.


pstaabp:
This is minor, so will merge.

PR #676, opened 2022-05-19
Fix pointer focus for the center of a circle.

drgrice1:
Curently the center point of a circle can not be focused by clicking on it with the mouse.  This is because the jsxgraph circle hasPoint method only returns true if the circle itself contains the given point, but not if the center point contains the given point.  So redefine the hasPoint method for a circle to fix that.

### DIFF
--- a/htdocs/js/apps/GraphTool/graphtool.js
+++ b/htdocs/js/apps/GraphTool/graphtool.js
@@ -664,6 +664,11 @@ window.graphTool = (containerId, options) => {
 			));
 			this.definingPts.push(center, point);
 			this.focusPoint = center;
+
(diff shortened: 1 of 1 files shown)


drgrice1:
To test this you can use the attached problem.  In that problem, graph a circle and then click on the point on the circle itself.  Then click on the center point again.  Without this pull request, focus will not go back to the center point when you click on it.  With this pull request it will.

Note that with or without this pull request you can get focus to the center point using tab or shift-tab (whichever is needed, depending on where focus is).

[Demo.pg.txt](https://github.com/openwebwork/pg/files/8729067/Demo.pg.txt)

Issue #679, opened 2022-05-24
AnswerHints misbehaving under certain circumstances (probably due to a Value::Error being called)


taniwallach:
In the following problem, when the second answer is set to `1/x` the AnswerHints code fails, the following errors are reported:
```
The evaluated answer is not an answer hash : ||.
Use of uninitialized value $credit{"AnSwEr0002"} in numeric eq (==) at (eval 1863) line 292.
Use of uninitialized value $credit{"AnSwEr0002"} in multiplication (*) at (eval 1863) line 308.
```
The second and third are from the `pg/macros/weightedGrader.pl` code.

When this is used with a weighted grader, due to that error, this part of the question is treated as having weight 1 (instead of 50 as intended by the code) so the allocation of partial credit is quite off. Without a weighted grader, the answer gets marked as incorrect but the warning about `The evaluated answer is not an answer hash : ||.` is still issued.

It seems that the division by zero when `$student->eval( x=>0 );` is being run is causing a Value::Error to be called , and apparently that leads to a `die` which eventually leads to there not being a answer hash as the output of the AnswerHints code. 

Ideally, I think that the AnswerHints code should avoid such a total failure. I would say that in the "worst" case, the original answer hash from before the post-filter runs could be the final result. Even better would be to try to collect the "hints" result ignoring those "cases" which fail (but to warn that some trouble was encountered).

~~I'm not sure how such a fix can be made to `pg/macros/answerHints.pl`.~~

A proposed fix is in https://github.com/openwebwork/pg/pull/680

---
```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
  "weightedGrader.pl",
  "answerHints.pl",
  "PGcourse.pl",
);

# Print problem number and point value (weight) for the problem
TEXT(beginproblem());

install_weighted_grader();

# Show which answers are correct and which ones are incorrect
$showPartialCorrectAnswers = 1;

# The answer for $ansD has exp(-y/n) so make most of range of y
# negative, but not too large in absolute value.

Context("Numeric");

########################################################################

$k = random( 2,4,1 );

$ans1 = Real(2);
$ans2 = Compute( "1/(x+$k)" );

BEGIN_TEXT

\( 1+1 = {} \) \{ $ans1->ans_rule(10) \}

$BR

Find the function of the form \( f(x) = \frac{1}{x+k} \)
such that \( f(0) = \frac{1}{$k} \).
$BR
[...]

PR #680, opened 2022-05-24
Patch AnswerHints

taniwallach:
Patch AnswerHints so that if a subroutine in the list "dies" there will still be a final result from the AnswerHints postfilter.
This is mean to address https://github.com/openwebwork/pg/issues/679.

### DIFF
--- a/macros/answerHints.pl
+++ b/macros/answerHints.pl
@@ -155,8 +155,8 @@ sub AnswerHints {
 						{
 							# Make the call to run the function inside an eval to trap errors
 							my $myResult = 0;
-							eval { $myResult = &$wrong($correct, $student, $ans); } or do {
-								$ans->{ans_message} = "error during AnswerHints processing";
+							eval { $myResult = &$wrong($correct, $student, $ans); 1; } or do {
+								warn "An error occurred in this problem.";
 								last;
 							};
 							if ($myResult) {
--- a/macros/answerHints.pl
+++ b/macros/answerHints.pl
@@ -113,66 +113,81 @@ Example:
 
 =cut
 
-sub _answerHints_init {}
+sub _answerHints_init { }
 
 sub AnswerHints {
-  return (sub {
-    my $ans = shift; $ans->{_filter_name} = "Answer Hints Post Filter";
-    my $correct = $ans->{correct_value};
-    my $student = $ans->{student_value};
-    Value::Error("AnswerHints can only be used with MathObjects answer checkers") unless ref($correct);
-    return $ans unless ref($student);
-    my $context = $correct->context;
-    my $hash = $context->{answerHash};
-    $context->{answerHash} = $ans;
-    my $processPreview = $correct->getFlag('answerHintsProcessPreview', 0);
-    $context->{answerHash} = $hash;
+	return (
+		sub {
+			my $ans = shift;
+			$ans->{_filter_name} = "Answer Hints Post Filter";
+			my $correct = $ans->{correct_value};
+			my $student = $ans->{student_value};
+			Value::Error("AnswerHints can only be used with MathObjects answer checkers") unless ref($correct);
+			return $ans unless ref($student);
+			my $context = $correct->context;
+			my $hash    = $context->{answerHash};
+			$context->{answerHash} = $ans;
+			my $processPreview = $correct->getFlag('answerHintsProcessPreview', 0);
+			$context->{answerHash} = $hash;
 
-    while (@_) {
-      my $wrongList = shift; my $message = shift; my @options;
-      ($message,@options) = @{$message} if ref($message) eq 'ARRAY';
-      my %options = (
-        checkCorrect => 0,
-        replaceMessage => 0,
-	checkTypes => 1,
-        processPreview => $processPreview,
-        score => undef,
-        cmp_options => [],
-        @options,
-      );
-      next if $options{checkTypes} && $correct->type ne $student->type;
-      next if !$options{processPreview} && $ans->{isPreview};
-      $wrongList = [$wrongList] unless ref($wrongList) eq 'ARRAY';
-      foreach my $wrong (@{$wrongList}) {
-	if (ref($wrong) eq 'CODE') {
-	  if (($ans->{score} < 1 || $options{checkCorrect}) &&
-	      ($ans->{ans_message} eq "" || $options{replaceMessage}) ) {
-	      # Make the call to run the function inside an eval to trap errors
-	      my $myResult = 0;
-	      eval { $myResult = &$wrong($correct,$student,$ans); } or do {
-	        $ans->{ans_message} = "error during AnswerHints processing";
-	        last;
-	      };
-	      if ($myResult) {
-	        $ans->{ans_message} = $ans->{error_message} = $message;
-	        $ans->{score} = $options{score} if defined $options{score};
-	        last;
-	      }
-	  }
-	} else {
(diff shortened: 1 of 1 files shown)


macros/answerHints.pl
```diff
@@ -146,11 +146,16 @@ sub AnswerHints {
...
-	      &$wrong($correct,$student,$ans)) {
-	    $ans->{ans_message} = $ans->{error_message} = $message;
-	    $ans->{score} = $options{score} if defined $options{score};
-	    last;
+	      ($ans->{ans_message} eq "" || $options{replaceMessage}) ) {
+	      # Make the call to run the function inside an eval to trap errors
+	      my $myResult = 0;
+	      eval { $myResult = &$wrong($correct,$student,$ans); };
+	      if ( $@ ) { $myResult = 0; }
+	      if ( $myResult ) {
```
drgrice1, line 154: The space inside the parentheses is inconsistent with the style of the rest of the file.  Those spaces are not preferable in any case according to the coding style that I like!

macros/answerHints.pl
```diff
@@ -146,11 +146,16 @@ sub AnswerHints {
...
-	      ($ans->{ans_message} eq "" || $options{replaceMessage}) &&
-	      &$wrong($correct,$student,$ans)) {
-	    $ans->{ans_message} = $ans->{error_message} = $message;
-	    $ans->{score} = $options{score} if defined $options{score};
-	    last;
+	      ($ans->{ans_message} eq "" || $options{replaceMessage}) ) {
+	      # Make the call to run the function inside an eval to trap errors
+	      my $myResult = 0;
+	      eval { $myResult = &$wrong($correct,$student,$ans); };
+	      if ( $@ ) { $myResult = 0; }
```
drgrice1, line 153: This line is not needed.  If the `$wrong` method has an error, then the assignment will not occur, and `$myResult` will still be `0` from the assignment on line 151.  Testing reveals that this is the case.

drgrice1 (approved):
I suppose this is okay.

I have mixed feelings about this though.

Your example code in the linked issue should set `$tmp = 0` and then wrap the `$tmp = $student->eval(x => 0)` in a perl `eval`.  Then the answerHint macro will work as expected and give the message that the condition requested in the problem is not satisfied.  This pull request will hide this error, and make it harder to see that the hints are not working in the intended way.

On the other hand, it doesn't hurt to catch errors from special cases don't matter that are accidentally missed by the author.

macros/answerHints.pl
pstaabp: Perhaps perltidy could be run on this file.  I figure eventually we'll run it on all files.

pstaabp: Perhaps perltidy could be run on this file.  I figure eventually we'll run it on all files.

drgrice1: I agree.  I would like to see that as well.

pstaabp (approved):
Fixes the error in the posted problem.  I would recommend the cleanups that @drgrice1 pointed out.

macros/answerHints.pl
taniwallach: Dropped the space in the second commit.
perltidy applied in the third commit, but I needed to drop `--xci` and `--nlop` from `.perltidy` temporarily, as the version on my Debian machine does not support those options.

macros/answerHints.pl
taniwallach: Thanks. Dropped in commit 2 of the PR. However, I added a `or do` case which sets a fixed error message (in the results message box).

taniwallach:
@drgrice1 @pstaabp - Thanks. Perltidy changes are in the third commit.

> Your example code in the linked issue should set `$tmp = 0` and then wrap the `$tmp = $student->eval(x => 0)` in a perl `eval`. Then the answerHint macro will work as expected and give the message that the condition requested in the problem is not satisfied.

Thanks. I fixed the real problem (was more detailed, but the core error was the occurrence of a divide by zero error for a student answer at the initial condition point). I had not considered the need to such until this occurred.

I thought it would be helpful to document this in the Wiki, but the Wiki problems technique page about AnswerHints does not even mention the use of a subroutine test, so explaining there that evaluation at a test point could potentially lead to such an issue seems overboard there.

> This pull request will hide this error, and make it harder to see that the hints are not working in the intended way.

Good point. I added a fixed "error message" (as a result table message) for when the `eval` fails. That will make the existence of a problem visible to the end user (who may report it) but prevents the post filter from dying and messing up the allocation of weighed credit, etc.

I tried unsuccessfully to use the value of `$@` to retrieve the internal error message ("Divide by zero"). Using `warn ref($@);` could get that to show up as a warning message at the bottom, but I did not manage to get it out to put into the results table. It seems my understanding of what a failed `eval` is putting into `$@` is just not sufficient to dig it out. Ideas for how to do that are welcome.

drgrice1 (approved):
The perltidy result is a little different with the remove options added back in, but it will do for now.

I am not sure that I agree with displaying the "error during AnswerHints processing" as an answer message.  I think of those messages as informative messages for the student.  I think it might be better if that were a simple warn statement so that it appears with the information about reporting the issue to the professor and such.  This is really an issue with the problem.  In any case, this is will do for now also.

I am not sure why `$@` is not being set exactly.  It most likely has to do with the `$SIG{__WARN__}` handling inside the translator though.  The way that you dealt with it seems to be working well though.

pstaabp:
This looks good now.  I agree with @drgrice1 about the message.  We could say something along the lines of "An error occurred in this problem.  Please inform your instructor."

taniwallach:
> This looks good now. I agree with @drgrice1 about the message. We could say something along the lines of "An error occurred in this problem. Please inform your instructor."

Sounds good to me, but I would like to somehow mention AnswerHints in the message, to provide a hint to whoever debugs it of where to look. How about: "An error occurred in this problem. Please inform your instructor. (die in AnswerHints subroutine)"?

drgrice1 (changes requested):
I found an issue with the `eval` approach.  See my comment.

macros/answerHints.pl
```diff
@@ -113,59 +113,81 @@ =head1 AnswerHints()
...
+				$wrongList = [$wrongList] unless ref($wrongList) eq 'ARRAY';
+
+				foreach my $wrong (@{$wrongList}) {
+					if (ref($wrong) eq 'CODE') {
+						if (   ($ans->{score} < 1 || $options{checkCorrect})
+							&& ($ans->{ans_message} eq "" || $options{replaceMessage}))
+						{
+							# Make the call to run the function inside an eval to trap errors
+							my $myResult = 0;
+							eval { $myResult = &$wrong($correct, $student, $ans); } or do {
```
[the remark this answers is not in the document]
taniwallach: @drgrice1 Thanks! That is a serious bug. Code patched as suggested and that fixes it.

taniwallach:
> I am not sure that I agree with displaying the "error during AnswerHints processing" as an answer message. I think of those messages as informative messages for the student. I think it might be better if that were a simple warn statement so that it appears with the information about reporting the issue to the professor and such.

Upon additional consideration, I changed the error message to a `warn` and the text to "An error occurred in this problem."

Since it is now a `warn` it will trigger the standard message above the problem: "Warning -- There may be something wrong with this question. Please inform your instructor including the warning messages below." Since the `warn` is from inside `answerHints.pl` the message below the problem provides an indication to the staff where to look for the bug.

drgrice1 (approved):

pstaabp (approved):

PR #681, opened 2022-05-26
Fix ww3 mq-editor preview button code causing console log error in ww2 gateway quizzes.

drgrice1:
In a gateway quiz since the form by the name `problemMainForm` does not exist, the webwork 3 code that looks form the preview button of a specific problem to click when enter is pressed in a MathQuill answer box returns an empty list.  This causes a console error message.


pstaabp:
Is this supposed to occur if the preview button is clicked in a gateway quiz?  I'm not seeing an error.

pstaabp:
Is this supposed to occur if the preview button is clicked in a gateway quiz?  I'm not seeing an error.

drgrice1:
No.  It does not happen when you click on the preview button.  It occurs if the focus is in a MathQuill answer box, and then you hit enter.

drgrice1:
After testing this with webwork3 itself I realized that this wasn't working there either.  The Element.closest method does not return a list.  It just returns the closest element satisfying the selector or null if that isn't found.  Previously I believe I was using a querySelectorAll which does return a list, and when I changed it, I forgot to drop the index.

pstaabp (approved):

Issue #675, opened 2022-05-16
Definition of factorial in PG can lead to what WW sees as an infinite loop and causes CPU load which may sometimes persist until a restart or manual intervention


taniwallach:
I ran into a problem where my server was showing very high CPU usage for a long time. It seems to have resulted from an Apache process which got into an overly long calculation while processing a problem render request and for some reason the system failed to kill that process.

The root cause seems to be that the definition of `factorial` has no upper bound on the (non-negative) integer whose factorial it will attempt to calculate in a `while` loop. This issue currently allows intentional or accidental denial-of-service attacks against a webwork server (particularly those with a small number of CPU cores), and I think we should try to prevent that possibility for the future.

Entering `171!` in an answer which is expecting a `Formula` gave me the response `Can't convert an Infinity to a constant` (as does `n!` for larger values of n) but does not cause an error where a `Real` is expected. `170!` does not trigger that message. That seems to indicate that n=170 is essentially the largest value for which Perl is obtaining a (finite) floating point number as the result of the calculation, so there is no reason to run the calculation for values over 171.

Where a `Real` is expected:
  - `21!!` gives the error message `Factorial can only be taken of (non-negative) integers`, as the code already determines that the result of `21!` is not something it can process. 
  - `20!!` gets the Apache process into a calculation it cannot complete (100% CPU on one core until the timeout occurs). I got the error `Timeout after processing this problem for 60 seconds. Check for infinite loops in problem source` also in the `Message` column for the answer where `20!!` was entered. 
  - `12!!` leads to a quite noticeable delay during processing.

I propose that we change the `_eval` method in `pg/lib/Parser/UOP/factorial.pm` to avoid overly long calculations by making some change before the `while` loop. However, I am not sure what the best approach is. Some options I can think of are:
  - `$self->Error("Factorial can only be taken of integers <= 170")  if ( $n > 170 );`
  - `$n = 171 if ( $n > 170 );`
  - `return( 9 ** 9 ** 9) if ( $n > 170 ); # Perl infinity`
The third option seems best as it will create an "infinite" value for all n >= 171, which is what would apparently be happening at present for any values which a server does manage to handle before the timeout. Even better would be to just the "Perl infinity" without running a calculation, but I'm not sure how that can be done.

@drgrice1 @dpvc @pstaabp @mgage - What do you think is the best approach to take?

---

[...]

PR #682, opened 2022-05-26
Limit loop to compute factorial to n<=170, for n>170 return Perl Inf

taniwallach:
Fixes the issue from https://github.com/openwebwork/pg/issues/675 by returning Perl `Inf` for any integer n>170, and using the current approach for integer values of n in [0,170]. That seems to be the consensus on the best patch for the problem itself. Non-integervalues will return the same error message as before.

### DIFF
--- a/lib/Parser/UOP/factorial.pm
+++ b/lib/Parser/UOP/factorial.pm
@@ -25,6 +25,7 @@ sub _eval {
   my $n = shift; my $f = 1;
   $self->Error("Factorial can only be taken of (non-negative) integers")
     unless $n =~ m/^\d+$/;
+  return Inf if $n > 170;
   while ($n > 0) {$f *= $n; $n--}
   return $f;
 }
--- a/lib/Parser/UOP/factorial.pm
+++ b/lib/Parser/UOP/factorial.pm
@@ -25,7 +25,7 @@ sub _eval {
   my $n = shift; my $f = 1;
   $self->Error("Factorial can only be taken of (non-negative) integers")
     unless $n =~ m/^\d+$/;
-  return Inf if $n > 170;
+  return $self->Package("Infinity")->new() if $n > 170;
   while ($n > 0) {$f *= $n; $n--}
   return $f;
 }
--- a/lib/PGEnvironment.pm
+++ b/lib/PGEnvironment.pm
@@ -62,7 +62,8 @@ sub new {
 		# Load from the conf file.
 		$self->{pg_dir} = $ENV{PG_ROOT};
 
-		my $defaults_file = $self->{pg_dir} . "/conf/pg_defaults.yml";
+		my $defaults_file = "$self->{pg_dir}/conf/pg_defaults.yml";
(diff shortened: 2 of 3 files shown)


lib/Parser/UOP/factorial.pm
```diff
@@ -25,6 +25,7 @@ sub _eval {
   my $n = shift; my $f = 1;
   $self->Error("Factorial can only be taken of (non-negative) integers")
     unless $n =~ m/^\d+$/;
+  return Inf if $n > 170;
```
dpvc, line 28: I'm wondering if this should produce a MathObject Infinity rather than a Perl `Inf`.  (My comment on the original discussion used `Inf` to make it work like the other ones in Alex's tests.)

I get error messages about `Inf` being a bareword for this line.  I think it may need to be

```suggestion
  return $self->Package("Infinity")->new() if $n > 170;
```

dpvc (changes requested):

pstaabp:
This also seems like we should have a unit-test for this.  I wrote a number of unit tests last summer and hoped to do more.  Let me see if I can pull one together to add to this PR.

lib/Parser/UOP/factorial.pm
taniwallach: I tried the proposed change now.

Even with that change putting `171!` inside a n input box for a Formula results in the error message `Can't convert an Infinity to a constant`.

`Compute( "171!" );` also triggers that error for either option of the return value for n>170 (and also when the prior version of the code is used).

As such, it seems to me that keep the return value as a Perl infinity (which it would have been in the past) is a more certain way of maintaining backwards compatibility, and saves on any overhead of creating an object here when it has no purpose in the end.

I do see a warning about the bareword from `perl -C lib/Parser/UOP/factorial.pm` but it does not seem to interfere when used inside pg.

However, if you still think the change is advisable, I will make it.

pstaabp:
I just wrote a factorial unit test and added as a PR to your branch @taniwallach.

lib/Parser/UOP/factorial.pm
dpvc: > Even with that change putting 171! inside a n input box for a Formula results in the error message Can't convert an Infinity to a constant.

I am not able to reproduce that.  It works correctly for me (reporting an entered value of `infinity` and that the value is incorrect.  No error message is produced.

taniwallach: I upgraded my version of PG. The Docker container had a somewhat old version. Now it works properly. Sorry.

taniwallach: Note: Peter's PR had the recommended change in it.

pstaabp (approved):

lib/Parser/UOP/factorial.pm
taniwallach: That is now merged, so `$self->Package("Infinity")->new()` is not the return value of n>170.

PR #683, opened 2022-05-26
In contextFraction only use \textstyle for mixed numbers.

somiaj:
useTextstyle controls if the fraction's TeX output is displayed
  using \textstyle or not. By default this is turned on to
  match current behavior. It can now be turned off for users
  who want to display fractions without \textstyle.

### DIFF
--- a/macros/contextFraction.pl
+++ b/macros/contextFraction.pl
@@ -907,11 +907,16 @@ sub string {
 sub TeX {
   my $self = shift; my $equation = shift; shift; shift; my $prec = shift;
   my ($a,$b) = @{$self->{data}}; my $n = "";
+  my $textstyle = '';
   return "$a" if $b == 1;
-  if ($self->getFlagWithAlias("showMixedNumbers","showProperFractions") && CORE::abs($a) > $b)
-    {$n = int($a/$b); $a = CORE::abs($a) % $b; $n .= " " unless $a == 0}
+  if ($self->getFlagWithAlias("showMixedNumbers","showProperFractions") && CORE::abs($a) > $b) {
+    $n = int($a/$b);
+    $a = CORE::abs($a) % $b;
+    $n .= ' ' unless $a == 0;
+    $textstyle = '\\textstyle';
+  }
   my $s = ""; ($a,$s) = (-$a,"-") if $a < 0;
-  $n .= ($self->{isHorizontal} ? "$s$a/$b" : "${s}{\\textstyle\\frac{$a}{$b}}")
+  $n .= ($self->{isHorizontal} ? "$s$a/$b" : "${s}{$textstyle\\frac{$a}{$b}}")
     unless $a == 0 && $n ne '';
   return "$n";
 }


Alex-Jordan:
I have something like this on my school's server. Mine straight up removes the `\textstyle` and lets the output be what it would naturally be given where it is placed. However then it looks bad when it's display math and it prints a mixed number. So if it's showing mixed numbers, it still includes the `\textstyle`.

For me, I have too many exercises where I wanted the `\textstyle` to just go away and I wouldn't want to go through and set a context flag in all those exercise files. I'm not opposed to this change, I'd just leave my customized `contextFraction.pl` in place. But I thought I'd share what I do about this.

drgrice1:
I do the same thing as @Alex-Jordan.  I use a custom version of this macro, and remove the `\textstyle`.

somiaj:
I agree, and would also prefer that textstyle wasn't used by default, though in the case of mixed fractions it is preferred. I did think of disabling it by default, and then turning it on if showMixedNumbers is set, and it might be worth updating it to do that.

I just went this route to not change the default behavior but allow users to be able to change this without having to use their own copy of the macro.

somiaj:
I updated the pull request. Adding an option that many might not know about or use may not be that useful in the long run. Instead I updated this to only include `\textstyle` in mixed numbers, otherwise the fraction's output will be formatted based on `\displaystyle`. This seems to be what most are doing, and this will allow other users who don't want to have a local copy of this macro get this behavior. If it doesn't get accepted, Users can just run a local copy of the macro like everyone who has commented seems to be doing anyways.

macros/contextFraction.pl
```diff
@@ -907,11 +907,16 @@ sub string {
...
   my ($a,$b) = @{$self->{data}}; my $n = "";
+  my $textstyle = '';
   return "$a" if $b == 1;
-  if ($self->getFlagWithAlias("showMixedNumbers","showProperFractions") && CORE::abs($a) > $b)
-    {$n = int($a/$b); $a = CORE::abs($a) % $b; $n .= " " unless $a == 0}
+  if ($self->getFlagWithAlias("showMixedNumbers","showProperFractions") && CORE::abs($a) > $b) {
+	$n = int($a/$b);
+	$a = CORE::abs($a) % $b;
+	$n .= ' ' unless $a == 0;
+	$textstyle = '\\textstyle';
```
drgrice1, line 916: The indentation of these lines is inconsistent with the rest of the file.  This file uses a 2 space indentation scheme, and these lines are indented with a single tab.

Unfortunately, this file (and most of PG) uses an indentation scheme that is inconsistent with what I usually promote.  That being to use tab indentation instead of spaces.  Tab indentation does not force the indentation amount on whoever is viewing the file.  Spaces do.

In any case, mixing tab and space indentation is even more troublesome.  If someone has their tabstop set to anything other than 4, then this will not line up correctly in the scope of the space indented file.

somiaj:
Tabs changed to being 4 spaces to match surrounding indentation.

PR #685, opened 2022-05-26
Revert to the mqeditor toolbar focus hack previously used with WeBWorK and PG 2.16.

drgrice1:
It seems that Safari does not properly honor the specifications for the FocusEvent.RelatedTarget, and the new method does not work for that lame browser (at least the current version).

### DIFF
--- a/htdocs/js/apps/MathQuill/mqeditor.js
+++ b/htdocs/js/apps/MathQuill/mqeditor.js
@@ -84,8 +84,11 @@
 			{ id: 'text', latex: '\\text', tooltip: 'text mode (")', icon: 'Tt' }
 		];
 
+		answerQuill.hasFocus = false;
+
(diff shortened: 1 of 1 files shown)


drgrice1:
I would prefer that this pull request is not merged.  I was happy to get rid of the setTimeout hack.

pstaabp:
This works for me on the latest Safari.   We could merge this in and if Safari is upgraded and the bug fixed, we could revert back to your post-2.16, pre #685 code.

taniwallach:
I cannot test this - no access to Safari.
However, I think that it is not wise to release 2.17 using an approach known not to work on Safari.
I recommend merging this, and setting up a new PR to change back to the FocusEvent.RelatedTarget based code, pending for when it works in Safari.

PR #690, opened 2022-06-04
Correct value for electron volt in Units.pm

### LINKED ISSUE #677: Error in Units: electronVolt factor

duffee:
Correct exponential for factor of eV
fixes #677

Adds definitions for kev, MeV and GeV

Remove t/README because it has been superseded by t/README.md


pstaabp (approved):
The splitting up of the code into different PRs is the way to go with this.  This looks good.

Alex-Jordan (approved):
Approving. In a future commit, I might review all of the natural constants in Units.pm and replace them with closer to full precision. So for example I might change 1.6022 to 1.602176634 based on [NIST](https://physics.nist.gov/cgi-bin/cuu/Value?evj). But I don't think that should hold this up.

PR #692, opened 2022-06-05
Migrate test suite to Test2

duffee:
This is a Work In Progress to show how the PG test suite would look if it were migrated to [Test2](https://metacpan.org/pod/Test2::V0).  @pstaabp asks for a reason to justify the move and I think the clearest way of talking about the issues is to have a diff we can all refer to.

This starts out with the tests for **Units.pm** which started the conversation.  It uses my preferred style of a Test module in **t/lib/** to set the environment and updates the README.md

The test for **contextTrigDegrees.pl** exposed some existing errors either in how the macro was being loaded or how the module runs.  Comments have been left and failing tests wrapped in a TODO block.

More migrated tests will be added to this PR as I find the time.


duffee:
Some reasons for using Test2  
1. It is more compact.  It can test nested data structures with a smaller, neater description of the data
2. it understands that hash keys are randomly ordered and how to check for hash keys that should not be there
3. it pulls in the ecosystem of testing tools for warnings, exceptions, mocks and classes into one suite.
4. it takes advantage of 15 years of experience with testing and makes changes that would break Test::More.
5. It's close enough that migrating from Test::More isn't too onerous (they are mostly drop-in replacements).
6. It is developed by the author of Test::More as the path for future development.

I was told that it could run the subtests in random order to flush out tests that depend on state but I haven't been able to verify that recently.  It does change the random seed every day.

t/README.md
```diff
@@ -1,72 +1,27 @@
-# Unit Tests for PG
+# Welcome to WeBWorK
```
pstaabp, line 1: I would keep this as relevant to PG instead of WeBWorK in that these should be testing the PG side of things.  We are trying to further split the WeBWorK side (UI) from the PG.

pstaabp: This README.md file seems to be identical to the one at the top level of PG.  Perhaps we should revert to some language about running the tests.

pstaabp:
Overall, the switch over to Test2 seems to be the right way to go and this branch is a good start to doing so.  

I got the t/units/basic_module.t to succeed with `prove -vl t/units/basic_module.t`, but not with the electron_volts.t file.  Just realized that this doesn't include your updates to these units.  Eventually, we will need to get all of the changes into develop to test this.

t/README.md
duffee: My mistake in the splitting of the PR.  I'll find it.

duffee:
@pstaabp All the test files have been migrated to Test2.  Of them all, **t/macros/tableau.t** was the most problematic because of the circular references in MathObjects prevent the deep comparisons.  Most of the tests are the same, except for a lot of `->string`.  None have been removed and they all pass now.  I've tried to tidy it up the best I can, but there will be a few format inconsistencies that fresh eyes will spot.  Most of the tests have been organized into subtests based on the comments in the original file.  You'll have to go through and check that the labels and the groupings make sense.

**t/macros/math_objects_basics.t** has examples of migrating `throws_ok` to the `like( dies {} )` pattern.  Everything else was fairly easy.

Once #690 has been merged into PG-2.17, I'll remove the version checking code from **t/units/electron_volts.t**

PR #693, opened 2022-06-13
Fix the prevention of creating degenerate objects with keyboard contols.

drgrice1:
When graphing a parabola (via any of the tools that do so) or a cubic it is possible to create degenerate objects.  The code that is supposed to prevent that wasn't quite working right.  The jsxgraph board object needs to be updated after setting the position of a point programatically.

To test this graph any of those objects and then move one of the defining points of the object around with the keyboard.  For example, graph a three point quadratic and try to move one point onto the same vertical line as another point.  That should not be possible.


pstaabp (approved):

drgrice1:
You can test this with https://github.com/openwebwork/pg/files/8729067/Demo.pg.txt

PR #694, opened 2022-07-07
Tweak the styles for labels, text inputs, selects, and textareas in problems.

drgrice1:
This removes the font size and from these elements.  It also removes the font-family that was set on selects.  The font family still applies to textareas and text inputs.


pstaabp (approved):

taniwallach (approved):
Tested and fixes the size issue. Sticking to the default font for label and select is also a change for the better.
@drgrice1 Thanks for handling this.

PR #696, opened 2022-07-13
Make knowls wait until the page css has loaded before initializing.

drgrice1:
This fixes issue [webwork2#1739](https://github.com/openwebwork/webwork2/issues/1739).

In the PG problem editor, the page in the iframe is not displayed until
after the page loads.  As such the javascript getComputedStyle method
fails.  The same things happens with the graphtool.  The same solution
is now implemented for knowls.

### DIFF
--- a/htdocs/js/apps/Knowls/knowl.js
+++ b/htdocs/js/apps/Knowls/knowl.js
@@ -15,6 +15,11 @@
 	};
 
 	const initializeKnowl = (knowl) => {
+		if (getComputedStyle(knowl)?.display === '') {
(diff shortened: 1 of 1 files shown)


drgrice1:
When testing this make sure to remember to run `npm install` or `./generate-assets` in the `pg/htdocs` directory.

somiaj (approved):
I can confirm this makes solution knowls work for me again in the problem editor.

mgage:
This patch works for me also

somiaj (approved):

PR #700, opened 2022-07-19
Add the package-lock.json file here too.

drgrice1:
There are no known issues here as there were for webwork2, but this is a precautionary measure.


drdrew42 (approved):
*you* get a package-lock! and *you* get a package-lock!

PR #702, opened 2022-07-22
Migrate current tests to Test2

duffee:
Squashed commits after messing up the rebase onto the latest develop branch for #692 .
Recreated the PR and these are the commit messages rescued from the original branch.

tableau.t
    PITA!  The circular references in MathObjects made
    the deep comparisons impossible, so all MathObjects
    were stringified for comparisons.

    WARN and DEBUG message sub redefinitions were removed.
    A "validator" and a stringify sub were added to get around
    some of the comparison issues.

    Organized into subtests.

math_objects_basics.t
    Convert `throws_ok` to the like/dies construct
    and the tests pass.
    Arranged sections in subtests

Three test files migrated from Test::More to Test2 without modification
    (other than moving the environment setup from the script to Test::PG)

This expands on the original t/README.md on the details of testing.

t/context
    Minor changes to migrate tests in t/context from Test::More to Test2
    There were few tests in these files.

    toltype_digits.t organized into subtests by topic

    Use the Test::PG environment with Test2

    Find that the original test produced unexpected results
    and the tests that expose the incorrect behaviour have
    been wrapped in a TODO block

Added unit tests for Units.pm
    Tests to validate the correctness of #677
    Unit tests for Units.pm electron volts
    Tests for parserNumberWithUnits.pl with basic energy conversions

    Adding a Test::PG module in t/lib/ to take the place of t/build_PG_envir.pl


duffee:
@pstaabp I've removed the check for PG-2.17 in **t/units/electron_volts.t**, updated **t/README.md** and your new test **t/math_objects/factorial.t** to use Test2.  The whole test suite is now consistent and mostly organized.

The only things that I can't seem to fix are the repeated calls to `check_score` in **units/basic_parser.t** and `cos(90)` the **todo** block at the end of **contexts/trig_degrees.t**.  Any help would be welcome.

pstaabp:
@duffee Now that 2.17 is official, we'll take a look at this for develop.  I'll see if I can figure out the `check_score` issue.

drgrice1 (changes requested):
This pull request needs some work.  See my comments.

All tests in t/macros pass on the develop branch, but many are failing with this pull request.  That will also need to be fixed before this can be merged.

t/lib/Test/PG.pm
```diff
@@ -0,0 +1,108 @@
...
+
+=head2 TODO
+
+=head3 Quiet warnings in F<t/contexts/trig_degrees.t>
+
+The functions sin, cos and atan2 are redefined and generate warnings.
+C<delete> them from the symbol table before loading the macro.
+
+F<macros/contextTrigDegrees.pl> declares C<$deg> twice, it being
+a required file, the package scope isn't heeded by the second C<my $deg>.
```
drgrice1, line 61: On this ... all we need to do is delete the second declaration of `$deg`.  The package key word does not provide scope.  The first declaration is visible throughout the file regardless of there begin any later package declarations.

t/macros/tableau.t
```diff
@@ -43,25 +47,34 @@ unless (%context && $context{current}) {
...
 my $B = Real(0);
 
-is($A, $B, "test zeroLevel tolerance");
-ok($A == $B, "test zeroLevel tolerance with ok");
+subtest 'test zeroLevel tolerance' => sub {
+	is($A->value, within($B->value, $B->getFlag('zeroLevel')), 'test zeroLevel tolerance');
+	ok($A == $B, 'test zeroLevel tolerance with ok');
+
+	my $real_object = object {
+		prop isa => 'Value::Real';
```
drgrice1, line 66: I realized that the reason some of these tests are failing is because you used `isa`.  `isa` is not in the version of Test2::Suite that is included in the Ubuntu repositories (version 0.000129).

In any case, this should be using `blessed` and not `isa` as that is a more specific and accurate test (and is in version 0.000129).  `isa` only tells you if the given object is an instance of the given class (any class derived from the given class will pass the test).  However, `blessed` tells you more.  It tells you that the object is specifically blessed as that class, not just derived from.

drgrice1 (commented):
After looking at this closer I see that the lib/Test/PG.pm module that was added here should be dropped, and the method in build_PG_envir.pl should be used instead.  The problem is that lib/Test/PG.pm module is not a proper module, and is a hack.  The file declares the Test::PG package, and then improperly switches to the main namespace without actually defining anything that is worthy of being called a module in the Test::PG namespace.

The build_PG_envir.pl needs some work too, but it is closer to being correct.  The main things that it does incorrectly is that it uses `require` to load `PG.pl` and the tests use `require` to load it.  Both of those should be done with `do`.  There are some other details that should be fixed with it as well though.

I have a branch built on top of #709 that does everything I have mentioned here, and gets the tests into order.

t/units/basic_parser.t
```diff
@@ -0,0 +1,202 @@
...
+
+
+subtest 'Verify classes and methods' => sub {
+    isa_ok $joule, 'Parser::Legacy::NumberWithUnits';
+    can_ok $joule,
+        [ qw/cmp splitUnits getUnitNames getUnits TeXunits cmp_parse adjustCorrectValue
+             add_fundamental_unit add_unit string TeX / ],
+        'Can we NumberWithUnits';
+
+    ok my $evaluator = $joule->cmp($Nm), 'Get an AnswerEvaluator';
```
drgrice1, line 73: There is another problem here.  The `cmp` method does not take a single argument like this.  It only takes hash key value pairs.  I think you might be confusing the `cmp` method with the `evaluate` method of the AnwerHash class.

drdrew42:
Closed in favor of #726

PR #705, opened 2022-08-05
Fix the mathview button for essay answers.

drgrice1:
This did not get updated to the new versions of bootstrap and fontawesome.

This fixes the other part of https://github.com/openwebwork/webwork2/issues/1774.


dlglin (approved):
Works as advertised.

drdrew42 (approved):
long live mathview? 😬

PR #707, opened 2022-08-10
Sig fig macro

sfiedle1:
Adding capability for a significant figure context.


pstaabp:
This was supposed to be work done between @sfiedle1  and myself before submitting to the PG repository.  I'm closing for now.

PR #710, opened 2022-08-20
Add a linter workflow to check that files are "perltidy".

drgrice1:
At this point it only checks that perltidy runs without changing the perl files.  In the future it will also lint other files.
    
A script to run perltidy on the source is added for convenience.  It is `bin/run-perltidy`.

Note that this pull request is broken into two commits.  The first adds the workflow and script.  The second is purely the result of running perltidy on the code base.

I have enabled the github actions for this repository so that it should run when this pull request is made, and you will be able to see the result of it in action.  Once this is merged, you should be able to active actions on your personal fork and then it will run each time you push a branch to your fork.  You will probably want to disable the coverage action though, since that doesn't work (and needs to be updated or removed).


pstaabp (approved):
All in favor of this. Wondering about the timing of bringing this in.  Since it touches nearly all files, we'll need to be careful since this will have a ton of merge conflicts with other PRs.  They could be tough to figure out.

drdrew42 (approved):
👍 

Sure, this touches a lot of files, but resolving merge conflicts with existing PRs shouldn't cause that much trouble. "Is the change whitespace/newline?" I think it'll be pretty clear what belongs to this PR and what doesn't...

PR #715, opened 2022-09-01
fix unionTables PTX output

Alex-Jordan:
This fixes an issue with unionTables and valid PTX output. It went unnoticed because I'm sure up to now there have been few instances of using a unionTables problem in a PTX project.


drgrice1 (approved):
You would know what is needed here.

Do you need this to be a hotfix to main?

Alex-Jordan:
I don't think it's needed as a hotfix. It only generates a warning in PTX, and someone who needs the warning to go away can just grab this `unionTables.pl`.

drgrice1:
I will go ahead and merge this.  It is a small change, and only affects PTX output.

PR #718, opened 2022-09-09
update the names of tmp folders for LaTeX images

Alex-Jordan:
This is not important. But recent investigations had me seeing tmp folders with "tikz" in the name that had nothing to do with tikz. This changes that.

### DIFF
--- a/lib/LaTeXImage.pm
+++ b/lib/LaTeXImage.pm
@@ -175,7 +175,7 @@ sub footer {
 sub draw {
 	my $self = shift;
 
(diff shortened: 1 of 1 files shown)


drgrice1 (changes requested):
This is a temporary directory that is deleted as soon as this method is finished.  So the name doesn't matter much at all.  There certainly is no reason to make the name depend on some condition.  Just change "tikz" to "latex" or something.  That works for tikz as well as that is still latex.

Alex-Jordan:
I force pushed the simpler thing that you suggested.

drgrice1 (approved):
Looks good. Thanks.

drgrice1 (approved):

PR #719, opened 2022-09-09
correct 'LaTeXImage' to 'PGlateximage'

Alex-Jordan:
I think this one is worthy of a hotfix. If you have a PGlateximage object, you currently cannot pass it directly to `image()` as you can with a `PGtikz` object or a `WWplot` object. You still have to use `image(insertGraph())`. I think this is just an oversight we made when changing the name of `LaTeXImage.pl` to `PGlateximage.pl`. I'd like to be able to tell people (and documentation) you do not need `insertGraph()`.

If you agree with this as a hotfix, what is the right way to put this into `develop` as well?


drgrice1 (approved):
This looks good.  I am okay with this being a hotfix.  Although this has been this way through the entire release of WeBWorK 2.16!

PR #723, opened 2022-09-10
Fix styles for tables output by begintable/endtable in PGbasicmacros.pl.

drgrice1:
When the begintable method is called, instead of using the deprecated border attribute, instead add the `pg-table` css class.  The css for that class is defined in htdocs/js/apps/Problem/problem.scss`.

This is a small change for this particular table.  Much more like this needs to be done in general to make PG generate valid html without deprecated attributes.

See https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8128#p19764 for a discussion on this particular issue.

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -2756,7 +2756,7 @@ sub begintable {
 		|| $displayMode eq 'HTML_asciimath'
 		|| $displayMode eq 'HTML_LaTeXMathML'
 		|| $displayMode eq 'HTML_img') {
-		$out .= "<TABLE BORDER='1' STYLE='text-align:center;'>\n"
+		$out .= '<table class="pg-table">';
 	}
 	else {
 		$out = "Error: PGbasicmacros: begintable: Unknown displayMode: $displayMode.\n";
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
This fixes the error mentioned. 

I just looked at the nicetables.pl macro.  It seems to use more modern css styling.  Are there any other tables around in other macros?

drdrew42 (approved):

Alex-Jordan:
niceTables.pl is poorly coded and should be rebuilt from scratch someday to have the same functionality. It was maybe my first coding project since a high school Pascal course and I just brute forced my way through it without any attempt at modularity.

But yes, it uses CSS for borders in HTML.

macros/PGbasicmacros.pl
```diff
@@ -2756,7 +2756,7 @@ sub begintable {
 		|| $displayMode eq 'HTML_asciimath'
 		|| $displayMode eq 'HTML_LaTeXMathML'
 		|| $displayMode eq 'HTML_img') {
-		$out .= "<TABLE BORDER='1' STYLE='text-align:center;'>\n"
+		$out .= '<table class="pg-table">'
```
drdrew42, line 2759: missing semicolon -- wouldn't perltidy complain about this?

drgrice1: Yeah, probably.  It wasn't there before. I didn't notice that.  I will add it.

drgrice1: Thanks for noticing that.  By the way, perltidy would automatically add that semicolon!

PR #726, opened 2022-09-13
Update the unit tests to use a full PG environment

drgrice1:
This builds on #709 and #702.  Since my reviews on #702 have not received any attention, consider this a replacement of that pull request.  The commits from that pull request are contained in this one.

The main thing this does is improves the `t/build_PG_envir.pl` file to use the PG environment from #709 to create a complete translator environment for use in the unit tests.  All of that will be needed for unit tests to properly work.

Also a unit test for testing the PGtikz.pl macro is added.


pstaabp:
Want to point out that if your external programs are not the defaults in `conf/pg_config.dist.yml` to create a `conf/pg_config.yml` file with overrides.  The new tikz test failed for me because my pdflatex is not in the standard location.

drgrice1:
Yeah, the pg_config.dist.yml will need to be set correctly for your system for the new tikz test to work.  Unfortunately, I don't think there is any way around that.

drdrew42 (approved):

PR #727, opened 2022-09-13
Fix graphing of the third point for the GraphTool cubictool.

drgrice1:
Currently if you graph three consecutive points where each point is one unit to the left of the previously graphed point, then when the third point is graphed there will be an error that causes the graph tool to become mostly unresponsive.  The problem was an undefined variable being used in phase3 of that tool.

You can test this by doing what is described above with the four point cubic tool in the attached demo problem.
[Demo.pg.txt](https://github.com/openwebwork/pg/files/9556818/Demo.pg.txt)

### DIFF
--- a/htdocs/js/apps/GraphTool/cubictool.js
+++ b/htdocs/js/apps/GraphTool/cubictool.js
@@ -231,7 +231,7 @@
 					// Get a new x coordinate that is to the right, unless that is off the board.
 					// In that case go left instead.
 					let newX = this.point3.X() + gt.snapSizeX;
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
Instead of being unresponsive, I got a `undefined variable x` error and then the graph tool wouldn't accept any other points.  This PR though fixes this.

drdrew42 (approved):
I suppose GraphTool isn't seeing wide usage yet, but perhaps a hotfix is in order?

pstaabp:
Agreed.  We should include this as a hot fix.

PR #728, opened 2022-09-13
Add interactive 3D graphs using plotly JavaScript library

somiaj:
Adds a macro that uses the plotly JavaScript library to create parametric surface plots.

https://plotly.com/javascript/3d-surface-plots/

This could be a possible replacement of the LiveGraphics3D java app that no longer functions, but is currently just an option to include interactive 3D graphics in a pg problem. This is just a draft. Sharing in case others want to use it and to get feedback. Its features are fairly limited at the moment and it won't produce hardcopy TeX output (looking into a way to achieve this -- plotly has a python library as well that could be used for this if javascript cannot be run on the server to create a .png output).

Here is an example problem that creates multiple 3D graphs:

```
DOCUMENT();
loadMacros(
    'PGstandard.pl',
    'PGML.pl',
    'plotly3D.pl',
    'PGcourse.pl'
);

$gr[0] = Graph3D(
    height => 300,
    width  => 300,
    title  => 'A',
    xFunc  => sub { $_[0]*cos($_[1]) },
    yFunc  => sub { $_[0]*sin($_[1]) },
    zFunc  => sub { $_[0]**2 },
    uMin   => 0,
    uMax   => 2,
    vMin   => 0,
    vMax   => 2*pi + 0.1,
);
$gr[1] = Graph3D(
    height => 300,
    width  => 300,
    title  => 'B',
    zFunc  => sub { $_[0]**2 - $_[1]**2 },
);
$gr[2] = Graph3D(
    height => 300,
    width  => 300,
    title  => 'C',
    xFunc  => sub { $_[0]*cos($_[1]) },
    yFunc  => sub { $_[0]*sin($_[1]) },
    zFunc  => sub { $_[0] },
    uMin   => 0,
    uMax   => 2,
    vMin   => 0,
    vMax   => 2*pi + 0.1,
);
$gr[3] = Graph3D(
    height => 300,
    width  => 300,
    title  => 'D',
    zFunc  => sub { exp(-$_[0]**2 - $_[1]**2) },
    uMin   => -2,
    uMax   => 2,
    vMin   => -2,
    vMax   => 2,
);

BEGIN_PGML
Below are graphs for the following functions of two variables:

A) [``f(x,y) = x^2 + y^2``]

B) [``f(x,y) = x^2 - y^2``]

C) [``f(x,y) = \sqrt{x^2 + y^2}``]

D) [``f(x,y) = e^{-x^2 - y^2}``]

[@ $gr[0]->Print @]*
[@ $gr[1]->Print @]*
[@ $gr[2]->Print @]*
[@ $gr[3]->Print @]*

END_PGML
ENDDOCUMENT();
```

### DIFF
--- a/macros/graph/plotly3D.pl
+++ b/macros/graph/plotly3D.pl
@@ -1,0 +1,626 @@
+
+=head1 NAME
+
+plotly3D.pl - Adds Graph3D, an object for creating 3D parametric curves
+and 3D parametric surface plots using the plotly JavaScript library.
+https://plotly.com/javascript/
+
+=head1 DESCRIPTION
+
+Loading this macro adds the Graph3D method which creates a 3D
+graph object. The graph object can be configured by a list of
+options of the form "option => value" (see below).
+
+    $graph = Graph3D(options);
+
+Use the addCurve method to add a parametric curve to the graph.
+The following adds a helix to the graph. The first array is
+the parametric functions x(t), y(t), and z(t), and the second
+array is the bounds with the optional number of points to plot.
+
+    $graph->addCurve(['3*cos(t)', '3*sin(t)', 't'], [0, 6*pi, 150]);
+
+Use the addFunction method to add a two variable function surface
+to the graph. The following adds the function f(x,y) = x^2 + y^2.
+The first input is the function in terms of x and y, followed
+by the x-bounds and y-bounds with the optional number of points to plot.
+Note the total points computed is the product of the two numbers given.
+
+    $graph->addFunction('x^2 + y^2', [-4, 4], [-4, 4]);
+
+Use the addSurface method to add a parametric surface to the graph.
+The following adds a sphere of radius 3. The first array is the
+parametric functions x(u,v), y(u,v), and z(u,v), followed by the
+u-bounds and v-bounds with optional number of points to plot.
+Note the total points computed is the product of the two numbers given.
+
+    $graph->addSurface(
+        ['3*sin(v)*cos(u)', '3*sin(v)*sin(u)', '3*cos(v)'],
+        [0, 2*pi, 30],
+        [0, pi, 30]
+    );
+
+Output the graph in PGML using the Print method.
+
+    [@ $graph->Print @]*
+
+Multiple curves surfaces can be added with the appropriate methods.
+
+=head1 PARAMETRIC CURVES
+
+The addCurve method takes two arrays as input followed by a list
+of options. The first array is three parametric functions, followed
+by the minimum, maximum, and optional number of points to plot.
+If the number of points is not given, it defaults to 100.
+
+    $graph->addCurve(
+        [xFunction, yFunction, zFunction],
+        [tMin, tMax, tCount],
+        options
+    );
+
+The additional options are given in a 'option => value' format.
+The current available options (and defaults) are:
+
+  width      => 5,       The width/thickness of the curve.
+
+  colorscale => 'RdBu',  The colorscale for the curve, which is a heatmap
+                         based on the z-value of the curve.
+
+  opacity    => 1,       The opacity of a curve between 0 and 1.
+
+  funcType   => 'jsmd',  How to interpret the parametric functions (see below).
+
+  variables  => ['t'],   The variable to use in the JavaScript function.
+
+=head1 PARAMETRIC SURFACES
+
+The addSurface method takes three arrays as input followed by a list
+of options. The first array is three parametric functions, followed by
+the minimum, maximum, and optional number of points for each of the
+two variables. If the number of points is not given, it defaults to 20.
+
+    $graph->addSurface(
+        [xFunction, yFunction, zFunction],
+        [uMin, uMax, uCount],
+        [vMin, vMax, vCount],
+        options
+    );
+
+The additional options are given in a 'option => value' format.
+The current available options (and defaults) are:
+
+  colorscale => 'RdBu',      The colorscale for the curve, which is a heatmap
+                             based on the z-value of the surface.
+
+  opacity    => 1,           The opacity of a curve between 0 and 1.
+
+  funcType   => 'jsmd',      How to interpret the parametric functions (see below).
+
+  variables  => ['u', 'v'],  The variables to use in the JavaScript function.
+
+=head2 FUNCTIONS
+
+The addFunction method takes a string, which is a function f(x,y), followed by
+two arrays which give the x-bounds and y-bounds, with optional number of points
+to plot. This is a wrapper for addSurface in which the variables are x and y.
+See addSurface for the list of options. funcType => 'data' cannot be used with
+functions, use addSurface directly to plot data.
+
+    $graph->addFunction(
+        'zFunction',
+        [xMin, xMax, xCount],
+        [yMin, yMax, yCount],
+        options
+    );
+
+=head1 COLORSCALES
+
+The colorscale colors the points of the curve/surface based on the z-value
+of the points. The colorscale can be one of the following predefined names:
+
+    'BdBu', 'YlOrRd', 'YlGnBu', 'Portland', 'Picnic', 'Jet', 'Hot'
+    'Greys', 'Greens', 'Electric', 'Earth', 'Bluered', or 'Blackbody'
+
+You can also define a custom colorscale as a list of colors for values
+ranging from 0 to 1. For example the default RdBu is the following colorset
+(note this must be a string since the array is passed to javascript):
+
+    "[[0, 'rgb(5,10,172)'], [0.35, 'rgb(106,137,247)'],
+      [0.5, 'rgb(190,190,190)'], [0.6, 'rgb(220,170,132)'],
+      [0.7, 'rgb(230,145,90)'], [1, 'rgb(178,10,28)']]"
+
+A colorset can have any number of color points between 0 and 1. To make
+the plot a single color, set the color for 0 and 1 to be the same:
+
+    "[[0, 'rgb(0,200,0)'], [1, 'rgb(0,200,0)']]"
+
+=head1 FUNCTION TYPES
+
+The functions to generate the plot can be either mathematical, JavaScript,
+Perl, or raw data, and this can be controlled using the funcType => type
+option in addCurve or addSurface methods. The valid types are:
+
+  jsmd    This is the default type, in which the functions are converted
+          from math formulas into JavaScript functions to generate the
+          plot. This should accept standard mathematical notation with
+          some exceptions: Multiplication must be an explicit "*".
+          "ucos(v)" is not be accepted, but "u*cos(v)" will. JavaScript
+          considers "-u^2" not well defined, instead use "-(u^2)".
+
+  js      The functions are interpreted as raw JavaScript functions. The
+          functions will be passed the defined variables and return a
+          single value.
+
+  perl    The functions are interpreted as Perl subroutines. The functions
+          will be passed the appropriate number of inputs, and return a
+          single value.
+
+  data    The functions are interpreted as a nested array of data points to
+          be sent directly to plotly to plot. This is useful for static plots
+          in which the points do not need to be generated each time.
+
+=head1 Graph3D OPTIONS
+
+Create a graph object: $graph = Graph3D(option => value)
+The valid options are:
+
+  height      The height and width of the div containing the graph.
+  width
+
+  title       Graph title to print above the graph.
+
+  style       CSS style to style the div containing the graph.
+
+  bgcolor     The background color of the graph.
+
+  image       Image filename to be used in hardcopy TeX output.
+              If no image is provided, the hardcopy TeX output
+              has a message that image must be viewed online.
+
+  tex_size    Size of image in hardcopy TeX output as scale factor from 0 to 1000.
+              1000 is 100%, 500 is 50%, etc.
+
+  tex_border  Put (1) or don't put (0) a border around image in TeX output.
+
+  scene       Add a JavaScript scene configuration dictionary to the plotly layout.
+              Example: scene => 'aspectmode: "manual", aspectratio: {x: 1, y: 1, z: 1},
+              xaxis: { range: [0,2] }, yaxis: { range: [0,3] }, zaxis: { range: [1,4] }'
+              https://plotly.com/javascript/3d-axes/ for more examples.
+
+=cut
+
+sub _plotly3D_init {
+	ADD_JS_FILE('https://cdn.plot.ly/plotly-latest.min.js', 1);
+	PG_restricted_eval("sub Graph3D {new plotly3D(\@_)}");
+}
+
+our $plotlyCount = 0;
+
+package plotly3D;
+
+sub new {
+	my $self  = shift;
+	my $class = ref($self) || $self;
+
+	$plotlyCount++;
+	$self = bless {
+		id         => $plotlyCount,
+		plots      => [],
+		width      => 500,
+		height     => 500,
+		title      => '',
+		bgcolor    => '#f5f5f5',
+		style      => 'border: solid 2px; display: inline-block; margin: 5px; text-align: center;',
+		scene      => '',
+		image      => '',
+		tex_size   => 500,
+		tex_border => 1,
+		@_,
+	}, $class;
+
+	return $self;
+}
+
+sub addSurface { push(@{ shift->{plots} }, plotly3D::Plot::Surface->new(@_)); }
+sub addCurve   { push(@{ shift->{plots} }, plotly3D::Plot::Curve->new(@_)); }
+
+sub addFunction {
+	my $self = shift;
+	my $func = shift;
+	my $b1   = shift;
+	my $b2   = shift;
+	my %opts = @_;
+	my @vars = ($opts{variables}) ? @{ $opts{variables} } : ('x', 'y');
+	my $type = $opts{funcType} || '';
+	if ($type eq 'perl') {
+		$self->addSurface([ sub { $_[0] }, sub { $_[1] }, $func ], $b1, $b2, %opts);
+	} elsif ($type eq 'data') {
+		Value::Error('Functions cannot use data. Use addSurface directly.');
+	} else {
+		$self->addSurface([ @vars, $func ], $b1, $b2, variables => [@vars], %opts);
+	}
+}
+
+sub TeX {
+	my $self = shift;
+	my $size = $self->{tex_size} * 0.001;
+	my $out  = ($self->{tex_border}) ? '\fbox{' : '\mbox{';
+	$out .= "\\begin{minipage}{$size\\linewidth}\\centering\n";
+	$out .= ($self->{title}) ? "{\\bf $self->{title}} \\\\\n" : '';
+	if ($self->{image}) {
+		$out .= &main::image($self->{image}, tex_size => 950);
+	} else {
+		$out .= '3D image not avaialble. You must view it online.';
+	}
+	$out .= "\n\\end{minipage}}\n";
+
+	return $out;
+}
+
+sub HTML {
+	my $self  = shift;
+	my $id    = $self->{id};
+	my $width = $self->{width} + 10;
+	my $title = ($self->{title}) ? "<strong>$self->{title}</strong>" : '';
+	my $plots = '';
+	my @data  = ();
+	my $count = 0;
+	my $scene = ($self->{scene}) ? "scene: { $self->{scene} }," : '';
+
+	foreach (@{ $self->{plots} }) {
+		$count++;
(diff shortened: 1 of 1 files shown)


somiaj:
Added functionality to use static data (vs generate it) and provide an image file for TeX output. I have looked into the LiveGraphics3D.pl macro, and don't see an easy way to make that macro use this instead of the old java app. I'm using this currently and it is working good, so it could be an alternative for 3D graphics. There are currently some OPL problems that are no longer functional that could be switched to this if the desire was there.

taniwallach:
I have not yet tested this, but the idea of providing easy to generate interactive 3D surface plots seems very nice. I would say that for static plots - it certainly does make sense to provide precalculated data.

Another option for interactive 3D in WeBWorK is CalcPlot3D:  https://c3d.libretexts.org/CalcPlot3D/index.html . Monica VanDieren has spoken about her use of CalcPlot3D with WeBWorK via an iFrame embedding, where the URL was generated to include parameters based on parameters chosen in the PG code. There are slides which mention the techniques used at https://meetings.ams.org/math/jmm2021/mediafile/Handout/Paper2767/JMM2021-vandieren-Webwork%2Bflipgrid%2Bcalcplot3D.pdf 

---

I just spent some time looking into the LiveGraphics3D issue, and found this PR and am posting my thoughts here, as it seems an suitable place to consider how to "fix" the OPL problems which depend on functionality dropped in WeBWorK 2.17, and what is the best manner for PG to handle them.

It turns out that the total removal of LiveGraphics3D support from WW 2.17 (done in https://github.com/openwebwork/webwork2/pull/1553 ) may not have been strictly necessary, as it no longer really depended on either a Java applet or Flash. Back in 2015 a JavaScript replacement was coded to allow the OPL problems which had 3D objects using LiveGraphics3D to work again (without needing Java). However, continuing to maintain the converter in https://github.com/goehle/webwork2/blob/7c21e4ca204382d0de7b9915dc728dd927c086a8/htdocs/js/apps/LiveGraphics/liveGraphics.js just to support a small number of legacy problems using the LiveGraphics3D file format does not seem necessary or wise. Thus, in the long-term the removal of that code from the webwork codebase would eventually have probably become necessary.

The JavaScript support for the LiveGraphics3D format was developed by Geoff Goehle in https://github.com/openwebwork/webwork2/pull/587 and https://github.com/openwebwork/pg/pull/211

From what I can tell his approach replaced the Java applet by converting the Mathematica code from the "object" file into the X3D format (a public standard https://www.web3d.org/x3d/what-x3d ) in a manner focused on use by the free x3dom JavaScript framework. However, the conversion was being done each time such a problem was loaded. It seems we would be better off just providing "converted data" directly and loading the relevant tool (x3dom) to display it.

To the best of my understanding, LiveGraphics3D supported more general 3-D objects than surface plots, so I'm not sure that plotly can provide all the functionality that used to be supported. Maybe the isosurface support would suffice? https://plotly.com/javascript/3d-isosurface-plots/ 

Although, it may be possible to build on the work Geoff did to convert the LiveGraphics3D files into data supported by plotly, I doubt that it is worth the effort to investigate plotly as an option for a "general fix" for the OPL problems which depend on LiveGraphics3D at present. For the existing problems it is probably far simpler to just use the output of the existing converter to x3dom format rather than put in the effort to create a converter which outputs data/code in some format plotly can support.

Bottom line: It seems to me that we would be better off by doing a one-time conversion into the "x3d" structure needed and providing that directly in the OPL as something to be loaded into the HTML version of the problems, and having these problems request the additional JS and CSS files needed to use x3dom.

What is the best mechanism which would allow storing the x3d object data in a separate file and have it loaded either by PG (on the server side) or directly by the browser for inclusion? I suspect it is more convenient to let the browser handle the loading of the "data file" and then put the entire "structure" into a prepared placeholder in the HTML DOM on the client side by JavaScript.

@somiaj @drgrice1 @pstaabp @drdrew42 What do you think of the general idea or replacing the LiveGraphic3D data files by "x3d" files? How do you think it should be implemented?

---

I looked at the problem Library/Union/setMVlevelsets/levels-4/levels-4.pg in a WW 2.16 Docker container, and copy-pasted the entire <x3d> block from that problem into a modified version of https://examples.x3dom.org/example/x3dom_helloWorld.html changed to use the CDN source as appears in https://github.com/x3dom/dist but made manual changes to the opening <x3d> tag from the sample file instead of using the one from the copy-pasted data. I could then see and manipulate the 3-D "graph".

I take that a simple proof-of-concept that a dump of the output from the current conversion code should suffice to provide the necessary data for rendering the desired interactive objects without ongoing use of `htdocs/js/apps/LiveGraphics/liveGraphics.js`.

---

Could the code in htdocs/js/apps/LiveGraphics/liveGraphics.js be easily used to build a script to automatically do the conversion of the LiveGraphics3D files (those in the OPL) into the relevant x3d XML data and save that into a file?

Is it worth the effort to automate the process? It could be that we could just manually open the relevant 50 or so OPL problems (on WW 2.16) and manually copy-past the relevant data into files to put into the OPL by hand

---

Note: x3dom seems to be available via npm (but a somewhat old version) so there is certainly no need to include the library in the webwork repositories as was done in the past.

Note: Geoff's code apparently only depended on Flash as a fallback for when the JavaScript approach failed. (See https://github.com/openwebwork/webwork2/commit/93fa74591b016628d5c85dd1090ef6dfe631fc3d )

drgrice1:
@taniwallach:  Note that the "converter" method was also not working anymore, and that also relies on obsolete techniques.

taniwallach:
> @taniwallach: Note that the "converter" method was also not working anymore, and that also relies on obsolete techniques.

@drgrice1 Thanks. I was not aware that it "broke". Would dumping the X3D "DOM" structure on a WeBWorK 2.16 server into a file (everything between the opening`<x3d>` tag and the closing `</x3d>` tag) suffice to allow reading that file (say via an AJAX request) and placing it into the problem where it would have gone before? A revised LiveGraphics3D.pl could then use PGalias to put the file in a web-accessible location, and the browser client could place it where needed. Alternately, it could just be put straight into the generated HTML of the problem. One of those options seems to me to be the path of least resistance to fixing those OPL problems.

somiaj:
I do like CalcPlot3D, though would prefer to be able to just use their javascript library in a problem vs an iframe. I hadn't realized there was some work to use CalcPlot3D in webwork, might have saved me some time and their surface graphs are a bit more standard for what is seen in Calculus III texts.

I went with plotly since I like matplotlib and it has a similar feel, was well documented, and I was able to get something up and running for my class this semester fairly quickly. Since I haven't used LiveGraphics3D, I'm unsure of all of its capabilities, but I do agree that to get the OPL problems up and running again, the easiest approach would be preferable, and plotly might not be that.

I have thought about extending the capabilities to use plotly to create other graph types, since it has a wide range of graphs it can create, both 2D and 3D, but the graphs are more geared towards data visualization.

lahvak:
Just today I was experimenting with another option for including static interactive 3d objects in WebWorK: [asymptote](https://asymptote.sourceforge.io/) can generate 3d interactive images in html files, like for example [this one](https://stuff.lahvak.space/calc3/notes/surfaces/cusp.html).  They can be easily included in a WeBWorK problem in an iframe.  You can just include them in the problem's directory like png images. Asymptote can also create a pdf version to include in the hardcopy.

Theoretically it would be possible to even generate these dynamically on the server, similarly to tikz images, but I think that would be too slow and resource hungry, asymptote used gpu to create these and takes quite a bit of time.

drgrice1:
It seems that you are correct @taniwallach.  The javascript method still does work.  I guess I didn't test that like I thought I did.  I will put in a pull request to reinstate that.

drgrice1:
I agree with @pstaabp.  Although the `LiveGraphics3D.pl` macro is fixed, a better replacement is needed.  This may be a good way to go.  Another option would be to use the JSXGraph library that the `GraphTool.pl` macro uses.  JSXGraph supports 3D graphs as well.  It is better for interactive graphing.  Plotly is better for data visualization.  So the choice depends somewhat on what is desired.

somiaj:
I have updated how this work slightly by moving adding a surface into its own method. This currently allows adding multiple surfaces. Here is the example problem updated to fit the changes.

```perl
DOCUMENT();
loadMacros(
    'PGstandard.pl',
    'PGML.pl',
    'plotly3D.pl',
);

$gr[0] = Graph3D(
    height => 300,
    width  => 300,
    title  => 'Graph A',
);
$gr[0]->addSurface(
    xFunc  => sub { $_[0]*cos($_[1]) },
    yFunc  => sub { $_[0]*sin($_[1]) },
    zFunc  => sub { $_[0]**2 },
    uMin   => 0,
    uMax   => 2,
    vMin   => 0,
    vMax   => 2*pi + 0.1,
);
$gr[1] = Graph3D(
    height => 300,
    width  => 300,
    title  => 'Graph B',
);
$gr[1]->addSurface(zFunc  => sub { $_[0]**2 - $_[1]**2 });
$gr[2] = Graph3D(
    height => 300,
    width  => 300,
    title  => 'Graph C',
);
$gr[2]->addSurface(
    xFunc  => sub { $_[0]*cos($_[1]) },
    yFunc  => sub { $_[0]*sin($_[1]) },
    zFunc  => sub { $_[0] },
    uMin   => 0,
    uMax   => 2,
    vMin   => 0,
    vMax   => 2*pi + 0.1,
);
$gr[3] = Graph3D(
    height => 300,
    width  => 300,
    title  => 'Graph D',
);
$gr[3]->addSurface(
    zFunc  => sub { exp(-$_[0]**2 - $_[1]**2) },
    uMin   => -2,
    uMax   => 2,
    vMin   => -2,
    vMax   => 2,
);

BEGIN_PGML
Below are graphs for the following functions of two variables:

A) [``f(x,y) = x^2 + y^2``]

        [@ $gr[0]->Print @]*

B) [``f(x,y) = x^2 - y^2``]

        [@ $gr[1]->Print @]*

C) [``f(x,y) = \sqrt{x^2 + y^2}``]

        [@ $gr[2]->Print @]*

D) [``f(x,y) = e^{-x^2 - y^2}``]

        [@ $gr[3]->Print @]*

END_PGML
ENDDOCUMENT();
```

pstaabp:
This looks nice (much better than the current live3d code) and the interaction works quite well.  I'm guess we can also use this for 3D line graphs and perhaps 3D vector fields.

@drgrice1 asked about using JSXgraph underneath as another option.  Seems like we should test both before making a decision for the direction to go with interactive 3D graphs.

somiaj:
@pstaabp There might also be room for both, as they do slightly different things. Also plotly could be extended to barcharts, scatter plots, histograms, piecharts, and much more depending on what people need from it. Though a lot might depend on what people want/need and who is going to put in the effort.

I am continuing to improve/develop this as I'm currently using it for my Calculus III course and so far it is working well, creates nice graphs that are helping student visualize some common surfaces in 3D. I have added the ability to add curves to the 3D plot. I have also added the ability to use JavaScript functions client side instead of perl functions server side to generate the plots.

I'll leave this here as I keep improving it for my use, though probably won't need to be merged unless others start using it. If anyone starts using this and needs features added, just let me know.

somiaj:
Yet again updated this. I switched to using JavaScript functions by default to save on server load, and added a converter from math formulas to JavaScript so the parametric curves/surfaces can be written easier. Also simplified the input to give an array of functions vs having to define each and every option. Here is the new test case.

```perl
DOCUMENT();
loadMacros(
    'PGstandard.pl',
    'PGML.pl',
    'plotly3D.pl',
);

$gr[0] = Graph3D(height => 300, width => 300, title => 'Graph A');
$gr[0]->addSurface(['u*cos(v)', 'u*sin(v)', 'u^2'], [0, 2], [0, 2*pi]);

$gr[1] = Graph3D(height => 300, width => 300, title => 'Graph B');
$gr[1]->addSurface(['u', 'v', 'u^2 + v^2'], [-2, 2], [-2, 2]);

$gr[2] = Graph3D(height => 300, width => 300, title => 'Graph C');
$gr[2]->addSurface(['u*cos(v)', 'u*sin(v)', 'u'], [0, 2], [0, 2*pi]);

$gr[3] = Graph3D(height => 300, width => 300, title => 'Graph D');
$gr[3]->addSurface(['u', 'v', 'exp(-(u^2 + v^2))'], [-2, 2], [-2, 2]);

$gr[4] = Graph3D(height => 300, width => 300, title => 'Graph E');
$gr[4]->addSurface(
    ['2*sin(v)*cos(u)', '2*sin(v)*sin(u)', '2*cos(v)'],
    [0, 2*pi],
    [0, pi],
    colorscale => 'Greys',
    opacity    => 0.3
);
$gr[4]->addCurve(
    ['2*sqrt(1 - (t/(6*pi))^2)*cos(t)', '2*sqrt(1 - (t/(6*pi))^2)*sin(t)', 't/(3*pi)'],
    [-6*pi, 6*pi, 400],
    width => 7
);

BEGIN_PGML
Below are graphs for the following functions of two variables:

A) [``f(x,y) = x^2 + y^2``]

        [@ $gr[0]->Print @]*

B) [``f(x,y) = x^2 - y^2``]

        [@ $gr[1]->Print @]*

C) [``f(x,y) = \sqrt{x^2 + y^2}``]

        [@ $gr[2]->Print @]*

D) [``f(x,y) = e^{-x^2 - y^2}``]

        [@ $gr[3]->Print @]*

E) Curve wrapped around a sphere.

        [@ $gr[4]->Print @]*

END_PGML
ENDDOCUMENT();
```

somiaj:
Moved the macro into the new location, and this is no longer a draft. I'm currently using this macro in my classes, but it might take others using it to help spot issues or improve it. For my use case it works well to create interactive 3D graphs for my Calc III course.

macros/graph/plotly3D.pl
```diff
@@ -0,0 +1,626 @@
...
+
+  scene       Add a JavaScript scene configuration dictionary to the plotly layout.
+              Example: scene => 'aspectmode: "manual", aspectratio: {x: 1, y: 1, z: 1},
+              xaxis: { range: [0,2] }, yaxis: { range: [0,3] }, zaxis: { range: [1,4] }'
+              https://plotly.com/javascript/3d-axes/ for more examples.
+
+=cut
+
+sub _plotly3D_init {
+	ADD_JS_FILE('https://cdn.plot.ly/plotly-latest.min.js', 1);
```
drgrice1, line 194: @somiaj:  This should not load `plotly` from a cdn.  It should be installed via npm with `npm install plotly.js-dist-min`.  The third-party-assets.json file should list the appropriate cdn for it.  It seems that would be `https://cdn.jsdelivr.net/npm/plotly.js-dist-min/plotly.min.js`.  That gives the system administrator the choice of serving it directly (which many prefer -- I certainly do) or serving it from a cdn.  It is also a bad idea to use `plotly-latest.min.js`.  Instead use a specific version.   You can use the actual latest version at this point for now, but this may cause problems later.  Always specify versions.

macros/graph/plotly3D.pl
```diff
@@ -0,0 +1,626 @@
...
+
+	return "\n" . <<END_OUTPUT;
+<div style="width: ${width}px; $self->{style}">
+	$title
+	<div id="plotlyDiv$id" style="width: $self->{width}px; height: $self->{height}px;"></div>
+</div>
+<script>
+document.addEventListener("DOMContentLoaded", function() {
+	$plots
+	var plotlyLayout$id = {
```
drgrice1, line 287: This shouldn't use `var`.  It should be `const` instead.

macros/graph/plotly3D.pl
```diff
@@ -0,0 +1,626 @@
...
+sub genJS {
+	my $self = shift;
+	return '' unless ($self->{funcType} =~ /^js/);
+	my $id    = shift || 1;
+	my $count = shift || 1;
+	my $vars  = join(', ', @{ $self->{variables} });
+	my $JSout = <<END_OUTPUT;
+	var xData${id}_$count = [];
+	var yData${id}_$count = [];
+	var zData${id}_$count = [];
```
drgrice1, line 458: There is no `var` in modern javascript.

drgrice1 (commented):
This was merged before it was properly reviewed.  There are some changes needed.  Please make the requested changes and put in another pull request.

drgrice1:
This macro will also need some more improvements to work for webwork3.

macros/graph/plotly3D.pl
```diff
@@ -0,0 +1,626 @@
...
+
+package plotly3D;
+
+sub new {
+	my $self  = shift;
+	my $class = ref($self) || $self;
+
+	$plotlyCount++;
+	$self = bless {
+		id         => $plotlyCount,
```
drgrice1, line 208: Using a simple count for html ids is certainly not good enough.  If multiple problems use this macro on the same page, then this will create id clashes.

drgrice1: You should use `$main::PG->getUniqueName('plotly3D')` to get an id that will be unique for this problem, and different  from any other problems that might be on the page that also use the macro.  That uses `PG_alias` and incorporates things into the returned name that make sure this is the case.

drgrice1:
This is really injecting a lot of javascript into the page.  It would be better to move that javascript into an actual javascript file in htdocs.  It should be possible to pass in the data by injecting a minimal amount of javascript into the page.  See the way the parserGraphTool.pl macro does this.  Although, I think I inject to much into the page with that macro, and intend to do that differently.  I think we need modules!

macros/graph/plotly3D.pl
somiaj: I set the version to be 2.18.2, the latest version in `package.json` and updated the cdn to match that version. Any reason to use jsdelivr over the suggested CDN on the plotly.js github page of `https://cdn.plot.ly/plotly-2.18.2.js`. Assuming this is a more generic CDN vs one just for plot.ly. Also do I need to do anything with the `package-lock.json` file?

drgrice1: For consistency it is best to use the jsdelivr one.  It will be the same as the one installed by npm.

macros/graph/plotly3D.pl
somiaj: Thanks, I still get confused on when I can use `const` vs `let` when I want to be able to modify the variables. Since I'm just pushing to the array, `const` seems to work. I was just following the plotly docs, many places have not updated things correctly.

macros/graph/plotly3D.pl
somiaj: This didn't quite work, as the name includes `-` which won't work in variable names, but `$main::PG->getUniqueName('plotly3D') =~ s/-/_/gr` seems to work under my initial tests.

macros/graph/plotly3D.pl
drgrice1: When you run `npm install plotly.js-dist-min` it will automatically add what is needed to the package-lock.json file.  You will need to commit that and include that in your pull request.

macros/graph/plotly3D.pl
drgrice1: Use `const` anytime you never assign to the variable again.  Use `let` if you need to assign to it again after initialization.  That is the only criteria.

macros/graph/plotly3D.pl
drgrice1: Yes, you need to do the `s/-/_/gr`.  I forgot to mention that.

macros/graph/plotly3D.pl
drgrice1: By the way, hopefully you did not manually edit the `package.json` file to add `ploty`.  You should just run `npm install plotly.js-dist-min` as I said.  That will add what is needed to both the `package.json` file and the `package-lock.json` file.  By default it will use the latest version, which is 2.18.2.

somiaj:
I'll have to look at what you did to try to move some of the javascript out of the page. It might take me a bit to figure out how to do this nicely, since from my viewpoint, a lot of the javascript is dependent on what the function the user wants to plot is.

Is there a macro where the javascript could be moved to the `<head></head>` of the page, or by modules, would this be a way we can generate a .js file that is then loaded?

Anyways, I created a pull request with the other requested changes, we can move the conversation over there.

drgrice1:
No javascript should be added to `<head></head>`.  Never do that.

Don't worry about this for now if it is too much work.  I can take a look at it at some point.  Eventually we will want to convert all of our javascript to using modules.  It is better than using deferred javascript as it makes it possible to avoid polluting the global namespace.

drgrice1:
By the way, that is the worst part of the javascript in this macro.  It horrendously pollutes the global namespace.  That is something that really does need work.

somiaj:
Even though all the javascript is inside functions() that are run on page load?

drgrice1:
Yeah, I see now.  It is all inside a function.  So that is good.

PR #732, opened 2022-09-30
Restore the functionality of the LiveGraphics3D.pl macro.

drgrice1:
It turns out @taniwallach was correct in that the javascript conversion method does work.  See the discussion in #728.

Note that the usage of the use_javascript_for_live3d option has been removed.  That is just how it works.  Anything else will fail anyway.

This perhaps should be a hotfix?  Vote please.

### DIFF
--- a/htdocs/js/apps/LiveGraphics/liveGraphics.js
+++ b/htdocs/js/apps/LiveGraphics/liveGraphics.js
@@ -1,0 +1,913 @@
+// liveGraphics.js
+// This is a javascript based replacement for the LiveGraphics3D java library
+// 
+// This program is free software; you can redistribute it and/or modify it under
+// the terms of either: (a) the GNU General Public License as published by the
+// Free Software Foundation; either version 2, or (at your option) any later
+// version, or (b) the "Artistic License" which comes with this package.
+// 
+// This program is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+// Artistic License for more details.
+
+var LiveGraphics3D = function (container, options) {
+    var my = this;    
+    
+    // define x3d container and scene
+    var x3d = $("<x3d/>").appendTo(container)
+	.css('width',options.width+'px')
+	.css('height',options.height+'px')
+	.css('border','none')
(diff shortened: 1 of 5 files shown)


drgrice1:
By the way, here is a list of all of the OPL problems that use the LiveGraphics3D.pl macro.
[live-graphics-3d-problems.txt](https://github.com/openwebwork/pg/files/9684990/live-graphics-3d-problems.txt)

lahvak:
I think it should be a hotfix, there are probably plenty of instructors in the same situation as I was: I am reusing all my homework set from last year in my Calc 3 class.  After publishing the last one, emails started coming in that two of the problems do not work.  I was able to install the necessary javascript files on our server, and it indeed seems to work fine, did not have any further complaints since Wednesday.  I came here last night to open an issue on it only to discover that the discussion is already going on :).

drgrice1:
Actually, here is a more complete list of OPL problems that use the LiveGraphics3D.pl macro.  This adds problems that use that macro indirectly via some FortLewis macros in the OPL.
[live-graphics-3d-problems.txt](https://github.com/openwebwork/pg/files/9685324/live-graphics-3d-problems.txt)

somiaj (approved):
Tested a few of the surface plots and they all seem to be working with this. Probably worth a hotfix, as this makes a bunch of currently broken problems functional again, and shouldn't mess with anything except problems that use this library.

A side comment is an improvement would be to draw the coordinate axes. There are numbers in each axis direction on the surface plots, but no axes drawn, which makes reading the graph a bit awkward (though this might be more than the fix of just getting functionality back and I don't recall what the original java applet did with coordinate axes).

drgrice1:
This pull request restored the functionality of the macro to what it was before. It does not attempt to make improvements on it.

drgrice1:
In fact what you are asking seems to be something the problems need to implement. Some of them do. I see axes on some problems.

somiaj:
Ahh okay, so the feature is there, it just isn't being used. The problems I looked at weren't using axes, but I only looked at some of the list you provided.

drgrice1:
I don't really know much about this. All of the OPL problems are listed in the previously attached file.  Some use a different technique that seems to yield nicer imagery.

taniwallach:
I also would like this to also be a hotfix.

pstaabp (approved):
Looks good and hot-fix seems right for this.

PR #736, opened 2022-10-04
Add a script to perltidy PG problem files

drgrice1:
The script just runs perltidy but it preprocesses the code in the file with the default_preprocess_code of the translator.  After that the code is valid perl that perltidy works on.  Then it postprocesses the file to change the code back to PG (with BEGIN_TEXT/END_TEXT, BEGIN_PGML/END_PGML, etc).

The script accepts all arguments that perltidy does.  Note that by default the script attempts to use the .perltidyrc file in the PG root directory.  If the PG_ROOT environment variable is not set or the script can't find the PG root directory for some reason, then perltidy defaults are used.  You can also explicitly use a different profile by passing the perltidy -pro argument.

### DIFF
--- a/bin/perltidy-pg.pl
+++ b/bin/perltidy-pg.pl
@@ -1,0 +1,105 @@
+#!/usr/bin/env perl
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2022 The WeBWorK Project, https://github.com/openwebwork
+#
+# This program is free software; you can redistribute it and/or modify it under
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
Tried this on a large number of my own problems.  Seems work well.

Since this is optional, I would agree to merge in.  @drgrice1 I know you mentioned whether or not we should add it.

Also, we may down the road require that all submission to the OPL be perl-tidied.

drdrew42 (approved):

PR #738, opened 2022-10-07
Load pgAuxiliaryFunctions in niceTables.pl to make the max() function…

### LINKED ISSUE #735: niceTables.pl depends on PGauxiliaryFunctions.pl
The `DataTable` subroutine in niceTables.pl calls the `max` function, which is defined in PGauxiliaryFunctions.pl, so if a PG file loads niceTables.pl but doesn't load PGauxiliaryFunctions.pl (or PGstandard.pl, which in turn loads PGauxiliaryFunctions.pl) then any calls to DataTable will result in an error.

The easy fix is to load PGauxiliaryFunctions.pl from within the `DataTable` subroutine.  Is that the right approach?

dlglin:
… available

This fixes #735 

MWE is attached.
[TestNiceTables.pg.txt](https://github.com/openwebwork/pg/files/9735238/TestNiceTables.pg.txt)

### DIFF
--- a/macros/niceTables.pl
+++ b/macros/niceTables.pl
@@ -205,6 +205,9 @@ CSS SYNTAX PRIMER
 
 =cut
 
+# ^uses loadMacros
+loadMacros("PGauxiliaryFunctions.pl");
+
 sub _niceTables_init { };    # don't reload this file
 
 sub DataTable {


drgrice1 (approved):
Looks good.

pstaabp (approved):

PR #741, opened 2022-10-23
Update the unit test workflow so that it works without a docker image

drgrice1:
This removes the dependence of the unit tests on a personal docker image.  All references to eltendor's personal docker image are removed.
    
The workflow has been renamed from coverage to unit-tests.  Unit tests are what we need.  Code coverage is of lesser importance to us at this point.  Furthermore, the code coverage may never work well for PG.  It completely misses the code in the macro files which is a rather severe shortcoming for pg as a large portion of the code is there.  Also the Devel::Cover module does not play well with the WWSafe module now used by some of the unit tests.
    
The docker/pg-no-ww.Dockerfile has been renamed to docker/pg.Dockerfile and updated to use the same approach (an Ubuntu 22.04 base image) as the unit test workflow.  There is no need to reference webwork in the file name.  The docker/test-docker.Dockerfile has been removed.  That is redundant.
    
The linter workflow has been switched to using the ubuntu-22.04 runner image for consistency.  Note that the ubuntu-latest image is Ubuntu 20.04.  Also switch from actions/checkout@v2 to actions/checkout@v3 (also used by the unit-tests workflow).  The v2 action uses node 12, and this causes a GitHub action warning about node 12 being deprecated in actions.  The v3 action used node 16.
    
Also update the cpanfile.  It was missing some dependencies.  In addition we require a version of Test2::V0 that has check_isa (so 0.000139 or newer), so add the version requirement.

Note that this is built on top of #724.  So this should be considered after that is merged.


pstaabp:
When running
```
docker run -it --rm -v `pwd`:/opt/webwork/pg pg prove -r t
```

I got the following error:
```
t/tikz_test/tikz_image.t ........ The pdf file was not created. at /opt/webwork/pg/lib/LaTeXImage.pm line 252.
Image file production failed. at /opt/webwork/pg/lib/LaTeXImage.pm line 278.
t/tikz_test/tikz_image.tCan't print graph to /opt/webwork/pg/htdocs/tmp/images/8aee17ab-0dfb-3125-b694-557f304cc6af___4731cff0-6a93-350c-a6f2-8817d2d83785.svg<BR> at /opt/webwork/pg/lib/PGcore.pm line 723.
```

Opening a shell in the docker and then running `prove -r t` produced the same error.  I thought it might be a permission issue, but the user in the shell in the docker is root, so shouldn't have an issue.

pstaabp:
Finally dug into this.  I had a `conf/pg_config.yml` file that had programs for my Mac (mainly pdflatex) and docker was using that config to run the tikz code and failing.  removing this and using the default worked. Everything runs clean now.

pstaabp (approved):

drgrice1:
Yeah, the docker build mounts the local pg directory.   It doesn't download a clean pg clone.  So the contents of your pg directory can affect the tests.

Issue #740, opened 2022-10-20
Automatic correct answer for answers with units


pkohl:
Hey folks,

Short version:  Entering an answer with units as 0 with an extra m/A appended to the units appears to always result in a green box, even if the answer isn't supposed to be zero.

Long version:  We found what appears to be a way to get a green box for any problem that requests an answer with units, if those units aren't compound units.  So, it applies if a problem asks for a number with units of V, T, or N, but not one that asks for an answer with units of m/s.

As I understand things, Webwork interprets A as Angstroms, so if one were to append m/A to an answer, it wouldn't change the dimensional correctness of the answer, but it would add several factors of 10 the number, which should result in the answer being evaluated as incorrect.  And it usually does, unless one puts in the number zero.  So if the answer to a problem is supposed to be "5 V", "0 V*m/A" gets a green box.  Unless we're doing something wrong with our code, this appears to be reproducible.  I'm including a couple screenshots of the answer reports and the code for one of the offending problems.

I've had issues with the difference between numbers and strings before, especially when small numbers are involved, and I tried massaging the answer variable with Real(), Compute(), and Compute(uc()).  It doesn't look like NumberWithUnits will take the output from Real() or Compute(uc()).  It does look like it'll take the output from Real(), but it doesn't fix the problem.  Still, I'm definitely not sure that this isn't user error.  Any suggestions would be appreciated.

<img width="705" alt="WebworkZeroBug_PHGN100" src="https://user-images.githubusercontent.com/29390081/197022780-5680ddc3-df09-424f-88be-bcd4cc95a6be.PNG">
<img width="731" alt="WebworkZeroBug_B" src="https://user-images.githubusercontent.com/29390081/197022782-8904e530-d3bb-47cb-b354-eb4a43b5f788.PNG">
<img width="677" alt="WebworkZeroBug_A" src="https://user-images.githubusercontent.com/29390081/197022784-0b990168-74c2-4619-b9b6-6d21fce40040.PNG">

Code example: 

DOCUMENT();
loadMacros(
"PGstandard.pl",
"PGgraphmacros.pl",
"parserPopUp.pl",
"PGML.pl",
"MathObjects.pl",
"parserMultiAnswer.pl",
"scaffold.pl",
"PGcourse.pl",
"PGessaymacros.pl",
"parserNumberWithUnits.pl",
"PGchoicemacros.pl",
"Constants.pg",
"parserRadioButtons.pl",
"Functions.pg"
);
[...]

PR #742, opened 2022-10-23
Fix the adjustment of unit answers for comparison.

drgrice1:
Instead of adjusting the correct answer by the correct factor over the student factor, instead adjust the student answer by the the student factor over the correct factor.  Furthermore, only do that temporarily while the comparison is made, and then revert it.

The adjustment is made at a much lower level now.  Instead of doing it in the `cmp_parse` method, do it in the `compare` method which is the overload method for the `<=>` operator.  That is the lowest level of evaluation in the MathObjects answer processing sequence.  This means that it is after the display values for the student answers are set.

Some additional unit tests for this have been added in t/units/basic_parser.t.  Also the "todo" tests that are related to this are now actual tests.

This fixes issue #740.

### DIFF
--- a/t/units/basic_parser.t
+++ b/t/units/basic_parser.t
@@ -32,9 +32,8 @@ use lib "$ENV{PG_ROOT}/lib";
 use Parser::Legacy::NumberWithUnits;
 use Units;
 
-loadMacros("parserNumberWithUnits.pl");
-
-Context("Numeric");
+loadMacros('parserNumberWithUnits.pl');
+loadMacros('parserFormulaWithUnits.pl');
 
 # define some basic objects
 my $joule             = NumberWithUnits(1, 'J');
@@ -44,7 +43,7 @@ my $energy_base_units = NumberWithUnits(1, 'kg*m^2/s^2');
 subtest 'Verify classes and methods' => sub {
 	isa_ok $joule, 'Parser::Legacy::NumberWithUnits';
 	can_ok $joule, [
-		qw/cmp splitUnits getUnitNames getUnits TeXunits cmp_parse adjustCorrectValue
+		qw/cmp splitUnits getUnitNames getUnits TeXunits cmp_parse
 			add_fundamental_unit add_unit string TeX /
 		],
 		'Can we NumberWithUnits';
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
A question based on your description: Suppose the coded correct answer is `1 amu` (which is on the order of 1E-27 kg). Do I understand right that this will be converted to roughly 1E-27 kg before comparison with a student answer (that has also been translated to kg)? If so, would a student answer of `0 amu` or `0 kg` be accepted, with 1E-27 well below the zero tolerance level?

Alex-Jordan:
I haven't done any coding for this issue. But in my head I was outlining an approach where comparison would be made twice. Once using the left operand's units scale, and once using the right operand's units scale. And it would have to pass both tests to count as equal. (You've made me see a third comparison would be natural as well, reducing both operands to the fundamental units scale.) Is there any advantage to an approach like that over what you have done here?

drgrice1:
I switched it to just adjusting the student answer, and not changing the correct at all.  The amu example was problematic.  I am not sure about doing multiple comparisons.  There is a challenge in determining which to trust.

Alex-Jordan:
If we do comparison using the author's unit scale, at least then we can always put responsibility on the author to use a reasonable unit scale given the answer, and/or to adjust context flags like tolType and zeroTolLevel.

drgrice1:
Just to be clear, you mean doing what it is now doing, right?  That is adjusting the student answer to that of the correct (the authors) answer.

Alex-Jordan:
Yes, that's what I mean. Affirming that what you changed it to is going to be good, even if not as thorough as the double comparison I had been considering. The double comparison may not have any real advantage over what you have now.

drgrice1:
After doing some more testing, it seems that the comparison now used by this pull request always gives the correct result.  The others might also be correct, but they are also the ones that may not be correct.  At least in the testing and analysis I have done so far.

Alex-Jordan:
To clarify, when you wrote "correct" just now, do you mean correct behavior from WeBWorK/PG? (There are a few other meanings for "correct" in this thread.)

drgrice1:
Yeah, when I referred to the "correct result" I meant the compare method returns the right thing.  That is 0 if the answers are equal, and nonzero if they aren't.

t/units/basic_parser.t
```diff
@@ -153,34 +150,63 @@ subtest 'Check display methods' => sub {
...
+
+subtest 'Check possible answers compared to the number 0.005 T' => sub {
+	my $correct = NumberWithUnits('0.005 T');
+
+	is check_score($correct, '0.005 T'),        1, '0.005 T is correct';
+	is check_score($correct, '5*10^-13 T*m/A'), 1, '5*10^-13 T*m/A is correct';
+	is check_score($correct, '0 T*m/A'),        0, '0 T*m/A is incorrect';
+};
+
+subtest 'Check possible answers compared to the formula 0.005 T' => sub {
```
pstaabp, line 183: Isn't this subtest exactly the same as the one above?

PR #745, opened 2022-11-15
Wrap solutions and hints in <div>...</div>.

drgrice1:
Don't just start a <p> tag, and never end it.  This removes the <p> tag and wraps solutions and hints in properly started and ended <div> tags.

In addition to making this output valid html, this ensures that knowl contents are placed immediately after a hint or solution.

This doesn't fix the bigger issue of the $PAR usage.  Problems in the OPL use this frequently.  The problem is that $PAR starts a <p> tag, but never ends it.  This results in invalid html.

### DIFF
--- a/macros/PGbasicmacros.pl
+++ b/macros/PGbasicmacros.pl
@@ -1186,7 +1186,7 @@ sub SOLUTION {
 	return "" if $solution_body eq "";
 
 	if ($displayMode =~ /HTML/) {
-		TEXT($PAR, knowlLink(SOLUTION_HEADING(), value => $solution_body, type => 'solution'));
+		TEXT('<div>', knowlLink(SOLUTION_HEADING(), value => $solution_body, type => 'solution'), '</div>');
 	} elsif ($displayMode =~ /TeX/) {
 		TEXT(
 			"\n%%% BEGIN SOLUTION\n"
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Yeah I noticed in at least one browser's DOM, there are hidden input fields inside the `p` that contains the solution. That shouldn't be.

I haven't literally tested this. As a cheap substitute for now, I edited a DOM to put a close `p` tag where this puts one. When I do that, I see new problems. That leads to this `p` having `div`s inside it (in the DOM) after I activate the knowl, which is not valid HTML either. (Although does a DOM have to express valid HTML?) Is this just a symptom of my cheap substitute for actually testing this?

I'm wondering if a `p` is really the right container for the "Solution" clickable in the first place.

drgrice1:
> Yeah I noticed in at least one browser's DOM, there are hidden input fields inside the `p` that contains the solution. That shouldn't be.

Yes, I have noticed that as well.  In fact I noticed that while working on the PreTeXt javascript.  I was working on making MathQuill work and the hidden inputs inside that `p` were causing a problem because that gets removed by PreTeXt's javascript.  That was this summer, and that is in fact when I created this branch, and for that purpose.

> I haven't literally tested this. As a cheap substitute for now, I edited a DOM to put a close `p` tag where this puts one. When I do that, I see new problems. That leads to this `p` having `div`s inside it (in the DOM) after I activate the knowl, which is not valid HTML either. (Although does a DOM have to express valid HTML?) Is this just a symptom of my cheap substitute for actually testing this?

The `p` having a `div` inside won't happen with this pull request.  Editing the DOM directly is tricky.

What about the knowl content is not valid HTML?  I am actually rather certain that it is.

> I'm wondering if a `p` is really the right container for the "Solution" clickable in the first place.

Yeah, that is probably true.  However, it is valid HTML.

drgrice1:
I see.  So the knowl contents end up being placed inside the `<p>` tags.  So it seems the correct thing here is to use a `<div>`.

drgrice1:
I switched to using `<div>`'s instead.

Alex-Jordan (approved):

pstaabp (approved):

drdrew42 (approved):

PR #749, opened 2022-11-19
Fix an errant regex check that doubly base64 encoded Geogebra applet states.

drgrice1:
I disovered this thanks to an observation of @drdrew's while attempting to diagnose an issue with a Geogebra problem written by another author. It turns out that one part of the problem was the applet state being doubly base64 encoded.


drdrew42 (approved):

drdrew42:
I would advocate for a hotfix here...

pstaabp (approved):

PR #751, opened 2022-11-23
Fix an errant regex check that doubly base64 encoded Geogebra applet states (hotfix)

drgrice1:
I disovered this thanks to an observation of @drdrew's while attempting to diagnose an issue with a Geogebra problem written by another author. It turns out that one part of the problem was the applet state being doubly base64 encoded.

### DIFF
--- a/macros/AppletObjects.pl
+++ b/macros/AppletObjects.pl
@@ -137,12 +137,12 @@ sub insertAll {
 
 	my $base_64_encoded_answer_value;
 	my $decoded_answer_value;
-	if ($answer_value =~ /<XML|<?xml/i) {
+	if ($answer_value =~ /<\??xml/i) {
 		$base_64_encoded_answer_value = encode_base64($answer_value);
 		$decoded_answer_value         = $answer_value;
 	} else {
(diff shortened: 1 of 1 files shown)


drdrew42 (approved):

drdrew42:
Thanks! 🙏

pstaabp:
I'll take @drdrew42 approval as enough, unless you want more testing.  If more testing is requested, what pg problem does this fix the problem for?

drgrice1:
Here is the problem that was causing the issue:
[AP_Lab1_problem4.pg.txt](https://github.com/openwebwork/pg/files/10122628/AP_Lab1_problem4.pg.txt)

Without this pull request the state is double encoded when answers are submitted.

pstaabp:
With that problem, upon submission the values in the geogebra table stay. Values were being cleared out (on the develop branch), so I think this may have been the issue and may be fixed.

However, any answers in the geogebra table are not being submitted as WeBWorK answers.  Should they be?

drgrice1:
They should not be submitted as webwork answers.  That is not implemented in the problem.

pstaabp (approved):

Issue #743, opened 2022-10-23
Should `A` stand for angstrom or ampere?


Alex-Jordan:
In `Units.pm`, the unit `A` is for angstrom, 10e-10 meters. Also `amp` is one of the fundamental units.

I searched the OPL (OPL, Contrib, and Pending folders) and I found no problems at all that use either of the above units. My grep regex in the search was `WithUnits(.*\W\(A\|amp\)\W`, which I think was sufficient but maybe is not. It found a few instances like `NumberWithUnits("$A","hr");` but nothing with the `A` actually being used as a unit.

Of course any such regex is not 100% conclusive, if the unit is hidden in a variable.

Anyway, it seems to me that `A` is much more commonly used as the abbreviation for ampere. (That is how it was being thought of recently when a student entered an answer of `0 T m/A` and discovered a bug.)

Should we change the meaning of `A` to ampere? It feels 1000x more likely to be used by a student to mean ampere than to mean angstrom. If there is ever a WW exercise that needs angstrom, well first of all `A` is not even the right character for that. It should be `Å`. But second of all, that exercise with its context could be coded to override the meaning of `A`.

pkohl:
If it doesn't break a bunch of things and isn't too hard to do, I'd be in favor of the change.  The SI symbol for the ampere is A (it's one of the seven SI base units), and I agree that it is much more likely for end users to mean ampere than angstrom if they put in that symbol.  

That said, it's also not that hard to work around if changing the symbol would break things.

duffee:
I also think `A` should stand for ampere.

The only references to Angstroms in the OPL are help files:
- OpenProblemLibrary/macros/FortLewis/AnswerFormatHelp.pl
- OpenProblemLibrary/Michigan/units.html

The question is whether you want a gradual transition or you like the Big Bang approach.

Either way, the first step is to add the new unit for Angstrom to Units.pm so that the help files in WW and OPL can be updated when the new unit is available in production.

I've just opened PR #753 to add some new units.  The new angstrom could be added to that PR for convenience or separated out in its own PR for faster merging to main.

pstaabp:
We've merged in some changes (listed above) to handle angstroms.

PR #753, opened 2022-12-12
Added new units to Units.pm

duffee:
This PR adds the Becquerel, the measure of radioactivity, and extends the second, metre, electron-volt, Coulomb, ampere and tesla to mostly smaller prefixes for electric and magnetic units.  These are used in the areas of electromagnetism and nuclear physics with a couple for problems on optics and large power projects.  Tests have been added for the new units in **t/units/**.  It also corrects the spelling of **mili** to **milli** (double L).

When writing the test for adding new units with an optional hashref, the tests failed using syntax that I use in working problems.  I've wrapped them in exception handling blocks and marked them with TODOs so that they pass, but are available for examination.

One point that might warrant discussion is the milliampere.  I've added it as `mamp` which is consistent, but unappealing as it's not obvious what it is.  I've also added `mA` which is in line with SI convention.

Units added:
microsecond, nanosecond
picometre, femtometre
tera electron-volt
Megawatt, milliwatt
milliCoulomb, microCoulomb, nanoCoulomb
milli-ampere
milli-tesla
Becquerel

### DIFF
--- a/t/units/conversions.t
+++ b/t/units/conversions.t
@@ -119,7 +119,7 @@ subtest 'Check astronomical units' => sub {
 		'parsec conversion';
 };
 
-subtest 'Recently added electrical units' => sub {
+subtest 'Additional electrical units' => sub {
 	is multiply_by(1000, evaluate_units('eV')), { evaluate_units('keV') }, 'kilo-electron volt conversion';
 	is multiply_by(1E6,  evaluate_units('eV')), { evaluate_units('MeV') }, 'mega-electron volt conversion';
 	is multiply_by(1E9,  evaluate_units('eV')), { evaluate_units('GeV') }, 'giga-electron volt conversion';
--- a/t/units/conversions.t
+++ b/t/units/conversions.t
@@ -1,0 +1,143 @@
+#!/usr/bin/perl -w
+
+use Test2::V0 '!E', { E => 'EXISTS' };
+
+die "PG_ROOT not found in environment.\n" unless $ENV{PG_ROOT};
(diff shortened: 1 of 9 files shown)


t/units/conversions.t
```diff
@@ -0,0 +1,143 @@
...
+
+	is multiply_by(299792458, evaluate_units('m/s')), { evaluate_units('c') }, 'speed of light conversion';
+	is { evaluate_units('c*yr') }, { evaluate_units('light-year') }, 'light year';
+
+	my $todo = todo 'use within() to fudge factor in 9th decimal place';
+	is multiply_by(cos($second_arc) / sin($second_arc), evaluate_units('AU')), { evaluate_units('parsec') },
+		'parsec conversion';
+};
+
+subtest 'Recently added electrical units' => sub {
```
pstaabp, line 122: Using "Recently" will seem strange when running this test in a couple of years.  Perhaps "Additional electrical units"

pstaabp:
Just to be clear, this doesn't changes the ampere unit to 'A'. correct?  

I think you mentioned in #743 that this PR could be updated to include it.

duffee:
Correct - this PR doesn't change the ampere unit to `A`.  I think that an alternative or alias for Angstrom needs to be added first, so that people have a chance to migrate to the new unit.

I have added the milliamp as both `mamp` and `mA` in anticipation that the ampere will eventually change from `amp` to `A`.

pstaabp (approved):

drgrice1:
@pstaabp:  I thought you said that all of these tests were working.  I guess I should have tried it myself.  They aren't working.

PR #755, opened 2022-12-22
Add the parserRadioMultiAnswer.pl macro.

drgrice1:
This macro in essence combindes the parserRadioButtons.pl and parserMultiAnswer.pl macros.  It basically allows the creation of answers that are radio button answers where each radio button has child answers.

The point of this type of answer is for problems where different types of answers are expected in different cases.  For example, a problem that asks a student to solve a system of linear equations.  The possible answers are of course a single solution, an infinite number of solutions, or no solutions.  So using this macro you could create an answer that has three radio buttons.  If the solution is a single solution, then the single solution radio part would have answer boxes in it for the student to enter the solution.  If there are an infinite number of solutions, then the radio button for that part would have answer boxes to enter the parameterization for the solution set.  If there are no solutions, then that part would not have answer boxes (no need).  So the student would see something like:

* A. The unique solution is x = ___, y = ___.
* B. There are an infinite number of solutions parameterized by x = ___, y = ___.
* C. There are no solutions.

A zip file with a few example problems is attached.
[examples.zip](https://github.com/openwebwork/pg/files/10290700/examples.zip)

### DIFF
--- a/htdocs/js/apps/RadioMultiAnswer/RadioMultiAnswer.js
+++ b/htdocs/js/apps/RadioMultiAnswer/RadioMultiAnswer.js
@@ -1,0 +1,171 @@
+// ################################################################################
+// # WeBWorK Online Homework Delivery System
+// # Copyright &copy; 2000-2022 The WeBWorK Project, https://github.com/openwebwork
+// #
+// # This program is free software; you can redistribute it and/or modify it under
+// # the terms of either: (a) the GNU General Public License as published by the
+// # Free Software Foundation; either version 2, or (at your option) any later
+// # version, or (b) the "Artistic License" which comes with this package.
+// #
+// # This program is distributed in the hope that it will be useful, but WITHOUT
+// # ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// # FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+// # Artistic License for more details.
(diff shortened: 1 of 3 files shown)


pstaabp:
Overall, this looks great.  Here's something that I ran into though.  I've attached a problem that uses a checker -- it's an eigenvalue problem and wanted to check for any vector parallel to the right answer.  

[real-eigens-2by2.pg.txt](https://github.com/openwebwork/pg/files/10298779/real-eigens-2by2.pg.txt)

If the correct answer or scales of the eigenvectors are put in, the answer is correct, however, if the second option is chosen (single eigenvalue with 2 eigenvectors), I get the error: `The evaluated answer is not an answer hash : ||.`

drgrice1:
@pstaabp:  The error is caused by line 81 in your problem.  You have `if ($radio_cor ne $radio_stu) { return; }`.  That should be `if ($radio_cor ne $radio_stu) { return 0; }`.  The checker should return a numeric value from 0 to 1.

pstaabp:
That fixed it.

pstaabp (approved):

drgrice1:
I should also point out that $radio_stu and $radio_cor are actually numeric values.  So they should be compared with `==` or `!=` instead of `eq` or `ne`.

drgrice1:
It seems the documentation in the macro is incorrect on this.  I will fix that.

PR #757, opened 2022-12-24
move PGsort from Translator.pm to PGbasicmacros.pl

pstaabp:
Writing tests for PG using num_sort or lex_sort fails because these depend on PGsort originally defined in Translator.pm.

This moves PGsort into PGbasicmacros.pl to make testing simpler.


drgrice1 (approved):
Looks good.  PGsort still works with the move.

PR #758, opened 2022-12-30
PreTeXt output for an array answer

Alex-Jordan:
Prior to this commit, an array_ans is defaulting to HTML output when the display mode is PTX. This gives a simple PTX `fillin` element instead.

### DIFF
--- a/lib/Value/AnswerChecker.pm
+++ b/lib/Value/AnswerChecker.pm
@@ -507,6 +507,7 @@ sub format_matrix {
 	my $displayMode = $self->getPG('$displayMode');
 	$array = [$array] unless ref($array->[0]) eq 'ARRAY';
 	return $self->format_matrix_tex($array, @_) if ($displayMode eq 'TeX');
+	return $self->format_matrix_PTX($array, @_) if ($displayMode eq 'PTX');
 	return $self->format_matrix_HTML($array, @_);
 }
 
@@ -532,6 +533,17 @@ sub format_matrix_tex {
 	$tex .= join('\cr' . "\n", @rows);
 	$tex .= '\end{array}\right' . $close . '\)';
 	return $tex;
+}
+
+sub format_matrix_PTX {
+	my $self  = shift;
+	my $array = shift;
+	my ($rows, $cols) = (scalar(@{$array}), scalar(@{ $array->[0] }));
+	my $ptx = '<fillin';
+	$ptx .= qq( rows="$rows") if $rows > 1;
+	$ptx .= qq( cols="$cols") if $cols > 1;
+	$ptx .= '/>';
+	return $ptx;
 }
 
 sub format_matrix_HTML {


Alex-Jordan:
This fails the perltidy check, but I did tun perltidy before committing. And I run it now and get no diff. 

```
cass317m4138:Value alex.jordan$ ~/pg/bin/run-perltidy AnswerChecker.pm 
/Users/alex.jordan/pg/bin/run-perltidy: line 6: shopt: globstar: invalid shell option name
cass317m4138:Value alex.jordan$ git diff AnswerChecker.pm
cass317m4138:Value alex.jordan$ 
```

Do you see what I'm doing wrong? (I've always gotten that "invalid shell option name" message before too.)

drgrice1:
The `globstar` shell option is most likely the problem.  It seems that option is not supported by the version of `bash` on OSX (at least from a quick google search on this).  You may need to install `bash` via brew.  See https://apple.stackexchange.com/questions/291287/globstar-invalid-shell-option-name-on-macos-even-with-bash-4-x.  That is a five year old post, so it may not be relevant though.

In any case, the `globstar` shell option is what makes the script recurse into subdirectories.  So when you run the script and that option is not working it is not recursing into the `lib/Value` directory.  When I run the script on this pull request it is showing changes to `lib/Value/AnswerChecker.pm`.  You could try running perltidy directly on that file with `perltidy --pro=.perltidyrc -b -bext='/' lib/Value/AnswerChecker.pm`.  That doesn't need any of the shell options set in the script.

Note that the `bin/run-perltidy` script does not take any arguments (like filenames) at this point.  So running `~/pg/bin/run-perltidy AnswerChecker.pm` won't do anything particular for the AnswerChecker.pm file.  It just tries to run perltidy on everything via the `**` glob which recurses into directories if the `globstar` shell option is set.  It would be a nice to update the script to accept file names though.  It would be convenient to be able to run perltidy on a single file (or glob of files) without having to type in the above perltidy command with all of the options.

Alex-Jordan:
That command you provided worked, and I force pushed. Have not yet tried installing bash via brew, but I'll keep that in mind for next time.

One thing I still don't understand is how my recent pull request #752 managed to use `bini/run-perltidy` successfully under the same conditions.

drgrice1:
That pull request only modifies files at depth one in the directory structure.  The bin/run-perltidy script runs the command `perltidy --pro=$INPUTDIR/.perltidyrc -b -bext='/' $INPUTDIR/**/*.p[lm] $INPUTDIR/**/*.t`.

When the globstar shell option is in effect the `**` part of `$INPUTDIR/**/*.p[lm]` and `$INPUTDIR/**/*.t` makes those match all `.pl`, `.pm`, and `.t` files at all depths inside `$INPUTDIR` (which is the pg root directory).  So it will match `macros/*.pl` and `lib/Value/AnswerChecker.pm`.

If globstar is not in effect then the `**` part will only match the first directory level.  So it will still match `macros/*.pl` and get all of the files in your other pull request, but it won't match `lib/Value/AnswerChecker.pm` for this pull request.

pstaabp:
> That command you provided worked, and I force pushed. Have not yet tried installing bash via brew, but I'll keep that in mind for next time.

I've installed bash using brew and these scripts now work.  MacOS has a very old version of bash installed, but they have switched to zsh for the built-in shell, which is fine, but not helpful when needing to run bash scripts.

drgrice1 (approved):
Looks good.  I haven't tested the PreTeXt output with this, but you know what is needed there.  I assume it works for that.

Issue #759, opened 2023-01-01
PGML single quote behavior


Alex-Jordan:
In chasing an issue, I've built the following MWE:

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "PGML.pl",
);

TEXT(PGML(q!It's here.!));

BEGIN_TEXT
\{PGML(q!It's here too.!)\}
END_TEXT

BEGIN_PGML
[@ PGML(q!It's not here.!) @]*
END_PGML

ENDDOCUMENT();
```

This should print:
```
It's here.
It's here too.
It's not here.
```

But the last line is not printed. Something about the single quote, in a string argument to `PGML()`, being executed within a `[@...@]` block is not working.

Oddly, if you add a second single quote to the string then you will see the ouptut with two single quotes. For example
```
BEGIN_PGML
[@ PGML(q!It''s not here.!) @]*
END_PGML
```
will let you see `It''s not here.` in output. If you go up to 3 single quotes or any odd number of single quotes, it's gone again. But it faithfully shows the right number of single quotes when it is an even number.

Alex-Jordan:
At first it looked like this is some edge case use from using `PGML::Format2` inside a `[@...@]` block. But I think it is a more significant issue with PGML parsing. Consider:

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "PGML.pl",
);

$content = q!it's a string!;

BEGIN_PGML
[$content]

[@$content@]

[@ q!it's a string! @]
END_PGML

ENDDOCUMENT();
```

I only see "it's a string" come out twice. I'm guessing that at some point, a string constructor like `q!...!` is not accounted for and something in the parsing expects that single quote to have a sibling.

Alex-Jordan:
Maybe it's a problem that the contents of `[@...@]` is parsed at all, instead of just being executed?

dpvc:
You are correct that there is parsing going on within the `[@ ... @]`.  In particular, quotes are expected to be balanced, so that you can do something like `[@ "@]" @]` and not match the quoted `@]`.  I wasn't thinking about the `q!...!` constructs, and the parser is not flexible enough to handle that (the entire PGML block is tokenized before anything is processed, but since the character used for the `q` can change, PGML can't tokenize the start and end delimiters because they aren't known up front, and we don't want to always match `q(.).*?\g{-1}`, as that would also be done *outside* of `[@...@]` where it shouldn't be (it also makes handling the token splitting a bit more complicated).

[...]

PR #760, opened 2023-01-03
Add support for q, qq, and qr string quoting operator within [@...@] PGML blocks (#759)

dpvc:
This PR provides for proper handling of `q!...!`, `qq!...!` and `qr!...!` strings in `[@ ... @]` blocks in PGML.  This fixes the problem reported in #759 by @Alex-Jordan, and his examples can be used as test cases.

The quoted strings are handled by a new `Quoted()` method that modifies the token list (`$self->{split}`) directly by hand (rather than going through the usual token processing loop) since the `q` delimiter charter is not known when the PGML string is initially split into tokens.


Alex-Jordan:
Thanks Davide, this is great and will address the issue that started this (which has to do with PreTeXt's automated translation of PTX to PG.) I was thinking about how those perl `q` operators work, and there are at least five differences with how this would behave.

1. The `.` at line 46 is not supposed to be a word character or white space.
2. Or, if the `q` (`qq`, `qr`) is immediately followed by white space, then the next character is the delimiter, and now it may be a word character.
3. If the delimiter is among `([{` then the closer is supposed to be the corresponding character from `)]}`. (I checked and something like `q[foo[` does not work.)
4. Nesting of regular delimiters is still respected. For example `q{{foo}}` makes "{foo}".
5. The would-be terminating character may be escaped. This is even true with regular quoting like `"fo\"o"`. While this is supposed to be true for perl in general, it seems that in PG, the Translator is already having trouble with something like that example.

1 through 3 seem like they can be addressed with small changes to this PR. I'll give it a try. I thought I'd check with you first to see if anything I wrote here sounds wrong.

4 seems hard. It doesn't impact the PTX use case though, where `q` will be used with some non-delimiter character. I only mention it to see if some magic occurs to you.

5 seems moot if Translator is already not handling that the way base perl would handle it.

dpvc:
Alex, after I made the PR yesterday, I was thinking about it further, and already had realized I needed to avoid white space for the delimiter, but I appreciate the fuller details about the possibilities.  I also wonder about whether there should be some restriction on what comes before the `q` as well, like a word break.

You are right that 1 through 3 should be straight-forward.  4 is not that hard, and I have written delimiter-balancing code many times, so that should not be a problem.  The `Quote()` method just needs to take that into account.  I'll work something up for that.

For 5, that is due to PG's doubling of all backslashes.  The `PGML::Format()` removes the doubles, so I believe that inside `BEGIN_PGML/END_PGML`, you should be able to do things like `[@ "fo\"o" @]`, but the `PGML()` function calls `PGML::Format2()` which first doubles all backslashes like PG does and then calls `PGML::Format()` on the result.  That was designed so that you can use `PGML()` in macro files that aren't processed with PG's backslash doubling.  But if you use `PGML()` in a PG file directly, it will end up getting quadrupled backslashes instead.  So you would want to use `PGML::Format()` directly if you need this within `.pg` files rather than `.pl` files, or use something like `PGML('[@ "fo~~"o" @])`, since the `~~` will ben turned into a single backslash before the PGML command is executed.  I understand the need for the backslash doubling, but it sure does cause problems at times.

So I think the backslash stuff for 5 does need to be handled.  It already is for single and double quotes, but not for `q` or the others.

dpvc:
I've pushed an update that should take care of your five issues.  See if that works better.

pstaabp:
On testing the first MWE of @Alex-Jordan of #759, I'm getting the following errors:
```
Use of uninitialized value $code in concatenation (.) or string at [PG]/lib/WeBWorK/PG/Translator.pm line 1351
from within PGML::Eval called at line 895 of [PG]/macros/core/PGML.pl
from within PGML::Parse::replaceCommand called at line 1327 of [PG]/macros/core/PGML.pl
from within PGML::Format::Command called at line 1192 of [PG]/macros/core/PGML.pl
from within PGML::Format::string called at line 1371 of [PG]/macros/core/PGML.pl
from within PGML::Format::html::Indent called at line 1185 of [PG]/macros/core/PGML.pl
from within PGML::Format::string called at line 1172 of [PG]/macros/core/PGML.pl
from within PGML::Format::format called at line 1816 of [PG]/macros/core/PGML.pl
from within PGML::Format called at line 1826 of [PG]/macros/core/PGML.pl
from within PGML::Format2 called at line 15 of tmpEdit/setGitHub_759/test1.pg.peter.tmp
```

and this is the same error I see on develop.  I doubled checked that I have the new code.  There's a Quoted method that is being called.

drgrice1:
I am seeing the same warnings that @pstaabp saw.  Specifically for
```perl
BEGIN_PGML
[@ PGML(q!It's not here.!) @]*
END_PGML
```
and
```perl
BEGIN_PGML
[@ q!it's a string! @]
END_PGML
```
from @Alex-Jordan's examples.

dpvc:
I've pushed another commit that should resolve the issues @pstaabp and @drgrice1 are seeing.  Sorry about that; they were due to a last minute change, and I also fixed another issue that I found while testing those problems.

drgrice1:
Your latest commit fixed the errors with @Alex-Jordan's examples.  However, this pull request is causing an issue with variable substitution in PGML as in the following example:
```
$a = 5;
BEGIN_PGML
[`[$a]`]
END_PGML
```

drgrice1:
Actually, you just need `[$a]` and not the PGML math mode delimiters.

pstaabp:
@drgrice1 I'm not seeing a problem with the variable substitution.  It looks like this seem to fix the issues above.  Here's the file I just ran:
```
DOCUMENT();
loadMacros("PGstandard.pl","PGML.pl");

$a = 5;

BEGIN_PGML
[$a] [`[$a]`]
END_PGML

ENDDOCUMENT();
```

drgrice1:
So the issue is specific to the letter "q".

drgrice1:
Use
```
$q = 5;

BEGIN_PGML
[$q]
END_PGML
```
to see the issue.

dpvc:
> So the issue is specific to the letter "q".

Yes, I see why.  I was afraid there might be a situation like that, but didn't think of this one.  I will think about how to address it.  It might be enough to disallow `]` and `@` as `q` delimiters, but probably better to not make the delimiter part of the pattern.

dpvc:
OK, I've made changes that should allow handling of `[$q]` and also things like `[@ $q @]` that would have failed as well.

dpvc:
OK, I see the failed unit test and will make another commit, but am out of time right now.

dpvc:
OK, made the changes.

drgrice1:
Your latest changes do fix the `$q` issue!  Thanks.

dpvc:
I've modified the regex used for the q-strings to make it work better in some edge cases (in particular, if the `q` directory follows `[@` with no space).  This should also be more efficient, since it avoids the lookbehind regex.

pstaabp (approved):
This looks good now.

drgrice1 (approved):
This looks pretty good now.  Thanks for fixing the issues, and working on this to begin with.

dpvc:
It took a bit to get it right, but glad that this seems to do the trick.

PR #761, opened 2023-01-04
Macro with random names and pronouns.

pstaabp:
This is a new macro that is based on the randomName functionality in the PCCmacros.pl file.  This extends the idea there to include the handling of pronouns and includes `they` in the possible pronoun list. 

This was originally submitted to the OPL, however with the restructuring of macros, this is being moved here.

[simple_example.pg.txt](https://github.com/openwebwork/pg/files/10345987/simple_example.pg.txt)

Is a simple example that using this macro. There is also a test file included. 

Note: this is based on #757 to handle testing.


pstaabp:
Added an init subroutine.

pstaabp:
> Note that the same thing is wrong with PCCmacros.pl, most of the macros you brought over from the OPL, and unfortunately a few of the existing PG macros.

We realized some cleanup was going to be needed.

Alex-Jordan:
Thanks for this. Could you add methods for possessive adjectives (his, her, their), possessive pronouns (his, hers, theirs) and object pronouns (him, her, them)? Each with capitalized variants.

pstaabp:
> Thanks for this. Could you add methods for possessive adjectives (his, her, their), possessive pronouns (his, hers, theirs) and object pronouns (him, her, them)? Each with capitalized variants.

Good idea.  Will do.

pstaabp:
Updated this with @Alex-Jordan 's suggestions.  Also updated a sample pg problem with this functionality. 

[simple_example.pg.txt](https://github.com/openwebwork/pg/files/10572766/simple_example.pg.txt)

pstaabp:
Just added another method to randomly get a last name.  It would be great if someone could test.

Alex-Jordan:
I used this recently and a need arose that I had to code my way around. It made me wish there were these enhancements:

```
($alice, $bob) = RandomPerson(n => 2);   # guaranteed distinct
```

```
$stooges = {Larry => 'he', Moe => 'he', Curly => 'he'};
$stooge = RandomPerson(names => $stooges);
```

What do you think about adding these?

pstaabp:
@Alex-Jordan  These looks like great additions.  I've written a bunch of problems needing unique names and the first one `n=>2` would simplify my problems.  I'll work on these.

pstaabp:
@Alex-Jordan updated the macro with the ability to generate multiple Person objects with unique names.  I generated your example above with:
```
@stooges = randomPerson(names => [['Larry', 'he'], ['Moe', 'he'], ['Curly', 'he']])
```
instead of the hashref since the keys don't appear in a consistent order.  Also, if you don't want to specify pronouns, you can do 
```
@stooges = randomPerson(names => ['Larry', 'Moe', 'Curly']);
```
which will randomly assign he/she/they to `Person`s.   I also decided to pass in arguments to `Person` as a hash instead of a hashref.  If you call that directly, you'll have to make that change. 

I have a general question.  I used the `NChooseK` method from `List.pm`.  It's written as an object-oriented method, so can't use it standalone.  It seems silly to have inside this macro in case others may want it.  Should I add to `PGbasicMacros.pl` since the random functions are there?

somiaj:
@pstaabp, I would love to see NchooseK as a basic macro. The one I use I grabbed from `ui/PGchoicemacros.pg`. It says it is DEPERICATED, but I really don't see why, it is a very useful macro, and is already stand alone.

Alex-Jordan:
Do you have the same issue with the `$first_names` hash ref that is there by default?

pstaabp:
Here is an updated problem with a bunch of functionality from this macro. 

[pronouns.pg.txt](https://github.com/openwebwork/pg/files/11012042/pronouns.pg.txt)

I should also have pointed out that I have renamed the `pronoun` and `Pronoun` functions to `subj_pronoun` and `Subj_pronoun` to be consistent with the other methods.  If these names are too long, I could shorten to `subj_p` or `subj_pro` for example.

pstaabp:
> Do you have the same issue with the `$first_names` hash ref that is there by default?

Yes. But I just added a sort to the names on the keys to be consistent.  I was wondering if I should code this up with arrays or array refs instead.

pstaabp:
> @pstaabp, I would love to see NchooseK as a basic macro. The one I use I grabbed from ui/PGchoicemacros.pg. It says it is DEPERICATED, but I really don't see why, it is a very useful macro, and is already stand alone.

It's nearly the same method I used.  I think I'll put it in PGbasicmacros.pl.  We may want to rename it to indicate that it selects k random numbers.  As is, I would guess it would be the function n!/(k!(n-k)!)), which it is not.

somiaj:
> It's nearly the same method I used. I think I'll put it in PGbasicmacros.pl. We may want to rename it to indicate that it selects k random numbers. As is, I would guess it would be the function n!/(k!(n-k)!)), which it is not.

Might also be nice to change the parameters it is sent. I use it a lot as `(a,b,c,d)[NchooseK(4,2)]`, which I think is similar to how you are using it. Maybe make it a subset function. `Subset($arrayref, $n)` which returns a new array which is subset of the original array in a random order. Then the old function `NchooseK($n,$k)` is just `Subset([0..$n-1], $k)`.

The one difference I noticed between the `List.pm` and `PGchoicemacros.pl` version is error checking, the `PGchoicemacros.pl` just returns the full list in a random order if `$k > $n`, instead of erroring out. Not sure which is better, but in this way the `Subset($arrayref, $n)` could just return the full array in a random order if `$n` is bigger than the number of elements.

Alex-Jordan:
> I should also have pointed out that I have renamed the pronoun and Pronoun functions to subj_pronoun and Subj_pronoun to be consistent with the other methods. If these names are too long, I could shorten to subj_p or subj_pro for example.

I'll be happy to use whatever you settle on for the method names. For input, if I were making them I would call them like 

```
[@$a->subject@] gave [@$a->possessive@] shirt to [@$b->object@] and now it was [@$b->possession@]
```

Note "subject", "object", and "possession" are nouns, and "possessive" is an adjective.

somiaj:
> Maybe make it a subset function. `Subset($arrayref, $n)` which returns a new array which is subset of the original array in a random order.

Thinking about this a bit more, `random_subset($n, @array)` might be a better name to match the other random function names. Also unsure if an `@array`, and array ref, or both would be the preferred input.

Maybe we should move the discussion of such a random function to a new pull request?

Alex-Jordan:
If you are sorting keys for the default list of random names, why not sort keys for a user defined list too?

I think logically, it's a hash and not an array. I may be too wary of assigning special meaning to particular indices in an array. like this does (index 0 means the name, index 1 means the object pronoun).

But actually I'm not sure it is universally logical to have the hash values be things like "he", "she", "they". Stop me if I'm overthinking, but if it were like:

```
%object_pronouns = {
   male => 'he',
   female => 'she',
   nonbinary => 'they'
}

$first_names = {
   Peter => {gender => 'male'},
   Alex => {},
  ...
```

"Alex" is not necessarily male. The absence of "gender" could mean you grab a random gender from `keys %object_pronouns` and store that in the Person object. Or for Peter you just store that "male" in the Person object. Then get all the pronouns downwind from that gender value using hashes like `%object_pronouns` and other short hashes for other pronouns.

FWIW, this approach could make this more portable to other languages with other pronoun schemes.

Just sharing my thoughts. I'm not requesting you change things. I will happily use this however it works.

somiaj:
@pstaabp Here is my suggested function, <https://github.com/somiaj/pg/tree/random-subset-macro>, I can make a pull request if you and others things this will be a useful addition.

Alex-Jordan:
I kept thinking about this today, and maybe it shouldn't be an array or a hash reference....maybe it should be an array of hash references.

```
@first_names = (
    {name => 'Alex', gender => 'female'},
    {name => 'Alex', gender => 'male'},
    {name => 'Alice', gender => 'female'},
    {name => 'Blake', gender => 'nonbinary'},
    {name => 'Bob', gender => 'male'},
    ...
);
```

Then you could have a name repeated, with a different gender. And as an array, no worries about the hash keys being in random order.

pstaabp:
@somiaj open up a PR and let's discuss this there.  Having two conversation going on here is difficult.

pstaabp:
I have updated again:
- added the `random_subset` method.
- dropped the `pronoun` of the name of methods. 
- updated the sample problem (attached)

[simple_example.pg.txt](https://github.com/openwebwork/pg/files/11068723/simple_example.pg.txt)

I'm on the fence about using the input format similar to this that @Alex-Jordan proposed: 
```
randomNames(names => [
    {name => 'Alex', pronoun => 'she'},
    {name => 'Alex', pronoun => 'he'},
    {name => 'Alice', pronoun => 'she'},
    {name => 'Blake', pronoun => 'they'},
    {name => 'Bob', pronoun => 'he'}]);
```

versus just using an arrayref of array refs: 
```
randomNames(names => [
   ['Alex','she'],
   ['Alex','he'],
   ['Alice','she'],
   ['Blake','they'],
   ['Bob','he']);
```

which is is simpler, but less clear.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,525 @@
...
+		@_
+	);
+
+	# if the names are passed in.
+	return
+		map { Person->new(name => $_->[0] // $_, pronoun => $_->[1] // list_random('he', 'she', 'they')) }
+		@{ $options{names} }
+		if ($options{names});
+
+	if ($options{n} == 1) {
```
Alex-Jordan, line 289: If n is 1 (or assumed to be 1) then it enters this block of the conditional code, and the keys of the hash are not being sorted before a random selection.

I think this will be moot once you settle on how to store the big list of names as an array ref.

pstaabp: Good catch.  I'll rethink a better way to store the names.

pstaabp:
Just updated to now sort the keys of the names upon loading the macro--so it is only done once.  If this seems too much, I can rewrite to make the original names an array instead of a hash.

Alex-Jordan:
My opinion (which you can ignore if you want) is to make it be:

```
[
    {name => 'Alex', pronoun => 'she'},
    {name => 'Alex', pronoun => 'he'},
    {name => 'Alice', pronoun => 'she'},
    {name => 'Blake', pronoun => 'they'},
    {name => 'Bob', pronoun => 'he'}
    ...
]
```

That much is in the file, so no concern about it being less convenient for authors.

For an author-defined list, they could pass the above structure. Or for their convenience, you could allow them to pass:

```
namelist => [
    ['Moe',  'he'],
    ['Larry', 'he'],
    ['Curly', 'he'],
]
```

And code checks the ref on the first entry, if it's ARRAY or HASH. If ARRAY, construct the array of hash refs with something like:

```
map{ {name => $_->[0], pronoun => $_[1]} } ( @{$namelist} )
```

pstaabp:
I think allowing authors either method sounds good to me. I'll update and push soon.

pstaabp:
Updated to include arrayrefs of either an arrayref or hashref for including names for the `randomPerson` method.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,554 @@
+
+=head1 NAME
+
+randomNamesPronouns.pl - Load macros for random names.
+
+=head2 SYNOPSIS
+
+ loadMacros('randomNamesPronouns.pl');
```
Alex-Jordan, line 8: Should this have only one space character of indentation? Not two like elsewhere? I do not know POD style standards enough to know.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,554 @@
...
+returns an array of 5 Person objects with unique names.
+
+=item * C<names =E<gt> arrayref> returns an array of specified Person objects.
+
+If the arrayref is in the form of C<[['name1','pronoun1'],['name2','pronoun2'],...]> then
+an array of Person objects of the names and pronouns specified.
+
+Example:
+
+  my ($a,$b,$c) = randomPerson(names => [['Bart','he'], ['Lisa','she'], ['Matty','they']]);
```
Alex-Jordan, line 272: I'm confused by this. I would expect `randomPerson(names => [['Bart','he'], ['Lisa','she'], ['Matty','they']]);` to return a single random person from among those three names. Not an array of all three, unless I also pass `n => 3`.

Alex-Jordan: And likewise in the POD that follows.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,554 @@
...
+
+=item * C<n =E<gt> k> returns an array of k Person objects with unique names.
+
+Example
+
+  randomPerson(n=>5)
+
+returns an array of 5 Person objects with unique names.
+
+=item * C<names =E<gt> arrayref> returns an array of specified Person objects.
```
Alex-Jordan, line 265: Should this just say something about replacing the default list of names/pronouns with your own list? It does not seem like passing a `names` option should be described as returning an array, if n is 1.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,554 @@
...
+	}
+}
+
+=head3 randomLastName
+
+This returns a random last name based on popular last names in the United States.  Example
+
+  $p = randomLastName();
+
+Note: it is just a string, and doesn't have the pronoun's that the Person object does.
```
Alex-Jordan, line 335:
```suggestion
Note: it is just a string, and doesn't have the pronouns that the Person object does.
```

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,554 @@
...
+and verb conjugation.
+
+=cut
+
+package Person;
+
+sub new {
+	my ($class, %opts) = @_;
+	my @v = grep { $opts{pronoun} eq $_ } qw/he she they/;
+	die "The pronoun must be either he, she or they. You passed in $opts{pronoun}"
```
Alex-Jordan, line 374:
```suggestion
	die "The pronoun must be either 'he', 'she' or 'they'. You passed in '$opts{pronoun}'"
```

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,554 @@
...
+  $p2 = new Person(name => 'Max', pronoun => 'they');
+
+  $p1->verb('find');
+
+returns 'finds'
+
+  $p2->verb('find')
+
+returns 'find'
+
```
Alex-Jordan, line 526: There's a double newline here.

Alex-Jordan:
I think I miscommunicated something earlier. When I proposed something like:

```
$stooge = randomPerson(names => [['Moe', 'he'], ['Larry', 'he'], ['Curly', 'he']]);
```

What I meant was `$stooge` ends up being one of those three author-provided people. Or you could do like:

```
($stooge1, $stooge2) = randomPerson(n => 2, names => [['Moe', 'he'], ['Larry', 'he'], ['Curly', 'he']]);
```

Based on the code and the POD, I think you understood it to means something else where the author definitely uses all three names. But then I'm not sure why they are using the macro instead of just hard coding the names and pronouns where they get used.

Alex-Jordan:
The original list of names in PCCmacros.pl came from unioning the sets of actual first names from several classes one term (my classes and two colleagues' classes). And also throwing in the first names of all full time PCC math faculty at that time. 

Once this is released, we shouldn't mess with the default list of names if we don't have to. So my question is do we really want this to be the list of names that gets shipped? Or do we want to try with more intent to capture representative names? Maybe all those original student names are as good as it gets for an empirical sampling of name diversity. But having the PCC math faculty in there is not so representative. Those names are `Alex Carl Carly Chris Dave Dennis Diane Emiliano Emily Farshad Henry Jeff Jessica Kandace Ken Kim Matthew Michael Michele Peter Phil Rebecca Ronda Ross Scot Shane Stephanie Stephen Tammy Virginia Wendy Will`.

pstaabp:
I could pull a top 100 list of names from the Social Security pages.

dlglin:
> I could pull a top 100 list of names from the Social Security pages.

This could work for American names.  How would this translate to other countries?

Here are some thoughts for future work that shouldn't preclude merging the current version:

- Can we allow individual installations to customize the dictionary of names without editing the macro?  One approach would be to separate the dictionary into a separate file, and then have a config option pointing to which dictionary to use.  That way people could create a list of names applicable to them.  We could even pack in multiple dictionaries if it makes sense.
- Is it even possible to localize the actual code?  Do different languages handle pronouns similarly enough that this would make sense?  Could this be modified so that someone using a different language could add their own attributes to names if the current attributes are insufficient?

Alex-Jordan:
> I could pull a top 100 list of names from the Social Security pages.

I'm afraid that will end up lacking diversity. Like, even more out of proportion than actual demographics. You'll end up with like 95 of those names having Anglo Saxon heritage, even with far less than 95% of the population having Anglo Saxon heritage.

pstaabp:
Made the recommended changes.

macros/misc/randomNamesPronouns.pl
drgrice1: I personally prefer four space indentation for code blocks in POD.  To me one or two spaces does not give enough offset.

macros/misc/randomNamesPronouns.pl
drgrice1: I agree with @Alex-Jordan on this.  This should only return 1 unless 3 are specifically asked for.

Also remove `my` from these in the POD.  Remember that this documentation is for PG problem authors.  If `my` is used in a PG problem it will most likely not work (depending on the situation).  I cut an pasted this code into a problem and then used it in a PGML section, and got warnings about `$a` being undefined.

drgrice1 (commented):
There are a few other things I see.

Also, please address the issues that @Alex-Jordan pointed out.

pstaabp:
> I'm afraid that will end up lacking diversity. Like, even more out of proportion than actual demographics. You'll end up with like 95 of those names having Anglo Saxon heritage, even with far less than 95% of the population having Anglo Saxon heritage.

Thoughts on where to pull from?   Looking at the top 1000-list at the SSA, the bottom third looks much more diverse to me. (Whatever that means).  That's a possibility.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,557 @@
...
+sub new {
+	my ($class, %opts) = @_;
+	my @v = grep { $opts{pronoun} eq $_ } qw/he she they/;
+	die "The pronoun must be either he, she or they. You passed in $opts{pronoun}"
+		if scalar(@v) != 1;
+	die 'The field name must be passed in' unless defined($opts{name});
+	my $self = {
+		_name    => $opts{name},
+		_pronoun => $opts{pronoun}
```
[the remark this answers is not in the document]
pstaabp: This was just a coding pattern that I've used in the past to do a form of privateness.  I can update.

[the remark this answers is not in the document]
drgrice1: I know.  I have seen you use it before.  However, the last time that I recall seeing it was in webwork3 code, and it was javascript.  This is a somewhat common (but outdated) convention in javascript.  It it not a perl coding convention though.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,557 @@
...
+
+=back
+
+=cut
+
+sub randomPerson {
+	my %options = (
+		n => 1,
+		@_
+	);
```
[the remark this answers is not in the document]
pstaabp: Fixed.

pstaabp:
@dlglin I missed your comment above about separate file.  I had tried to store the names in another file originally, but I believe because of the safe, this isn't allowed in a macro.  I think to get around this is to make an actually perl module to load the file and then add the Module to the list of allowed modules would do this.  I thought this was overkill for this macro, but could do this.

pstaabp:
@Alex-Jordan I missed what you listed above about using `randomPerson` to actually pull a single person randomly.  This makes more sense.

macros/misc/randomNamesPronouns.pl
pstaabp: I misunderstood what you meant earlier.  This makes more sense and have updated this.

macros/misc/randomNamesPronouns.pl
pstaabp: I'm definitely in the 2 space camp, but can update to 4 spaces if we want to standardize.

macros/misc/randomNamesPronouns.pl
pstaabp: fixed.

macros/misc/randomNamesPronouns.pl
pstaabp: fixed.

macros/misc/randomNamesPronouns.pl
pstaabp: fixed.

pstaabp:
I think I made all of the changes except for the list of names.

pstaabp:
Updated the list of names. In the  POD, I described where they originated.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,617 @@
...
+sub randomPerson {
+	my %options = (n => 1, @_);
+
+	# if the names are passed in.
+	if ($options{names}) {
+		my @persons;
+		for my $p (@{ $options{names} }) {
+			if (ref $p ne 'HASH') {
+				push(@persons,
+					Person->new(name => $p->[0] // $p, pronoun => $p->[1] // list_random('he', 'she', 'they')));
```
drgrice1, line 376: `$p->[0] // $p` for the name is not good.  The only reason this works is because macros do not enforce `strict refs`.  If `use strict` were properly in effect this would be an error in the case that `$p` is not an array reference.  My suggested `randomPerson` method in the other comment does this correctly.

drgrice1 (commented):
More changes are needed, and some restructuring to simplify things.

pstaabp:
merged @Alex-Jordan PR and thanks for updating with @drgrice1 comments--learning about the `wantarray`functionality.  I knew I could cleanup `randomPerson` even more and that helped as well as the refactor to an array.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,648 @@
...
+
+    @p = randomPerson(n => 2, names => [[ Bart => 'he' ], [ Lisa => 'she' ], [ Matty => 'they' ]]);
+
+=back
+
+=cut
+
+sub randomPerson {
+	my %options = (n => 1, @_);
+	my @persons = ();
```
drgrice1, line 361: Please drop the `= ()` part.  In Perl hashes and arrays are always created initialized to an empty hash or array.  Adding `= ()` does nothing except to add code clutter.

drgrice1 (approved):
This is looking pretty good now.  One minor quibble in the code, but other than that I think this pull request is ready to go now.

Alex-Jordan:
I had two more thoughts. You could add this block t the verb conjugation logic to handle most verbs that end in `y`:

```
} elsif (substr($plur, -1) eq 'y') {
		$sing = $plur;
		$sing =~ s/y$/ies/;
		return ($self->{pronoun} eq 'they' ? $plur : $sing);
```

And then I realized there is one verb where 3rd person past tense depends on singular/plural, and you could add:

```
sub werewas {
        return shift->verb('were', 'was');
}
```

macros/misc/randomNamesPronouns.pl
pstaabp: yep.

pstaabp:
I wonder about adding the 'y' ending.  There's verbs like say and pay that conjugate normally that would have to be a special case.  Again, we could add another case, but where do we stop.

pstaabp:
I changed my mind and added the exception for -'ay' words.   Also, made the changes directly and missed your PR @Alex-Jordan.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,662 @@
...
+
+=cut
+
+sub verb {
+	my ($self, $plur, $sing) = @_;
+	if (defined($sing)) {
+		return ($self->{pronoun} eq 'they' ? $plur : $sing);
+	} elsif (substr($plur, -1) eq 's' || substr($plur, -2) eq 'ch' || substr($plur, -2) eq 'sh') {
+		return ($self->{pronoun} eq 'they' ? $plur : $plur . 'es');
+	} elsif (substr($plur, -2) ne 'ay' && substr($plur, -1) eq 'y') {
```
Alex-Jordan, line 603: This handles `pay/pays`, but what about `key/keys`, `deploy/deploys`, and `buy/buys`? The exception to the `y->ies` rule should look for `[aeou]y`.

pstaabp: Added code to handle these as well as well as test cases.

pstaabp:
Updated the code again and tested also that adding names to `@namespronouns` in `PGcourse.pl` overrides the default list in the macro--this can be used for updating the names list locally.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,662 @@
...
+    $p2 = new Person(name => 'Max', pronoun 'they');
+    $p2->verb('fly', 'flies');
+
+returns 'fly'
+
+=cut
+
+sub verb {
+	my ($self, $plur, $sing) = @_;
+	if (defined($sing)) {
```
drgrice1, line 599: Here is an idea that you could consider.  If you change this to
```suggestion
	if ($self->can($plur)) {
		return $self->$plur;
	} elsif (defined($sing)) {
```
and then perhaps change the methods below to be named by the third person plural form, for example rename `dodoes` to just `do`, then you could call `$person->verb('do')`.

Alex-Jordan:
There is a problem (?) with reserved words like "is" and "are". So I didn't
name those methods that way. But maybe you are suggesting using verb('do')
instead of a special method?

drgrice1:
I was suggesting that the `areis` method be renamed to `are`, and then you could either call `$person->verb('are')` or `$person->are`.  Both would give the same thing.  The first would really just call the second.  Or if you want the `areis` method name, then you could just alias it with `are` and have it both ways.

Alex-Jordan:
I assumed that there would be an issue with naming a subroutine `is`, `are`, or `do` because these showed up in a list of perl keywords I found online. Is that not the case? That is the only reason I named it `areis` instead of one or the other.

I just tried each one and didn't see an immediate problem with `is` or `are`, but `do` led to a perl error `ERRORS from evaluating PG file: 'do "file"' trapped by operation mask`.

Alex-Jordan:
I didn't try `do` as a method, just as a subroutine in a .pg problem. It might behave differently.

drgrice1:
I tested changing all of the special methods to the third person plural form including "dodoes" to "do", and it worked with no warnings.  Although perlcritic does give a warning about the "do" subroutine being a homonym for a builtin keyword.  That is not the only builtin keyword that PG has a homonym for though ("cmp" is one for example).  There is a difference in how an object method is called and how a direct subroutine is called.  For an object method as this would be it would be called as `$person->do` which works, and a direct subroutine as you tested would be called as just `do()` which would be interpreted not as the subroutine call, but as the keyword `do` and that is trapped in the safe zone.

macros/misc/randomNamesPronouns.pl
pstaabp: This is a good idea.

pstaabp:
I'll update this with the single verb method names `->do` `->are`, ...  Perhaps  I also just thought of providing a capitalized version of the verb as well.

pstaabp:
One more update with this discussion.  @drgrice1 I had to add the code suggested above for the `$self->can($plur)` in a different order to avoid infinite recursion.

drgrice1:
Yeah, I saw that.  I realized the problem after I posted the code.

drgrice1:
There is one another problem with my idea.  That is with the other "Person" package methods that might also be verbs.  I.e., 'object', 'subject', and 'name'.  For example, consider the sentence `[$person->Subject] [$person->verb('object')] to being called a frog.`  With the current code this gives "He him to being called a frog" instead of "He objects to being called a frog."  So to continue with this, there probably will need to be a special case in the `verb` method for those.

pstaabp:
> For example, consider the sentence [$person->Subject] [$person->verb('object')] to being called a frog.

Great example sentence.  I'll put in special cases for these.

pstaabp:
@drgrice1 I'm thinking of switching line 611 to 
```
} elsif ($self->can($plur) && !(grep {$_ eq lc($plur) } qw/object subject name/)) {
```
and tests fine, but wondering if this is the right way to go with this.  I thought just using regexp as well.

drgrice1:
That looks good.  Using a regexp would work also.  I don't think it matters to much which you go with.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,729 @@
...
+
+    loadMacros('randomNamesPronouns.pl');
+
+=head2 DESCRIPTION
+
+C<randomNamesPronouns.pl> provides a randomName function that generates a random
+name with subject, possessive, adjectives and object pronouns
+with and without capitilization and verb conjugation.
+
+Note: This idea and the names were taken from the C<PCCmacros.pl> C<RandomName>
```
Alex-Jordan, line 16: Now that the names don't come from the PCC macro, we can leave this out.

```suggestion
This idea was taken from the C<PCCmacros.pl> C<RandomName>
```

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,729 @@
...
+subroutine to extend to the handling of pronouns. The first names were generated by randomly
```
Alex-Jordan, line 17:
```suggestion
subroutine to extend to the handling of pronouns.
```

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,729 @@
...
+selecting 200 names from the United States Social Security Administration's top
```
Alex-Jordan, line 18:
```suggestion
The first names (aka given names) were generated by randomly selecting
200 names from the United States Social Security Administration's top
```

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,729 @@
...
+1000 list (actually 1000 per gender) from 2005.
+The additional names associated with they pronouns were taken from a L<wikipedia list
+of public trans people|https://en.wikipedia.org/wiki/List_of_transgender_people>
+who use they pronouns. The last names were generated from a list of top 100 names
```
Alex-Jordan, line 22:
```suggestion
who use they pronouns. The last names (aka family names) were generated from
```

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,729 @@
...
+in the United States in 2020.
```
Alex-Jordan, line 23:
```suggestion
 a list of top 100 names in the United States in 2020.
```

macros/misc/randomNamesPronouns.pl
Alex-Jordan: i repeatedly found myself reading "first names" to mean the first several names on the list below. Here are some edits to make it more clear what is meant.

macros/misc/randomNamesPronouns.pl
```diff
@@ -0,0 +1,729 @@
...
+and then call the randomPerson subroutine
+
+    $p1 = randomPerson();
+
+The variable C<$p1> is now a C<Person> object with methods to access the names, pronouns
+and verb conjugation.  It is can be used within a problem as
+
+    BEGIN_PGML
+    [$p1->name] travels 1.5 miles to school.
+    After school, [$p1->subject] [$p1->gogoes] to work.
```
Alex-Jordan, line 40:
```suggestion
    After school, [$p1->subject] [$p1->go] to work.
```

Alex-Jordan:
I opened a PR to @pstaabp 's branch that edits POD and regenerates the surname pool. ("Jordan" ended up in there, but I it was random.)

Not to drag this out, but I just have a few more thoughts.
* Do we really need `Verb('say')` when an author can use `verb('Say')`? if `say` were stored as a variable, OK. But it seems like an author will always be typing the verb they want.
* One thing that is bugging me is having to write `[$p1->name] travels 1.5 miles to school` instead of just `[$p1] travels 1.5 miles to school`. My perl/PG/WeBWorK is not thorough enough to understand how I would get `[$p1]` to print as a string instead of a hash reference. But I guess it can be done, since MathObjects do this.

somiaj:
> * One thing that is bugging me is having to write `[$p1->name] travels 1.5 miles to school` instead of just `[$p1] travels 1.5 miles to school`. My perl/PG/WeBWorK is not thorough enough to understand how I would get `[$p1]` to print as a string instead of a hash reference. But I guess it can be done, since MathObjects do this.

Would adding `->string` and `->TeX` methods like math objects work here, which is what PGML will call (though unsure if it also checks if you are a mathobject first and if there is a way to trick that into working for this macro).

Alex-Jordan:
On a Hail Mary I did add a `string` method, but it didn't work. Somewhere something recognizes actual Math Objects before it tries to use a string method on them. I haven't worked out where that happens. I spent a moment trying to make PGML recognize `Person` objects, but I stopped, realizing that if it can be done, it should work more broadly than just with PGML.

somiaj:
Looking at the code it seems that MathObject's know to return `->string` method if they are in a scalar context, so it should be possible to get this object to also do the same thing (or use `->name` method in the right context). It doesn't appear that the `->string` method is explicitly called outside of math mode (look at the `recplaceVariable` sub in `PGML.pl`). I don't know enough about perl to make that happen though.

drgrice1:
The `Person` package is not a MathObject as it does not derive from the `Value` package.  So the `->string` method won't work for it as it does for a MathObject.

Alex-Jordan:
We get that, but are wondering how to make a `Person` object be handled like a MathObject.

I guess we could make "Person" be a new kind of math object, maybe subclassed from String but with all the methods we have here. 

```
[$A] has 1 lb of gold in [$A->possessive] hands.  
[$B] has 2 lb of feathers in [$B->possessive] hands.  
Who is carrying more weight? [_]{$B}{20}
```

somiaj:
@Alex-Jordan It appears that MathObjects overload `""` with the `->stringify` method. It is possible to also overload the stringify output for a Person object, but this cannot be done inside the macro since overload is masked in the safe compartment. One way around this is to create a base `StringifyOverload` class that is loaded outside the safe compartment, then `Person` (and possibly other classes) can use this as a base class to override string output via a stringify method.

I came up with the following and was able to get the desired effect of `$person` producing the name in a string context. To use this, put the following in the file `pg/lib/StringifyOverload.pm` and configure `webwork2/conf/defaults.conf` to load it. Then use this as the base class of Person.

```perl
package StringifyOverload;
use overload
	fallback => 1,
	'""' => 'stringify';

sub stringify { return shift; }

1;
```

I had to add `fallback => 1` to get this to work since PGML checks if a string is empty, `eq ''`, and fallback is needed to use the overloaded stringify output in string comparisons. This also has the effect a `$person` can be compared to a string using the persons name.

pstaabp:
Wasn't too hard to make a MathObject of type `Value::String`.  Now we can write `[$p]` instead of `[$p->name]`.

pstaabp:
A few other changes after discussion:
1. fixed the error that @Alex-Jordan found for `randomPerson(n=>1, names =>[['Bart','he']]);`
2. made the pronouns be more flexible--possibly add another pronoun.  (Note: this works great for getting the pronoun forms, but noticed that the verb method is still hard coded for these three.
3. Remove the `Verb` method, since capitalization can be handled by capitalizing the input like `->verb('Fly')` 
4. changed the macro name to `randomPerson.pl`.  (Note: I did this in a separate commit, so it's easier to see the above changes.  Also, documentation updated to new macro name)

Alex-Jordan:
This PR may be a record for longest comment thread :)

I just opened a PR (https://github.com/pstaabp/pg/pull/15) to Peter's branch that adds syntax sugar.

pstaabp:
Thanks @Alex-Jordan.  I think that is a nice addition--much simpler way to write problems. and now will update my sample problems to reflect this.

Alex-Jordan:
Maybe this is not worth changing, but I find it awkward to have this:
```
my $PRONOUNS = {
	subjects   => [ 'he',  'she',  'they' ],
	possessive => [ 'his', 'her',  'their' ],
	possession => [ 'his', 'hers', 'theirs' ],
	object     => [ 'him', 'her',  'them' ]
};
```

and then later, to correlate `'she'` with `'her'` etc by using those grep code lines. Since everything is predicated on the subject pronoun, what do you think about:

```
my %PRONOUNS = (
        he => { possessive => 'his', possession => 'his', object => 'him },
        she => { possessive => 'her', possession => 'hers', object => 'her },
        they => { possessive => 'their', possession => 'theirs', object => 'them },
);
```

I think this would make things more clear later in the code. Furthermore, this is more easily extensible. If you want to add `it =>  { possessive => 'its', possession => 'its', object => 'it }`, you can just assign `$PRONOUNS{'it'}`. But the current way would require inserting an entry into each of the four array references.

Alex-Jordan:
There is now an open PR to @pstaabp's branch that makes it easy to add an additional pronoun family (it, its, its, it;  xe, xer, xers, xim; etc). FWIW I'd put a fork in this once that is resolved.

pstaabp:
Merged in.  I think this is ready to go.

drgrice1 (approved):
Looks good.  Good work @pstaabp and @Alex-Jordan.

drgrice1:
I think we can count 3 approvals here with the @pstaabp, @Alex-Jordan, and myself.  I will merge it.

christianp:
I found this issue because your abstract for EAMS caught my attention. Can I interest you in [this dataset of US name frequency data](https://github.com/numbas/numbas-extension-random-person/blob/master/datasets/us.json) that I use in the Numbas "random person" extension? It's derived from the data published by US Social Security: https://www.ssa.gov/oact/babynames/limits.html

pstaabp:
We pulled the names from the U.S. Social Security site for 2005 (with expectation that those born then are now college-age).  I like that you have other name databases--UK and France it looks like.

PR #764, opened 2023-01-18
Fix student answer strings in hardcopy.

drgrice1:
Now that the preview_latex_string is used for hardcopy, the character used in the TeX verb for strings causes a problem (it was chr(128)). This switches to using the same character that used for the different display modes as is used in lib/String.pm.

You can use Library/MontanaState/FL/2.2B83Nested2.pg to test this.  Submit an answer, and generate a hardcopy with student answers.

I have had this sitting here for a while, but this just came up on the forums (see https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8197).


pstaabp (approved):
This fixes the issue noted.

drdrew42 (approved):

PR #768, opened 2023-01-23
Macro for nondecimal bases

pstaabp:
This provides a macro from converting decimals to and from non-decimal bases.   Although not necessary, the maximum base is 16 (hexadecimal).  Also, a colleague of mine uses base-12 with T and E as the digits ten and eleven in that base, so there is an optional array of digits to use.  

A test file is provided and to run it enter (from PG_ROOT): `prove -v t/macros_math/nondecimal_base.t` 

Also, a sample PG is provided: 
[nondecimal.pg.txt](https://github.com/openwebwork/pg/files/10484004/nondecimal.pg.txt)

There are plenty of examples in the test file and the sample PG file.


dpvc:
Although the code works, it is not as efficient as it could be.  In particular, the conversion to base 10 can be improved in several ways.  Your current loop uses two multiplications, one addition, and two assignments for each digit processed, but it can be done with one multiplication, one addition, and one assignment (see below).  But the real inefficiency is in using `grep` to find the base-10 value for a given digit.  Note that this will compare the given digit to *every* digit, even after it finds the correct one, and so that is a lot of looping that gets done for each digit in the original value.  One option would be to join the digits into a single string and use `index()` to find the position (and so the base-10 value) of a given digit, but there is a better way, which is to use a lookup hash that maps the digits back to their base-10 values.  I give an example in the code below.

To eliminate one multiplication and one assignment from the loop (and also the building of a new list from the `reverse()` call), you can evaluate the original number as a polynomial using Horner's method.  Thinking of *d*₁*d*₂...*d*ₙ as an *n*-digit base *b* number, we can write the base-10 equivalent as *d*₁*b*ⁿ⁻¹ + *d*₂*b*ⁿ⁻² + ... + *d*ₙ₋₁*b* + *d*ₙ, which can be rewritten (by factoring out successive *b*s) as (...((*d*₁*b* + *d*₂)*b* + *d*₃)*b* + ... + *d*ₙ₋₁)*b* + *d*ₙ.

That means you can compute the value by taking the first digit, multiplying by *b* and adding the next digit, then multiply that sum by *b* and add the next digit, and so on.  That is one multiply, one addition, and one assignment per digit (rather than two multiplies and assignments), and doesn't require you to reverse the digit list.

The following example code implements this approach, using the digit lookup table rather than the `grep` loop.

``` perl
sub convertBase {Base::convert(@_)}

package Base;

#
#  The standard digits, pre-built so it doesn't have to be done each time the conversion is called
#
our $digits16 = [ '0' .. '9', 'A' .. 'F' ];
our $digit16 = { map { ($digits16->[$_], $_) } (0 .. scalar(@$digits16) - 1) };

sub convert {
  my $value = shift;

  #
  # Get the conversion options
  #
  my %options = (
    from => 10,
    to => 10,
    digits => $digits16,
    @_
  );
  my $from = $options{'from'};
  my $to = $options{'to'};
  my $digits = $options{'digits'};

  die "The digits option must be an array of characters to use for the digits"
    unless ref($digits) eq 'ARRAY';

  #
  # The highest base the digits will support
  #
  my $maxBase = scalar(@$digits);

  die "The base of conversion must be between 2 and $maxBase"
    unless $to >= 2 && $to <= $maxBase && $from >= 2 && $from <= $maxBase;

  #
  # Reverse map the digits to base 10 values
  #
  my $digit = ( $digits == $digits16 ? $digit16 :
                { map { ($digits->[$_], $_) } (0 .. $maxBase - 1) } );

  #
  #  Convert to base 10
  #
  my $base10;
  if ($from == 10) {
    die "The number to convert must consist only of digits: 0,1,2,3,4,5,6,7,8,9" 
      unless $value =~ m/^\d+$/;
    $base10 = $value;
  } else {
    foreach my $d (split(//, $value)) {
      die "The number to convert must consist only of digits: " . join(',', @$digits[0 .. $from - 1])
        unless defined($digit->{$d});
      $base10 = $base10 * $from + $digit->{$d};
    }
  }
  return $base10 if $to == 10;

  #
  #  Convert to desired base
  #
  my @base = ();
  do {
    my $d = $base10 % $to;
    $base10 = ($base10 - $d) / $to;
    unshift(@base, $digits->[$d]);
  } while $base10;

  return join('', @base);
}
```

This uses a package namespace to make the `$digits16` and `$digit16` values not be global, but also only have to be computed once (there are other ways to do that, but this makes it possible to modify these lists from within a problem, if needed).

It also handles the `options` hash by using a default hash rather than the `//` operator, since this makes the allowed options and their defaults more apparent.  The `digits` option is allowed to be longer than 16, so this gives the option of using bases bigger than 16 without modifying the code (the `$digits16` value could also be larger to make that the default).

The check for valid digits is done in the conversion loop rather than using a dynamically created regex, which is more expensive (also, you haven't quoted regex special characters, so someone could potentially create an invalid or unexpected regex by using things like `*` or `?` as digits).

Finally, the conversion from base 10 to another base uses an array that is joined at the end rather than creating multiple strings, though both are about the same efficiency.

In my timing tests, the conversion to base 10 is about 4 times faster using Horner's method with the loop hash, skipping all the validity checking to just concentrate on the algorithm, and even with the checking, it is still 3 times faster.  The conversion from base 10 to an alternate base is only marginally faster with the arrays rather than the strings.

I changed the `Value::Error` calls to plain old `die`, as `Value::Error` does that in the end anyway, and it seems odd to use a `Value` call for something that is not MathObject-specific.  But that is a personal preference.  In any case, you don't need `return` with `Value::Error`, as `Value::Error` throws an error condition and stops the execution of the function at that point.

-----

I note that your examples all use `contextArbitraryString.pl`.  That can lead to some issues (e.g., leading or trailing spaces in the student answer can be a problem, as I recall).  It might be better to make an actual context that handles numbers in alternate bases.  I once wrote a hexadecimal context, which was made available in [this post](https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4181#p11832), that might serve as a starting point.  The `hex()` and `main::spf($self->value, '%X')` functions would need to be replaced by the conversion to and from base 10, but the rest of it should probably be OK as is.  There would need to be a context flag indicating the base being used, so that you could change the base to the one you wanted.

This has the advantage of being able to do arithmetic in the given base, while the `LimitedBase` context could prevent operations, as is the case for `LimitedHex` in the `contextHex.pl` linked to the post above.  I'm sure there are more features that could be added to the context.  That was just a quick one I put together as an example.

-----

Of course, these are just suggestions.  You can take them or leave them as you see fit.

macros/math/nondecimal_base.pl
```diff
@@ -0,0 +1,147 @@
...
+
+    $x = 47;
+    $x2 = convertBase($x, to => 2);
+
+    $y16 = convertBase('2EF9', from => 16);
+
+This can be used in problems with base conversion.
+
+    $x = random(200,500);
+    $x_16 = Real(convertBase($x, to => 16));
```
dpvc, line 32: You should make it clear that this requires the `ArbitraryStrings` context.

macros/math/nondecimal_base.pl
```diff
@@ -0,0 +1,147 @@
...
+
+    BEGIN_PGML
+    Convert the number [$x] (in decimal) to base-16
+
+    [$x] = [___]{$x_16} [`_{16}`]
```
dpvc, line 37: Probably don't want the space between `{$x_16}` and the following `[`.

t/macros_math/nondecimal_base.t
```diff
@@ -0,0 +1,114 @@
...
+
+	is convertBase(853,  to => 12), '5B1',  'convert to base 12';
+	is convertBase(2023, to => 12), '1207', 'convert to base 12';
+	is convertBase(1678, to => 12, digits => [ '0' .. '9', 'T', 'E' ]), 'E7T',
+		'convert to base 12 using non-standard digits';
+
+	is convertBase(5752,  to => 16), '1678', 'convert to base 16';
+	is convertBase(41446, to => 16), 'A1E6', 'convert to base 16';
+};
+
```
dpvc, line 44: Should there be tests that include both `to` and `from`?  The code allows that.  Of course, they must both use the same `digits`, so I'm not sure if that is a limitation that needs addressing or not.

dpvc (changes requested):

dpvc:
Also, one could use [Math::Base::Convert](https://metacpan.org/pod/Math::Base::Convert) for converting bases.

t/macros_math/nondecimal_base.t
pstaabp: I thought of this and for some unknown reason didn't add such a test.  I will.

macros/math/nondecimal_base.pl
pstaabp: Yes.  very true.

pstaabp:
@dpvc Appreciate the feedback.  I was thinking of getting a functional conversion first rather than an efficient one (thinking that efficiency wouldn't really matter for ww problems), but this is great. 

> Also, one could use [Math::Base::Convert](https://metacpan.org/pod/Math::Base::Convert) for converting bases.

I looked into this and decided that this would be simple enough to do "in house" instead of requiring another package. 

I was thinking of creating a context.  I'll look into your example from the forum.

pstaabp:
@dpvc I'm starting to convert it to a context using your Hex example and just pushed current changes.  What I was thinking was, for example, I could create a context and the set the base, then perform arithmetic (at least +,-,*).  

However, the following

```
Context('NondecimalBase');
setBase(5);
$a = Compute('240+113');
```
correctly parses 240 in base 5, 113 in base, however also adds 240+113 in base 10 (353), then tries to convert 353 from base 5 and I get an error.  Do I need to override +,-,* ?

dpvc:
Sorry for not getting back to you earlier on this.  I looked at it, but didn't have time to write initially.

I haven't tried changing anything, but my firth thought is that you are trying to do conversion both to and from the base all in the `context::NondecimalBase::Number` class that replaces `Parser::Number`, and you have not replaced `Value::Real` for this context.  These two need to work in concert to implement the new base.  

The MathObjects library consists of two packages, `Parser` and `Value`.  The `Parser` classes implement the parsing of a string into its abstract syntax tree, while the `Value` packages implement the different mathematical objects and the actions you can take with them (like adding them, and so on).  The `Value` package was originally intended to be able to be used separately from `Parser` for computational purposes, but they have become rather intertwined over time, but the original separation into parsing and computation is still present.  

Your `context::NondecimalBase::Number` is the parsing portion of the alternative-base setup, and it should recognize a number in the alternative base (as it currently does), but its `eval()` method should produce a (subclass of) `Value::Real` that is designed to do the computation for those numbers.  That subclass will inherit all the functionality of an actual `Value::Real`, and you simply need to override its output functions in order to have it display its underlying real in the required base.

The current `eval()` method doesn't return a `Value::Real`, but rather a string, and that is not going to operate properly.   So the first step to making this work is to convert `context::Hex::Hex` to `context::NondecimalBase::Real` with the output routines converting the base-10 real that is its value to the required base output.  Then make your `eval()` function in `context::NondecimalBase::Number` return the `Package('Real')` as it did in the Hex context.

I will make more comments in the code itself, but that, I think, is the main issue.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,273 @@
+
+=head1 NAME
+
+contextHex.pl - Implements a MathObject class and context for integers
+                in hexadecimal notation.
```
dpvc, line 5: You'll need to change these lines, of course.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,273 @@
...
+	sub setBase     { context::NondecimalBase::setBase(@_); }
+	sub convertBase { context::NondecimalBase::convert(@_); }
+}
+
+###########################################################################
+
+package context::NondecimalBase;
+
+# defines the base for the context.
+our $base = 10;
```
dpvc, line 66: This should be a flag in the context rather than a global value.  As it is here, you can't have two separate context's with different bases.  For example, if you wanted to ask for a value to be converted to base 2 and to base 16 in the same problem, you would not be able to do that.  All the information about the context that can vary should be in the context instance, not in package globals, which should only be used for constant values.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,273 @@
...
+
+use Data::Dumper;
+#
+#  Initialize the contexts and make the creator function.
+#
+sub Init {
+	my $context = $main::context{NondecimalBase} = Parser::Context->getCopy("Numeric");
+	$context->{name} = 'NondecimalBase';
+	$context->{parser}{Number} = 'context::NondecimalBase::Number';
+	# $context->{value}{Real} = 'context::Hex::Hex';
```
dpvc, line 81: This override is necessary to make this work.  See may comments on the PR itself.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,273 @@
...
+	$context->{pattern}{number} = '[0-9A-F]+';
+	$context->functions->disable('All');
+	$context->operators->remove('^');
```
dpvc, line 84: This was removed because it was redefined below (as exclusive or).  If you are not going to add the bitwise operations, you can leave `^` as is, and not remove it.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,273 @@
...
+	#   '^'  => {precedence => .5, associativity => 'left', type => 'bin', string => ' ^ ',
+	#            class => 'context::Hex::BOP::hex', eval => sub {$_[0] ^ $_[1]}},
+	#   '>>' => {precedence => .4, associativity => 'left', type => 'bin', string => ' >> ',
+	#            class => 'context::Hex::BOP::hex', eval => sub {$_[0] >> $_[1]}},
+	#   '<<' => {precedence => .4, associativity => 'left', type => 'bin', string => ' << ',
+	#            class => 'context::Hex::BOP::hex', eval => sub {$_[0] << $_[1]}},
+	#   '~'  => {precedence => 6, associativity => 'left', type => 'unary', string => '~',
+	#            class => 'context::Hex::UOP::not'},
+	# );
+	# $context->{precedence}{Hex} = $context->{precedence}{special};
```
dpvc, line 101: You will need to use

``` perl
$context->{precedence}{NondecimalBase} = $context->{precedence}{special};
```

in order to make sure that operations involving your new `context::NondecimalBase::Real` will be handled by your code (rather than bu the other MathObject involved in the operation).

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,273 @@
...
+	my $digits = $options{'digits'};
+	print Dumper "from: $from; to: $to";
+
+	die "The digits option must be an array of characters to use for the digits"
+		unless ref($digits) eq 'ARRAY';
+
+	#
+	# The highest base the digits will support
+	#
+	my $maxBase = scalar(@$digits16);
```
dpvc, line 130: Since you are using `$digits` as the digits, you need to use `@$digits` here, not `@$digits16`.  When not specified by the user, `$digits` will be `$digits16`, but if they use `digits => ['0', '1']`, for example, you need to restrict them to base 2, while if they use `digits => ['0' .. '9', 'A' .. 'Z']`, they can use bases higher than 16.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,273 @@
...
+
+	die "The base of conversion must be between 2 and $maxBase"
+		unless $to >= 2 && $to <= $maxBase && $from >= 2 && $from <= $maxBase;
+
+	#
+	# Reverse map the digits to base 10 values
+	#
+	my $baseBdigits = { map { ($digits->[$_], $_) } (0 .. $from - 1) };
```
dpvc, line 138: While this works, it does mean you are making this array over and over again as you convert, even when that isn't necessary in most cases.  If the user doesn't specify the `digits` option, you can use the `$digit16` mapping without having to rebuild it each time.  That is the point of `$digit16`.  The code I provided includes a test here for whether `$digits == $digits16` and uses `$digit16` in that case for efficiency.  I recommend you do that here.

In fact, since you are now using a context, you could have the digits be a context flag as well, and use `setBase()` to set both the base and the allowed digits.  Then you can create the reverse mapping in `setBase()` and only do it once, even if alternative digits are specified.  You could then make `convert()` take a context as an argument so that you can read the digits from it (or use them as the default digits) and the reverse map.  It is also possible to subclass the Context class and make `convert()` a method of the context itself.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,273 @@
...
+}
+
+# set the base for the context.
+
+use Data::Dumper;
+
+sub setBase {
+	my $b = shift;
+	return Value::Error('The base must be greater than 1 and less than or equal to $max_base')
+		unless $b >= 2 && $b <= scalar(@$digits16);
```
dpvc, line 178: There is no `$max_base` variable, and if you are using `$digits16`, you know the highest is 16.

But since your `convert()` allows for specifying a different set of digits, it may be that you want to allow the digits to be specified here.  Then you can tell what the base is and can check that you have the right number of digits (rather than the other way around), and can create the digits array and reverse mapping once instead of doing it in `convert()` each time.  The base, the digits, and the  reverse mapping can be stored as properties in the context itself, rather than as globals.  If you set them up here, you won't need `$digits16` or $digit16` any more (just use `['0' ... '9', 'A' .. 'F']` as the default).

You are right to set the `$context->{pattern}{number}` based on the actual digits (but should only use the ones up to the base requested), though you will need to do `$context->update` after doing so.

If you subclass the Context object to add `convert()` as a method, you should also make `setBase()` a method as well.  Otherwise, you will need to pass the context whose base you are setting as an argument to `setBase()`.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,273 @@
...
+	$value = context::NondecimalBase::convert($value, from => $base);
+	return $self->SUPER::new($equation, $value, $ref);
+}
+
+#
+#  Return the value of the number in its given base.
+#
+sub eval {
+	$self = shift;
+	return context::NondecimalBase::convert($self->{value}, to => $base);
```
dpvc, line 209: This should be removed so that the `eval()` returns the proper `Value::Real` subclass (in the following line), not a string.  Then make sure that `$context->{value}{Real}` is properly set above to `context::NondecimalBase::Real`.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,273 @@
...
+	$self->Package('Real')->make($self->context, $self->{value});
+}
+
+###########################################################################
+#
+#  A replacement for Value::Real that handles hexadecimal integers
+#
+package context::Hex::Hex;
```
dpvc, line 217: Change this to `context::NondecimalBase::Real` and fix the comment above.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,273 @@
...
+our @ISA = ('Value::Real');
+
+#
+#  Stringify and TeXify in hex notation
+#
+sub string {
+	my $self = shift;
+	return main::spf($self->value);
```
dpvc, line 225: This is where

``` perl
return context::NondecimalBase::convert($self->{value}, to => $base);
```

should be.

dpvc (changes requested):
I've made some extensive comments, here, that I hope will help you get this working.

macros/contexts/contextNondecimalBase.pl
pstaabp: Finally getting some time to work on this.  I had pulled the base as a flag in the context and then was thinking of using `setBase` to do this as well (and will need to do something similar for digits), but it seems that `setBase` has no access to the context.   My latest try is to call setBase like:

```
context::NondecimalBase->setBase(5);
```
and then in the `setBase` function, I try to get the context via:

```
sub setBase {
	my ($name, $base) = @_;
	my $context = $main::context{$name} = Parser::Context->getCopy($name);
}
```

but this is undefined.  I was looking at `contextAlternativeDecimal.pl` as an example.

macros/contexts/contextNondecimalBase.pl
pstaabp: > If you subclass the Context object to add convert() as a method, you should also make setBase() a method as well. Otherwise, you will need to pass the context whose base you are setting as an argument to setBase().

I think I'm also missing something here.  This would be a subclass to `Parser::Context` ?

pstaabp:
@dpvc when you get a chance to look now.  Lot's of cleanup, but adding isn't working right now.  I'm testing `240+113` each in base 5, and it seems like `Parser::BOP::add` is being called with the two number interpreted in decimal.  Do I need to override this to first convert from base b to 10 and then add?

dpvc:
> I have a class `context::NondecimalBase::BOP::add`

You do not need that, but I think you misunderstand the way it is supposed to work.  Your `context::NondecimalBase::Number` should produce a `context::NondecimalBase::Real` as the result of its `eval()` method (yours currently produces a string; see below).  The `context::NondecimalBase::Real` holds its value as a base-10 real, and all the computation is done by Perl on those base 10 values.  That is all handled by the inherited methods from `Value::Real`.  The only real purpose of `context::NondecimalBase::Real` is to convert the base-10 internal value to the proper base in its `string()` and `TeX()` methods.  That's it.  Your `string()` method is correct, but your `TeX()` one is not, as it takes the `string()` output and converts it *again* to the new base.  You simply want 

``` perl
sub TeX {
	my $self = shift;
	my $base = $self->{context}{flags}{base};
	return '\text{' . $self->string . '}';
}
```

for that.

To fix your `context::NondecimalBase::Number`, you need to remove the line

``` perl
	return context::NondecimalBase::convert($self->{value}, to => $base);
```

which (incorrectly) returns a string (the same string the student entered, which was converted to base 10 and then back, for no apparent reason).  That lets it fall through to 

``` perl
	$self->Package('Real')->make($self->context, $self->{value});
```

which produces the `context::NondecimalBase::Real` with the proper base-10 value.

I also think you may not have the right idea about having a subclass of the Context class, but that is a separate issue for now.  See if you can get the computation working first, and just set the `base` flag by hand for now.

pstaabp:
@dpvc thanks for the help. I think I have a better handle on this now.  If `$n` is a `context::Nondecimal::Real`, then `$n->value` is the base-10 representation of the number and `$n->string` is the base b representation.  

And addition is working.  That is `Compute('240+113')` using base-5 is 403.

dpvc:
> If `$n` is a `context::Nondecimal::Real`, then `$n->value` is the base-10 representation of the number and `$n->string` is the base b representation.

Exactly.

> And addition is working.

Great!  Good job.  The next step is to straighten out the context.  Let me know when you are ready to do that.

pstaabp:
Seems to be mostly working now.  I'm trying to build a limited version of the context as well so students aren't allowed to put in operations.

pstaabp:
I wouldn't call this "ready for prime-time".  Maybe ready for late-night viewing instead, but quite functional and passing all its tests!!!

I added a 'LimitedNondecimalContext' as well which (consistent with other packages) allows only numerical values.  Also, the code is cleanup and much better documentation. 

Here are two sample problems using these Contexts.  
[nondecimal.pg.txt](https://github.com/openwebwork/pg/files/11065630/nondecimal.pg.txt)
[nondecimal2.pg.txt](https://github.com/openwebwork/pg/files/11065631/nondecimal2.pg.txt)

somiaj:
I notice two things testing this out, first the example given with different letters for digits doesn't seem to work:

```
Context('NondecimalBase');
Context()->flags->set(base => 12, digits => [0..9,'T','E']);

$test = Compute("3TE");
```

I get an error that `'TE' is not defined in this context;`, and I think this is because the pattern for numbers has not been updated to deal with the new characters, `$context->{pattern}{number} = '[0-9A-F]+';`.

Second, if I pick two characters in the A-F range, I don't get any errors, but then I don't get the same characters back.

```
Context('NondecimalBase');
Context()->flags->set(base => 12, digits => [0..9,'B','D']);

$test = Compute("3BD");
BEGIN_PGML
[$test] = [_]{$test}
END_PGML
```

This gives me `3AB` instead of `3BD` as the output. I also see `3AB` as the correct answer not `3BD`.

pstaabp:
I had tested this with '8E', but E is by definition defined, but hadn't tested '8T'.  It appears that this is being parsed as 8*T, however '8E' is being parsed as the number '8E'.

pstaabp:
Another update with the error that @somiaj noticed is fixed. More testing I determined that the digits need to be updated any time the base is changed. (For example if we go from base 12 to base 16).  

I couldn't figure out how to run code when the `Context()->flags->set(...)` is called, so determined that the digits used need to be updated each time to ensure the correct digits are used.  If anyone has some ideas, so those two lines (noted the code) would only need to be run when the base changed.

pstaabp:
@dpvc Can you take a look at this again when you get a chance.

pstaabp:
Updated the code again.  Thanks a ton @dpvc. I knew I was in the ballpark, but clearly wasn't quite right. I think the other macros that I was using as models didn't create a new Context object like you suggested here.  

I decided to leave the `digits` option in the `convertBase` function since this can be used without the context, so the base and digits are not known.   I'm wondering if I should have separate `digits` options for the `to` and `from` base calls.  That is, should we have the ability to convert from base `b1` (with digits `d1`) to base `b2` with digits `d2`? 

I also have a new sample problem. 
[nondecimal.pg.txt](https://github.com/openwebwork/pg/files/11337727/nondecimal.pg.txt)

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,268 @@
...
+
+	Value::Error('The base must be set for this context') unless $context->{base};
+
+	$value = $context->fromBase($value);
+	return $self->SUPER::new($equation, $value, $ref);
+}
+
+sub eval {
+	$self = shift;
+	my $base = $self->{equation}{context}{base};
```
dpvc, line 249: Since `$base` is never used, this line should be removed.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,268 @@
...
+	my $base10 = 0;
+	for my $d (split('', $baseB)) {
+		die 'The number to convert must consist only of digits: ' . join(',', @$digits) unless defined($digit->{$d});
+		$base10 = $base10 * $b + $digit->{$d};
+	}
+
+	return $base10;
+}
+
+# A replacement for Parser::Number that acepts numbers in a non-decimal base and
```
dpvc, line 231: `acepts` => `accepts`

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,268 @@
...
+in non-decimal bases
+
+=head1 DESCRIPTION
+
+This context implements positive integers and some operations on integers in a non-decimal base
+greater than or equal to 2.  The numbers will be stored internally in decimal, though parsed
+and shown in the chosen base.
+
+In addition, basic integer arithemetic (+,-,*,^) are available for these numbers, however because it
+is assumed that all numbers are integers, division '/' is disallowed.
```
dpvc, line 14: It might be better to implement integer division (result is essentially `int(A/B)`) rather than remove it entirely.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,268 @@
...
+and then select the appropriate context -- one of the following:
+
+    Context('NondecimalBase')->setBase(5);
+
+
+will now interprets those in Compute and student answers in base 5.
+
+    $a = Compute('104');
+    $b = Compute('233');
+    $sum = Compute("$a+$b"); # this is the base-5 number 342 (decimal 97)
```
dpvc, line 32: There is no need to use `Compute()`, here, as you can just do `$sum = $a + $b` and avoid the multiple conversions to/from strings.  You could use

``` perl
$sum = Compute("104 + 233");
```

as a better example.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,268 @@
...
+
+For many problems, one may wish to not allow operators in the student answers.  Use
+'LimitedNondecimalBase' for this.
+
+    Context('LimitedNondecimalBase')->setBase(5);
+
+    $a = Compute('104');
+    $b = Compute('233');
+    $sum = Compute("$a+$b"); # There will be an error on this line now.
```
dpvc, line 41: Similarly, here the use of `$a` and `$b` are unnecessary, and cloud the issue.  Just

``` perl
$sum = Compute("104 + 233"); # There will be an error on this line now.
```

is more direct.   Note that

``` perl
$sum = $a + $b;
```

can still be performed, however.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,268 @@
...
+
+=head2 Sample PG problem
+
+A simple PG problem that asks a student to convert a number into base-5 may include:
+
+    Context('LimitedNondecimalBase')->setBase(5);
+
+    # decimal number picked randomly.
+    $a = random(130,500);
+    $a_5 = Compute(convertBase($a, to => 5));
```
dpvc, line 61: Actually, there is an easier way to do this.  Since the context has `Real` set to be a non-decimal real in the current base, using `Real($a)` is enough to get you the desired `$a_5` without converting by hand.  E.g.,

``` perl
$a = random(130, 500);
$a_5 = Real($a);
```

is all that is needed.  Or it could be

``` perl
$a_5 = Real(random(130, 500));
$a = $a->value;
```

if you happen to have the non-decimal real first.  E.g.,

``` perl
$a = Compute("104");
$b = Compute("233");
$sum_5  = $a + $b;      #  = 342
$sum_10 = $sum->value;  #  = 97
```


The upshot is that converting into the context's base can be done using `Real()` and out of it via `->value`.

dpvc (changes requested):
> I decided to leave the `digits` option in the `convertBase` function since this can be used without the context, so the base and digits are not known. I'm wondering if I should have separate digits options for the `to` and `from` base calls.

I don't think you want to do that.  Because the `setBase()` function can take either a numeric base (using the default digits) or an array of digits (with the base being given implicitly as the number of digits in the array), and since `convertBase()` passes `to` and `from` to `setBase()`, that means that `to` and `from` can be either a number (using the default digits) or an array of digits (specifying the base implicitly as the number of digits).  So you can specify digits already, and there is no need for a `digits` parameter here.  For example, you can do 

``` perl
convertBase(11010100, from => [0, 1], to => [0 .. 7]);
```

to convert from binary to octal.

If you did want a separate `digits` option, then yes, you would need a separate one for each.  But this is basically redundant, and a source of potential inconsistency, since the number of digits might not match the given base, so you would need to include extra error checking (whether the digits are an array, and whether their number matches the base, if it is given).  In your current code, since you use `$digits` in both cases, you will be converting to and from the same base, regardless of `to` and `from`, as long as they are not 10.

Because the numbers of digits determines the base, there is no reason to give both, which avoids a potential mismatch between the two.  Giving a numeric base rather than digits indicates that your are happy to accept the standard digits.

macros/contexts/contextNondecimalBase.pl
pstaabp: fixed.

macros/contexts/contextNondecimalBase.pl
pstaabp: fixed.

macros/contexts/contextNondecimalBase.pl
pstaabp: I was looking a `contextFraction.pl` as an example and thought that I would need to make 
````package context::NondecimalBase::BOP::divide;```
and define division to be `int(A/B)`.  And then (similar to `contextFraction.pl`), defined a `_eval` subroutine.  

I added: 
```
$context->operators->set(
		'/'  => { class => 'context::NondecimalBase::BOP::divide'},
		' /' => { class => 'context::NondecimalBase::BOP::divide'},
		'/ ' => { class => 'context::NondecimalBase::BOP::divide'}
	);
```
to the `new` sub starting on line 152.  I'm not getting the `_eval` called above.  

Is this the right way to go and I'm missing something or is there an easier way?

File: macros/contexts/contextNondecimalBase.pl (hunk 18 above)
Comment (pstaabp) [reply macros/contexts/contextNondecimalBase.pl:32]: Makes sense!

File: macros/contexts/contextNondecimalBase.pl (hunk 19 above)
Comment (pstaabp) [reply macros/contexts/contextNondecimalBase.pl:41]: Again, makes sense.

File: macros/contexts/contextNondecimalBase.pl (hunk 20 above)
Comment (pstaabp) [reply macros/contexts/contextNondecimalBase.pl:61]: Again, I think I understand now. Thanks.

**Comment by pstaabp** [comment]
@dpvc About using digits, I see now that putting an array ref in for to or from does what is needed, so I agree, we don't need that and will adjust the documentation.  Thanks for all the other suggestions above. 

I had a question about defining `/` as integer division.  I asked above. 

Lastly, I'm thinking of a student putting in answers.  Let's say the answer is a base 5 and there are digits other than 0..4 in the answer, like '456'.  The error is `Variable '56' is not defined in this context`.  I think this is because on line 186:
```
$self->{pattern}{number} = '[' . join('', @$digits) . ']+';
```

It would be a nicer error message to use the one from the convert subroutine: `The number to convert must consist only of digits: ...`

To do this, if we change line 186 to allow any alphanumeric character, I think the convert one would be thrown.

File: macros/contexts/contextNondecimalBase.pl (hunk 17 above)
Comment (dpvc) [reply macros/contexts/contextNondecimalBase.pl:14]: It's hard to tell without seeing the actual implementation of the `context::NondecimalBase::BOP::divide`.  Can you post the package here?

Also, you may need to override the `div` method in `context::NondecimalBase::Real` to do the integer division.  The original is

https://github.com/openwebwork/pg/blob/a29b85f259fac1924688a2a6f8b80b9715c4bc55/lib/Value/Real.pm#L89-L93

so you probably only need to add `int()` around the result of the division in the last line of the function.

**Comment by dpvc** [comment]
> if we change line 186 to allow any alphanumeric character, I think the convert one would be thrown.

Yes, but you may cause other problems by doing that.  For example, in the (non-limited) context, if you have a variables `x` and `y` and want to write a binary expression like `1x + 0y`, you won't be able to do it, since `1x` and `0y` would be treated as numbers with a bad digit. 

Maybe just adding 0 through 9 would be sufficient?  Or maybe just uppercase alphas and digits might be safer.  Or having an `extraDigits` option for `setBase()` that could be set by the author to the things that a student might use as a digit, so there are no surprises about expression like that.

**Comment by dpvc** [comment]
PS, it is also possible to set the context to override that message and provide a different one.  Say, `'56' does not consist of valid digits and is not a valid variable name` or something like that.

**Comment by pstaabp** [comment]
@dpvc I think I have the integer division working nicely and cleaned up documentation and other things.

**Review by dpvc — CHANGES_REQUESTED** [review]
Sorry for the long delay in getting back to you on this.  I have concerns about how the digits are being handled that I mentioned below, plus one suggestion for an efficiency.

File: macros/contexts/contextNondecimalBase.pl [hunk 21]
```diff
@@ -0,0 +1,274 @@
...
+sub new {
+	my $self    = shift;
+	my $class   = ref($self) || $self;
+	my $context = bless Parser::Context->getCopy('Numeric'), $class;
+	$context->{name}           = 'NondecimalBase';
+	$context->{parser}{Number} = 'context::NondecimalBase::Number';
+	$context->{value}{Real}    = 'context::NondecimalBase::Real';
+	$context->functions->disable('All');
+	$context->constants->clear();
+	$self->{pattern}{number}               = '[' . join('', 0 .. 9, 'A' .. 'Z') . ']+';
```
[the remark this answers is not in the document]
pstaabp: 1. Even though in general it would be nice to extend to bases larger than 36, I'm wondering if this is needed for WeBWorK problems. 
2. if we add `'a'..'z'` to the `$context->{pattern}{number}`, we should be able to use lower case letter (and then the size of the base could grow to 62). 

I updated the `{pattern}{number}` like you suggested.

pstaabp:
@dpvc made changes again.  If you could take a look.

pstaabp:
@dpvc, could you take a look at this again?

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,280 @@
...
+contextNonDecimalBase.pl - Implements a MathObject class and context for numbers
+in non-decimal bases
+
+=head1 DESCRIPTION
+
+This context implements positive integers and some operations on integers in a non-decimal base
+greater than or equal to 2.  The numbers will be stored internally in decimal, though parsed
+and shown in the chosen base.
+
+In addition, basic integer arithemetic (+,-,*,/^) are available for these numbers.
```
dpvc, line 13: comma after `/`?

dpvc (approved):
Sorry, I thought I had done the review earlier.

Just a few typos in the comments.

macros/contexts/contextNondecimalBase.pl
```diff
@@ -0,0 +1,280 @@
...
+
+    loadMacros('contextNondecimalBase.pl');
+
+There are two contexts: C<NondecimalBase> and C<LimiitedNondecimalBase>, where the former
+allows operations between numbers and the latter only allows numbers. To use either
+first, one must set the base.  For example:
+
+    Context('NondecimalBase')->setBase(5);
+
+will now interprets those in Compute and student answers in base 5.
```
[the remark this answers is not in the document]
pstaabp: fixed

macros/contexts/contextNondecimalBase.pl
pstaabp: fixed.

Alex-Jordan:
I am reviewing this, and see some small non-controversial things. I think it easiest if I just open a PR with changes.

Possibly controversial: could the name be better? Strictly speaking it is not accurate, since you could use base ten with this. `contextAlternateBase.pl` or `contextBaseN.pl` are all that I thought of.

Alex-Jordan:
So I think I was able to do what I described. That is, make this happen:

```
Context('NondecimalBase')->setBase(5);
TEXT(Real(9));     # prints 14, which is nine in base five.
TEXT(Real('9'));   # raises an error because '9' is not a base 5 digit.
```

The question is if everyone thinks this is a good distinction to have. The TLDR of my post before was to consistently make strings be interpreted with the alternate base, and plain perl integers be interpreted base ten.

Alex-Jordan:
Let's forget what I was trying to do in the last few posts. While I can get `Real(9)` and `Real('9')` to be distinguished, it will not carry over easily to things like `List(8, 9)` and `List('8', '9')`, etc. The distinction is too fragile.

I opened a PR (https://github.com/pstaabp/pg/pull/22) to your branch with four commits:


* some light editing to the documentation. This includes clarifying when MathObject argument strings will have numerical substrings interpreted in base ten and when in the context's base.
* I added some strings you can use for `setBase`. The most extreme is `...->setBase('base64')` as a short way to declare the 64 digits for that base.
* Since this is all integer stuff, I added a modulo operator `%`.
* I changed the file name (and all the package naming etc) to BaseN instead of NonDecimalBase. If you want to keep that name, can just cut the last commit. Although the last commit also has perltidy, so if it is cut, it needs another perltidy application.

pstaabp:
Just merged in the suggestions from @Alex-Jordan and added a few more unit tests.

Alex-Jordan (approved):
Officially marking for approval.

macros/contexts/contextBaseN.pl
```diff
@@ -0,0 +1,330 @@
...
+
+    # decimal number picked randomly.
+    $a = random(130,500);
+    $a_5 = Real($a); # converts $a to base-5
+
+    BEGIN_PGML
+    Convert [$a] to base-5:
+
+    [$a] = [__]{$a_5}[`_5`]
+    END_PGML
```
Alex-Jordan, line 95: oops, this is almost a complete PG file but missing the last line.

Alex-Jordan:
I marked for approval. But then remembered one last thing I had thought to add. I opened a PR to your branch about that.

Beyond that, was there any consideration for case insensitivity? Like having hexadecimal, but letting 10AE equal 10ae? I know it could be more trouble than it's worth when you consider a digit set that contains both capital and lowercase letters. But assuming the only letters in the digit set were capital, you could imagine allowing case insensitivity. Maybe that is just enabling inattentive students...

pstaabp:
> I marked for approval. But then remembered one last thing I had thought to add. I opened a PR to your branch about that.
> 
> Beyond that, was there any consideration for case insensitivity? Like having hexadecimal, but letting 10AE equal 10ae? I know it could be more trouble than it's worth when you consider a digit set that contains both capital and lowercase letters. But assuming the only letters in the digit set were capital, you could imagine allowing case insensitivity. Maybe that is just enabling inattentive students...

Maybe on a case-by-case basis.  Clearly on your new base64 example, case sensitivity is important.

PR #769, opened 2023-01-27
Fix and deprecate unionInclude.pl

drgrice1:
Make the unionInclude.pl macro work again.  `fileName` is no longer in the environment.  Instead of having `fileName` and `probFileName` in the environment that were both always the same thing, there is now only `probFileName`.

However, this macro should not be used anymore, so it is also moved to deprecated.

There is also a minor fix for a warning in niceTables.pl that I keep seeing.  This was included in #709 and accidentally reverted in #724.


pstaabp (approved):
This does fix a warning from the macro and YES this should be deprecated. In order to actually work, problems need to be setup in the right way relative to others in the set.

PR #772, opened 2023-02-02
macro to check data to be stored in a db. 

pstaabp:
This is in conjunction with openwebwork/webwork2#1879  which provides functionality to store student data in a database to be retrieved for other problems.  See the webwork2 pr for more detail.


drgrice1 (commented):
This macro is missing the `_init_` method.

Perhaps you could find a way of implementing this that doesn't require the messy `~~&store_number_list` usage.  One way to do it that is perhaps a bit cleaner is to use a closure.  Change this to
```
sub store number_list {
    return sub {
        my (...) = @_;
        ...
        return ...;
    }
}
```
Then you could use it with `[___]{$ans->cmp(list_checker => store_number_list(), key => 'whatever')}`.

Another way to do this is to store the anonymous `sub` in a variable with
`$store_number_list = sub { your code };`.
Then you could use it with
`[___]{$ans->cmp(list_checker => $store_number_list, key => 'whatever')}`

Also it seems that there should be other methods added to store other types of information.  Why limit to only storing real numbers?

macros/core/externalData.pl
```diff
@@ -0,0 +1,41 @@
...
+=head1 NAME
+
+externalData.pl - Macro that reads and writes data from a problem to the database.
+This is useful for using data common to multiple problems.
+
+=cut
+
+sub store_number_list {
+	my ($correct, $student, $ansHash, $value) = @_;
+	my @errors = ();    # stores error messages
```
drgrice1, line 25: I just noticed this too.  Never define an array with `my @array = ();`.  It is just unnecessary.  Just use `my @array;`.  It is initialized to the empty array by default.  No need to explicitly add `= ()`.  The same goes for hashes.  Never do `my %hash = ()`.

macros/core/externalData.pl
```diff
@@ -0,0 +1,41 @@
...
+
+	# Make sure that the key is set
+	my $key = $ansHash->{key};
+	warn 'You must pass in the key to save the data.' unless $key;
+
+	# Get the answer name for this input and store the name and key.
+	my $ans_name = $ansHash->{ans_label};
+	push @{ $main::PG->{flags}{KEPT_EXTRA_ANSWERS} }, "_ext_data:$ans_name:$key";
```
drgrice1, line 33: Actually, this should be `RECORD_EXTRA_ANSWERS("_ext_data:$ansHash->{ans_label}:$key")`.  Your code is pretty much what the `RECORD_EXTRA_ANSWERS` method does.

pstaabp:
> Also it seems that there should be other methods added to store other types of information. Why limit to only storing real numbers?

I have plans for others.  I wanted to see some feedback on this to make sure I have a sound method first.

macros/core/externalData.pl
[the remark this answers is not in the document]
pstaabp: done.

pstaabp: I see that.

macros/core/externalData.pl
pstaabp: done.

pstaabp:
> This macro is missing the _init_ method.

oops.  added.

> Perhaps you could find a way of implementing this that doesn't require the messy ~~&store_number_list usage. One way to do it that is perhaps a bit cleaner is to use a closure. 

Yes.  This would definitely be better.

macros/core/externalData.pl
```diff
@@ -0,0 +1,123 @@
...
+
+		# Get the answer name for this input and store the name and key.
+		RECORD_EXTRA_ANSWERS("_ext_data:$ansHash->{ans_label}:numeric_list:$key");
+
+		# check that all numbers are real.
+		my $all_real = 1;
+		$all_real = $all_real && Value::isRealNumber($_) for (@$student);
+		push(@errors, 'One of the numbers is not a real number') unless $all_real;
+
+		return ($all_real ? scalar(@$student) : 0, @errors);
```
drgrice1, line 65: It seems that this would be a lot cleaner and readable if it were:
```suggestion
		for (@$student) {
			return 0, 'One of the numbers is not a real number' unless Value::isRealNumber($_);
		}

		return scalar(@$student), @errors;
```

Although I don't think you are using the correct check with `Value::isRealNumber($_)`.  That is unless you intend for formulas involving variables to be allowed as entries in the list.  With that check an answer of `1,2,x,x^2+4` will be counted as correct.  If this is intended then you will need to change https://github.com/openwebwork/webwork2/pull/1879 to accept that sort of answer.  An answer like the example given causes the warning `Argument " x" isn't numeric in addition (+) at /opt/webwork/webwork2/lib/WeBWorK/Utils/ProblemProcessing.pm line 485.`   I doubt you did intend to allow that.  So the check should be `Value::classMatch($_, 'Real')` instead.

macros/core/externalData.pl
```diff
@@ -0,0 +1,123 @@
...
+  END_PGML
+
+=cut
+
+sub store_scalar {
+	my $type = shift;
+	warn "The type $type is not a valid datatype" unless grep { $_ eq $type } qw(string real);
+	return sub {
+		my ($correct, $student, $ansHash, $value) = @_;
+		my @errors;    # stores error messages
```
drgrice1, line 112: `@errors` is not used.  Please delete.

drgrice1 (commented):
I see a few issues that need to be fixed.

macros/core/externalData.pl
```diff
@@ -0,0 +1,123 @@
...
+
+		# Make sure that the key is set
+		my $key = $ansHash->{key};
+		warn 'You must pass in the key to save the data.' unless $key;
+
+		# Get the answer name for this input and store the name and key.
+		RECORD_EXTRA_ANSWERS("_ext_data:$ansHash->{ans_label}:numeric_list:$key");
```
drgrice1, line 58: As per my comment on https://github.com/openwebwork/webwork2/pull/1879, remove the `:numeric_list:` type part of this.  That is not a good idea.  The same for `:string:` and `:real:` below.

macros/core/externalData.pl
```diff
@@ -0,0 +1,123 @@
...
+	return sub {
+		my ($correct, $student, $ansHash, $value) = @_;
+		my @errors;    # stores error messages
+
+		# Make sure that the key is set
+		my $key = $ansHash->{key};
+		warn 'You must pass in the key to save the data.' unless $key;
+
+		# Get the answer name for this input and store the name and key.
+		RECORD_EXTRA_ANSWERS("_ext_data:$ansHash->{ans_label}:$type:$key");
```
drgrice1, line 119: This checker doesn't actually check the answer for validity at all.  For a string answer that may be fine, but for a real number it should check that the answer entered is actually a real number.  In fact, since the two are really quite different, there should not be just a `store_scalar` method.  Instead there should be a `store_string` and `store_number` (or `store_real`) method.

macros/core/externalData.pl
drgrice1: Also, the `RECORD_EXTRA_ANSWERS` call should not be in the checker.  That also should be before the checker in the `store_number_list` method.

Of course all of these changes need to be made to any other methods in this file.

drgrice1: Ahh, I guess it needs to be here with this technique.  I just noticed you are obtaining the answer label from the answer hash.  Hmm.  This is probably not a good structure.  I don't think it is a good idea to call RECORD_EXTRA_ANSWERS in a checker.  That seems fragile.

macros/core/externalData.pl
pstaabp: This is intended for Reals not formulas with x.

macros/core/externalData.pl
pstaabp: Do you have any suggestions to improve this?

drgrice1: For now, just leave the `RECORD_EXTRA_ANSWERS` inside the checker.  It should be fine.  Just remove the data type specifier (`numeric_list`, `string`, `real`).  Trying to validate data types on the webwork side of things without the actual MathObject is unreliable at best.

drgrice1 (approved):
This is ready to go now I think.

PR #774, opened 2023-02-10
Rewrite of draggableProof.pl and draggableSubsets.pl

drgrice1:
This switches the draggableProof.pl and draggableSubsets.pl macros from using jquery.nestable to using sortablejs.  The sortablejs library works much better, and is a library that is maintained.  jquery.nestable has never been maintained, and was never even intended to be used in production.  In addition all dependence on jQuery has been removed.  The sortablejs library is the only third party dependency.

The draggableProof.pl and draggableSubsets.pl macros have been completely rewritten to fix issues seen with them.  Currently hardcopy is broken for answers (both student and correct).  The current macros do not support custom checkers, and due to limitations of the current design can not be extended to do so without this rewrite.

There were several options added to the macros.  These options allow for customization of various textual components of the javascript output. These options would be useful for writing a problem for another language utilizing the macro.

A "cmpOptions" option was added to both macros.  This can be used to conveniently pass options that will then be directly passed on to the cmp method.

The previous `Print` method of the macros is now the `ans_rule` method. The `Print` method still exists, but just calls the `ans_rule` method. The addition of the `ans_rule` method makes it so that you can use the macro in PGML via `[_]{$draggable}` where `$draggable` is a DraggableProof or DraggableSubset object.  Unfortunately, if you want problems to be compatible with older versions of PG you will still need to use the `Print` method (things done wrong!).

Note that the dependence of the macros on the PGchoicemacros.pl macro has also been removed.  However, the draggableProof.pl macro still loads it.  This is because some of the OPL problems that use the macro use the methods from the PGchoicemacros.pl macro without loading it directly.

The attached zip file contains a few demonstration problems, as well as a set export file containing all problems from the OPL and Contrib that use the draggableProof.pl macro.  No problems use the draggableSubset.pl macro at this point.
[DraggableProblems.zip](https://github.com/openwebwork/pg/files/10711146/DraggableProblems.zip)


pstaabp (approved):
Tested this and looks good.

PR #775, opened 2023-02-12
Rework the mqeditor toolbar postioning and behavior

drgrice1:
The toolbar is absolutely positioned relative to the position of the answerQuill input on the page that it corresponds to.  This means that the toolbar will stay with the input when the page scrolls.  If the problem is rendered in an iframe, then the toolbar is still always to the right of the input.  Care is taken to ensure that the toolbar doesn't go off the top or bottom of the page (unless the page is too short to accommodate the toolbar height -- see below).

The mqeditor toolbar is now scrollable when in short pages.  The toolbar does not collapse into multiple columns, and is always a single column. So it doesn't intrude into the problem when the page is short anymore than its single column 53px width (plus a 10 pixel gap to the right of the toolbar).

In order for the toolbar to be scrollable, it needs to be able to attain focus.  Otherwise it will disappear when the user tries to scroll it using the scrollbar.  So the toolbar is inserted into the page immediately after the MathQuill input.  This means that if the MathQuill input has focus and the tab button is pressed, then focus will go to the first button of the toolbar (and then on through the buttons when tab is pressed additional times).  If you tab all the way through the buttons, then focus will continue in the page to the next place it would have gone if the toolbar wasn't there.  Note that the toolbar is not in the reverse tab order when using shift-tab from somewhere after the MathQuill input (since it doesn't exist at that time).  This new tab order approach is probably more accessible in any case, since the toolbar now can be used with the keyboard.

Finally, if you don't want the toolbar to open for you at all, then type alt-t when any MathQuill input has focus.  The toolbar will hide, and will not open again for any MathQuill input.  If you type alt-t again in any MathQuill input, then the toolbar will open and thereafter will continue to open for all MathQuill inputs.  This setting is saved in local storage, and so will persist for any page on the site. I.e., alt-t typed in any MathQuill input toggles whether the toolbar is enabled or not.

Edit:  Instead of using alt-t (or some other key combination) to toggle toolbar status, this now uses a right-click dropdown menu.

Also, the mqeditor.css file is not mqeditor.scss.


Alex-Jordan:
Thanks, this sounds really nice. Would you expect that I can test this on 2.17? (By copying the three files, running npm install, restating apache, and clearing cache?)

Meanwhile, I am weighing spinning up a 2.18 (er, dev?) server. My usual dev server is sort of locked at 2.17 so i can experiment with things to apply to our production server.

drgrice1:
Yes, this should be compatible with webork 2.17.  Although, there are only two files to copy.  The third (mqeditor.css) will need to be deleted.

Alex-Jordan:
On a Mac, what should the equivalent be to alt-t? I have tried combinations of {shift,control,option,command} in Firefox and they do not work like alt-t is described. Note that command-t or command-shift-t opens a new tab in firefox, and option-t and option-shift-t makes make special characters † and ˇ.

Otherwise I find that it is mostly working. I found some cosmetic issues, and one iframe issue.

In regular WeBWorK, if the answer blank is near the top of the screen, and the screen is narrow enough to trigger the smallscreen layout, the palette is placed high up as in this screen shot:

Screen Shot 2023-02-12 at 2 13 32 PM


When using html2xml to render a problem, the div.quill-toolbar is taller than it needs to be, and the border extends down to the bottom of the screen as in this shot:

Screen Shot 2023-02-12 at 2 25 24 PM

Lastly when I put an html2xml based page as the src of an iframe like here:
https://spot.pcc.edu/~ajordan/temp.html
then when I put focus into the answer blank, there is a console error about the iframe-internal js trying to access `document` and the toolbar does not appear.

dlglin:
> Meanwhile, I am weighing spinning up a 2.18 (er, dev?) server. My usual dev server is sort of locked at 2.17 so i can experiment with things to apply to our production server.

With 2.18 running under mojolicious it should be possible to run an instance of 2.18 on the same server as a separate 2.17 instance.  I haven't tried this, but I can envision several different ways to accomplish it.

drgrice1:
@Alex-Jordan:  All of the issues you mentioned should be fixed with the changes I just added.

That is, except for the alt-t issue.  I believe that the key combination on OSX should be option-t.  However, it seems that 
OSX maps that to †, and apparently doesn't allow an override (why?).  Just to make sure, you did have a MathQuill input focused when you tried option-t correct?  On any browser, a MathQuill input must be focused (with the cursor blinking in it) in order for the key combination to work.  That probably is the case, but just want to make sure.

I chose alt-t because it doesn't interfere for Firefox or Chrome on (Ubuntu) linux.  In any case, I knew that this may need to be adjusted.  What is a good key combination that will work on (most) platforms and (most) browsers?

@dlglin:  2.17 and 2.18 can certainly be run together on the same system, 2.17 under apache2 with mod_perl and 2.18 under morbo on a different port.   At this point you can even use the same database (with care).

Alex-Jordan:
I can confirm that everything I mentioned is working now, except the alt-t feature.

For "alt-t" on OSX, maybe command-option-t can be the alternative. I checked Firefox, Chrome, Safari, and Edge and I do not find a conflict. Also with Voiceover, I do not find a conflict (most Voiceover commands use control-option with one or more other keys.)

And then I found this, which suggests command-option-t is what should be used: https://defkey.com/apple-macos-shortcuts#7773

Alex-Jordan:
I see, the `pretext-content` class is added to the `main` surrounding the `p` and then the CSS kicks in on `.pretext-content p`. I'll think of how to address that.

drgrice1:
Yeah, my mistake.  The class is actually on a containing `div` as you said.  Not directly on the `p` tag.

drgrice1:
> Here's another thing. At this page: https://pretextbook.org/examples/webwork/sample-chapter/html/section-4.html I can activate exercises 4, 5, and 6, and MathQuill is working. But with exercise 1, MathQuill never activates. There are no console errors. Do you have any insight to why the difference? All these exercises are using the same server, same host course.
> 
> On this page: https://pretextbook.org/examples/webwork/sample-chapter/html/section-1.html exercise 6 has MQ, but none of the others show it.

Does MathQuill work with those problems without the javascript from this pull request?  I can't find the hidden MaThQuIlL inputs for the answers in those problems.  If those aren't there, then the mqeditor doesn't active MathQuill for those inputs.

Alex-Jordan:
Good catch, I should have noticed the association with solutions and hints.

Is it `mqeditor.js` or something in `pg` somewhere that places those hidden inputs? Do we have any hope of placing them somewhere else closer to the original input elements?

Or could we put the hidden mathquill inputs inside their own div? That would make the browser realize it is time to close the p.

drgrice1:
This is already fixed for the next release of webwork.  That was the point of https://github.com/openwebwork/pg/pull/745.  I was having the problem you are having trying to get MathQuill to work for PreTeXt last summer, and that pull request is the result.  PG places the inputs, but it also places other things at the end of the problem that were being sucked into the `p` tags.  There is really no other way to do it because the inputs have to be placed after the rest of the problem is parsed so it knows what the answer rules are. Also, there is no guarantee that the solutions or hints will be at the end of the problem.  Those can be placed earlier in the problem as well, and then other things could be sucked into those `p` tags.  With the next release of webwork those will be `div` tags that are properly closed.

drgrice1:
@Alex-Jordan:  I added special handling to deal with the case that the MathQuill input is contained in a relatively positioned parent.  So now this should work for the PreTeXt problems that have the `pretext-content` style applied to a parent.

drgrice1:
Using a key combination to  toggle toolbar status seems to be troublesome.  On OSX it cannot be option-t which is the equivalent of alt-t.  It was suggested to use command-option-t, but the equivalent on other platforms is meta-alt-t.  That is a rather unintuitive key combination on other platforms.  The meta key on linux is the keyboard windows key if present.  Otherwise I don't even know what it would be (it can be configured).  On windows I don't know what it is, but according to MDN it is NOT the windows key.

So instead this now uses a right-click dropdown menu for this.

pstaabp:
Overall, this looks good on webwork2, however I noticed on the library browser (problem in an iframe) that the bottom of the buttons are cutoff:

image

Alex-Jordan:
But can you scroll the buttons? It's supposed to behave that way now.

drgrice1:
@Alex-Jordan:  We were still in the meeting when @pstaabp posted that.  I told him then what you posted, and he saw that it does scroll.

pstaabp (approved):

Alex-Jordan:
Following the meeting today, I straightened things out to apply this to the AIM 2.17 server. Right-clicking still does not work, but now there is a console error that explains why. Something is not right with dropdown.js, which is part of bootstrap. So now it seems to be a 2.17 versus develop clash. Do you see a way this can be safely applied to a 2.17 server?

drgrice1:
Yeah, bootstrap was updated for the develop branch.  I can see if I can figure out what has changed that doesn't work with the version of bootstrap used for 2.17.

drgrice1:
@Alex-Jordan:  I made some changes to the "right-click" menu that should fix the issue.  The changes are better for the develop branch also.  Bootstrap 5.1.3 didn't like not having the rest of the markup that I omitted.  Bootstrap 5.2.3 doesn't really like it either though, and as a result I had to use a hack to close the popup.  I added the markup, and removed the hack.  So this will work for both versions of Bootstrap.

Alex-Jordan:
Thanks for working on that, and all of this. It's working now in 2.17, in PTX embeddings.

drgrice1:
I added another small tweak.  Now if you press escape when the toolbar has focus, then the toolbar will close.  This makes it easier to move on to the next answer rule.

Alex-Jordan:
As this PR is right now, if I wanted to apply this to 2.17, is it still as simple as putting the two new files in place, deleting the third, etc? Just checking this is independent from #784 .

PR #777, opened 2023-02-16
Fix some HTML validation issues

drgrice1:
Specifically the following changes are made:

Remove the size attribute from hidden inputs created via NAMED_HIDDEN_ANS_RULE.  That is not a valid attribute for a hidden input.

Remove the type="text/css" attribute from style tags and type="text/javascript" attribute from script tags.  In modern html those should be omitted.

Remove the wrap attribute from textarea answer rules. The "virtual" value for that attribute hasn't been valid for a long time.  Just use the default "soft" value.

Remove the invalid autocorrect attribute from inputs for answer rules.

The debug and warning messages in PGcore should not add html formatting to the arrays of messages that are stored.  Let the caller deal with formatting it the in the desired way.  The only result of this is that we end up removing this html later almost every single time it is handled.
    
Also output better and valid html for translator errors.  A pre can not contain a <hr> tag.

This is just a collection of issues that I have been accumulating in a branch. It is time to put this in.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -321,7 +321,7 @@ sub NAMED_ANS_RULE {
 
 		# Note: codeshard is used in the css to identify input elements that come from pg
 		HTML => qq!<input type=text class="codeshard" size=$col name="$name" id="$name" aria-label="$label" !
-			. qq!dir="auto" autocomplete="off" autocorrect="off" autocapitalize="off" spellcheck="false" !
+			. qq!dir="auto" autocomplete="off" autocapitalize="off" spellcheck="false" !
 			. qq!value="$answer_value">!
 			. qq!<input type=hidden name="$previous_name" value="$answer_value">!,
 		PTX => qq!<fillin name="$name" characters="$col" />!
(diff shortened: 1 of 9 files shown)


macros/core/PGbasicmacros.pl
```diff
@@ -455,11 +455,9 @@ sub NAMED_ANS_BOX {
 		TeX        => qq!\\vskip $height in \\hrulefill\\quad !,
 		Latex2HTML =>
 			qq!\\begin{rawhtml}<TEXTAREA NAME="$name" id="$name" aria-label="$label" ROWS="$row" COLS="$col"
-		              WRAP="VIRTUAL">$answer_value</TEXTAREA>\\end{rawhtml}!,
-		HTML => qq!<TEXTAREA NAME="$name" id="$name" ROWS="$row" COLS="$col"
-		        WRAP="VIRTUAL">$answer_value</TEXTAREA>
-		        <INPUT TYPE=HIDDEN  NAME="previous_$name" VALUE = "$answer_value">
-		        !,
+				>$answer_value</TEXTAREA>\\end{rawhtml}!,
+		HTML => qq!<TEXTAREA NAME="$name" id="$name" ROWS="$row" COLS="$col">$answer_value</TEXTAREA>!
```
pstaabp, line 459: Should this TEXTAREA have the aria-label on it?

drgrice1: Yeah, it should.   In fact, a mistake that traces back to be0e7139c (a @goehle commit) is apparent here.  Looking at the code above, aria labels were added by mistake to the Latex2HTML code but not to the actual HTML code.  I will move that.  The Latex2HTML should not have that at all (and will soon be deleted anyway).

pstaabp (approved):

PR #778, opened 2023-02-19
Add the capability to set input value for radios and checks

drgrice1:
This is done for the parserRadioButtons.pl, parserCheckboxList.pl, and parserRadioMultiAnswer.pl macros.
    
The values for the choices are the student "original" answers which are displayed in the past answers table (or answer log).  Currently for this macro those are always B0, B1, etc.  This changes that so those can be customized to show more informative "original" answers in the results table.  Note that if using a custom checker, these are also the values that you would need to use for comparison there.

These values can be specified in two ways for parserRadioButton.pl and parserCheckboxList.pl.  One extends to the previous method of providing a label with a choice. For example, a choice could have been give as { label => text }.  To give a value with that now use { label => [ text, value ] }.  For a full example:

```
$radio = RadioButtons(
        [
                { A => [ 'answer 1', 'always first answer' ] },
                [
                        { B => [ 'answer 2', 'random order second answer' ] },
                        { C => [ 'answer 3', 'random order third answer' ] }
                ]
                { D => [ 'answer 4', 'always last answer' ] }
        ],
        1,
        displayLabels => 1
);
```
Note that with this approach the labels, text, and value are all randomized together and everything is shown in that random order.  So in the problem you will get for instance
* A. answer 1 (the value of this radio will be 'always first answer')
* C. answer 3 (the value of this radio will be 'random order third answer')
* B. answer 2 (the value of this radio will be 'random order second answer')
* D. answer 4 (the value of this radio will be 'always last answer')

Of course B and C might be switched.

The other way of specifying the values is to give a reference to an array containing values for each of the choices.  This works for all three macros.  For example,
```
$radio = RadioButtons(
        [
                'answer 1',
                [

                       'answer 2',
                        'answer 3',
                ]
                'answer 4',
        ],
        1,
        labels => 'ABC',
        values => [
                'always first answer',
                'random order second answer',
                'random order third answer',
                'always last answer'
        ]
);
```
Note that with this approach the values are still randomized with the choices, but the labels are not.  So in the problem you will get for instance
* A. answer 1 (the value of this radio will be 'always first answer')
* B. answer 3 (the value of this radio will be 'random order third answer')
* C. answer 2 (the value of this radio will be 'random order second answer')
* D. answer 4 (the value of this radio will be 'always last answer')

Of course the text for B and C might be switched and the values with them, but the labels B and C will always be with the corresponding text as shown.

Of course one should choose the values to not give the correct answer away.

This also moves the "uncheckable" javascript out of the macro and into a separate javascript file loaded via ADD_JS_FILE.  The javascript is updated to not pollute the global namespace.

Note that the NAMED_ANS_RADIO_EXTENSION of PGbasicmacros.pl needed to be extended to allow setting of the input `id` by the caller.  Previously "${name}_$value" was used.  The values set by the author could make those invalid id's.

Both that method and the NAMED_ANS_RADIO method also needed to be extended to allow the caller to add attributes to the input.  This is to that data attributes could be added that are used by the javascript.  Previosly this macro used a regular expression substitution to add an `onclick` handler.  Really what is needed is for the PGbasicmacros.pl macro to have a generic method to generate html tags with attributes.

To make this easier to deal with a new `tag` method was added to PGbasicmacros.pl.  This method is similar in usage to the `tag` method of Mojolicious::Plugin::TagHelpers.  It is intended to extend this usage to more of the methods in PGbasicmacros.pl and elsewhere.  Note that it does not include the capability to open a tag without closing it, and never should be extended to do so.  That promotes bad practice and leads to invalid html.

### DIFF
--- a/htdocs/js/apps/RadioButtons/RadioButtons.js
+++ b/htdocs/js/apps/RadioButtons/RadioButtons.js
@@ -1,0 +1,59 @@
+// ################################################################################
+// # WeBWorK Online Homework Delivery System
+// # Copyright &copy; 2000-2022 The WeBWorK Project, https://github.com/openwebwork
+// #
+// # This program is free software; you can redistribute it and/or modify it under
+// # the terms of either: (a) the GNU General Public License as published by the
+// # Free Software Foundation; either version 2, or (at your option) any later
+// # version, or (b) the "Artistic License" which comes with this package.
+// #
+// # This program is distributed in the hope that it will be useful, but WITHOUT
+// # ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// # FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+// # Artistic License for more details.
+// ################################################################################
+'use strict';
+
+(() => {
+	// Setup uncheckable radios.
+	const setupUncheckableRadio = (radio) => {
+		if (!radio.dataset.uncheckableRadioButton) return;
+		delete radio.dataset.uncheckableRadioButton;
+
+		if (radio.checked) radio.dataset.currentlyChecked = '1';
+
+		radio.addEventListener('click', (e) => {
+			if (radio.dataset.shift && !e.shiftKey) {
+				radio.dataset.currentlyChecked = '1';
+				return;
+			}
+			const currentlyChecked = radio.dataset.currentlyChecked;
+			if (currentlyChecked) {
+				delete radio.dataset.currentlyChecked;
+				radio.checked = false;
+			} else {
+				radio.dataset.currentlyChecked = '1';
+			}
+		});
+	};
+
+	// Deal with uncheckable radios already in the page.
+	document.querySelectorAll('input[type="radio"]').forEach(setupUncheckableRadio);
+
+	// Deal with radios that are added to the page later.
+	const observer = new MutationObserver((mutationsList) => {
+		for (const mutation of mutationsList) {
+			for (const node of mutation.addedNodes) {
+				if (node instanceof Element) {
+					if (node.tagName.toLowerCase() === 'input' && node.type.toLowerCase() === 'radio')
+						setupUncheckableRadio(node);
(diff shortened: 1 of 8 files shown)


drgrice1:
Some examples demonstrating the usage of this new capability are attached.
[examples.zip](https://github.com/openwebwork/pg/files/10775499/examples.zip)

drdrew42:
Thanks for including test problems, that's incredibly useful!

Just a question regarding the formatting of `rma-with-radio-sub.pg`
is the sub-part of answer A supposed to be indented?

Screen Shot 2023-02-22 at 5 28 34 PM

drgrice1:
Yes, it should be indented.  Check that your static assets are up to date by running `npm install` or `./generate-assets` in `pg/htdocs`.  The css generated by that has effect here.

drdrew42:
OMG of course.

drdrew42:
This works great, but there is some unexpected re-sorting of responses in parserCheckboxList

drgrice1:
@drdrew42:  So it turns out that the sorting is in fact all about the null byte hack.  The problem is the hack (it is even called that) on lines 989 - 999 of `lib/WeBWorK/PG/Translator.pm` that is specifically for check boxes.

drgrice1:
@drdrew42:  The order is now fixed.  This required changing how extended response groups are stored.  Instead of using a hash, an array is used.  The translator hack needed to be modified to accommodate this.  This removes one small part of the null byte hack.  Unfortunately answers are still passed in as null byte separated strings, and end up being converted back to that later.  In the middle there is no reason for that.

drgrice1:
Actually, it turns out this is a much bigger part of removing the null byte hack than I thought.  The rest is on the webwork2 (and standalone renderer) side of things.  If you drop the null byte joining done by Form.pm, and then add a little joining for the past answers table then all of the null byte stuff is gone.  Well, aside from a lot of unnecessary code in PG that can be deleted.

pstaabp (approved):

PR #779, opened 2023-02-20
Make it possible to pass options to the cmp method with PGML

drgrice1:
Currently passing options to the cmp method using PGML answer methods is not possible.  So instead you have to resort to using the basic PG approach of calling ans_rule or ans_array, and then calling the cmp method yourself.  For example to use a custom checker with a matrix object you must do something like:
```
$ans = Matrix([ [ 1, 2, 3 ], [ 4, 5, 6 ] ]);

BEGIN_PGML
Matrix answer: [@ $ans->ans_array @]*
END_PGML

ANS($ans->cmp(
	checker => sub { ... }
));

```

This pull request makes it so that you can pass those options using the PGML answer syntax.  This can be done for example with:
```
$ans = Matrix([ [ 1, 2, 3 ], [ 4, 5, 6 ] ]);
$ans_cmp_opts = {
	checker => sub { ... }
};

BEGIN_PGML
Matrix answer: [_]*{answer=>$ans, cmp_options=>$ans_cmp_opts}
END_PGML
```
You can also use: `Matrix answer: [_]*{$ans}{5}{NEW_ANS_NAME()}{$ans_cmp_opts}`.  Unfortunately, the order is answer, width, name, cmp_options, and that can't be changed for compatibility reasons.  So you  have to pass an answer name with this approach which should always be `NEW_ANS_NAME()` or a scalar variable whose value is the result of calling that method.

This new option is most useful for using a custom checker with a multitude of answer types like matrices (as demonstrated above), parserRadioButtons.pl answers, and I am sure there are many others.  I have implemented `cmpOptions` for some macros like `parserGraphTool.pl` and `draggableProof.pl` (in #774), but this gives a much better way that doesn't need to be implemented for each MathObject value type answer.

Edit:  Now you can also add space to the begining when using the hash options.  For example, `[_]*{ answer => $ans, width => 5, cmp_options => $ans_cmp_opts }` or `[_]*{$ans}{5}{ cmp_options => $ans_cmp_opts }`.

### DIFF
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -779,7 +779,7 @@ sub terminateOptions {
 	$self->{block} = $self->{block}{prev};
 	$self->{block}->popItem;
 	$block = $self->{block}->topItem;
-	if ($options =~ m/^[a-z_][a-z0-9_]*=>/i) {
+	if ($options =~ m/^\s*[a-z_][a-z0-9_]*\s*=>/i) {
 		my %allowed = (map { $_ => 1 } (@{ $block->{options} }));
 		my ($options, $error) = PGML::Eval("{$options}");
 		$options = {}, PGML::Warning "Error evaluating options: $error" if $error;
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -1270,7 +1270,7 @@ sub Answer {
 	my $rule;
 	$item->{width} = length($item->{token}) - 2 if (!defined($item->{width}));
 	if (defined($ans)) {
-		if (ref($ans) =~ /CODE|AnswerEvaluator/) {
+		if (ref($ans) eq 'CODE' || (ref($ans) eq 'AnswerEvaluator' && !Value::isValue($ans->{rh_ans}{correct_value}))) {
 			if (defined($item->{name})) {
 				$rule = main::NAMED_ANS_RULE($item->{name}, $item->{width});
 				main::NAMED_ANS($item->{name} => $ans);
@@ -1279,6 +1279,8 @@ sub Answer {
 				main::ANS($ans);
 			}
 		} else {
+			$ans = $ans->{rh_ans}{correct_value} if ref($ans) eq 'AnswerEvaluator';
+
 			unless (Value::isValue($ans)) {
 				$ans = Parser::Formula($item->{answer});
 				if (defined($ans)) {
@@ -1304,10 +1306,12 @@ sub Answer {
 			$rule = $ans->$method(@options);
 			$rule = PGML::LaTeX($rule);
 			if (!(ref($ans) eq 'parser::MultiAnswer' && $ans->{part} > 1)) {
+				my @cmp =
+					ref($item->{answer}) eq 'AnswerEvaluator' ? $item->{answer} : $ans->cmp(%{ $item->{cmp_options} });
 				if (defined($item->{name})) {
-					main::NAMED_ANS($item->{name} => $ans->cmp(%{ $item->{cmp_options} }));
+					main::NAMED_ANS($item->{name} => @cmp);
 				} else {
-					main::ANS($ans->cmp(%{ $item->{cmp_options} }));
+					main::ANS(@cmp);
 				}
 			}
 		}
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -372,7 +372,7 @@ sub Verbatim {
 sub Answer {
 	my $self  = shift;
 	my $token = shift;
-	my $def   = { options => [ "answer", "width", "name", "array" ] };
+	my $def   = { options => [ "answer", "width", "name", "cmp_options" ] };
 	$def->{hasStar} = 1 if $token =~ m/\*$/;
 	$self->Item("answer", $token, $def);
 }
@@ -1305,9 +1305,9 @@ sub Answer {
 			$rule = PGML::LaTeX($rule);
 			if (!(ref($ans) eq 'parser::MultiAnswer' && $ans->{part} > 1)) {
 				if (defined($item->{name})) {
-					main::NAMED_ANS($item->{name} => $ans->cmp);
+					main::NAMED_ANS($item->{name} => $ans->cmp(%{ $item->{cmp_options} }));
 				} else {
-					main::ANS($ans->cmp);
+					main::ANS($ans->cmp(%{ $item->{cmp_options} }));
 				}
 			}
 		}


Alex-Jordan:
> So instead you have to resort to using the basic PG approach of calling ans_rule or ans_array, and then calling the cmp method yourself.

This is specific to answer inputs arrays though, right? With the asterisk?

For regular answer arrays, I have always gotten away with like:

```
$ans = Real(1);
$cmp = $ans->cmp(checker => sub { ... });

BEGIN_PGML
Answer: [_]*{$cmp}
END_PGML
```

And the custom checker works. In fact I just tested that I can do this:

```
Context("Matrix");

$ans = Matrix([ [ 1, 2 ], [ 3, 4] ]);
$cmp = $ans->cmp(checker => sub {return 1;});

$realans = Real(1);
$realcmp = $realans->cmp(checker => sub {return 1;});


BEGIN_PGML
Matrix answer: [_]*{$cmp}

Real answer: [_]{$realcmp}
END_PGML

ENDDOCUMENT();
```

And both custom answer checkers are in play (always awarding credit). It's just that the star for making an array of inputs is not being honored.

Alex-Jordan:
So I think your improvement is good. And also there may be a bug where the logic to determine array dimensions that goes into effect when `[_]*` is used is not considering that the argument may be a stored `cmp()` routine instead of a MathObject.

drgrice1:
This is not specific to array answers.  I mentioned parserRadioButtons.pl.  Those are not array answers, and this works for those.  Currently you can not do
```
$radio = RadioButtons(...);
$cmp = $radio->cmp(...);
BEGIN_PGML
[_]{$cmp}
END_PGML
```
That does not output the radio buttons.

The point is that any object that does some sort of special output for its ans_rule method does not work without this pull request. The examples you demonstrate are all cases where no special output is needed.

drgrice1:
> So I think your improvement is good. And also there may be a bug where the logic to determine array dimensions that goes into effect when `[_]*` is used is not considering that the argument may be a stored `cmp()` routine instead of a MathObject.

I am not sure what you mean by this?  Are you saying this pull request causes a bug?  Or there is an existing bug?

drgrice1:
I think I understand your comment now.  Are you saying you think that it is a bug that array answers don't work by passing the result of the  `cmp` method as it is?  Unfortunately, that is not a bug.  That is simply not possible.  When the result of the `cmp` method is passed, PGML has no way of getting the original `Value` object that has the `ans_array` method needed to output the array answer.  The result of the `cmp` method is an `AnswerEvaluator` object which does not give access to the original answer object, and does not have an `ans_rule` or `ans_array` method.

Alex-Jordan:
I'm saying I think it is a current bug that with my example, I do not see an answer array even though I used `[_]*` with an answer that is a matrix. (I acknowledge that I didn't pass the actual matrix; I passed its `cmp()` routine.)

Maybe it is wrong to call it a bug. Maybe `[_]{}` was never meant to be passed a `cmp()` routine in the braces. It's what I have done for a long time now, following a recommendation from Davide iirc.

So while this PR gives a good way to pass options to `cmp()` in PGML, I'd like to also make my example work as expected and produce an array of answer inputs. Maybe this has nothing to do with PGML, and something like `ans_array()` could benefit from inspecting if the argument is already a cmp() routine.

Alex-Jordan:
The `AnswerEvaluator` object knows what the correct answer is though. So if `ans_array()` is called from an AnswerEvaluator object, the code in `ans_array()` should be able to work out what dimensions to use.

Or so my thinking goes without having looked at any code yet.

drgrice1:
I think that you are misunderstanding how this works.  The correct answer is not the same as the original object.  Matrix objects have a special ans_array method.  They don't just use the generic `ans_array` method in PGbasicmacros.pl.

Also, clearly PGML is intended to accept an `AnswerEvaluator` object (or even a CODE reference) via `[_]{$cmp}` or the code wouldn't be there to handle it.  However, that is not designed to work with `ans_array`.

drgrice1:
By the way, this is all handled by the `Answer` method of the `PGML::Format` package in `PGML.pl`.  So that is the place to look to see how this all works.

drgrice1:
Also see the `ans_array` method of the `Value::Matrix` package in `lib/Value/AnswerChecker.pm` to see the special handling that is done to output matrices in array format.

dpvc:
I think Alex's request actually is possible.  If `$ans` is an `AnswerEvaluator` created by a MathObject, then `$ans->{rh_ans}{correct_value}` should be the original MathObject that created the evaluator, so you could check if that exists, and if so, use its `ans_rule` or `ans_array` method to create the answer blank or answer array, and still use `main::ANS($ans)` (or the named version of that) to attach the evaluator to the blank(s).  One could refactor some of the code for handling a MathObject so that it could be reused for both situations, but it looks like it would be doable.

drgrice1:
That should work for `Matrix` objects, but wouldn't work for all `MathObject` answers.  The `RadioButtons` object and many that use similar techniques do not save themselves as the `$ans->{rh_ans}{correct_value}`.

drgrice1:
Well, I take that back.  I guess it would work for `RadioButtons`.

drgrice1:
It seems that works with the `GraphTool` object created by my `parserGraphTool.pl` macro as well.  I will look into implementing that, unless you want to do it @dpvc.

dpvc:
> Well, I take that back. I guess it would work for `RadioButtons`.

Yes, it is part of the `cmp()` method for the MathObject base class, and so any MathObject subclass will inherit it automatically, unless they override the `cmp()` method and don't call the `SUPER::cmp()` method.  It looks like only 5 files in the `pg/macros` directory override `cmp()` and all but `parserMultiAnswer.pl` call the super method.  MultiAnswer's `cmp()` method returns an ARRY of answer evaluators, so couldn't be used as in Alex's example anyway, and are handled specially by PGML in any case.

> I will look into implementing that, unless you want to do it

I don't have time to work on it at the moment, so be my guest.

drgrice1:
Are you looking at the macros in the develop branch?  I am seeing six macros that override the `cmp` method.  Those are parserGraphTool.pl, parserRadioButtons.pl, draggableProof.pl, draggableSubsets.pl, parserMultiAnswer.pl, and parserRadioMultiAnswer.pl.  Furthermore, parserGraphTool.pl and parserRadioButtons.pl are the only ones that call the `SUPER::cmp` method.  Although draggableProof.pl and draggableSubsets.pl will after #774.  parserRadioMultiAnswer.pl works similar to parserMultiAnswer.pl and creates a custom `AnswerEvaluator` object, and so would need special handling much like parserMultiAnswer.pl.  Although it has the `cmpOpts` option that allows for doing this sort of thing.

drgrice1:
I have added the change to make the `[_]*{$cmp}` approach work as desired.  So now you can do
```
$radio = RadioButtons(...);
$cmp = $radio->cmp(...);
BEGIN_PGML
[_]{$cmp}
END_PGML
```
or
```
$ans = Matrix(...);
$cmp = $ans->cmp(...);
BEGIN_PGML
[_]*{$cmp}
END_PGML
```

drgrice1:
I left the `cmp_options` that I originally added in this pull request as an alternate approach that can be used.  That may still be useful.

dpvc:
> Are you looking at the macros in the develop branch?

Yes, but not fully up to date. I didn't have the `parserRadioMultiAnswer.pl` file.  

> Furthermore, parserGraphTool.pl and parserRadioButtons.pl are the only ones that call the `SUPER::cmp` method

Yes, I admit to not looking ad the draggable files very closely (I was out of time), but they seemed to call other objects' `cmp()` methods, so I counted that as getting the base `cmp()` functionality.

On a quick scan of your recent changes, I don't see anything wrong, so I approve the changes.

drgrice1:
Yeah, draggableProof.pl and draggableSubset.pl don't technically override the Value `cmp` method, because they are not Value objects.  Although they will be after #774.

somiaj:
One nice thing I can see about being able to use `$cmp` AnswerEvaluator directly is being able to apply Pre/Post filters to the object, which cannot be done by just sending options to cmp(). Would it be worth it to extend your new cmp_options additions to maybe accept filters as well? Though might not be worth the effort since you have updated things to use the AnswerEvaluator directly.

drgrice1:
I don't think that would be worth it.  That can be done using the AnswerEvaluator directly, and would be a mess to try to do as an option like the `cmp_options` approach.

drdrew42:
Just to clarify here before testing, in addition to sequentially providing parameters:
`Matrix answer: [_]*{$ans}{5}{NEW_ANS_NAME()}{$ans_cmp_opts}`, 
it is possible to instead provide a labeled hash: 
`Matrix answer: [_]*{answer => $ans, width => 5, name => NEW_ANS_NAME(), cmp_options => $ans_cmp_opts}`, 
or eschewing all others -- store the cmp code in a scalar and provide it as:
`Matrix answer: [_]*{$cmp}`

Alex-Jordan:
Will PGML do any parsing/processing of the text inside curly braces of these:

```
[_]*{$ans}{5}{NEW_ANS_NAME()}{$ans_cmp_opts}
```

```
[_]*{answer => $ans, width => 5, name => NEW_ANS_NAME(), cmp_options => $ans_cmp_opts}
```


Like if you had:
```
[_]*{answer=>$ans, cmp_options=>{...}}
```
with brackets inside the `...` (either capping some array reference, or as part of some string), will it work as expected?

drgrice1:
> 

Yes. All of those work. Also I want to clarify that none of those ways of passing the answer are new. The only new thing is cmp_options and that when an AnswerEvaluator object is passed for a Value object, the Value object ans_rule or ans_array method is used.

dpvc:
Note also that you can also combine the forms, as in

```
[_]{$ans}{cmp_options => {...}}
```

rather than having to use

```
[_]{answer => $ans, cmp_options => {...}}
```

or

```
[_]{$ans}{5}{cmp_options => {...}}
```

rather than

```
[_]{answer => $ans, width => 5, cmp_options => {...}}
```

In particular, one need not worry about `name => NEW_ANS_NAME()` in any of these cases.

drgrice1:
I didn't realize you could mix the methods like that.  I see how it works in the code though.  They are parsed as separate `options` blocks.  Nice.

drgrice1:
By the way, you do need to be careful with where you put spaces when you are using the hash method to pass the answer parameters.  If you do
`[_]*{ answer => $answer, width => 5, cmp_options => { checker => $checker } }`
or
`[_]*{answer => $answer, width => 5, cmp_options => { checker => $checker } }`
it will not work.  However if you do
`[_]*{answer=> $answer, width => 5, cmp_options => { checker => $checker } }`
then it will work.

The important things are that there is no space after the brace that starts an option block, and there is no space in between the first hash key given and the `=>`.  After that spaces are fine.

@dpvc:  Would you mind a tweak to the regular expression on line 782 of PGML.pl in the `terminateOptions` method to allow spaces in the above mentioned places?  Or would that cause other problems?  In my testing it seems fine, and I can't think of any problems it would cause.

drgrice1:
The tweak specifically is to change that line from
`if ($options =~ m/^[a-z_][a-z0-9_]*=>/i) {`
to
`if ($options =~ m/^\s*[a-z_][a-z0-9_]*\s*=>/i) {`

dpvc:
@drgrice1, I don't have any problem with that.

drgrice1:
@dpvc:  Cool.  I have added that.

pstaabp (approved):
This is a nice change.

drdrew42 (approved):
👍

PR #780, opened 2023-02-22
Remove CRLF line endings when a PG file is loaded.

drgrice1:
Currently a pg file that has CRLF line endings fails to render.  This fixes that by removing such line endings in the default_preprocess_code method.

An alternate approach that would also work would be to do this when the translator opens a file in the source_file method.  However, this would not catch cases when a source string that contains CRLF line endings is provided instead of a source file.  That may not be a real concern though?

I am not exactly sure what changed that caused files with CRLF line endings to fail.  They work for PG 2.17, but do not for the develop branch.

### DIFF
--- a/lib/WeBWorK/PG/Translator.pm
+++ b/lib/WeBWorK/PG/Translator.pm
@@ -1403,6 +1403,8 @@ sub PG_answer_eval {
 sub default_preprocess_code {
 	my $evalString = shift // '';
 
+	$evalString =~ s/\r\n/\n/g;
+
 	# BEGIN_TEXT, END_TEXT, and the others that follow must occur on a line by themselves.
 	$evalString =~ s/\n\h*END_TEXT[\h;]*\n/\nEND_TEXT\n/g;
 	$evalString =~ s/\n\h*END_PGML[\h;]*\n/\nEND_PGML\n/g;


drgrice1:
Attached is a PG problem with CRLF line endings for testing.
[crlf-file.pg.txt](https://github.com/openwebwork/pg/files/10804447/crlf-file.pg.txt)

drdrew42:
The re-work of WeBWorK::PG::Local caused this issue.

Originally, this was the code:
```perl
	if (ref($translationOptions->{r_source}) ) {
		# the source for the problem is already given to us as a reference to a string
		$source = ${$translationOptions->{r_source}};
	} else {
	    # the source isn't given to us so we need to read it
	    # from a file defined by the problem

		# we  grab the sourceFilePath from the problem
		$sourceFilePath = $problem->source_file;

	    # the path to the source file is usually given relative to the
	    # the templates directory. Unless the path starts with / assume
	    # that it is relative to the templates directory

	    $sourceFilePath = $ce->{courseDirs}->{templates}."/"
	    	         .$sourceFilePath unless ($sourceFilePath =~ /^\//);
	    #now grab the source
		eval {$source = readFile($sourceFilePath) };
		$readErrors = $@ if $@;
	 }
```

The specific call is `eval {$source = readFile($sourceFilePath) };`, where `readFile` replaced CRLF with LF and `readFile` is provided by WeBWorK::Utils

Now, in develop, WeBWorK::PG line 122 `eval { $translator->source_file($sourceFilePath) };` reads the file and stores the result directly in the translator hash without replacing CRLF.

somiaj:
Note, files with CRLF didn't fully work in 2.17. They did render find in the problem viewer, but they don't render in the library browser (and other places such as the set editor which render problems similar to the library browser) and just return, "There was an error rendering this problem!". I haven't had time to test if this fixes that issue in develop as well.

drdrew42 (approved):
Agreed, this is definitely where the find/replace for pg files belongs... whenever a problem is processed, the preprocessor will clean it up.

pstaabp (approved):

PR #783, opened 2023-03-05
Lock the MathQuill version installed by npm (hotfix).

drgrice1:
With breaking changes coming to MathQuill, PG-2.17 needs to be locked into using the current version of MathQuill.  This locks PG-2.17 to the WeBWorK-2.17 tag in our MathQuill repository.

Obviously this needs to be a hotfix.  It can't be anything else.


drgrice1:
I guess this really isn't needed if we make sure that everyone starts using `npm ci` instead of `npm install`.  The package-lock.json file is used with `npm ci` and the sha hash in the package-lock.json file for PG-2.17 is exactly for the commit tagged by the tag added in this commit.

PR #784, opened 2023-03-08
Update mqeditor for recent mathquill updates

drgrice1:
This updates the mqeditor to work with changes to MathQuill.  See [openwebwork/mathquill](https://github.com/openwebwork/mathquill).

Make sure to run `npm install` to test this.  Don't just run `./generate-assets` or `npm run generate-assets`.  In fact, you don't even need to run `./generate-assets`.  So you can run `npm install --ignore-scripts` to just install the updated MathQuill library.

Also note that you will need to do a hard refresh on a page that loads MathQuill in order to make the new MathQuill javascript load.  This is something we need to think about for javascript updates for those things installed by npm in general.  In most cases updated third party libraries use the same file names.

### DIFF
--- a/htdocs/js/apps/MathQuill/mqeditor.js
+++ b/htdocs/js/apps/MathQuill/mqeditor.js
@@ -35,8 +35,11 @@
 			restrictMismatchedBrackets: true,
 			sumStartsWithNEquals: true,
 			supSubsRequireOperand: true,
-			autoCommands: 'pi sqrt root vert inf union abs',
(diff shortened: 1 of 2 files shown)


pstaabp:
Getting a strange error on this.  On any problem, the first answer box is rendering as a span, and the rest are regular inputs.  No mathquill is coming up.  Here's a screenshot:
image

I did `npm install`.

drgrice1:
This pull request is now ready to go.

pstaabp (approved):

PR #785, opened 2023-03-08
Add unicode degree units.

drgrice1:
Add °C, °F, °K (shouldn't that just be K), and ° as alternate units for degC, degF, degK (again should this just be K), and deg.

Note that the correct SI unit for Kelvin is just K not °K (since 1967).  So shouldn't K be added for that?

Also add the plural forms of angstrom.

The unit tests are updated to handle these units.

Edit:  There are some additional changes now.

A basic alias system for units is now implemented.  Aliases can be added for a unit via the 'aliases' key.  The value should
be an array of aliases.  These are added as keys to the %known_units hash and the 'aliases' key deleted at compile time.

Also fix some perlcritic issues in the Units.pm, and clean it up a bit. 

The "degree Kelvin" units have been removed from the help files, but not from the actual units allowed (for now).

Also change Centrigrade to Celsius in the help files.  What is a Centrigrade?  Also, a Centigrade is not correct either.  It has been Celsius since 1948.

### DIFF
--- a/lib/Units.pm
+++ b/lib/Units.pm
@@ -18,20 +18,6 @@ use utf8;
 # there can be only one / in a unit.
 # powers can be negative integers as well as positive integers.
 
-# These subroutines return a unit hash.
-# A unit hash has the entries
-#      factor => number   number can be any real number
-#      m      => power    power is a signed integer
-#      kg     => power
-#      s      => power
-#      rad    => power
-#      degC   => power
-#      degF   => power
-#      degK   => power
-#      mol	  => power
-#	   amp	  => power
-#	   cd	  => power
-
 # Unfortunately there will be no automatic conversion between the different
 # temperature scales since we haven't allowed for affine conversions.
 
@@ -43,7 +29,7 @@ our %fundamental_units = (
 	'rad'    => 0,
 	'degC'   => 0,
 	'degF'   => 0,
-	'degK'   => 0,
+	'K'      => 0,
 	'mol'    => 0,    # moles, treated as a fundamental unit
 	'amp'    => 0,
 	'cd'     => 0,    # candela, SI unit of luminous intensity
@@ -77,7 +63,11 @@ our %known_units = (
 		'factor' => 1,
 		'degC'   => 1
 	},
-	'°C' => {    # unicode "\x{00B0}C"
+	"\x{00B0}C" => {    # Unicode degree symbol with text C
+		'factor' => 1,
+		'degC'   => 1
+	},
+	"\x{2103}" => {     # Unicode degree Celsius symbol
 		'factor' => 1,
 		'degC'   => 1
 	},
@@ -85,18 +75,29 @@ our %known_units = (
 		'factor' => 1,
 		'degF'   => 1
 	},
-	'°F' => {    # unicode "\x{00B0}F"
+	"\x{00B0}F" => {    # Unicode degree symbol with text F
 		'factor' => 1,
 		'degF'   => 1
 	},
-	# FIXME: Shouldn't this just be K?
+	"\x{2109}" => {     # Unicode degree Fahrenheit symbol
+		'factor' => 1,
+		'degF'   => 1
+	},
+	'K' => {
+		'factor' => 1,
+		'K'      => 1
+	},
 	'degK' => {
 		'factor' => 1,
-		'degK'   => 1
+		'K'      => 1
 	},
-	'°K' => {    # unicode "\x{00B0}K"
+	"\x{00B0}K" => {    # Unicode degree symbol with text K
 		'factor' => 1,
-		'degK'   => 1
+		'K'      => 1
+	},
+	"\x{212A}" => {     # Unicode Kelvin symbol
+		'factor' => 1,
+		'K'      => 1
 	},
 	'mol' => {
 		'factor' => 1,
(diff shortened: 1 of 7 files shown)


Alex-Jordan:
Do you want to also have?
https://www.compart.com/en/unicode/U+2103
https://www.compart.com/en/unicode/U+2109

Who knows where students copy-paste from.

A agree that it is just "K". Probably students will often write it with the degree symbol though, so it seems helpful to have both.

drgrice1:
I have one question on how this should be in the code.  When @duffee implemented the unicode for Angstroms he used the actual unicode character for the unit key, and added the unicode perl string in a comment.  Should the unicode perl string be used directly?  As to functionality for PG, they both work the same.  However, I have noticed display issues in my text editor in some cases with using the literal unicode character (due to monospace font usage).  In fact this occurs with ℃ (`\x{2103}` in perl).  Since that character occupies more width than other characters in the monospace font the end `'` is shown over the end of the character, and looks odd.  When you focus that character it shows it over the ending `'`.

Alex-Jordan:
I think to avoid confusion, it would be best to escape all non-ASCII characters. There is an even worse issue that can happen, if you open the file with an editor/viewer where the font simply has nothing to show for that character.

drgrice1:
I agree. I will change to that.

lib/Units.pm
```diff
@@ -43,7 +43,7 @@ our %fundamental_units = (
 	'rad'    => 0,
 	'degC'   => 0,
 	'degF'   => 0,
-	'degK'   => 0,
+	'K'      => 0,
```
Alex-Jordan, line 46: Do you want to make the corresponding change in the comments above this? (Or maybe just remove that comment block?)

Alex-Jordan:
I may be getting carried away, but there are many places that could use [µ](https://www.compart.com/en/unicode/U+00B5) and [Ω](https://www.compart.com/en/unicode/U+2126).

I think the main thing is to ask if a student might use these for input. I'm not sure.

lib/Units.pm
drgrice1: Thanks for pointing that out.  I didn't see that.  I decided to delete that comment block.  I don't see a need to have the code also in a comment block.

drgrice1:
µ and Ω could certainly come up.  This could get really involved though.

There is microfarad (uF), which is µF (unicode mu and letter F) or ㎌ (U+338C).

There is ohm, which is is Ω.

We also have kohm (kilo-ohm), which could be kΩ (letter k and unicode omega) or ㏀ (unicode U+33C0).

The there is Mohm (mega-ohm).  I didn't Google this, but I bet there is a unicode for this too.

lib/Units.pm
```diff
@@ -77,13 +63,41 @@ our %known_units = (
...
+	},
 	'degK' => {
 		'factor' => 1,
-		'degK'   => 1
+		'K'      => 1
+	},
+	"\x{00B0}K" => {    # Unicode degree symbol with text K
+		'factor' => 1,
+		'K'      => 1
+	},
```
duffee, line 97: I strongly disagree with allowing students to use the incorrect unit.  It stores up problems for their futures.

drgrice1: I don't disagree with you.

Alex-Jordan:
For now, we could limit the unicode additions to angstrom and temperatures.

Later, I think we need some sort of alias system. So instead of five things for angstrom, there would only be one, and the others would all point to Å for example. And then we replace "um" with "µm" and have "um" as an alias. ㏀ could be an alias for "kΩ", etc.

duffee:
My long desire was for displaying µ and Ω, so that I could use `\($resistance\)` in the problem text where `$resistance` is a `NumberWithUnits` and not have to worry about the unit.  My preference for input is `u` and `ohm` because I'm a keyboard person and I'm thinking of the student working on her phone at the bus stop.  Nothing against those that can greek up their input, I'm just not that good. ;)

I'm not a big fan of ㏀ for display as it's hard to read.

Alex-Jordan:
An alias system would let input (either from the problem author or from the student) such as "um", "ohm", and even "㏀" feed into the MathObject displaying itself using "µm", "Ω", and "kΩ".

For the student side, I expect the special characters to come less from their keyboard, and more from copy-pasting from some web page they are referencing. (And that web page may be using a font where "㏀" is more readable.)

So I'm in support of developing an alias system, but I am not sure how large of a project it might be. If it's too large, the long-hoped-for incorporation of unit parsing into MathObjects directly might be a better use of our time.

lib/Units.pm
```diff
@@ -1,998 +1,789 @@
...
+		m      => 3
+	},
+	cup => {
+		factor => 0.000236588,
+		m      => 3
+	},
+	cups => {
+		factor => 0.000236588,
+		m      => 3
+	},
```
duffee, line 237: You probably want to make `cups` an alias of `cup`

lib/Units.pm
```diff
@@ -1,998 +1,789 @@
...
+		s      => -2
+	},
+	bar => {
+		factor => 100000,
+		m      => -1,
+		kg     => 1,
+		s      => -2
+	},
+	mbar => {    # megabar
+		factor => 100,
```
duffee, line 502: comment for `mbar` should be millibar, not megabit.  factor is correct.

lib/Units.pm
drgrice1: I missed `cups` and `dalton`.  Thanks for pointing those out.

Yeah, unnecessary quoting with fat commas bugs me.  I couldn't leave it.

lib/Units.pm
drgrice1: Thanks for looking this over and finding these things.  I have fixed that mistake.

PR #786, opened 2023-03-12
Fix a possible error that can occur when the MathQuill toolbar is removed.

drgrice1:
This the 'transitionend' event can occur multiple times, which can result in the toolbar tooltips being disposed multiple times as well. This causes an error that is shown in the console.

This can be rather easily reproduced by setting focus to a MathQuill answer box and then holding down the tab button, letting focus cycle through the toolbar and beyond.


pstaabp (approved):
fixes the error

PR #787, opened 2023-03-13
Plotly3D code improvements.

somiaj:
Add a local copy of plotly.js version 2.18.2 to the collection of node_modules installed on the server to be able to host the javascript locally instead of using a CDN.

Use PG->getUniqueName('plotly3D') to get a unique string for the variable names instead of relying on incrementing the variables based on the number of graphs used. This allows this to avoid conflicts in tests or pages that have more than a single problem on the screen.

### DIFF
--- a/macros/graph/plotly3D.pl
+++ b/macros/graph/plotly3D.pl
@@ -3,7 +3,7 @@
 
 plotly3D.pl - Adds Graph3D, an object for creating 3D parametric curves
 and 3D parametric surface plots using the plotly JavaScript library.
-https://plotly.com/javascript/
+L<https://plotly.com/javascript/>
 
 =head1 DESCRIPTION
 
@@ -11,7 +11,8 @@ Loading this macro adds the Graph3D method which creates a 3D
 graph object. The graph object can be configured by a list of
 options of the form "option => value" (see below).
 
-    $graph = Graph3D(options);
+    loadMacros('plotly3D.pl');
+    $graph = Graph3D(options => value);
 
 Use the addCurve method to add a parametric curve to the graph.
 The following adds a helix to the graph. The first array is
@@ -123,14 +124,14 @@ of the points. The colorscale can be one of the following predefined names:
     'Greys', 'Greens', 'Electric', 'Earth', 'Bluered', or 'Blackbody'
 
 You can also define a custom colorscale as a list of colors for values
-ranging from 0 to 1. For example the default RdBu is the following colorset
-(note this must be a string since the array is passed to javascript):
+ranging from 0 to 1. For example the default RdBu is the following colorscale
+(note this must be a string since the array is passed to JavaScript):
 
     "[[0, 'rgb(5,10,172)'], [0.35, 'rgb(106,137,247)'],
       [0.5, 'rgb(190,190,190)'], [0.6, 'rgb(220,170,132)'],
       [0.7, 'rgb(230,145,90)'], [1, 'rgb(178,10,28)']]"
 
-A colorset can have any number of color points between 0 and 1. To make
+A colorscale can have any number of color points between 0 and 1. To make
 the plot a single color, set the color for 0 and 1 to be the same:
 
     "[[0, 'rgb(0,200,0)'], [1, 'rgb(0,200,0)']]"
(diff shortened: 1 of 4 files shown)


macros/graph/plotly3D.pl
```diff
@@ -191,21 +191,18 @@ =head1 Graph3D OPTIONS
 =cut
 
 sub _plotly3D_init {
-	ADD_JS_FILE('https://cdn.plot.ly/plotly-latest.min.js', 1);
+	ADD_JS_FILE('node_modules/plotly.js-dist-min/plotly.min.js', 0, { defer => undef });
```
drgrice1, line 194: Did you mean to add the `defer`ed attribute?  You didn't have that before.  That will mess things up if your javascript is not designed for that.  Although, your javascript should be designed for that if it is not.

macros/graph/plotly3D.pl
```diff
@@ -265,26 +262,25 @@ sub HTML {
...
-	my $count = 0;
 	my $scene = ($self->{scene}) ? "scene: { $self->{scene} }," : '';
 
 	foreach (@{ $self->{plots} }) {
-		$count++;
-		$plots .= $_->HTML($id, $count);
-		push(@data, "plotlyData${id}_$count");
+		my $thisId = $_->{id};
+		$plots .= $_->HTML;
+		push(@data, "plotlyData_$thisId");
```
drgrice1, line 270: There is no need for `$thisId`.  Just change this to:
```suggestion
		$plots .= $_->HTML;
		push(@data, "plotlyData_$_->{id}");
```

macros/graph/plotly3D.pl
somiaj: I forget why I didn't have it there, but following your examples it seemed to be preferred, and in my tests problem it seems to work just fine, so I added it back.

macros/graph/plotly3D.pl
somiaj: fixed.

macros/graph/plotly3D.pl
drgrice1: All javascript should be loaded deferred.  Either that or loaded at the end of the page body.  No javascript should be loaded in `head` unless it is deferred or async.

somiaj: Ahh, it might be my first version of this didn't use `document.addEventListener("DOMContentLoaded", function() {` to run the javascript when the page was loaded that was why I removed it. But now all my javascript is only run on page load.

drgrice1:
I added one more small change to my pull request to this branch.  That is I updated the method used for loading the javascript to work for the webwork3 rendering approach in which the javascript runs after the page has loaded.  As such the DOMContentLoaded event has already run and the plotly javascript will also have already been loaded.  So for that the initialization method in this macro needs to just run.

drgrice1 (approved):
Looks good.  Thanks for making the requested changes.

pstaabp:
Just for documentation, these are additional improvements to #728

pstaabp:
Heads up, but this will need an `npm ci` and not just a `generate-assets.js` for testing (and after this is merged into develop) since there is a new js package added.

pstaabp:
After doing the above things, including clearing the cache, I'm getting the following plot for the example you posted on the #728:

image

It appears that plotly.min.js is loading correctly, but doesn't seem to be rendering the plot.

drgrice1:
@pstaabp:  Which example from #728 are you using?  You aren't using the example in the original comment are you?  Make sure you are using the updated example there.

drgrice1:
Scroll down, and use the last one in the comments there.

pstaabp:
Of course I used an old problem.  The updated one at the bottom of #728 works great.

macros/graph/plotly3D.pl
```diff
@@ -11,6 +11,7 @@ =head1 DESCRIPTION
 graph object. The graph object can be configured by a list of
```
pstaabp, line 11: Add the link above on line 6: `L<https://plotly.com/javascript>`

macros/graph/plotly3D.pl
```diff
@@ -144,21 +145,73 @@ =head1 FUNCTION TYPES
...
+          );
+
+          This plots a surfacing using the points (x1,y1,z1), (x2,y2,z2), ...,
+          and (x9,y9,z9). The addSurface method requires bounds, but they are not
+          used, so [0,0] needs to be included, but is ignored. Using the perl method
+          to first generate the arrays, then copying the result and using the data
+          method can be useful to speed up rendering of nonrandomized plots.
+
 
 =head1 Graph3D OPTIONS
```
pstaabp, line 217: It would look a little better if the line below this (218) used monospace 
`C<$graph = Graph3D(options =E<gt> value)>`

somiaj: What tool do you use to render the POD, so I can see how it is formatted in HTML?

drgrice1:
@somiaj:  Probably the best way is to run the `bin/dev_scripts/generate-ww-pg-pod.pl` script.  That generates the POD in the same way that the POD is generated for the wiki (as that is the script that generates the POD for the wiki).

somiaj:
Can't seem to get the script to work. I installed `libpod-parser-perl` (which caused the script to run without errors) and then ran:

```
./generate-ww-pg-pod.pl -v --pg-root=/opt/webwork/pg --webwork-root=/opt/webwork/webwork2 -o ~/POD
```

And it creates three files, `~/POD/index.html`, `~/POD/pg/index.html`, and `~/POD/webwork2/index.html`. Unsure what I'm missing to make it traverse through the files to build the html docs.

macros/graph/plotly3D.pl
```diff
@@ -184,28 +237,33 @@ =head1 Graph3D OPTIONS
...
+              the aspect ratio, and view range of the 3D axes. The scene is a string
+              which contains a JavaScript dictonary to pass to plotly. Example:
+
+              scene => 'aspectmode: "manual",'
+                       . 'aspectratio: {x: 1, y: 1, z: 1},'
+                       . 'xaxis: { range: [0,2] },'
+                       . 'yaxis: { range: [0,3] },'
+                       . 'zaxis: { range: [1,4] }'
+
+              See https://plotly.com/javascript/3d-axes/ for more examples.
```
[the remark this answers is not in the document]
somiaj: Thanks for the POD improvements.

drgrice1:
You don't need to pass the `--pg-root` and `--webwork-root`  options if the usual PG_ROOT and WEBWORK_ROOT environment variables are set.

I am not sure what is going wrong for you.  If I cut an paste the command line that you posted it runs fine, and generates all of the webwork2 and pg POD.

The only thing I can think of is to check that all of the perl modules are available that are used by the script.  Those are Pod::Usage, Getopt::Long, Pod::Simple::XHTML, File::Find, File::Path, IO::File, Pod::Find, File::Basename, and POSIX.

somiaj:
Okay, figured it out. The parser doesn't like symbolic links for `/opt/webwork/pg` and `/opt/webwork/webwork2`.

macros/graph/plotly3D.pl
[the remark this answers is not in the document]
somiaj: Using a link `L<https://plotly.com/javascript/3d-axes/>` doesn't appear to work here, since this is in a code block.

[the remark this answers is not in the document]
drgrice1: It probably shouldn't be in a code block.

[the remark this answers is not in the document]
drgrice1: All of the options should really be listed with `=item`.  This method that we commonly use of indenting to show text options is a shortcut, but is not the correct way to do this.  I am guilty of this as well.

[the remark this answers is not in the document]
somiaj: Thanks for the pointer.

pstaabp (approved):

PR #788, opened 2023-03-15
Nice tables

Alex-Jordan:
This makes it so all the CSS specifications can be hash references, like `captioncss => {color => 'blue'}` instead of `'color: blue;'`. The older way of using the strings still works.

And I updated the POD.

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -159,7 +159,9 @@ See below to override for an individual row.
 =head3 HTML output
 
 Each css property setting should be a hash reference.
-For example,  C<{font-family =E<gt> 'fantasy', text-decoration =E<gt> 'underline'}>.
+For example,  C<{'font-family' =E<gt> 'fantasy', color =E<gt> 'red'}>.
+If a key has a dash character, it needs to be in quotes. Alternatively,
+you may uses a javascript flavor of CSS key like C<{fontFamily =E<gt> 'fantasy'}>
 
 =over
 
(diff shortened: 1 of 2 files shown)


macros/ui/niceTables.pl
```diff
@@ -405,7 +361,44 @@ =head2 Options for ROWS
 
 =head2 Options for COLUMNS
 
-Column styling is handled indirectly for now, mostly through the texalignment option above.
+Column styling is handled indirectly for now, mostly through the C<texalignment> option above.
+
+=head2 Deprecations
+
```
pstaabp, line 367: I think it may be helpful to include the alternative here.  I'm guessing instead of `tablecss => 'font-family: fantasy; text-decoration: underline;'>` to use
```
tablecss => { fontFamily => 'fantasy', textDecoration => 'underline' }
```

pstaabp:
This is some other bugfixes/additions to #752

pstaabp:
@Alex-Jordan you mention deprecating `tex=>` on a cell (bottom of the POD).  You included it in your `niceTables.pg` test file.  Is this for legacy purposes?

macros/ui/niceTables.pl
Alex-Jordan: Up at line 161 describes the new way to to declare CSS. Did you see that there?

Alex-Jordan:
> @Alex-Jordan you mention deprecating tex=> on a cell (bottom of the POD). You included it in your niceTables.pg test file. Is this for legacy purposes?

Well, it's because I only did the official deprecation with a recent commit, and did not review the sampler file. I'll remove that from the sampler.

macros/ui/niceTables.pl
pstaabp: I did see that and know that you can put them in either way.  I read this as the camelCase version was the preferred method and if that is the case, we should use examples in the preferred way.  If it doesn't matter, let's leave it.

Alex-Jordan: Either `fontFamily` or `'font-family'` is OK to use. If I had to make a choice, I would support `'font-family'` because that is the actual CSS key. The other is just a convenience that is consistent with the js convention for referencing a CSS key. But yes, both are meant to be supported. And both are new at the same time, not being deprecated.

PR #789, opened 2023-03-20
Add random_subset macro to PGbasicmacros.pl.

somiaj:
Takes the function `NchooseK` from `PGchoicemacros.pl` or `List.pm` and add it to `PGbasicmacros.pl` with a new name `random_subset($n, @array)` which returns a randomly ordered array containing `$n` elements from the provided `@array`. Optionally, an array ref can be used instead, `random_subset($n, $array_ref)`.

This function is useful when wanting to get a random list of unique items from a given array or to reorder the array. Moving it to `PGbasicmacros.pl` places it with the other random functions and makes it available to all problems without having to load the full `PGchoicemacros.pl`. I find many uses for this function when randomizing problems, and this was also mentioned in #761, where it was suggested renaming it, since `NchooseK` is also `n!/(k! (n-k)!)` instead of a subset.

### DIFF
--- a/macros/core/PGauxiliaryFunctions.pl
+++ b/macros/core/PGauxiliaryFunctions.pl
@@ -15,12 +15,12 @@
 
 =head1 NAME
 
-A set of auxiliary functions that are often used in PG problems.  
+A set of auxiliary functions that are often used in PG problems.
 
 
 =head1 DESCRIPTION
 
-This macro creates the following functions that are available for PG: 
+This macro creates the following functions that are available for PG:
 
 	step($number)
 	ceil($number)
@@ -52,7 +52,8 @@ sub _PGauxiliaryFunctions_init {
 
 	Usage: step(x);
 
-returns the step function (or Heaviside function) with jump at x=0.  That is when x<0, it returns 0, when x>=0 the function returns 1. 
+returns the step function (or Heaviside function) with jump at x=0.  That is when x<0, it returns 0,
+when x>=0 the function returns 1.
 
 Example:
 
@@ -71,9 +72,9 @@ sub step {    # Heaviside function (1 or x>0)
 
 	Usage: ceil(x);
 
-returns the ceiling function of x.  This rounds up to the nearest integer.  
+returns the ceiling function of x.  This rounds up to the nearest integer.
 
-Examples: 
+Examples:
 
 	ceil(3.14159) returns 4
 	ceil(-9.75) return -9
@@ -91,9 +92,9 @@ sub ceil {
 
 	Usage: floor(x);
 
-returns the floor function of x.  This rounds down to the nearest integer.  
+returns the floor function of x.  This rounds down to the nearest integer.
 
-Examples: 
+Examples:
 
 	floor(3.14159) returns 3
 	floor(-9.75) return -10
@@ -107,13 +108,13 @@ sub floor {
 	$out;
 }
 
-=head3 max function 
+=head3 max function
 
 =pod
 
 	Usage: max(@arr);
 
-returns the maximum of the values in the array @arr. 
+returns the maximum of the values in the array @arr.
 
 Example
 
@@ -134,13 +135,13 @@ sub max {
 
 }
 
-=head3 min function 
+=head3 min function
 
 =pod
 
 	Usage: min(@arr);
 
-returns the minimum of the values in the array @arr. 
+returns the minimum of the values in the array @arr.
 
 Example
 
@@ -163,13 +164,13 @@ sub min {
 
 # round added 6/12/2000 by David Etlinger. Edited by AKP 3-6-03
 
-=head3 round function 
+=head3 round function
 
 =pod
 
 	Usage: round(x);
 
-returns integer nearest x. 
+returns integer nearest x.
 
 Example:
 
@@ -184,17 +185,17 @@ sub round {
 	$out;
 }
 
-=head3 Round function 
+=head3 Round function
 
 =pod
 
 	Usage: Round(x);
 
-returns integer nearest x.  
+returns integer nearest x.
 
 	Usage: Round(x,n);
 
-returns the number rounded to n digits.  
+returns the number rounded to n digits.
 
 Example:
 
@@ -208,19 +209,19 @@ sub Round {
 	elsif (@_ == 2) { $_[0] > 0 ? Round($_[0] * 10**$_[1]) / 10**$_[1] : Round($_[0] * 10**$_[1]) / 10**$_[1] }
 }
 
-=head3 lcm function 
+=head3 lcm function
 
 =pod
 
 	Usage: lcm(@arr);
 
-returns the lowest common multiple of the array @arr of integers.  
+returns the lowest common multiple of the array @arr of integers.
 
 Example:
 
-	 lcm(3,4,5,6) returns 60.
+	lcm(3,4,5,6) returns 60.
 
-Note: it checks for an empty array, however doesn't check if the inputs are integers. 
+Note: it checks for an empty array, however doesn't check if the inputs are integers.
 
 =cut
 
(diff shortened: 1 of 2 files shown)


macros/core/PGbasicmacros.pl
```diff
@@ -1300,6 +1305,30 @@ sub list_random {
...
+	if ($n > scalar(@set)) {
+		warn 'random_subset: The first input must be smaller than equal to the number of elements.';
+		$n = scalar(@set);
+	}
+
+	my @out;
+	while (@set && @out < $n) {
+		push(@out, splice(@set, random(0, $#set, 1), 1));
+	}
+	@out;
```
drgrice1, line 1329: This should be `return @out;`.  See the other comment.

macros/core/PGbasicmacros.pl
```diff
@@ -1300,6 +1305,30 @@ sub list_random {
 	return $li[ random(1, scalar(@li)) - 1 ];
 }
 
+sub random_subset {
+	my ($n, @set) = @_;
+	@set = @{ $set[0] } if (scalar(@set) == 1 && ref($set[0]) eq 'ARRAY');
+
+	unless ($n =~ /^\d+/ && $n > 0) {
+		warn 'random_subset: The first input must be a positive integer';
+		return ();
```
drgrice1, line 1314: There is no need for the parentheses here.  `return;` is equivalent to `return ();`.

drgrice1 (changes requested):
On initial glance I see that there are some perlcritic issues that need to be dealt with.  No reason to add new perlcritic issues to the massive list of issues already in pg.

Alex-Jordan:
Thanks, this is such a useful macro to have easy access to.

For a few reasons, I suggest allowing `$k` to be `0` and returning an empty array. The simpler reason is that the existing `NchooseK` function does this. Another reason is that I can imagine scenarios where `$k` is random and might be 0, and you wouldn't want to have to code separately for that case.

Alex-Jordan:
Sorry, I wrote `$k` as in "n choose k", but it is `$n` in the code here.

somiaj:
@Alex-Jordan I went back and forth on this, currently it does return an empty array if `$n = 0`, but it also gives a warning. I'll just remove the warning.

drgrice1:
I think @Alex-Jordan is right.  This would be better in PGauxiliaryFunctions.pl.  In fact, PGsort and its subsidiary routines should also be there.  @pstaabp put those in the wrong place, and that should have been caught when he moved that in from Translator.pm.  PGbasicmacros.pl is really for html generation methods, particularly answer rule creation.  That is what everything else in that file does with perhaps a few other out of place exceptions.  Other out of place exceptions include `random`, `non_zero_random`, and such.  Those should probably also be in PGauxiliaryFunctions, but it would probably be a bad idea to move those at this point since they have been there since 2003 (at least).

somiaj:
I see `random_subset` similar to `list_random` and it depends only on `random`, so it would make sense to have all these functions altogether in the same location. But we will break other things to move them all over (I see 493 problems in the OPL that contain `PGbasicfunctions.pl` but don't contain `PGauxiliaryFunctions.pl`).

Anyways, I'll move the function if that is the preferred consensus to put it in `PGauxiliaryFunctions.pl`, (update incoming).

somiaj:
I just noticed that `PGchoicemacros.pl` doesn't pull (or use) any other macro in `PGauxiliaryFunctions.pl`, and I think I saw some problems in the OPL that use `PGchoicemacros.pl` with only `PGbasicmacros.pl`. So though `NchooseK` and `shuffle` have been depreciated for a while, there is a small chance they won't work with this change.

To be safe should I add `loadMacros('PGauxiliaryFunctions.pl')` to `PGchoicemacros.pl`, or just leave the two functions there alone?

drgrice1:
At this point, as I said, it would probably be a bad idea to move `random`, although it would fit better in `PGauxiliaryFunctions.pl`.  Most problems use `random`, and I am sure many of them do not load `PGstandard.pl` or directly load `PGauxiliaryFunctions.pl`.

There are 160 problems in the OPL and Contrib (and 19 more in Pending) that use PGsort.  All of them load `PGstandard.pl`, and so also load `PGauxiliaryFunctions.pl`.  However, there are also some macros that use PGsort.  The macros do not load `PGstandard.pl` (and shoudn't) or `PGauxiliaryFunctions.pl` (and should).

For NchooseK and `PGchoicemacros.pl`, add the `loadMacros` call.

In general I think that we need to do some work in going through macros, and ensuring that macros load other macros if they use the methods from those other macros.

macros/core/PGauxiliaryFunctions.pl
```diff
@@ -532,5 +533,45 @@ sub fact {
 	P($_[0], $_[0]);
 }
 
+=head3 random_subset function 
+
+=pod
+
+	Usage: random_subset($n, @set);
+
+This function returns a randomly ordered array of $n elements taken from the array @set,
```
pstaabp, line 542: I would add this selects randomly from the set without replacement.

pstaabp (approved):

macros/core/PGauxiliaryFunctions.pl
```diff
@@ -532,5 +539,45 @@ sub fact {
...
+	if ($n > scalar(@set)) {
+		warn 'random_subset: The first input must be smaller than or equal to the number of elements.';
+		$n = scalar(@set);
+	}
+
+	my @out;
+	while (@set && @out < $n) {
+		push(@out, splice(@set, random(0, $#set, 1), 1));
+	}
+	return @out;
```
Alex-Jordan, line 579: Maybe this should be:

```
return wantarray ? @out : \@out;
```

somiaj: To me that seems like it would require another input `$wantarray` to the function, which I don't think is worth adding for that.

somiaj: Sorry, my mistake, this is a perl internal, and seems like a useful change. Thanks

Alex-Jordan:
One consequence of the `wantarray` change I suggested. If you do like 

```
$x = random_subset(1, ['a'..'d']);
```

Then `$x` will come out like `['a']`, not like `'a'`. I think this is OK, because the name of the subroutine is random "subset". But anyone trying to use it that way and expecting to get like `'a'` will have to understand this.

drgrice1 (approved):
This looks good to me now.  I think we should leave the other discussed methods in PGbasicmacros.pl for now.  We can think more on that for the next release.

macros/core/PGauxiliaryFunctions.pl
```diff
@@ -266,57 +268,59 @@ sub gcf {
...
-Examples: 
+Examples:
 
 	random_coprime([1..9],[1..9]) may return (2,9) or (1,1) but not (6,8)
 	random_coprime([-9..-1,1..9],[1..9],[1..9]) may return (-3,7,4), (-1,1,1), or (-2,2,3) but not (-2,2,4)
 
-Note: in the example above (-2,2,3) is valid because not all three share a factor greater than 1.  If you don't want to 
-allow pairs of numbers to be coprime, see random_pairwise_coprime. 
+Note: in the example above (-2,2,3) is valid because not all three share a factor greater than 1.
+If you don't want to allow pairs of numbers to be coprime, see random_pairwise_coprime.
```
Alex-Jordan, line 312: This is an old mistake, but i saw it since it's in the diff. I think this line should be:
```suggestion
If you don't want to allow pairs of numbers to have a common factor, see random_pairwise_coprime.
```
If there is no more to edit in this PR, you could accept my code suggestion here to fix this.

Alex-Jordan (approved):

macros/core/PGauxiliaryFunctions.pl
somiaj: thanks, I'll add a small fix.

drgrice1:
Wow, three approvals.  That is a record for recent times.  Frequently we don't even obtain the informally required 2 approvals.

I will merge this.

PR #790, opened 2023-03-22
add DropDown variant of PopUp

Alex-Jordan:
This follows https://github.com/openwebwork/pg/discussions/782.

Here is a small test file, demonstrating the original `PopUp()`, but also `DropDown()` with and without a custom prompt.

```
DOCUMENT();
loadMacros("PGstandard.pl", "PGML.pl", "parserPopUp.pl",);

$p = PopUp(['a'..'d'], 2);
$d1 = DropDown(['a'..'d'], 2);
$d2 = DropDown(['a'..'d'], 2, placeholder => 'Choose one:');

BEGIN_PGML
[_]{$p}  
[_]{$d1}  
[_]{$d2}
END_PGML

ENDDOCUMENT();
```

I plan to also add a `TrueFalse()` subroutine as well. Maybe I will name it `TFPopUp` or `PopUpTF` to be safer from a namespace collision. It would just take one argument (true/false) and effectively make `PopUp(['True', 'False'], answer)` except in static output forms (TeX and PTX) it won't print anything. (Elsewhere instructions in the exercise are already asking something like "Is the following true or false?")

Any feedback on the `TrueFalse()` method also welcome here.

### DIFF
--- a/macros/parsers/parserPopUp.pl
+++ b/macros/parsers/parserPopUp.pl
@@ -37,9 +37,10 @@ The difference between C<PopUp()> and C<DropDown() >is that in HTML,
 the latter will have an unselectable placeholder value. This value is '?'
 by default, but can be customized with a C<placeholder> option.
 
-C<DropDownTF()> is like C<DropDown> with options being "True" or "False".
-In this case, C<correct> is initerpreted as a perl boolean. Except strings
-like "false" and "F" are recognized as false. Also, in static output (PDF, PTX)
+C<DropDownTF()> is like C<DropDown> with options being localized versions of
+"True" and "False". 1 is understood as "True" and 0 as "False". The initial
+letter of the localized word is understood as that word if those letter are
+different. All of this is case-insensitive. Also, in static output (PDF, PTX)
 the menu is not printed. It is assumed that context makes the menu redundant.
 
 By default, the choices are left in the order that you provide them,
@@ -135,9 +136,9 @@ sub new {
 	shift if Value::isContext($_[0]);    # remove context, if given (it is not used)
 	my $choices     = shift;
 	my $value       = shift;
-	my $options     = shift;
-	my $placeholder = $options->{placeholder} // 0;
-	my $static      = $options->{static}      // 1;
+	my %options     = @_;
+	my $placeholder = $options{placeholder} // '';
+	my $static      = $options{static}      // 1;
 	Value->Error("A PopUp's first argument should be a list of menu items")
 		unless ref($choices) eq 'ARRAY';
 	Value->Error("A PopUp's second argument should be the correct menu choice")
@@ -193,7 +194,7 @@ sub MENU {
 	my $size        = shift;
 	my %options     = @_;
 	my @list        = @{ $self->{choices} };
-	my $placeholder = shift @list if $self->{placeholder};
+	my $placeholder = $self->{placeholder};
 	my $menu        = "";
 	$name = main::NEW_ANS_NAME() unless $name;
 	my $answer_value = (defined($main::inputs_ref->{$name}) ? $main::inputs_ref->{$name} : '');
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Now with `DropDownTF()`. New sample:

```
DOCUMENT();
loadMacros("PGstandard.pl", "PGML.pl", "parserPopUp.pl",);

$p = PopUp(['a'..'d'], 2);
$d1 = DropDown(['a'..'d'], 2);
$d2 = DropDown(['a'..'d'], 2, placeholder => 'Choose one:');
$t = DropDownTF(1);
$f = DropDownTF(0);

BEGIN_PGML
[_]{$p}  
[_]{$d1}  
[_]{$d2}

Assess each statement as true or false.  
a. [`1+1=2`] [_]{$t}
a. [`1+1=3`] [_]{$f}

END_PGML

ENDDOCUMENT();
```

macros/parsers/parserPopUp.pl
```diff
@@ -225,6 +251,30 @@ sub MENU {
...
+	return parser::PopUp->new($choices, $value, \%options);
+}
+
+#
+# TrueFalse() variant of PopUp()
+#
+
+sub DropDownTF {
+	my ($self, $value, %options) = @_;
+	my $sanitized_value = $value ? 'True' : 'False';
```
pstaabp, line 273: These should be localized with `maketext`.  and just below this.

pstaabp:
I think this looks good.  I know we talked about using a new name for this, so we don't have conflicts, but I feel in X years, we will say "why do we have two popup names"?   Ah, the wonders that is WeBWorK.

macros/parsers/parserPopUp.pl
```diff
@@ -225,6 +251,30 @@ sub MENU {
...
+	$sanitized_value = 'False' if (defined $value && $value =~ /^\s*(false|f)\s*$/i);
```
dpvc, line 274: If you localize this, the regular expression gets harder.  If one assumes that true and false have different first letters in all languages, then if you have the localized "True" in `$true` and the localized "False" in `$false`, then you might be able to do something like

``` perl
sub DropDownTF {
	my $self    = shift;
	my $value   = shift;   ### you will need to call the localization on this value to get the localized T/F
	my %options = (
	  static => 0,
	  @_
	);
	my $sanitized_value = ({
	  lc(substr($true, 0, 1)) => $true,
	  lc($true) => $true,
	  1 => $true,
	  lc(substr($false, 0, 1)) => $false,
	  lc($false) => $false,
	  0 => $false,
	})->[lc($value)];
	Value->Error("The value should be one of $true or $false") unless defined $sanitized_value;
	return parser::PopUp->DropDown([ $true, $false ], $sanitized_value, %options);
}
```

Here we use a HASH to list the valid true/false values and look up the given value (once it is localize).  If it isn't one of the allowed options, we give an error.

I'm hoping `substr()` handles UTF-8 properly so that if there is a multi-character UTF value, you get the proper number of bytes.

dpvc (changes requested):
I've made a number of suggestions.  you can ignore them if you wish.

Alex-Jordan:
Thanks @dpvc , I understood your suggestions and I think i applied all of them.

macros/parsers/parserPopUp.pl
```diff
@@ -256,23 +257,42 @@ sub MENU {
...
+	my $false        = main::maketext('False');
+	my %sanitization = (
+		lc($true)  => $true,
+		1          => $true,
+		lc($false) => $false,
+		0          => $false
+	);
+	if (lc(substr($true, 0, 1)) ne lc(substr($false, 0, 1))) {
+		$sanitization{ substr($true,  0, 1) } = $true;
+		$sanitization{ substr($false, 0, 1) } = $false;
```
dpvc, line 291: I think the substrings need to be lower-cased, here.  Other than that, I think it looks good.

macros/parsers/parserPopUp.pl
```diff
@@ -256,23 +257,42 @@ sub MENU {
...
+	}
+	my $sanitized_value = $sanitization{ lc($value) };
```
dpvc, line 293: Oh, one more thing, `$value` is already lower-cased in line 276 above, so no need to `lc()` it again here.

dpvc (approved):

Alex-Jordan:
Thanks @dpvc , I made some changes. I think I wanted both upper and lowercase T/F as keys. Does that seem right to you?

dpvc:
> I think I wanted both upper and lowercase T/F as keys. Does that seem right to you?

No.  You would have had that if you left in the `lc()` around the value and added it around the two `substr()` calls.  Since an upper case T/F will be lower-cased to in both the keys and the value, they will match with either case.  The way you have it now, you can only use `true` not `True` or `TRUE` as a correct answer for true, so you would need to use more keys for those.  Folding everything to lower-case reduces the need for upper-case or mixed-case versions of the keys.  That was one of the points of doing it that way initially.

Alex-Jordan:
@dpvc I see that the `uc` keys are unnecessary. But are you sure about "True" and "TRUE"? I seem to be able to use those as correct answers the way things are now. All keys are `lc`'d, and values for the hash are "True" or "False". The hash is applied to the `lc` of the correct value.

dpvc:
Sorry, I forgot that you did the `lc()` then you initially got the `$value, even though I had pointed that out early (it was a long day).  You are good now.

pstaabp (approved):

PR #791, opened 2023-03-25
Add a numberLine graphing mode to the graphtool.

drgrice1:
An interval tool is added that is designed to work with this graphing mode.  Note that the extreme left and right ends of the graph are considered points at infinity.  So both bounded and unbounded intervals can be graphed with this tool.
    
Endpoints can be made to be include or excluded with an include/exclude tool that works like the previous solid/dash tool for curves.  It is a bit different though as it applies per point, not for the entire object.
    
Some examples of the answer format for this graph object are "{interval,(-4,5]}", "{interval,(-infinity,10)}", or "{interval,[5,infinity)}".
    
If the option useBracketEnds is set to 1, then the endpoints are displayed as parentheses and brackets instead of open and closed dots (which is the default).

A couple of examples of usage are attached.
[Examples.zip](https://github.com/openwebwork/pg/files/11070431/Examples.zip)

Note there are also some other minor fixes and tweaks that are included that I have had sitting around.


Alex-Jordan:
Wow, thanks for this. It works as advertised. I can start the basics for coding some new exercises.

We also use brackets and parentheses here, so I'm happy to see you implemented those. Could the weight of those be heavier? The bracket and the parenthesis (at its thickest point) appear thin compared to the shaded interval. Maybe they should be thicker compared with the tick marks. 

I'm wondering if the interval itself should be thicker. You are probably reusing the thickness on lines in 2D, but it could be appropriate to be thicker here.

When I mouse over the right end, I see "(infty)". (But mousing over the left end, I see "-∞".

Unrelated to this PR. While trying to examine thickness of things, I used the web browser zoom. The svg element for the graph stayed fixed at 398px wide. Is it possible to make it possible for users to enlarge the image?

drgrice1:
> We also use brackets and parentheses here, so I'm happy to see you implemented those. Could the weight of those be heavier? The bracket and the parenthesis (at its thickest point) appear thin compared to the shaded interval. Maybe they should be thicker compared with the tick marks.

Yeah, I think the brackets and parenthesis are to thin also.  The problem is that they are actually text, and not an svg object.  So it is limited as to what can be done.  I tried adding a css style of `font-weight: bold`, but that didn't seem to help much.  I will try to see if there is anything else that can be done.
 
> I'm wondering if the interval itself should be thicker. You are probably reusing the thickness on lines in 2D, but it could be appropriate to be thicker here.

The thickness of the interval line is actually more than the thickness of lines for the 3D graph mode.  I could see about making them thicker.

> When I mouse over the right end, I see "(infty)". (But mousing over the left end, I see "-∞".

That is odd.  I see "∞" on the right.  I decided to use MathJax (and switched the usual coordinate hints to that as well).  I wonder if it is not typesetting correctly for you for some reason.

> Unrelated to this PR. While trying to examine thickness of things, I used the web browser zoom. The svg element for the graph stayed fixed at 398px wide. Is it possible to make it possible for users to enlarge the image?

Do you mean the graph div?  When I zoom in with browser zoom, the graph does enlarge.  It is still reported as 398px, but a pixel becomes a larger unit.

drgrice1:
I added `font-weight:900` to the bracket and parenthesis end points.  It helps a little.

drgrice1:
I increased the sizes of things by another notch now.

drgrice1:
Perhaps this will allay some of your concerns on size.  I added a new fullscreen button to the toolbox.

Alex-Jordan:
I like the thicker things, and I think they are good enough now.

I realized I had my browser set to zoom text only. My bad. The image as a whole zooms as expected now. But the fullscreen button is nice too.

I'm still getting `(infty)` after hovering on the right, but oddly getting `-∞` just fine after hovering on the left.

Screen Shot 2023-03-25 at 10 03 28 PM

At line 427, is there a missing `() => ` that I see in the nearby lines?

htdocs/js/apps/GraphTool/graphtool.js
```diff
@@ -402,7 +419,18 @@ window.graphTool = (containerId, options) => {
 	gt.snapRound = (x, snap) => Math.round(Math.round(x / snap) * snap * 100000) / 100000;
 
 	gt.setTextCoords = options.showCoordinateHints
-		? (x, y) => gt.current_pos_text.setText(`(${gt.snapRound(x, gt.snapSizeX)}, ${gt.snapRound(y, gt.snapSizeY)})`)
+		? (x, y) => {
+			if (options.numberLine) {
+				const bbox = gt.board.getBoundingBox();
+				const xSnap = gt.snapRound(x, gt.snapSizeX);
+				if (xSnap === bbox[0]) gt.current_pos_text.setText(() => '\\(-\\infty\\)');
+				else if (xSnap === bbox[2]) gt.current_pos_text.setText('\\(\\infty\\)');
```
Alex-Jordan, line 427:
```suggestion
				else if (xSnap === bbox[2]) gt.current_pos_text.setText(() => '\\(\\infty\\)');
```

macros/parsers/parserGraphTool.pl
```diff
@@ -100,109 +100,158 @@ =head1 GRAPH OBJECTS
...
 
-=item texSize (Default: texSize => 400)
+=item texSize (Default: C<< texSize => 400 >>)
 
 This is the size of the graph that will be output when a hard copy of the problem is generated.
 
+=item numberLine (Default: C<< numberLine => 0 >>)
+
+If set to 1, then the graph will show both the horizontal and vertical axes.  This is the
+default. If set to 0, then only the horizontal axis will be shown, and the graph can be
```
Alex-Jordan, line 220: Are your descriptions for `1` and `0` reversed here?

drgrice1:
> At line 427, is there a missing `() => ` that I see in the nearby lines?

Good catch.  It is odd that it was still working for me.  That made it hard for me to find the problem.

macros/parsers/parserGraphTool.pl
drgrice1: Yep.  Thanks.

drgrice1:
@Alex-Jordan:  By the way, there is another new macro that you should take a look at regarding PTX output.  That is the parserRadioMultiAnswer.pl macro.  I added a case for PTX output to that one, but I was not sure what should be output, so it is probably not right.  I have been meaning to ask you to take a look at it to determine what it should output.

pstaabp:
Overall, this looks great.  Do the endpoints need to be on integers?  (This might be a general graphTool question).  I was looking through the docs, but didn't see anything. 

I often give students problems where answers are not integers to remind them that there are non-integer answers.   I could imagine have answers on multiples of 1/2 or 1/3 perhaps and use the minorTickX option.

drgrice1:
The endpoints do not have to be integers, but there are limitations on what they can be.  With the defaults snap size, the coordinates must be integers.  You can set the snap size to 0.5, for example, and then students will be able to graph any multiple of that.  1/3 would be trouble as there would probably be rounding issues.  You could try it though.

The point of the graphtool is to introduce graphing concepts to students.  It is not to remind them that fractions exist.

drgrice1:
@pstaabp:  I did a little more work on the number line graphing mode to make it be a little better with rational step sizes.  Here is an example problem.  Try it with seed 1187 for example, to see it with a snap size of 1/3.
[SolveLinearInequalityRationalSolution.pg.txt](https://github.com/openwebwork/pg/files/11074288/SolveLinearInequalityRationalSolution.pg.txt)

Note there are some issues when the distance between two snap points is too small.  You will notice the arrow at infinity disappears until the other point is moved far enough away.  Unfortunately there isn't much that can be done about that.  JSXGraph detects that there isn't enough room for the array, and hides it.

drgrice1:
Yes. It would take some modifications to WeBWorK to allow for graphs instead.

pstaabp:
> @pstaabp: I did a little more work on the number line graphing mode to make it be a little better with rational step sizes. Here is an example problem. Try it with seed 1187 for example, to see it with a snap size of 1/3.

Not too bad.  Nice to know it's capable.  thanks.

pstaabp (approved):

Alex-Jordan:
I wrote some problems that use this and put them into a PreTeXt book. I was expecting it to basically work, and it does! Try one of these exercises, like #61:

Exercise Group: Exercise Group 
https://spot.pcc.edu/math/orcca/ed3/html/section-comparison-symbols-and-notation-for-intervals.html#p-812

After clicking "Activate", I was expecting that if I clicked the button to show the correct answer, nothing would happen. To my surprise, a correct answer graph appears. That is thanks to you, and code you wrote once upon a time here:

https://pretextbook.org/js/0.2/pretext-webwork/2.17/pretext-webwork.js

I must have known this at one time, but forgotten, since before now I haven't tried seriously using the GraphTool with PreTeXt. However, it produces a square graph, not using the numberline dimensions. Is there a quick patch I could apply to `pretext-webwork.js` that would make the correct answer appear with the same dimensions as the prompt graph?

Also I will need to adjust the PTX static output I put in #792 in the situation when `showInStatic` is 1. It should _show_ nothing, but acknowledge there was a question there that has an answer. The symptom I need to fix is that with those exercises in that link above, before you click to activate an exercise, you can click a knowl to show the answer. It is not showing anything for the graphing part of the exercise. It should (for now) at least show something like `{interval, [4,infinty)}`. [And maybe one day show the answer too.]

drgrice1:
@Alex-Jordan:  The css style for the number line is needed.  It looks like it will need to be added after line 608.  The css class for a number line is `graphtool-number-line`.  All of the graphtool styles are in pg/htdocs/js/apps/GraphTool/graphtool.css.  There are probably some other things in there that are needed.  Some of the styles there affect functionality.

pstaabp (approved):

Alex-Jordan:
I asked an instructor to test a new batch of problem files, including some that use the number line graphing tool. Her feedback includes:

> It took me many attempts to figure out how to use the tools to graph on the number line. Some instructions would be helpful.

Have you thought about how there could be any sort of help mechanism that explains how to use tools? This tool is perhaps the first one where it is not patently obvious how to use it. I don't have a suggestion, just wondering if I can avoid writing usage instructions into each problem file.

It was not obvious to her how to exit fullscreen mode (by using ESC). When you go to fullscreen mode, could the fullscreen button turn into an exit-fullscreen button?

She found that if you start an interval and move the mouse to make it an infinite interval, you might end up with your mouse a little outside of the graph. The arrowhead is there, so you go ahead and click to punch in the interval. But clicking outside the graph, the interval you were constructing disappears. And this was frustrating. I'm not sure how best to deal with that or what is even possible. But I notice that if I start an interval and click on 0, then move the mouse to, say, 2, then move the mouse down outside the graph, then I can click and nothing at all happens. That is, the interval I was constructing does not disappear. If this behavior were the same for the infinite interval, that would work.

Alex-Jordan:
> The css style for the number line is needed. It looks like it will need to be added after line 608. The css class for a number line is graphtool-number-line. All of the graphtool styles are in pg/htdocs/js/apps/GraphTool/graphtool.css. There are probably some other things in there that are needed. Some of the styles there affect functionality.

Thanks, I applied the height restriction to the `graphtool-number-line` class. I _think_ that is all that is needed. For the graph that a user actually works with, all of the styling should be coming from the WW server. This "correct answer" graph is special. Do I have that right?

drgrice1:
Yeah, that is right.  You are loading the graphtool css file from the server.  So you only need to have style for the "correct answer" graph since you don't use the attempts table.

drgrice1:
> Have you thought about how there could be any sort of help mechanism that explains how to use tools? This tool is perhaps the first one where it is not patently obvious how to use it. I don't have a suggestion, just wondering if I can avoid writing usage instructions into each problem file.

I have thought about this, but am not sure what form is best for this.  Part of the reason that I haven't pursued it further is because none of my students have had any issue figuring out the existing tools.  I do anticipate this being an issue with the interval tools though.  One thing that I have thought of doing is adding a sort of live caption below the graph tool that updates through the phases of the active tool.  A simpler option would be to just add a helpLink file.

> It was not obvious to her how to exit fullscreen mode (by using ESC). When you go to fullscreen mode, could the fullscreen button turn into an exit-fullscreen button?

The fullscreen button actually does become an exit fullscreen button.  If you click on it in fullscreen, then it exits full screen.  However, the text of the button doesn't change to reflect that.  I tried to make it that way, but there is a problem in detecting state changes and making the text reliable show the correct thing.  I will try to find a way to make that work.

> She found that if you start an interval and move the mouse to make it an infinite interval, you might end up with your mouse a little outside of the graph. The arrowhead is there, so you go ahead and click to punch in the interval. But clicking outside the graph, the interval you were constructing disappears. And this was frustrating. I'm not sure how best to deal with that or what is even possible. But I notice that if I start an interval and click on 0, then move the mouse to, say, 2, then move the mouse down outside the graph, then I can click and nothing at all happens. That is, the interval I was constructing does not disappear. If this behavior were the same for the infinite interval, that would work.

I knew you would bring this up.  First, there are two layers of focus for the graph tool.  The the graph board itself can have focus, and the `div` that contains the graph board and the tool buttons can have focus.  As soon as the mouse leaves the graph board, JSXGraph stops receiving mouse motion events, but if you are still inside the containing `div`, then the tool as a whole still has focus. When you click somewhere inside the containing `div`, the tool remains focused, but if you click outside of that focus goes elsewhere.  That can not be changed, as it would break the entire focus/blur system and natural tab order.  So what is happening is that when you click to the right or left of the graph board but just off of it, that is also outside of the containing div.  So the tool loses focus and the active tool is deactivated.  But when you click below the graph board but still in the button box area, the tool remains focused, and the active tool remains active.  That also can't be changed because that also would break the focus/blur system.  However, what could be done is to add a little padding on the outer div (maybe 10 or 15 pixels) and perhaps a "visually appealing" border on that.  Then as long as a click is within that outer border the tool can remain active.  This gives a little buffer where you can click to the left and right, and the tool won't lose focus.  However, it will still be the case that if you click outside of that the tool will deactivate.

There is another issue with the interval tool that I found.  I adding a "layer" toggle to defining points of graph objects when the object is focused.  The reason for this is because otherwise interval endpoints are visible in DOM order.  By that I mean that if one interval is graphed, and then another interval is graphed after that, the end points of the second interval will always be on top regardless of which one has focus.  The controls still work, and the focused interval is the one that the mouse and keyboard move.  However, it is a little visually confusing when the point being controlled is visually behind another inactive point.  The "layer" toggle doesn't actually change the layer, but just sets the layer to what it was before.  However, just setting the layer causes JSXGraph to remove the node from the DOM and reinsert it.  This moves the node later in the DOM, and for SVG this makes it visually in front.  Unfortunately, (in Chrome, but not in Firefox) this changes the natural tab order and makes it so that tabbing through the graphed objects does not work right.  As a result, I am going to have to remove the layer toggle, and that will make it so that when multiple objects are graphed, the active object will be visually behind later graphed objects.

PR #792, opened 2023-03-26
let PTX print GraphTool graph; add option to not print GraphTool grap…

Alex-Jordan:
…h in static output modes

Before this, PTX is printing nothing for a GraphTool graph. Now it will include the image files (tex, pdf, png, svg; packaged into a tgz).

Also I added an option to suppress printing the graph for static output (hardcopy and PTX). If the exercise says "Plot the line y=2x+1" or "Plot the interval (1,∞)" you don't necessarily want to print the blank graph on hardcopy. You can if you want to, and that is the default. This is more of a concern for PTX than hardcopy, where all those blank graphs could end up taking a lot of space in a printed book.

This may clash some with #791, as it has at least one adjacent edit. I can resubmit this after #791 is merged.

### DIFF
--- a/macros/parsers/parserGraphTool.pl
+++ b/macros/parsers/parserGraphTool.pl
@@ -203,6 +203,11 @@ passed as parameters directly to the GraphTool object's cmp() method.
 
 This is the size of the graph that will be output when a hard copy of the problem is generated.
 
+=item showInStatic (Default: 1)
+
+In "static" output forms (TeX, PTX) you may not want to print the graph if it is just taking
+space. In that case, set this to 0.
+
 =back
 
 =cut
@@ -271,6 +276,7 @@ sub new {
 		yAxisLabel          => 'y',
 		ariaDescription     => '',
 		showCoordinateHints => 1,
+		showInStatic        => 1,
 		availableTools      =>
 			[ 'LineTool', 'CircleTool', 'VerticalParabolaTool', 'HorizontalParabolaTool', 'FillTool', 'SolidDashTool' ],
 		texSize => 400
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Note most of the diff here is just a change in indentation. The real diff is a new POD entry,  adding the `showInStatic` option to the blessing, a small edit to some code commentary, and the necessary adjustment to the conditional logic about display modes.

drgrice1 (approved):
This looks good.  It does have minor conflicts with #791, but they are easy to remedy.

I personally think that if the problem asks to graph something like the line `y=2x+1` it is still best to print the graph with the grid.  That is like providing graph paper for the student, and is a good aid in graphing the line.  It is much better than asking the student to construct their own (usually quite sloppy) axes.  However, this is okay as the default still does print the graph, and I can see that in PTX output it may not be desired in some cases.

macros/parsers/parserGraphTool.pl
```diff
@@ -610,27 +616,28 @@ sub constructJSXGraphOptions {
 }
 
 # Produce a hidden answer rule to contain the JavaScript result and insert the graphbox div and
-# javacript to display the graph tool.  If a hard copy is being generated, then PGgraphmacros.pl
-# is used to generate a printable graph instead.  An attempt is made to make the printable graph
-# look as much as possible like the JavaScript graph.
+# javacript to display the graph tool.  If a hard copy is being generated, then PGtikz.pl is used
```
pstaabp, line 619: javacript -> javascript

pstaabp (approved):
Looks good.  Found one typo (that wasn't even yours)

macros/parsers/parserGraphTool.pl
drgrice1: Actually, it should be `JavaScript`.  Thanks for pointing that out.  That was that way before.

drgrice1:
> Looks good. Found one typo (that wasn't even yours)

I am pretty certain it was my typo.   Just not from this pull request.

macros/parsers/parserGraphTool.pl
drgrice1: I fixed this typo in #791.  So no need to fix it here and cause more conflict with that pull request.

pstaabp:
I'm going to merge this and then @drgrice1 will need to fix an conflicts.

Alex-Jordan:
I think I used bad technique here with the nested if. Nested ifs can be
hard to read.

The logic would be better as something like:

If TeX or PTX && !showInStatic
...
Elif TeX or PTX && showInStatic
...
Else

drgrice1:
It is fine.  It is actually better the way you did it.  It is not hard to read really.  What is worse is checking a conditional more than is needed.  That has runtime efficiency issues.  In addition, your proposed `if ... elsif ... else ...` would require an empty block.  That is not good practice either.

PR #793, opened 2023-03-27
Fix algorithm for Euler's totient function

d-torrance:
There were two issues:

* We were stopping the for loop too soon and missing the case where the number in question is a prime squared.  For example, if the input is 9, then we want to check if both 2 and 3 are divisors, but we stopped after 2 because 3^2 < 9 is false.
* According to Euler's product formula, we only multiply by (1 - 1/p) one time for each prime factor p, but for example, when the input was 4, we were multiplying by (1 - 1/2) twice, once for each iteration of the while loop.  So we move this operation outside the while loop.

### DIFF
--- a/macros/contexts/contextInteger.pl
+++ b/macros/contexts/contextInteger.pl
@@ -184,11 +184,12 @@ sub phi {
 	$self->Error("Cannot phi on Zero.") if $a == 0;
 	$result = $a;
 	$n      = $a;
-	for (my $i = 2; ($i**2) < $n; $i++) {
+	for (my $i = 2; ($i**2) <= $n; $i++) {
+		next unless ($n % $i == 0);
 		while ($n % $i == 0) {
 			$n      /= $i;
-			$result -= $result / $i;
 		}
+		$result -= $result / $i;
 	}
 	$result -= $result / $n if $n > 1;
 	return $result;
--- a/macros/contexts/contextInteger.pl
+++ b/macros/contexts/contextInteger.pl
@@ -187,7 +187,7 @@ sub phi {
 	for (my $i = 2; ($i**2) <= $n; $i++) {
 		next unless ($n % $i == 0);
 		while ($n % $i == 0) {
-			$n      /= $i;
+			$n /= $i;
 		}
 		$result -= $result / $i;
 	}


pstaabp:
Thanks for noticing this and putting in a fix.  You put this pull request into main.  Do you need this as a "hot fix" or can we get this into develop, which will make it into the next release this summer? 

If not, I checked this on develop and wrote a short test.

macros/contextInteger.pl
```diff
@@ -182,9 +182,11 @@ sub phi {
   my $a = abs(shift);
   $self->Error("Cannot phi on Zero.") if $a == 0;
   $result = $a; $n = $a;
-  for (my $i = 2; ($i ** 2) < $n; $i++) {
-    while ($n % $i == 0) {
-      $n /= $i;
+  for (my $i = 2; ($i ** 2) <= $n; $i++) {
+    if ($n % $i == 0) {
```
drgrice1, line 186: This would be better as
```suggestion
    next unless ($n % $i == 0);
```

pstaabp:
Just a note, the `contextInteger` macro is a bit of a mess.  Most of the functions use `$self->Error` and `$self` is not defined.  I'll clean this up, write some more tests and submit to develop.

d-torrance:
`develop` would be fine.  I'll admit I didn't read the docs to realize which branch to use.  :)

Should I close this PR?

drgrice1:
No need to close it. Just rebase onto develop with `git rebase origin/develop`, then force push with `git push --force`.  Finally, change the base here.

macros/contextInteger.pl
d-torrance: Thanks for the suggestion!  Fixed.

d-torrance:
Everything's moved on top of the `develop` branch now!

pstaabp (approved):
@d-torrance Thanks.  This looks good.

macros/contexts/contextInteger.pl
```diff
@@ -184,11 +184,12 @@ sub phi {
 	$self->Error("Cannot phi on Zero.") if $a == 0;
 	$result = $a;
 	$n      = $a;
-	for (my $i = 2; ($i**2) < $n; $i++) {
+	for (my $i = 2; ($i**2) <= $n; $i++) {
+		next unless ($n % $i == 0);
 		while ($n % $i == 0) {
 			$n      /= $i;
```
drgrice1, line 190: I am not sure why the Github perltidy action is not running on this pull request.  However, if it were running it would give an error here.  This line should be
```suggestion
			$n /= $i;
```

pstaabp: Good catch.  I can just commit that.

drgrice1:
Now the action is running.  Something with the base changing that prevented the workflows from running.

drgrice1 (approved):
This checks out. phi was giving the wrong value for many inputs, but those are correct with this pull request.

drgrice1:
Thanks @d-torrance.  I will merge this now.

macros/contexts/contextInteger.pl
d-torrance: Yeah, I thought that spacing was weird!  I didn't need to change that line to fix the code so I didn't mess with it.

drgrice1: The spacing was correct when the `$result -= $result / $i` was below it inside the while loop.  When that was moved out, the spacing needed to be removed.

We have a Github workflow the runs perltidy on the code.  If perltidy changes anything, then the workflow fails.  If the workflow fails we won't merge the pull request until the issue is fixed.  Webwork2 and PG code has historically been quite messy and hard to read.  This makes it difficult to move forward with future development.  Perltidy is a nice tool that helps to prevent that mess.

Issue #794, opened 2023-03-27
Error in PGML list numbering when using helpLink


dlglin:
Using the following code:
```
BEGIN_PGML
1. Item 1 [@helpLink("intervals")@]*
2. Item 2
END_PGML
```
I get the following output:
<img width="177" alt="Screenshot 2023-03-27 at 3 35 47 PM" src="https://user-images.githubusercontent.com/1794586/228072659-f67b441f-5c5c-46a8-87fc-6319b07c6e7e.png">

Note that the second list item is numbered 3.  A code block that doesn't call `helpLink` doesn't cause the issue, but I don't know if it applies to other functions as well.

somiaj:
The knowl the helplink uses to show/hide information is inside a list item. When adding the knowl link, a hidden list is created, `<li style="list-style: none;"><knowl div></div></li>`, hence the next list item will be item number 3.

pstaabp:
Fixed in #795

PR #795, opened 2023-03-28
Fix knowls in lists.

drgrice1:
When knowls are added in a list, instead of adding an `li` tag and inserting that into the list, add a `div` tag and insert that inside the existing `li` tag.  This results in the same appearance, is valid html, and does not affect numbering inside an ordered list.

This fixes issue #794.


somiaj (approved):
Fixes the issue in my tests too.

PR #797, opened 2023-03-31
Update the JSXGraph library to the latest version and other tweaks.

drgrice1:
JSXGraph is now at version 1.5.0.  This version has changed event handling.  The JSXGraph 'hit' event no longer sends the 'focusin' event, so that is now just handled directly in the grpahtool 'focusin' event handler.

In addition a feature added in this release is keyboard keydown and keydrag handling.  The graphtool implements its own keyboard event handling because that was not available before.  In addition the JXSGraph keyboard events are only sent for arrow keys.  That could be tapped into, but doesn't help much over the current implementation, and wouldn't work at all for other keys that the graph tool needs to act on.

There are also some changes in this pull request to make the new interval tools work better. These are summarized below.

Fix an issue caused by toggling the layer of defining points when an object is focused.  Toggling the layer messes up the natural tab order, and so it can't be done.  This has the down side that points that are graphed later are always above previously graphed points.  This is not a big deal for any of the previous graph objects since their points are not visible when not focused, however for intervals this is noticeable as the end points are always visible.

Add some padding to the outer graph tool container and a visual border. This adds space between the actual graph board and the outer container's boundary.  This makes it so that clicking to the right or left of the graph but still inside the outer container does not deactivate tools. This is important for graphing points at infinity for intervals. Currently a click to the right or left to easily deactivates the tool when you want to graph one of these points at infinity.

Change the text of the 'Fullscreen' button to 'Exit Fullscreen' when fullscreen mode is entered.  Of course, change that back to 'Fullscreen' when fullscreen mode is exited.

Remove the "bounded" quantifier from the tooltip for the interval tool, as well as from the css classes for its icons.  That was a remnant from an initial implementation that was going to have separate bounded and unbounded interval tools.

### DIFF
--- a/htdocs/js/apps/GraphTool/cubictool.js
+++ b/htdocs/js/apps/GraphTool/cubictool.js
@@ -115,6 +115,8 @@
 				},
 
 				pairedPointDrag(gt, e) {
+					if (e.type === 'keydown') return;
+
 					const coords = gt.getMouseCoords(e);
 					let left_x = this.X(), right_x = this.X();
 
--- a/htdocs/js/apps/GraphTool/graphtool.js
+++ b/htdocs/js/apps/GraphTool/graphtool.js
@@ -184,36 +184,13 @@ window.graphTool = (containerId, options) => {
 			gt.board.containerObj.tabIndex = -1;
 
 			gt.board.on('move', (e) => {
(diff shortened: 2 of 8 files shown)


pstaabp (approved):
Looks good.

PR #799, opened 2023-03-31
make graphtool provide answer fillin with PTX output

Alex-Jordan:
Much of the diff here is again indentation changes.

Previously, an `$out` variable is initialized before all the conditional code. Some blocks of the conditional code append to `$out`, and some simply replace it. After the conditional code, it returns `$out`, but it's confusing that some pathways have overwritten what it was initialized to be and others appended to that. So this moves the `return` commands to within each block, and I think it is more clear what each scenario is returning.

Also this helped me handle the `showInStatic -> 0` situation better for TeX. Without this commit, it was showing an answer blank in that situation (from the hidden answer field). There may be an issue that hidden answer fields are displayed in hardcopy output in the first place, but I did not pursue this.

And then the main thing that led to this is that now PTX produces its own form of "hidden input" field (regardless of `showInStatic`). Which is a `fillin` element inside a `p` with nothing else inside the `p`. It's a bit hackish but that is the convention we have. This way PTX will recognize that something was asking for an answer, and then for example in the answers at the back of the book, an answer will be listed. For now it will just be those strings like `{interval,(0,1)}` and that's OK.

### DIFF
--- a/macros/parsers/parserGraphTool.pl
+++ b/macros/parsers/parserGraphTool.pl
@@ -762,24 +762,24 @@ sub constructJSXGraphOptions {
 # to generate a printable graph instead.  An attempt is made to make the printable graph look
 # as much as possible like the JavaScript graph.
 sub ans_rule {
-	my $self = shift;
-	my $out  = main::NAMED_HIDDEN_ANS_RULE($self->ANS_NAME);
+	my $self         = shift;
+	my $answer_value = $main::envir{inputs_ref}{ $self->ANS_NAME } // '';
+	my $ans_name     = main::RECORD_ANS_NAME($self->ANS_NAME, $answer_value);
 
-	if ($main::displayMode =~ /^(TeX|PTX)$/) {
-		if ($self->{showInStatic}) {
-			return &{ $self->{printGraph} }
-				if defined($self->{printGraph}) && ref($self->{printGraph}) eq 'CODE';
+	if ($main::displayMode =~ /^(TeX|PTX)$/ && $self->{showInStatic}) {
+		return &{ $self->{printGraph} }
+			if defined($self->{printGraph}) && ref($self->{printGraph}) eq 'CODE';
 
-			my @size = $self->{numberLine} ? (500, 100) : (500, 500);
+		my @size = $self->{numberLine} ? (500, 100) : (500, 500);
 
-			my $graph = main::createTikZImage();
-			$graph->tikzLibraries('arrows.meta');
-			$graph->tikzOptions('x='
-					. ($size[0] / 96 / ($self->{bBox}[2] - $self->{bBox}[0])) . 'in,y='
-					. ($size[1] / 96 / ($self->{bBox}[1] - $self->{bBox}[3]))
-					. 'in');
+		my $graph = main::createTikZImage();
+		$graph->tikzLibraries('arrows.meta');
+		$graph->tikzOptions('x='
+				. ($size[0] / 96 / ($self->{bBox}[2] - $self->{bBox}[0])) . 'in,y='
+				. ($size[1] / 96 / ($self->{bBox}[1] - $self->{bBox}[3]))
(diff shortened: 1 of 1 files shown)


macros/parsers/parserGraphTool.pl
```diff
@@ -798,124 +797,129 @@ sub ans_rule {
...
+		$graph->tex($tikz);
+
+		my $out = main::image(
+			main::insertGraph($graph),
+			width    => $size[0],
+			height   => $size[1],
+			tex_size => $self->{texSize}
+		);
+		$out .= qq!<p><fillin name="$ans_name"/></p>! if ($main::displayMode eq 'PTX');
+		return $out;
```
drgrice1, line 915: This could be
```suggestion
		return main::image(
			main::insertGraph($graph),
			width    => $size[0],
			height   => $size[1],
			tex_size => $self->{texSize}
		) . ($main::displayMode eq 'PTX' ? qq!<p><fillin name="$ans_name"/></p>! : '');
```

drgrice1 (approved):
Looks good.

However, you might as well go all the way, and eliminate `$out` altogether!

drgrice1 (changes requested):
Oh, wait.  There is a problem here.  There was a reason that `main::NAMED_HIDDEN_ANS_RULE($ans_name)` was called at the beginning of the method and in any case.  That must be done.  That records the answer name, and if that is not done, then correct answers and student answers will not work correctly.

macros/parsers/parserGraphTool.pl
```diff
@@ -762,24 +762,23 @@ sub constructJSXGraphOptions {
 # to generate a printable graph instead.  An attempt is made to make the printable graph look
 # as much as possible like the JavaScript graph.
 sub ans_rule {
-	my $self = shift;
-	my $out  = main::NAMED_HIDDEN_ANS_RULE($self->ANS_NAME);
+	my $self     = shift;
+	my $ans_name = $self->ANS_NAME;
```
drgrice1, line 766: To fix the problem with the answer not being recorded you could change this to
```suggestion
	my $answer_value = $main::envir{inputs_ref}{ $self->ANS_NAME } // '';
	my $ans_name     = main::RECORD_ANS_NAME($self->ANS_NAME, $answer_value);
```
and then change line 923 as suggested below.  None of the other processing done by NAMED_HIDDEN_ANS_RULE is needed.

macros/parsers/parserGraphTool.pl
```diff
@@ -798,124 +797,129 @@ sub ans_rule {
...
+		return $out;
+	} elsif ($main::displayMode eq 'PTX') {
+		return qq!<p><fillin name="$ans_name"/></p>!;
+	} elsif ($main::displayMode eq 'TeX') {
+		return '';
 	} else {
 		$self->constructJSXGraphOptions;
-		my $ans_name = $self->ANS_NAME;
+		my $out = main::NAMED_HIDDEN_ANS_RULE($ans_name);
 		$out .= <<END_SCRIPT;
```
drgrice1, line 923: This is the other change needed:
```suggestion
		return main::tag('input', type => 'hidden', name => $ans_name, id => $ans_name, value => $answer_value)
			. main::tag(
				'input',
				type  => 'hidden',
				name  => "previous_$ans_name",
				id    => "previous_$ans_name",
				value => $answer_value
			) . <<END_SCRIPT;
```
and delete `return $out` below.

drgrice1 (commented):
This is one way to fix the issue.

drgrice1:
Hopefully you can sort out what is meant in that.  The two suggestions got merged together.

Alex-Jordan:
OK, I applied these edits and force pushed after doing some testing.

drgrice1 (approved):
Looks good.  Thanks for making those changes.

pstaabp (approved):

PR #800, opened 2023-04-04
Add a dynamic help system to the graph tool.

drgrice1:
Help messages are displayed as the user interacts with the graph tool.

There is a button that disables/enables help.  When clicked the button disables help for all graphtool graphs.  The setting is saved in local storage and so persists.  If clicked again, then help is enabled.

The bootstrap tooltips have been removed.  Instead those use the new help message box.  The tooltip messages are also bit more descriptive of the tool.

Instead of bootstrap modal dialogs for confirming object deletion or clearing the board, a sort of dialog in the content of the message box is used.

Button styles were added to replace the bootstrap button styles.

This makes the graph tool no longer dependent on bootstrap.

Note that this includes #797.

### DIFF
--- a/htdocs/js/apps/GraphTool/cubictool.js
+++ b/htdocs/js/apps/GraphTool/cubictool.js
@@ -156,9 +156,11 @@
 
 		CubicTool: {
 			iconName: 'cubic',
-			tooltip: '4-Point Cubic Tool',
+			tooltip: '4-Point Cubic Tool: Graph a cubic function.',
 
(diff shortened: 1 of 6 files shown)


pstaabp (approved):
This looks good.  However, the help text doesn't look like it is (or can be made) localized?  both PG side and javascript.

Alex-Jordan:
With three approvals (implicitly Glenn's) I'll merge this.

PR #801, opened 2023-04-04
Update the drag handlers for the graphtool graph objects.

drgrice1:
This takes advantage of the keyboard motion events that are now sent by JSXGraph to drag and move handlers as of version 1.5.0.  Now the graphtool drag and move handlers that previously only handled pointer events also handle keyboard motion events.

This also fixes an issue on touch screen devices where it is possible (and actually quite easy) to drag objects off the board.

Note that this obviously includes #797.  However, it is independent of #800, and (surprisingly) does not conflict with that pull request.

### DIFF
--- a/htdocs/js/apps/GraphTool/cubictool.js
+++ b/htdocs/js/apps/GraphTool/cubictool.js
@@ -115,6 +115,8 @@
 				},
 
 				pairedPointDrag(gt, e) {
+					if (e.type === 'keydown') return;
+
 					const coords = gt.getMouseCoords(e);
 					let left_x = this.X(), right_x = this.X();
 
--- a/htdocs/js/apps/GraphTool/cubictool.js
+++ b/htdocs/js/apps/GraphTool/cubictool.js
@@ -19,29 +19,6 @@
 				this.focusPoint = point1;
 			},
 
-			handleKeyEvent(gt, e, el) {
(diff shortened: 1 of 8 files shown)


pstaabp (approved):
Tested on a touchscreen device and works well--hadn't tried that before.  On keyboard device, seems to be able to handle everything as well.

Alex-Jordan:
I'm not sure what to test here. Even without this PR I can use the keyboard to move selected things around in a graph, so I think I do not understand what the new feature is.

drgrice1:
This isn't a new feature.  It just re-implements the mouse and keyboard handlers using new features of the JSXGraph library that were added in the latest version.  So you are testing that things work the same both with and without this pull request.  It is a cleaner implementation, because the mouse and keyboard drags are handled by the same methods.

The only thing that is new is that this restricts drags from being able to leave the board.  Previously on touch screen devices you could drag points off the board and leave them there which is not desirable.  This prevents that.  It can happen with a mouse drag, but not unless you set the snap sizes quite small relative to the board size (which you shouldn't do in general anyway).

Alex-Jordan:
OK, then with the three of us approving, I will merge this. I did not test with a touch screen device, but Peter did, and I assume you did as well. I will leave it to you to close #797 if you confirm that is appropriate.

Alex-Jordan:
Oh, I guess #797 closed automatically.

PR #802, opened 2023-04-04
Add a coordinateHintsType option to the graph tool.

drgrice1:
This option changes the way coordinate hints are shown.  By default the coordinates are displayed as decimal numbers accurate to five decimal places.  If this is set to 'fraction', then those decimals will be converted and displayed as fractions.  If this is set to 'mixed', then those decimals will be converted and displayed as mixed numbers.  For example, if the snapSizeX is set to 1/3, then what would be displayed as 4.66667 with the default 'decimal' setting, would be instead be displayed as 14/3 with the 'fraction' setting, and '4 2/3' with the 'mixed' setting.  Note that these fractions are typeset by MathJax.


pstaabp:
By the way, thanks for thinking about this.  It's always good to make sure that students have problems that aren't always integers. (They really don't like it though.)

pstaabp (approved):

Alex-Jordan (approved):
Thanks, this is nice. I tested and it's working with `decimal`, `fraction`, and `mixed`. With numberline or not.

Is that a continued fraction-based algorithm for converting the decimal?

I wondered if you'd ever want x-coordinates one way and y-coordinates some other way. Instead of `coordinateHintsType` there could be `coordinateHintsTypeX` and `coordinateHintsTypeY`. I'm not sure, since that would trade off a lot of extra typing when you want both to be, say, `fraction` for the rare occasion when you want like (fraction, decimal).

drgrice1:
> Is that a continued fraction-based algorithm for converting the decimal?

Yes, it is a continued fraction algorithm.  It is similar to the one implemented in the `contFrac` method of `contextFraction.pl`.  Although this implementation is not recursive like that one.  In fact, I rewrote the `contFrac` method to also use a loop instead of recursion in another branch that I will probably put in as a pull request as well.  Benchmarking shows it is at least twice as fast to use a loop than recursion.  It also carries less memory overhead, since it doesn't build up an array of past data that isn't even needed.

> I wondered if you'd ever want x-coordinates one way and y-coordinates some other way. Instead of `coordinateHintsType` there could be `coordinateHintsTypeX` and `coordinateHintsTypeY`. I'm not sure, since that would trade off a lot of extra typing when you want both to be, say, `fraction` for the rare occasion when you want like (fraction, decimal).

This could be done.  I will see if I can find a way to do it so that there is a single option that can be set as well as the dual x/y options.

drgrice1:
Now there are three options: `coordinateHintsType`, `coordinateHintsTypeX`, and `coordinateHintsTypeY`.  The latter two do the same as the former, but for the specified coordinate only.  If either of the latter two are undefined (the default), then the value of `coordinateHintsType` is used for that coordinate.

PR #803, opened 2023-04-05
record answer name even when answer input not shown in static

Alex-Jordan:
There was an issue with #790 and this should fix it. Now, even when the popup options are not shown in static output, the answer names are still recorded.

Also, for consistency with #799, I changed the option name from `static` to `showInStatic`.

Finally, previously I viewed that option as something internal. But now I described it in the POD after I found I wanted to hide the menu for static output in a PreTeXt project.

### DIFF
--- a/macros/parsers/parserPopUp.pl
+++ b/macros/parsers/parserPopUp.pl
@@ -41,7 +41,7 @@ C<DropDownTF()> is like C<DropDown> with options being localized versions of
 "True" and "False". 1 is understood as "True" and 0 as "False". The initial
 letter of the localized word is understood as that word if those letter are
 different. All of this is case-insensitive. Also, in static output (PDF, PTX)
-the menu is not printed. It is assumed that context makes the menu redundant.
+C<showInStatic> is 0. It is assumed that context makes the menu redundant.
 
 By default, the choices are left in the order that you provide them,
 but you can cause some or all of them to be ordered randomly by
@@ -89,6 +89,10 @@ You can use the PopUp menu object in MultiAnswer objects.  This is
 the reason for the pop-up menu's ans_rule method (since that is what
 MultiAnswer calls to get answer rules).
 
+There is one option, C<showInStatic>. It is 1 by default, except for
+C<DropDownTF> it is 0. This option controls whether or not the menu
+is displayed in a static output format (PDF hardcopy or PTX).
+
 =cut
 
 loadMacros('MathObjects.pl');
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Stand by, I have a new issue to fix here.

Alex-Jordan:
OK, this should be ready to check.

pstaabp:
So I hadn't tested the hardcopy output with this earlier.  On develop I have the following errors: 
```
Use of uninitialized value $text in substitution (s///) at line 1836 of [PG]/macros/core/PGML.pl
Use of uninitialized value $_[0] in join or string at line 2298 of [PG]/macros/core/PGbasicmacros.pl
```
and in the correct answers, it is missing the TF popup answers.  This PR fixes both of these.  

Is the answer names not being recorded the source of the warnings?  I was a little unclear about that statement.

macros/parsers/parserPopUp.pl
```diff
@@ -221,8 +224,9 @@ sub MENU {
 			$menu .= $escaped_item . '</li>' . "\n";
 		}
 		$menu .= '</var>';
-	} elsif ($main::displayMode eq "TeX") {
-		return unless $self->{static};
+	} elsif ($main::displayMode eq 'PTX' && !$self->{showInStatic}) {
+		$menu = qq(<fillin name="$name"/>);
```
drgrice1, line 228: I don't really like the approach of
```
if ($main::displayMode eq 'PTX' && $self->{showInStatic}) {
} elsif ($main::displayMode eq 'PTX' && !$self->{showInStatic}) {
}
```
that you are employing to avoid nesting if statements.
In my opinion it would be better to do
```
if ($main::displayMode eq 'PTX') {
    if ($self->showInStatic) {
    } else {
    }
}
```
I understand that you don't like the nested `if`s, but it is a natural language construct, and doesn't involve unnecessary conditional checking.  Indentation makes the nesting clear to me.

This is just my opinion, so no need to change it.

drgrice1 (commented):
This generally looks good.  Just a couple of suggestions.

macros/parsers/parserPopUp.pl
```diff
@@ -137,18 +141,18 @@ sub new {
 	my $choices     = shift;
 	my $value       = shift;
 	my %options     = @_;
-	my $placeholder = $options{placeholder} // '';
-	my $static      = $options{static}      // 1;
+	my $placeholder = $options{placeholder}  // '';
+	my $static      = $options{showInStatic} // 1;
```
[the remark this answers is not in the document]
Alex-Jordan: OK, I made these changes.

macros/parsers/parserPopUp.pl
Alex-Jordan: And I changed it to nest the ifs here.

Alex-Jordan:
Thanks for reviewing this. I made changes and force pushed.

drgrice1 (approved):
This looks good.  Thanks for making the suggested changes.

pstaabp (approved):

PR #804, opened 2023-04-06
set column halign correctly for HTML

Alex-Jordan:
In HTML, the default styling for a cell's text-align is center. So this was wrongly explicitly writing it to be center when the author wants center. And worse, it was failing to respect the author's wishes when then author wanted the column to be left-aligned.

PTX defaults to left-alignment in this situation, which probably confused me.

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -952,6 +952,8 @@ sub Row {
 			my $css = '';
 
 			# col level
+			$css .= css('text-align', 'left')
+				if ($cellAlign->{halign} eq 'l');
 			$css .= css('text-align', 'center')
 				if ($cellAlign->{halign} eq 'c');
 			$css .= css('text-align', 'right')
@@ -964,6 +966,7 @@ sub Row {
 				if ($cellAlign->{tex} =~ /\\itshape/);
 			$css .= css('font-family', 'monospace')
 				if ($cellAlign->{tex} =~ /\\ttfamily/);
+
 			if ($cellAlign->{tex} =~ /\\color(\[HTML\])?\{(.*?)[}!]/) {
 				$css .= css('color', ($1 ? '#' : '') . $2);
 			}


drgrice1:
Perhaps I am interpreting this incorrectly or using the macro in an unintended way, but with the following MWE on the develop branch the entries in the second table are centered, while with this pull request they are not.
```perl
DOCUMENT();

loadMacros('PGstandard.pl', 'PGML.pl', 'niceTables.pl');

BEGIN_PGML
[@ DataTable([
		[['Header Row', headerrow => 1], ['Header 1'], ['Header 2']],
		[['Data Row'], ['Data 1'], ['Data 2']]
	], align => 'ccc')
@]*
END_PGML

ENDDOCUMENT();

Alex-Jordan:
I'll look into it. It occurs to me that there may be different defaults in play with the CSS text-align in a table cell in the two arenas I've been using this: webwork2 and a PTX embedding.

Alex-Jordan:
OK, there is a weird thing with HTML tables and browser standards. It turns out that by default, the content of `<th>` elements are center-aligned and the content of `<td>` elements are left-aligned.

I think that default convention is not compatible with the way this macro is designed. If you declare `align => 'ccc'`, you mean for each column to have center alignment, not one thing for header cells, another for data cells. Although you could accomplish that by setting an aligmnent on certain cells to override the column alignment.

So I think that I should either:
1. change this to always explicitly style the `text-align` on all cells (rather than omit it when I think the default would be the right choice; OR
1.  act one way when it's a header cell, another way when it's a data cell. And omit the setting of `text-align` appropriately for each case.

Do you have a preference? I guess I would go with option 2, since I was trying to make the HTML be less cumbersome when possible.

Note that this macro uses a lot of inline styling. It could change to apply classes, and some separate CSS file styles everything. That could shorten the HTML where things like `<td style="text-align:right;">` becomes `<td class="r">`. I'm not sure it would be worth the work that would go into that though...

drgrice1:
I don't have a strong preference for either of the options that you gave.  Go with what is best for you.

One thing that I though of that has bearing on the behavior is not only HTML defaults, but also Bootstrap styles.  In webwork2 we always have Bootstrap on the page, and Bootstrap sets styles for generic table elements as well.

I think that the inline styles are fine for now.  I am not sure there would be much benefit for this macro to use classes and external css instead.  Although, usually I would say the opposite.  That approach is better when you want to apply styles uniformly to a table, but this macro allows for a per cell styling.  It seems that trying to use classes for that would end up having a close to one-to-one correspondence between css styles and classes.  On the other hand, with scss and Sass tools it might be possible to do things in a nice way.  For now stick with the inline styles though.

Alex-Jordan:
I updated this to declare the `text-align` for all cells. What is really going on with `th` and `td` defaults is confusing and this avoids any confusion.  Your sample problem works with this.

drgrice1 (approved):
This looks good.  This works for me now.

PR #805, opened 2023-04-06
Add macro fortlewis vectorfield2d

somiaj:
The `VectorField2D.pl` macro wasn't copied from the OPL, this adds the macro back, fixes a warning, and runs perltidy.

### DIFF
--- a/macros/graph/VectorField2D.pl
+++ b/macros/graph/VectorField2D.pl
@@ -1,7 +1,7 @@
 
 =head1 NAME
 
-VectorField2D.pl - Adds a vector filed graph to a WWPlot (from PGgraphmacros.pl) graphobject.
+VectorField2D.pl - Adds a vector field graph to a WWPlot (from PGgraphmacros.pl) graphobject.
 
 =head1 DESCRIPTION
 
@@ -83,11 +83,11 @@ Defines a single point to skip when creating the vector field.
 
 sub _VectorField2D_init { };    # don't reload this file
 
-loadMacros('PGgraphmacros.pl');
+loadMacros('MathObjects.pl', 'PGgraphmacros.pl');
 
 sub VectorField2D {
 	my %options = (
-		graphobject     => '',
+		graphobject     => undef,
 		Fx              => sub { return 1; },
 		Fy              => sub { return 1; },
 		xvar            => 'x',
--- a/macros/graph/VectorField2D.pl
+++ b/macros/graph/VectorField2D.pl
@@ -1,18 +1,12 @@
 sub _VectorField2D_init { };    # don't reload this file
 
-loadMacros("MathObjects.pl", "PGgraphmacros.pl");
+loadMacros('PGgraphmacros.pl');
 
 sub VectorField2D {
-
-###########################################
-	#
-	#  Set default options
-	#
-
 	my %options = (
-		graphobject     => $gr,
-		Fx              => Formula("1"),
-		Fy              => Formula("1"),
+		graphobject     => '',
+		Fx              => sub { return 1; },
+		Fy              => sub { return 1; },
 		xvar            => 'x',
 		yvar            => 'y',
 		xmin            => -5,
@@ -21,7 +15,7 @@ sub VectorField2D {
 		ymax            => 5,
 		xsamples        => 10,
 		ysamples        => 10,
-		vectorcolor     => "blue",
+		vectorcolor     => 'blue',
 		vectorscale     => 0.25,
 		vectorthickness => 2,
 		vectortipwidth  => 0.08,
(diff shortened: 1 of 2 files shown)


macros/graph/VectorField2D.pl
```diff
@@ -0,0 +1,113 @@
...
+	my $xstem;
+	my $ystem;
+	my $xmidtip;
+	my $ymidtip;
+	my $xleftbarb;
+	my $xrightbarb;
+	my $yleftbarb;
+	my $yrightbarb;
+
+	foreach my $i (0 .. $options{xsamples}) {
```
pstaabp, line 63: switch `foreach` to `for` here and below.

somiaj: This was just a copy of the previous macro, starting to clean up the way it is coded opens up a lot of other issues as well. As such I just refactored the whole macro to hopefully be a little more efficient. A bigger issue was the fact of storing the information about every arrow in a double array, which was never used.

somiaj:
Here is a test problem if people want something to test with this macro:

```perl
DOCUMENT();
loadMacros(
    'PGstandard.pl',
    'PGML.pl',
    'PGgraphmacros.pl',
    'VectorField2D.pl',
);
$refreshCachedImages = 1;

Context('Numeric')->variables->are(x => 'Real', y => 'Real');

($Fx[0], $Fy[0]) = (sub { return 5 * $_[1]; }, sub { return -3 * $_[0]; });
($Fx[1], $Fy[1]) = (Compute('5 x'), Compute('-3 y'));
($Fx[2], $Fy[2]) = (Compute('-5 y'), Compute('3 x'));
($Fx[3], $Fy[3]) = (sub { return -5 * $_[0]; }, sub { return 3 * $_[1]; });

# Graphs
for my $i (0..3) {
    $gr[$i] = init_graph(-5, -5, 5, 5, grid => [10, 10], axes => [0, 0], pixels => [500, 500]);
    $gr[$i]->lb('reset');
    for my $j (1..4) {
        $gr[$i]->lb(new Label(-4.7,   $j,  $j, 'black', 'center', 'middle'));
        $gr[$i]->lb(new Label(-4.7,  -$j, -$j, 'black', 'center', 'middle'));
        $gr[$i]->lb(new Label(  $j, -4.7,  $j, 'black', 'center', 'middle'));
        $gr[$i]->lb(new Label( -$j, -4.7, -$j, 'black', 'center', 'middle'));
    }
    $gr[$i]->lb(new Label(4.7, 0.2, 'x', 'black', 'center', 'middle'));
    $gr[$i]->lb(new Label(0.2, 4.7, 'y', 'black', 'center', 'middle'));
    VectorField2D(
        graphobject     => $gr[$i],
        Fx              => $Fx[$i],
        Fy              => $Fy[$i],
        xvar            => 'x',
        yvar            => 'y',
        xmin            => -5,
        xmax            => 5,
        ymin            => -5,
        ymax            => 5,
        xsamples        => 10,
        ysamples        => 10,
        vectorcolor     => "blue",
        vectorscale     => 0.05,
        vectorthickness => 2,
        vectortiplength => 0.65,
        vectortipwidth  => 0.15,
        xavoid          => 0,
        yavoid          => 0,
    );
}


BEGIN_PGML
Vector field graphs:

a) [``\vec{\bf F}(x,y) = \langle 5y, -3x \rangle``]

        [@ image($gr[0], width => 250) @]*

b) [``\vec{\bf F}(x,y) = \langle 5x, -3y \rangle``]

        [@ image($gr[1], width => 250) @]*

c) [``\vec{\bf F}(x,y) = \langle -5y, 3x \rangle``]

        [@ image($gr[2], width => 250) @]*

d) [``\vec{\bf F}(x,y) = \langle -5x, 3y \rangle``]

        [@ image($gr[3], width => 250) @]*
END_PGML
ENDDOCUMENT();
```

macros/graph/VectorField2D.pl
pstaabp: I guess we can just keep as is then.  On our todo list, we'd like to clean up all of these macros at some point.

somiaj: I already cleaned this one up and added some POD to it.

macros/graph/VectorField2D.pl
```diff
@@ -0,0 +1,163 @@
...
+
+Defines a single point to skip when creating the vector field.
+
+=back
+
+=cut
+
+sub _VectorField2D_init { };    # don't reload this file
+
+loadMacros('PGgraphmacros.pl');
```
drgrice1, line 86: This should be 
```suggestion
loadMacros('MathObjects.pl', 'PGgraphmacros.pl');
```
This was in the macro in the OPL and some problems depend on it.  For example, try Contrib/NAU/setCalcIII/DivRot.pg.  Without loading MathObjects.pl that problem gives an error about an undefined subroutine &main::Context.  Furthermore, the MathObjects.pl macro is a dependency of this macro, and so it should be loaded.  The `Value::isFormula` calls need it.

drgrice1 (commented):
Good improvements to the old macro.  There are a few issues though.

drgrice1:
By the way, attached is a set definition file with all problems from the OPL and Contrib that use this macro.
[setVectorField2D.def.txt](https://github.com/openwebwork/pg/files/11186416/setVectorField2D.def.txt)

macros/graph/VectorField2D.pl
```diff
@@ -0,0 +1,163 @@
...
+
+sub VectorField2D {
+	my %options = (
+		graphobject     => '',
```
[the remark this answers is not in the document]
somiaj: I changed this from `graphobject => $gr,`. Looking at this closer I realized that it could be possible that someone accidentally has this macro working because they called their graphobject `$gr`. Should this be something we worry about? (I have updated this to `undef` as you suggested for now.)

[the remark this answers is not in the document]
somiaj: Since all of the OPL problems seem to work with this, it should be fine.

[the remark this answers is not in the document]
drgrice1: I wouldn't worry about that.  None of the OPL problems do what you are suggesting.  So anyone else with problems that aren't public can fix the issue if that is the case.  I doubt there is much of that.  There are only a handful of problems that use this macro in the OPL, and I doubt there are many others out there.

drgrice1 (approved):
This looks good.  This will be needed to keep the handful of OPL problems that use this macro working.

pstaabp (approved):
Looks good.

PR #806, opened 2023-04-07
Update mathview to not use jQuery and to use modern javascript.

drgrice1:
This also fixes a bug in the current code.  The math in the math view popover was no longer being typeset.  I suspect this was caused by the recent bootstrap update and a change in the way that popovers work.  The math view code was attempting to typeset the code when the button is clicked.  That triggers the opening of the popover, but since it is not yet open, the typesetting does nothing.  This moves the typesetting request into the show.bs.popover event handler that bootstrap triggers when the popover is added to the DOM.

The typeahead code that I had previously commented out is now removed. I looked into finding a library that replaces bootstrap twitter's typeahead functionality, but was unable to find one that has the features that were utilized by math view.  In fact the latest version of bootstrap twitter's typeahead (finalized 8 years ago) had already removed some of the functionality that math view depended on.  If someone would like to investigate further, feel free.


drgrice1:
I forgot to mention that I moved the button and dropdown for changing the operations out of the popover title.  The reason for this is that was resulting in invalid html.  Bootstrap puts the title content into an `h3` tag.  The `div`s and `ul`s for the nav and dropdown are not valid in an `h3` tag.

pstaabp (approved):
A bit painful to use compared to mathquill, but looks fine.

drgrice1:
I now added a backdrop to the text area that maintains a copy of the textarea contents. When the textarea loses focus to the equation editor, then the backdrop is made visible and shows where the selection or cursor is. That is made lighter in color to hopefully not cause confusion as to what actually has focus.  This is basically the same as what I added for the MathQuill editor in this case.

I did not change the icons at this point.  Should that be done here?  Currently a different icon is used for the equation editor for a basic text type input answer, and for a essay textarea input.  A `fa-th` grid like icon is for the basic text type input, and the `fa-pencil` icon for the essay answer input.

Also, I did not add a clear button.  That might be nice here as well, but there isn't as much room for another button in this case.  I suppose the buttons could be moved to another row below the popovers input.

Alex-Jordan:
I tested and it works well, including with the "ghost cursor".

Do you understand MathView config well enough to prevent `inf` from being rendered as ∈f? It's a longstanding annoyance I had with MV. I tried to meddle with the config once, but I had to throw my hands up.

Alex-Jordan (approved):

drgrice1:
> Do you understand MathView config well enough to prevent `inf` from being rendered as ∈f? It's a longstanding annoyance I had with MV. I tried to meddle with the config once, but I had to throw my hands up.

I can look into it.  This only happens for MathView on regular text inputs (not essay type textarea MathViewers).  The reason it happens is that for regular text inputs MathView uses its `PGML` mode.  That means that it uses MathJax's "parse" mode (with backtick delimiters) instead of MathJax's "TeX" mode (with `\(` and `\)` delimiters).  So it parses things like `a/b` into `\frac{a}{b}`.  However, it also parses `inf` into `\in f`.  I am not sure how to fix that.  That might be a question @dpvc could answer.

Alex-Jordan:
Three approvals, so merging. (My comment about `\in f` is a matter for another day.)

PR #807, opened 2023-04-08
Give essay answers a MathQuill equation editor.

drgrice1:
This adds a button to the essay textarea that when clicked opens a collapse below the text area.  The collapse contains a MathQuill entry box and an insert button.  The user can type equations in the MathQuill entry, and then click "Insert" to inject latex wrapped in `\(` and `\)` into the essay textarea at the current cursor location.

This is perhaps not quite as nice as the usual MathQuill answer box for regular input type text answers, but at least provides something for essay type answers.  This is vaguely similar to what MathView does for essay type answers.

Note that PG already adds hidden latex MaThQuIlL_AnSwEr00?? blanks for essay answers.  Previously those were ignored as the corresponding AnSwEr00?? blank is not of type text and doesn't have the codeshard class.  That hidden field is used to identify the textareas in combination with the latexentryfield class that essay answers have. However, unlike regular text type inputs nothing is actually saved in the hidden latex field.

### DIFF
--- a/htdocs/js/apps/MathQuill/mqeditor.js
+++ b/htdocs/js/apps/MathQuill/mqeditor.js
@@ -193,10 +193,10 @@
 			insertButton.type = 'button';
 			insertButton.classList.add('btn', 'btn-primary', 'flex-grow-0', 'ms-2');
 			insertButton.textContent = 'Insert';
-			insertButton.addEventListener('click', () =>
-				insertAtCursor(answerQuill.input,
-					`\\(${answerQuill.mathField.latex().replace(/^(?:\\\s)*(.*?)(?:\\\s)*$/, '$1')}\\)`)
-			);
+			insertButton.addEventListener('click', () => {
+				const latex = answerQuill.mathField.latex().replace(/^(?:\\\s)*(.*?)(?:\\\s)*$/, '$1');
+				if (latex) insertAtCursor(answerQuill.input, `\\(${latex}\\)`);
+			});
 
 			const clearButton = document.createElement('button');
 			clearButton.type = 'button';
--- a/htdocs/js/apps/MathQuill/mqeditor.js
+++ b/htdocs/js/apps/MathQuill/mqeditor.js
@@ -85,7 +85,7 @@
 			button.setAttribute('aria-label', 'Equation Editor');
 
 			const icon = document.createElement('i');
-			icon.classList.add('fa-solid', 'fa-pencil');
+			icon.classList.add('fa-solid', 'fa-square-root-variable');
 			button.append(icon);
 
 			container.append(button);
@@ -147,6 +147,15 @@
 					`\\(${answerQuill.mathField.latex().replace(/^(?:\\\s)*(.*?)(?:\\\s)*$/, '$1')}\\)`)
 			);
 
+			const clearButton = document.createElement('button');
+			clearButton.type = 'button';
+			clearButton.classList.add('btn', 'btn-primary', 'flex-grow-0', 'ms-2');
+			clearButton.textContent = 'Clear';
+			clearButton.addEventListener('click', () => {
+				answerQuill.mathField.empty();
+				answerQuill.textarea.focus();
+			});
+
 			contents.append(cardHeader, cardBody);
 			collapse.append(contents);
 			innerContainer.append(collapse);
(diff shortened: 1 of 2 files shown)


drgrice1:
Attached is a problem to test this on (although any problem with an essay answer will work).  This problem will also work to test #806.
[essay-prob.pg.txt](https://github.com/openwebwork/pg/files/11183076/essay-prob.pg.txt)

pstaabp (approved):
Looks good.

Alex-Jordan:
It works nicely. I have some observations.
* Instead of a pencil icon, would something more suggesting of math be better? Like a x² or a ∑?
* Along with the insert button, would a clear button be helpful or just clutter? Presently I still cannot triple click to select all, but iirc that may be separately worked on.
* Say the cursor is somewhere in the essay text area, like in the middle of a paragraph. Then I move focus to the equation editor to type some math. Now I might not recall where the cursor was in the text area and I'm unsure if now is the right time to click "Insert". Is there any way to continue to display where the cursor was in the text area?

drgrice1:
> * Instead of a pencil icon, would something more suggesting of math be better? Like a x² or a ∑?
I used the same icon that MathView uses for this.  But, I think something like what you suggest would be better, probably also for MathView.  Although, not a unicode letter.  Those generally don't work that well for button icons.  A font awesome icon would be better.
> * Along with the insert button, would a clear button be helpful or just clutter? Presently I still cannot triple click to select all, but iirc that may be separately worked on.
I think there is room for two buttons.  Generally equations don't get that long.  Here also I used the same general design that MathView uses for this.
> * Say the cursor is somewhere in the essay text area, like in the middle of a paragraph. Then I move focus to the equation editor to type some math. Now I might not recall where the cursor was in the text area and I'm unsure if now is the right time to click "Insert". Is there any way to continue to display where the cursor was in the text area?
Unfortunately, there is not much that can be done with this, and MathView has this same problem.  The issue is that the textarea loses focus, and the cursor is hidden.  That is something handled natively by the browser, and you can't have focus in two places at the same time.

Alex-Jordan:
I tested this again. The "ghost cursor" in the textarea is nice. Kudos on figuring out how to make that happen. I tested on FF, Chrome, and Safari and it works well each time. Including with selected text.

One suggestion: when I click the Clear button, focus leaves the input area. But a user probably intends to start typing, so could focus go back to the input field at the end of the button click?

These are thoughts I had, possibly for separate development from this PR. When grading an essay question, you can leave comments in a textarea. Could this be added there as well? I have colleagues who don't know LaTeX and would use this there. In that setting, the √x button would fit well under the textarea.

Also, in that place to leave comments for students, there is a Preview button that generates a popup with the text but the math is rendered. Could that mechanism could be copied for the students too in the essay answer input textarea?

drgrice1:
> I tested this again. The "ghost cursor" in the textarea is nice. Kudos on figuring out how to make that happen. I tested on FF, Chrome, and Safari and it works well each time. Including with selected text.

I have used a similar technique before, but didn't want to do it.  It is tedious to get the css to work right for alignment of the "ghost cursor/selection" with the text in the textarea.  It seems to be working well though.

> One suggestion: when I click the Clear button, focus leaves the input area. But a user probably intends to start typing, so could focus go back to the input field at the end of the button click?

I added this.

> These are thoughts I had, possibly for separate development from this PR. When grading an essay question, you can leave comments in a textarea. Could this be added there as well? I have colleagues who don't know LaTeX and would use this there. In that setting, the √x button would fit well under the textarea.
> 
> Also, in that place to leave comments for students, there is a Preview button that generates a popup with the text but the math is rendered. Could that mechanism could be copied for the students too in the essay answer input textarea?

These should be possible.  I think that they should be a separate pull request (or two pull requests) though.

Alex-Jordan:
I tested again and I see the focus go back to the input after clicking Clear.

I noticed another thing. If the input is empty (or just space characters) then clicking Insert inserts a `\(\)`. Should it just do nothing instead when the string is empty? Not sure about if it is white space.

drgrice1:
Yeah, the MathView editor does the same, and I just left it.  I changed that now for the MathQuill editor.  So if the MathQuill input contains only whitespace, then nothing is inserted into the textarea when 'Insert' is clicked.

Should this be done with the MathView editor as well?  I could add that to this pull request since the MathView pull request was merged already.

PR #808, opened 2023-04-10
PTX checkbox

Alex-Jordan:
This is mainly to handle PTX output. The checkboxes will come out as either a PTX `ol`, `ul`, or `dl`. It depends on how labels are being handled by the problem.

Also I added lowercase letter labeling, lowercase Roman numeral labeling, and uppercase Roman numeral labeling. The existing capital letter labeling only works for up to 26 items, so I stuck with that cap.

### DIFF
--- a/macros/parsers/parserCheckboxList.pl
+++ b/macros/parsers/parserCheckboxList.pl
@@ -463,7 +463,8 @@ sub cmp_defaults {
 		list_type         => 'selection',
 		requireParenMatch => 0,
 		implicitList      => 0,
-		correct_choices   => $self->data
+		correct_choices   => $self->data,
+		correct_values    => [ map {"$_"} $self->value ]
 	);
 }
 
--- a/macros/parsers/parserCheckboxList.pl
+++ b/macros/parsers/parserCheckboxList.pl
@@ -547,9 +547,10 @@ sub CHECKS {
 
 	for my $i (0 .. $#{ $self->{orderedChoices} }) {
 		my $value = $self->{values}[$i];
-		my $tag   = $self->{orderedChoices}[$i];
-		$value = '%' . $value                                   if (grep { $i == $_ } @{ $self->{checkedI} });
-		$tag   = $self->labelFormat($self->{labels}[$i]) . $tag if $self->{displayLabels};
+		$value = '%' . $value if (grep { $i == $_ } @{ $self->{checkedI} });
+
+		my $tag = $self->{orderedChoices}[$i];
+		$tag = $self->labelFormat($self->{labels}[$i]) . $tag if $self->{displayLabels};
 		if ($i > 0) {
 			push(
 				@checks,
@@ -578,18 +579,17 @@ sub CHECKS {
 			my $marker = '';
 			$marker = "1" if $originalLabels eq '123';
 			$marker = "a" if $originalLabels eq 'abc';
-			$marker = "A"
-				if uc($originalLabels) eq 'ABC' && $originalLabels ne 'abc';
+			$marker = "A" if uc($originalLabels) eq 'ABC' && $originalLabels ne 'abc';
 			$marker = "i" if $originalLabels eq 'roman';
-			$marker = "I"
-				if uc($originalLabels) eq 'ROMAN' && $originalLabels ne 'roman';
-			$list_type  = qq(ol	 marker="$marker.");
+			$marker = "I" if uc($originalLabels) eq 'ROMAN' && $originalLabels ne 'roman';
+
+			$list_type  = qq(ol marker="$marker.");
 			$close_list = 'ol';
 		} elsif ($self->{localLabels} || ref $originalLabels eq 'ARRAY') {
 			$list_type  = 'dl width="narrow"';
 			$close_list = 'dl';
 			my %checks_to_labels = map { $checks[$_] => $self->{labels}[$_] } (0 .. $#{ $self->{orderedChoices} });
-			map {s/^(<li.*?>)/$1<title>$checks_to_labels{$_}<\/title>/g} @checks;
+			map { $_ =~ s/^(<li.*?>)/$1<title>$checks_to_labels{$_}<\/title>/gr } @checks;
 		}
 		$checks[0] = qq{<$list_type name="$name">\n$checks[0]};
 		$checks[-1] .= "</$close_list>";
--- a/macros/parsers/parserCheckboxList.pl
+++ b/macros/parsers/parserCheckboxList.pl
@@ -574,25 +574,27 @@ sub CHECKS {
 
 		# Do we want an ol, ul, or dl?
 		my $list_type      = 'ul';
-		my $marker         = '';
+		my $close_list     = 'ul';
 		my $originalLabels = $self->{originalLabels};
 		if ($originalLabels =~ m/^(123|abc|roman)$/i) {
-			$list_type = 'ol';
-			$marker    = "1" if $originalLabels eq '123';
-			$marker    = "a" if $originalLabels eq 'abc';
-			$marker    = "A"
+			my $marker = '';
+			$marker = "1" if $originalLabels eq '123';
+			$marker = "a" if $originalLabels eq 'abc';
+			$marker = "A"
 				if uc($originalLabels) eq 'ABC' && $originalLabels ne 'abc';
 			$marker = "i" if $originalLabels eq 'roman';
 			$marker = "I"
 				if uc($originalLabels) eq 'ROMAN' && $originalLabels ne 'roman';
+			$list_type  = qq(ol	 marker="$marker.");
+			$close_list = 'ol';
 		} elsif ($self->{localLabels} || ref $originalLabels eq 'ARRAY') {
-			$list_type = 'dl';
+			$list_type  = 'dl width="narrow"';
+			$close_list = 'dl';
 			my %checks_to_labels = map { $checks[$_] => $self->{labels}[$_] } (0 .. $#{ $self->{orderedChoices} });
 			map {s/^(<li.*?>)/$1<title>$checks_to_labels{$_}<\/title>/g} @checks;
 		}
-		$marker = ($marker ne '') ? qq{ marker="$marker."} : '';
-		$checks[0] = qq{<$list_type$marker name="$name">\n$checks[0]};
-		$checks[-1] .= "</$list_type>";
+		$checks[0] = qq{<$list_type name="$name">\n$checks[0]};
+		$checks[-1] .= "</$close_list>";
 
 		# Change math delimiters
 		@checks = map { $_ =~ s/\\\(/<m>/gr } @checks;
--- a/macros/parsers/parserCheckboxList.pl
+++ b/macros/parsers/parserCheckboxList.pl
@@ -50,7 +50,7 @@ checkbox.
 
 The values set as described above are the answers that will be displayed in the
 past answers table.  See the C<values> option below for more information.
-Problem authors are encourages to set these values either as described above, or
+Problem authors are encouraged to set these values either as described above, or
 via the C<values> option.  This is useful for instructors viewing past answers.
 
 By default, the choices are left in the order that you provide them, but you can
@@ -98,7 +98,7 @@ choices in their original order (though the C<< { label => text } >> form is
 preferred).
 
 Default: labels are the text of the choice when they don't include any special
-characters, and "Button 1", "Button 2", etc., otherwise.
+characters, and "Choice 1", "Choice 2", etc., otherwise.
 
 =item C<S<< values => array reference >>>
 
--- a/macros/parsers/parserCheckboxList.pl
+++ b/macros/parsers/parserCheckboxList.pl
@@ -463,8 +463,7 @@ sub cmp_defaults {
 		list_type         => 'selection',
 		requireParenMatch => 0,
 		implicitList      => 0,
-		correct_choices   => $self->data,
-		correct_values    => [ map {"$_"} $self->value ]
+		correct_choices   => $self->data
 	);
 }
 
--- a/macros/parsers/parserCheckboxList.pl
+++ b/macros/parsers/parserCheckboxList.pl
@@ -573,7 +573,7 @@ sub CHECKS {
 
 		# Do we want an ol, ul, or dl?
 		my $list_type      = 'ul';
-		my $close_list     = 'ul';
+		my $subtype        = '';
 		my $originalLabels = $self->{originalLabels};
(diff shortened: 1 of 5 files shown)


Alex-Jordan:
We want to move the PTX output from WeBWorK away from having `var` elements. Since it's related to the checkboxes multiple choice, I also added a commit here that changes `var` to `ul` for radio buttons.

pstaabp (approved):
Looks good.  Tested the lower case latin and roman labels.  I'm guessing the pretext is working the way you want.

macros/parsers/parserCheckboxList.pl
```diff
@@ -292,14 +300,24 @@ sub addLabels {
...
+		if uc($labels) eq 'ROMAN' && $labels ne 'roman';
+	$labels = [] if $labels eq 'text';
 
 	if (ref($labels) ne 'ARRAY') {
 		my $replace = $labels ne 'auto';
 		if (!$replace) {
-			for (@$choices) { $replace = 1 if $_ =~ m/[^-+.,;:()!\[\]a-z0-9 ]/i }
+			for (@$choices) {
+				$replace = 1 if $_ =~ m/[^-+.,;:()!\[\]a-z0-9 ]/i;
+			}
```
drgrice1, line 320: I don't like this change.  In fact some perl style guides say the way it was is preferred for one line blocks.  Note that if on one line the semicolon should be omitted.  See https://perldoc.perl.org/perlstyle.

macros/parsers/parserCheckboxList.pl
```diff
@@ -345,18 +364,27 @@ sub getCorrectChoices {
 		}
 
 		for (0 .. $#{ $self->{orderedChoices} }) {
-			if ($value eq $self->{orderedChoices}[$_] || $value eq ($self->{labels}[$_] || '')) {
+			if ($value eq $self->{orderedChoices}[$_]
+				|| $value eq ($self->{labels}[$_] || ''))
+			{
```
drgrice1, line 369: I don't like this change either.  If a list of conditionals fits nicely into the 120 character limit it looks better on one line.

macros/parsers/parserCheckboxList.pl
```diff
@@ -374,14 +402,17 @@ sub getCheckedChoices {
 		}
 
 		for (0 .. $#{ $self->{orderedChoices} }) {
-			if ($value eq $self->{orderedChoices}[$_] || $value eq $self->{labels}[$_]) {
+			if ($value eq $self->{orderedChoices}[$_]
+				|| $value eq $self->{labels}[$_])
+			{
```
drgrice1, line 407: Again, this is better on one line.

macros/parsers/parserCheckboxList.pl
```diff
@@ -345,18 +364,27 @@ sub getCorrectChoices {
...
+		main::PGsort(
+			sub {
+				$self->getIndexByValue($_[0]) < $self->getIndexByValue($_[1]);
+			},
+			@{ $self->{data} }
+		)
+	];
+
+	Value::Error('The correct choices must be among the label values')
+		unless @{ $self->{data} };
```
drgrice1, line 387: I personally prefer keeping this on line.  For the `statement if|unless conditional` construct, I generally prefer that to be on one line if it fits well within the 120 character limit.

macros/parsers/parserCheckboxList.pl
```diff
@@ -374,14 +402,17 @@ sub getCheckedChoices {
...
+			{
 				push(@{ $self->{checkedI} }, $_);
 				last;
 			}
 		}
 	}
 
-	Value::Error('The checked choices must be among the label values') unless @{ $self->{checkedI} };
+	Value::Error('The checked choices must be among the label values')
+		unless @{ $self->{checkedI} };
```
drgrice1, line 415: This also fits nicely onto one line.

macros/parsers/parserCheckboxList.pl
```diff
@@ -409,7 +441,8 @@ sub labelFormat {
 sub getIndexByValue {
 	my ($self, $value) = @_;
 	return -1 unless defined $value;
-	my ($index) = grep { $self->{values}[$_] eq $value } 0 .. $#{ $self->{values} };
+	my ($index) =
+		grep { $self->{values}[$_] eq $value } 0 .. $#{ $self->{values} };
```
drgrice1, line 445: I definitely do not like this change.  This without a doubt is more readable on one line.

macros/parsers/parserCheckboxList.pl
```diff
@@ -424,7 +457,8 @@ sub labelText {
 	my ($s, $e) = ('', '');
 	return $choice if scalar(@words) < 3;
 	do { $s .= shift(@words); $e = pop(@words) . $e if @words }
-		while length($s) + length($e) + 10 < $self->{maxLabelSize} && scalar(@words);
+		while length($s) + length($e) + 10 < $self->{maxLabelSize}
+		&& scalar(@words);
```
drgrice1, line 461: This is better on one line.

macros/parsers/parserCheckboxList.pl
```diff
@@ -456,8 +490,9 @@ sub cmp_preprocess {
 	my ($self, $ans) = @_;
 	if (defined $ans->{student_value} && @{ $ans->{student_value}->data }) {
 		$ans->{original_student_ans} = join(', ', map { $self->labelText($_) } @{ $ans->{student_value}->data });
-		$ans->{preview_latex_string} = $self->quoteTeX($ans->{original_student_ans});
-		$ans->{student_ans}          = $self->quoteHTML($ans->{original_student_ans});
+		$ans->{preview_latex_string} =
+			$self->quoteTeX($ans->{original_student_ans});
```
drgrice1, line 494: Definitely better on one line.

macros/parsers/parserCheckboxList.pl
```diff
@@ -532,8 +567,9 @@ sub CHECKS {
 	for my $i (0 .. $#{ $self->{orderedChoices} }) {
 		my $value = $self->{values}[$i];
 		my $tag   = $self->{orderedChoices}[$i];
-		$value = '%' . $value                                   if (grep { $i == $_ } @{ $self->{checkedI} });
-		$tag   = $self->labelFormat($self->{labels}[$i]) . $tag if $self->{displayLabels};
+		$value = '%' . $value if (grep { $i == $_ } @{ $self->{checkedI} });
+		$tag   = $self->labelFormat($self->{labels}[$i]) . $tag
+			if $self->{displayLabels};
```
drgrice1, line 572: Preferred the way this was.

macros/parsers/parserCheckboxList.pl
```diff
@@ -569,14 +628,31 @@ sub CHECKS {
 sub protect {
 	my ($self, $s) = @_;
 	return '' if !defined $s || $s eq '';
-	return main::MODES(TeX => $self->quoteTeX($s), HTML => $self->quoteHTML($s));
+	return main::MODES(
+		TeX  => $self->quoteTeX($s),
+		HTML => $self->quoteHTML($s)
+	);
```
drgrice1, line 634: I don't see the benefit of separating this onto multiple lines.

macros/parsers/parserCheckboxList.pl
```diff
@@ -552,15 +588,38 @@ sub CHECKS {
...
+		my $originalLabels = $self->{originalLabels};
+		if ($originalLabels =~ m/^(123|abc|roman)$/i) {
+			my $marker = '';
+			$marker = "1" if $originalLabels eq '123';
+			$marker = "a" if $originalLabels eq 'abc';
+			$marker = "A"
+				if uc($originalLabels) eq 'ABC' && $originalLabels ne 'abc';
+			$marker = "i" if $originalLabels eq 'roman';
+			$marker = "I"
+				if uc($originalLabels) eq 'ROMAN' && $originalLabels ne 'roman';
```
drgrice1, line 605: This is inconsistent.  Some of the `if` statements are on the same line, and some on the next.  This is a case where it would definitely improve readability to have all `if` statements on the same line.  Add an empty line after the last `$marker = ...` line if you want to prevent perltidy from trying to align those with the next `$list_type = ...` line.

drgrice1 (changes requested):
I don't like many of the changes here that seem to be intended to be some kind of code clean up.  Many of them don't really improve readability.  Some even do the opposite.

macros/parsers/parserCheckboxList.pl
drgrice1, line 572: If lines 568-572 were changed to
```perl
		my $value = $self->{values}[$i];
		$value = '%' . $value if (grep { $i == $_ } @{ $self->{checkedI} });

		my $tag = $self->{orderedChoices}[$i];
		$tag = $self->labelFormat($self->{labels}[$i]) . $tag if $self->{displayLabels};

```
It would be better organized, and the empty lines improve readability as well as prevent perltidy from aligning things badly.

drgrice1:
I see.  It sounds like the pg `.perltidyrc` settings didn't get used.  When I run perltidy on the file with those changes reverted, it doesn't make those changes.

drgrice1 (approved):
Looks good.  Sorry about that.

Alex-Jordan:
No, I messed up running pertidy. I still don't use the script because I haven't researched that glob issue. I use the command you suggested in a previous PR. I think I remember now I ran it from the wrong folder so it didn't have the path right to the config file. It gave an error, which I assumed meant it did not run. But now I get it: it ran anyway. Then I ran pertidy again properly, but I now understand that actually did nothing (or not much) since it had already run.

Alex-Jordan:
Another commit here in PGbascimacros.pl has to do with the PTX `title` element that is used in a description list item, which comes out of this PR.

While I was editing that file, I noticed an issue I should have fixed when doing all the recent niceTables work. It's a small thing, tacked in here in its own commit, within the `PTX_cleanup` routine.

Alex-Jordan:
This last commit is wrong. I have to be in meetings now though and will have to fix it later. I am trying to get the array of correct answer values to show up in the answer hash. Right now it only has the labels in the form of a correct answer string like "Choice 1, Choice 3, and Choice 5". But I'm hoping to get like `["B0", "B2", "B4"]`. There is a line that inserts `data` into the answer hash, but it comes out as `[null, null, null]`.

Alex-Jordan:
This line:

```
$_ = Value::makeValue($_, context => $context) for @{ $self->data };
```

is destroying the elements of `$self->{data}`. (If I comment that out, then the answer has changes from having `correct_choices => [null,null,null]` to having `correct_choices => `["B0","B2","B4"]`.)

Do you understand why that is happening? If you don't have a quick answer, wait until I can look into this using 2.18. So far I've just been doing PTX stuff associated with this using the AIM 2.17 server. If it's not obvious what the issue is, I will want to reproduce the issue with 2.18 before spending more time trying to figure it out.

macros/parsers/parserCheckboxList.pl
```diff
@@ -447,7 +463,8 @@ sub cmp_defaults {
 		list_type         => 'selection',
 		requireParenMatch => 0,
 		implicitList      => 0,
-		correct_choices   => $self->data
+		correct_choices   => $self->data,
+		correct_values    => $self->{values}
```
drgrice1, line 467: `$self->{values}` will be all values, not just the correct values.  `$self->data` is actually the `correct_values`.  What exactly are you trying for here?

Alex-Jordan:
>$self->{values} will be all values, not just the correct values.

This is why I wrote:

> This last commit is wrong.

I'm trying to get like `[B0 , B2 , B4 ]`. But I just get `[null,null,null]`. This is from the "raw" JSON output that pretext-webwork.js uses.

I will try everything with 2.18, I just need some time is all. I had meetings this afternoon and a family thing starting now through the evening.

drgrice1:
The reason that you get `[null, null, ...]` for the `correct_choices` in the raw format is that when the answer hash is passed back to webwork2 from pg the Translator calls `stringify_answers`.  This then calls `stringify_hash` on each answer hash in the hash of answers.  That goes through all the keys of the answer hash, and converts each one into a string.  If the key value is an array or hash (the value of `correct_choices` is an array), then `JSON->new->utf8->allow_unknown->allow_blessed->encode` is called on the key value.  Since the entries of the `correct_choices` array are blessed math objects the values can't be encoded, and so that method just returns `null`.

You noted that if you comment out line 259 (`$_ = Value::makeValue($_, context => $context) for @{ $self->data };`) of parserCheckbox.pl, then the answers come through.  You say this "destroys" the `correct_choices`.  It is not actually destroying the values, but converting them into math objects, and then the above occurs.  If that line is not done, then the entries of the `correct_choices` are left as Perl strings which work fine in the JSON encode call.  However, if you comment out that line the Value::List checker also ceases to work and the problems utilizing the macro become dysfunctional.  So that is not really an option.

Note that prior to WeBWorK/PG 2.16 the stringify_hash method would not have called JSON encode on the key.  Instead it would have set `$self->{$key} = "$self->{$key}";`.  In that case you would instead get something like "ARRAY(0x564da8610918)" instead of what you now get.  That would actually be the case regardless of if line 259 of parserCheckbox.pl were called.

One thing that you could do is change line 450 of parserCheckbox.pl from `correct_choices   => $self->data` to `correct_choices   => [ map { "$_" } $self->value ]`.  In other words stringify the elements already at this point.  Then JSON encode will work on those string values just as if you had comment out line 259.  However, if someone were to use this in an answer checker (which is inside the safe zone and before the translator stringfication occurs), those would be pure Perl strings, and not math objects, which may not be what is expected.

drgrice1:
Another option that may work is to add a `TO_JSON` method to math (Value) objects, and add `convert_blessed` to the stringify_hash JSON encode call.  Then that will call the `TO_JSON` method of any blessed object if it has one, and if not return `null` as before.  In addition, this wouldn't affect anything done before the `stringify_answers` call of the translator while still in the safe zone.

Alex-Jordan:
What about adding a new thing to the answer hash in this instance? So instead of changing `correct_choices => ...`, adding a new thing: `correct_values   => [ map { "$_" } $self->value ]`? That would solve things for my particular need.

drgrice1:
Sure.  That should work.

Alex-Jordan:
I replaced the last commit with that, and it is working for PTX. You can try 7.3 [here](https://spot.pcc.edu/~ajordan/temp/checkboxes/section-7.html). (I haven't tweaked things for other exercises on this page yet.)

drgrice1:
@Alex-Jordan:  I put a pull request in to your branch for this pull request with an alternate approach that I would like to be considered instead of adding the `correct_values` key.

drgrice1 (approved):
This looks good.

PR #809, opened 2023-04-13
Revised handling of PERSISTENCE_HASH data

taniwallach:
Persistent problem data - modify to retrieve from `$envir` and to allow update and get operations.
Paired with https://github.com/openwebwork/webwork2/pull/1940
The webwork2 side will now store this in a special field.

### DIFF
--- a/lib/PGcore.pm
+++ b/lib/PGcore.pm
@@ -73,10 +73,14 @@ sub new {
 		POST_HEADER_ARRAY => [],
 		#		PG_ANSWERS                => [],  # holds answers with labels # deprecated
 		#		PG_UNLABELED_ANSWERS      => [],  # holds unlabeled ans. #deprecated -replaced by PG_ANSWERS_HASH
-		PG_ANSWERS_HASH    => {},    # holds label=>answer pairs
-		PERSISTENCE_HASH   => {},    # holds other data, besides answers, which persists during a session and beyond
-		answer_eval_count  => 0,
-		answer_blank_count => 0,
+		PG_ANSWERS_HASH => {},    # holds label=>answer pairs
+
+		# Holds other data, besides answers, which persists during a session and beyond.
+		PERSISTENCE_HASH         => $envir->{PERSISTENCE_HASH} // {},    # Main data, received from DB
+		PERSISTENCE_HASH_UPDATED => {},    # Keys whose updated values should be saved by the DB
+
+		answer_eval_count            => 0,
+		answer_blank_count           => 0,
 		unlabeled_answer_blank_count => 0,
 		unlabeled_answer_eval_count  => 0,
 		KEPT_EXTRA_ANSWERS           => [],
@@ -538,15 +542,25 @@ sub record_unlabeled_array_name {
 }
 
 sub store_persistent_data {    # will store strings only (so far)
-	my $self    = shift;
-	my $label   = shift;
-	my @content = @_;
+	my ($self, $label, @values) = @_;
 	if (defined($self->{PERSISTENCE_HASH}->{$label})) {
 		warn "can' overwrite $label in persistent data";
 	} else {
-		$self->{PERSISTENCE_HASH}->{$label} = join("", @content);    #need base64 encoding?
+		$self->{PERSISTENCE_HASH_UPDATED}{$label} = 1;
+		$self->{PERSISTENCE_HASH}{$label}         = join("", @values);
 	}
 	$label;
+}
+
+sub update_persistent_data {    # will store strings only (so far)
+	my ($self, $label, @values) = @_;
+	$self->{PERSISTENCE_HASH_UPDATED}{$label} = 1;
+	$self->{PERSISTENCE_HASH}{$label}         = join("", @values);
+}
+
+sub get_persistent_data {
+	my ($self, $label) = @_;
+	return $self->{PERSISTENCE_HASH}{$label};
 }
 
 sub check_answer_hash {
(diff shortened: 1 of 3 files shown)


taniwallach:
This is initially intended to allow saving "state" of scaffolds so I can resume work on https://github.com/openwebwork/pg/pull/506 however it will support any other data a problem author might want to remain available.

I think it should provide a better mechanism to support file (image uploads) as envisioned in https://github.com/openwebwork/pg/pull/482 than in the `last_answer` field as currently done there.

The problem author decides when data gets stored into the `PERSISTENCE_HASH` and the webwork2 side will save it after the problem is rendered (and not only on submission of an answer).

lib/PGcore.pm
```diff
@@ -549,6 +552,18 @@ sub store_persistent_data {    # will store strings only (so far)
 	$label;
 }
 
+sub update_persistent_data {    # will store strings only (so far)
+	my $self  = shift;
+	my $label = shift;
+	$self->{PERSISTENCE_HASH}->{$label} = join("", @_);
```
drgrice1, line 558: The arguments should be unpacked on the first line (perlcritic says "Always unpack @_ first at line 555, column 1.  See page 178 of PBP.  (Severity: 4)").  The point is that this makes the expected arguments of the method much clearer.  Also, there is no need for the extra de-referencing arrow.  This should be
```suggestion
	my ($self, $label, @values) = @_;
	$self->{PERSISTENCE_HASH}{$label} = join('', @values);
```

lib/PGcore.pm
```diff
@@ -549,6 +552,18 @@ sub store_persistent_data {    # will store strings only (so far)
...
+}
+
+sub get_persistent_data {
+	my $self  = shift;
+	my $label = shift;
+	return ($self->{PERSISTENCE_HASH}->{$label});
```
drgrice1, line 564: Similar changes here.  This should be
```suggestion
	my ($self, $label) = @_;
	return $self->{PERSISTENCE_HASH}{$label};
```
perlcritic won't complain about this, but it is still better as it makes the expected arguments clear.  Note there is no need for the parenthesis around the return value.  The parenthesis don't add anything.  Note that parenthesis do not turn values into lists or arrays.  The calling context does that.

macros/PG.pl
```diff
@@ -307,7 +307,15 @@ sub ANS_NUM_TO_NAME {
...
-	$PG->store_persistent_data(@_);    #needs testing
+	$PG->store_persistent_data(@_);
+}
+
+sub update_persistent_data {
+	$PG->update_persistent_data(@_);
+}
+
+sub get_persistent_data {
+	return ($PG->get_persistent_data(shift));
```
drgrice1, line 318: Again, there is no need for the parentheses here.

macros/PG.pl
```diff
@@ -307,7 +307,15 @@ sub ANS_NUM_TO_NAME {
 }
 
 sub store_persistent_data {
-	$PG->store_persistent_data(@_);    #needs testing
+	$PG->store_persistent_data(@_);
```
drgrice1, line 310: It would be better to write this as
```suggestion
	my ($label, @values) = @_;
	$PG->store_persistent_data($label, @values);
```
perlcritic will complain about this, but furthermore it makes it clear what the expected arguments are.  This is something that needs to be changed throughout the pg code in numerous places, and has already cause much confusion on how to properly call methods.  It is simple to just pass `@_`, but forces one wanting to use the method to track down the correct arguments by tracing back through the code.

macros/PG.pl
```diff
@@ -307,7 +307,15 @@ sub ANS_NUM_TO_NAME {
...
+}
+
+sub update_persistent_data {
+	$PG->update_persistent_data(@_);
```
drgrice1, line 314: Same changes needed here.  Make the expected arguments clear.

drgrice1 (commented):
This generally looks good, but I have a few suggestions that I think clarify the usage of the new methods added.

lib/PGcore.pm
taniwallach: Thanks for all the comments and suggestions. All were adopted.
There is now also an additional hash `PERSISTENCE_HASH_UPDATED` used to signal to the front end which keys are new or have updated values. That will reduce unnecessary processing and database calls.

lib/PGcore.pm
```diff
@@ -538,17 +542,27 @@ sub record_unlabeled_array_name {
...
-	my $self    = shift;
-	my $label   = shift;
-	my @content = @_;
+	my ($self, $label, @values) = @_;
 	if (defined($self->{PERSISTENCE_HASH}->{$label})) {
 		warn "can' overwrite $label in persistent data";
 	} else {
-		$self->{PERSISTENCE_HASH}->{$label} = join("", @content);    #need base64 encoding?
+		$self->{PERSISTENCE_HASH_UPDATED}->{$label} = 1;
+		$self->{PERSISTENCE_HASH}->{$label}         = join("", @values);
```
drgrice1, line 550: There is one thing that I mentioned before that seems to have been forgotten.  That is the extra dereferencing arrows.  There is no need for the second level of arrows.  Change this to
```suggestion
		$self->{PERSISTENCE_HASH_UPDATED}{$label} = 1;
		$self->{PERSISTENCE_HASH}{$label}         = join("", @values);
```

taniwallach: Sorry - fixed now.

drgrice1 (approved):
This looks good.

lib/PGcore.pm
taniwallach: I will get this change in https://github.com/openwebwork/pg/pull/506 tomorrow.

PR #811, opened 2023-04-15
Initial checkin of sample problems and parsing script

pstaabp:
This branch provides sample problems originally in the OPL with embedded comments in a particular format that can be parsed with a provided script. This was originally discussed in openwebwork/webwork2#1936

The sample problems are in `doc/sample-problems` and there is README.md file with more information especially on the format of the the coding documentation.  These problems originally came from the OPL in the directory `FortLewis/Authoring/Templates`. I only brought over the PGML versions and have updated the code and the comments that were originally on the WeBWorK wiki. 

There is a script `bin/parse-prob-doc.pl` that currently parses the file and outputs the resulting HTML file.   This uses a Mojo::Template to do the formatting as HTML--I used this mainly because it lightweight, fast and is already an installed module.  The template is `bin/prob-template.mt`. 

This currently only takes a directory of pg files, parses and converts to HTML--outputting to a local directory.   I plan to make this recursive to generate the files in an identical directory structure.  I think it's quite crude now--perhaps just a proof of concept.  Other things that are needed

- [x] in each html file, include the full source of the pg file with documentation comments stripped (the script does the, but the template does not have this).  I was thinking either have a link to the source or perhaps a collapsible element in the html page. 
- [x] nicer styling with css in common file
- [x] an index page with links to each file (perhaps searchable)
- [ ] it may be better as a bash script for the file io with corresponding perl module for the parsing. 

Lastly, I think we should have the documentation automatically be moved to another repository (perhaps `pg-docs`) and use GitHub pages to serve the files.  The POD could also live here.  We should be able to have a GitHub action run to do this last step (perhaps not automatically, but somehow triggerable).


somiaj:
@pstaabp Here is my suggestion for the file format/parsing, <https://github.com/somiaj/pg/commit/ab550033d89a1deff9ea7ec96cd568d23c533a94>. The patch is very crude, I was mostly just testing out the fileParser, which to me is working like I envisioned. Also I think we need to remove all of those block comments from the code. I only updated a single problem for this, and won't spend more time on it unless you would like to go this direction.

pstaabp:
@somiaj I decided to have a marker that documentation blocks start with a `#:%name` and then options.  
I used your code as a base and I think simplified and adapted it. Now the parsing code can just walk through the code knowing where it is without the state variable.  

Let me know what you think about this?  All of the files in `doc/Misc` now use this new format and running:
```
perl /opt/webwork/pg/bin/parse-prob-doc.pl --problem_dir=/opt/webwork/pg/doc/sample-problems/Misc/ --out_dir=some_directory
```
creates a folder of html files.

somiaj:
I can see a possible issue (I think) with your changes in which wasn't in my version. Your `#:` blocks no longer have to be continuous (since they wait for the next `#:%` line to start the next block), so you could have markdown paragraphs coming from discontinuous blocks being put together without a new line between them, thus two things that you may want to be separate paragraphs get put together into a single one.

Another thought I have is a bit more tricky. I don't like the way the markdown blocks push against the `#:`. If I were writing these blocks to me the following just looks better.

```
#:%preamble
#: The following modules need to be loaded:
#:
#:   + PGML.pl
#:   + contextLimtedVector.pl
```

To me having that space at the start of `#:` does make things a bit easier to read/write. The trick party is to allow both `#:` and `#: `, and to deal with when that extra space should be removed before sending the code to the markdown processor (unsure how badly one space can mess up the parser).

pstaabp:
> I can see a possible issue (I think) with your changes in which wasn't in my version. Your #: blocks no longer have to be continuous (since they wait for the next #:% line to start the next block), so you could have markdown paragraphs coming from discontinuous blocks being put together without a new line between them, thus two things that you may want to be separate paragraphs get put together into a single one.

Well if they don't have a `#:` then the line will be added to the code block instead.  This would probably just add another blank like. 

> Another thought I have is a bit more tricky. I don't like the way the markdown blocks push against the #:. 

I don't think everything needs to be bumped up again the `#:`.  I just tested your example, and it works fine.  Any non `#:%` line would be interpreted as markdown with `#:` stripped from the beginning of the line  (Line 59 of the script) and should preserve space.

somiaj:
> Well if they don't have a `#:` then the line will be added to the code block instead. This would probably just add another that line to the code part.

What I mean is something like this:

```
#%section
#: description of some stuff
code
code
#: more description
more code
more code
```

The second `#:` line just gets tacked onto the previous description without a new line between them, thus markdown will see this as a single paragraph, and it should probably be seen as two paragraphs.

> I don't think everything needs to be bumped up again the `#:`. I just tested your example, and it works fine. Any non `#:%` line would be interpreted as markdown with `#:` stripped from the beginning of the line (Line 59 of the script) and should preserve space.

The space being persevered is the issue that may arise (since spaces mean things to markdown). If you just remove `#:` from the start of the lines, then if someone adds a space for readabilty reasons the markdown block after parsed would look something like:

```
 line1
 line2

     Indented line

 line3
```

So every line has a space in front if it. I think these should be stripped (But maybe the markdown already nicely deals with this so this is a non-issue)

somiaj:
Note, the first issue about non-continuous markdown blocks is not really an issue if we just make sure the files are formatted correctly to begin with, just something I noticed is all.

pstaabp:
Added more problems and updated the script to crawl through the `doc/sample-problems` directory and builds an html file for each problem.  There is also now a `categories` directory (in the output directory) with an index and a list of problems for each category.  

Added a list styling with bootstrap and code highlighting with [highlight.js](https://highlightjs.org).

Next step is to get this script working as a github action that will post to [openwebwork.github.io/pg-docs](openwebwork.github.io/pg-docs)

Another goal is to automatically link POD documentation for non-standard macros.

pstaabp:
I have a working version of the sample problems in a GitHub page.   TLDR: go to [https://pstaabp.github.io/pg-docs/](https://pstaabp.github.io/pg-docs/). 

Currently, I'm testing this on my repository, but the plan will be for this to go to openwebwork.github.io/pg-docs.  

I got this working by two GitHub actions in my pg-docs repository (right now in the `test-workflow` branch). 
* first action called "Create Sample Problems Docs"
  - copies the pg repository (currently the `sample-problems-doc` branch on my repository)
  - runs the `bin/parse-prob-doc.pl` script in `PG/bin` 
  - commits the generated html code. 
* second action called "Deploy static files" copies the contents to the GitHub pages site.  

Additionally, I added a copy to clipboard for each of the code blocks.  I will create a copy entire field to clipboard action as well.

pstaabp:
Added some more features. TLDR: https://pstaabp.github.io/pg-docs/

* the pod is now created and posted on the GitHub.io page. 
* the script parses each sample problem and automatically links to macro files (except PGStandard, PGML, ...) 

If this is looking reasonable, I'll move it to: https://openwebwork.github.io/pg-docs/

drgrice1 (commented):
There are a couple of things I notice that should be changed.

The `.mt` extension is not a standard extension for Mojolicious template files.  Please change those to `.html.ep` which is the standard for html Mojolicious template files.  Otherwise syntax highlighters for Mojolicious template files are not active by default.

Please make the parse-prob-doc.pl file executable.

drgrice1:
Also, clean up category_layout.mt (and rename it to category_layout.html.ep).  The formatting in that file is inconsistent with the formatting of the other template files.

drgrice1:
I just noticed that the actual filename you used is `catagory_layout.mt`.  Also fix the misspelled "catagory".

drgrice1:
I see where you got the `.mt` extension now.  I am accustomed to using Mojo::Templates via a Mojolicious app which uses `.ep` templates.  If used directly, they use the `.mt` extension.  So you can stick with that extension.

somiaj:
@pstaabp looked around a bit, and looking good.  Two things I noticed, first is I don't really like the syntax highlighting, it really isn't catered towards pg and PGML files.  I wonder if there is a way we could use the PGCodeMirror syntax highlighting to match the webwork2 editor.  Second lots of `TEXT(beginproblem());` lines that could be removed from the example files.

drgrice1:
Even before considering pg and PGML syntax, highlight.js does not seem to do well even for syntax highlighting of Perl.  It doesn't even have a different color for variables.

Not to mention that the color scheme that highlight.js uses has low foreground/background contrast, and is far from accessible.

pstaabp:
@drgrice1 Will cleanup the templates.  And remove `TEXT(beginproblem());` and unnecessary `Context('Numeric')` calls from examples.  

I was wondering if codeMirror might be too heavyweight for this since we are not doing editing, but the advantage is that we have the PGML formatting for this.  I'll try switching over to that.

somiaj:
@pstaabp Seems CodeMirror5 has a runmode addon, <https://codemirror.net/5/doc/manual.html#addon_runmode>, which can use CodeMirror to use its tools to syntax highlight a text box, but not invoke any of the editor functionality.

somiaj:
Looking closer, also check the option right under that one, `runmode/colorize.js`, which is described as:

> By default, it'll get all pre tags. Will read the data-lang attribute of these nodes to figure out their language, and syntax-color their content using the relevant CodeMirror mode

drgrice1:
@pstaabp: I added a pull request to your branch with some improvements to the layout.  The main change is instead of generating an index and another copy of that index for each of the categories, a single index file is generated in the root output location with all of it.  Bootstrap tabs are used for the categories instead of separate pages.

somiaj:
One issue that occurred to me with using CodeMirror is the `PG.js` and all of the codemirror stuff is in the `webwork2` repo, not the `pg` one.

drgrice1:
@pstaabp:  My pull request now also includes a switch from using highlight.js to using CodeMirror and its standalone runmode.  For now I just copied the PG.js file from webwork2 into docs/js.

drgrice1:
I noticed that since I switched to CodeMirror's runmode, it is removing all indentation in the problems.  I am not sure why it is doing that, but that is a problem.

drgrice1:
Actually, it is only removing indentation with actual tab characters.  I was doing some clean up and perltidying problems.  So you probably won't see this.  So we need to make sure to use spaces for indentation in problems.

somiaj:
Could the script that loads files also switch tabs for spaces just to be safe, `s/\t/    /g`?

pstaabp:
> Actually, it is only removing indentation with actual tab characters. I was doing some clean up and perltidying problems. So you probably won't see this. So we need to make sure to use spaces for indentation in problems.

@drgrice1  I'll 

> Could the script that loads files also switch tabs for spaces just to be safe, `s/\t/ /g`?

@somiaj good idea.

pstaabp:
Another update.  Instead of just a listing of sample problems by categories, there is a listing by problem techniques and subject area like on the wiki.  

Also, added more files--all of the current files on "sample Problems" from the wiki are here and still the "problem techniques" need to merge over.

drgrice1:
Please remove the "Precalc" directory and "Precalculus" subject/category.  Precalculus is not a mathematical "subject".  Precalculus is a "course" offered at many institutions that is designed to prepare students for Calculus and contains a mix of topics from actual mathematical subjects (mostly algebra and trigonometry).  Move the files in that directory to the correct location (in this case they are all algebra problems).  This also needs to be done with the OPL.  This makes it hard for instructors to find these problems because they are incorrectly categorized.

dlglin:
I thought that all of chapter/sections in the "Precalculus" subject in the OPL were aliases to other places.  I agree that "Precalculus" is not a subject in itself, but I don't mind using it as a pointer to things in their canonical homes.

pstaabp:
About the Precalc/Algebra problems, I have labelled most of those problems in both subject areas--similar to the pointer in the OPL. I tried to think about broad ways of labelling/categorizing.

drgrice1:
It is okay to have the "Precalculus" subject I guess, even though it is not a proper subject.  Just remove the directory, and move the files where they belong.

pstaabp:
> It is okay to have the "Precalculus" subject I guess, even though it is not a proper subject. Just remove the directory, and move the files where they belong.

This started with Paul Pearson's problems which had that directory structure. I'll update.

drgrice1:
@pstaabp:  I added another pull request to this branch with several improvements.  It fixes the broken layouts on narrow screens, switches to using Pandoc instead of Text::MultiMarkdown, adds some more problems, and cleans up many other problems.

pstaabp:
Another update including:
* default page for each of the doc views (css styling isn't great on sidebar though)
* default documentation for blank sections
* link to full problem for download in each page
* Move css to separate file
* Switch to a light code mirror theme
* Adding a few more sample problems and cleanup

drgrice1:
@pstaabp: I added another pull request to this branch that fixes some things that were broken in recent revisions, and fixes some issues in the graph tool problem documentation.

drgrice1:
I notices that there are several problems that put the answer rules into integrals that are calling `Context()->texStrings` and `Context()->normalStrings` before and after the tables with the answer rules are created.  Those should be removed.  The calls are not needed for what is done there, and in fact don't do anything there.

drgrice1:
In fact `Context()->texStrings` and `Context()->normalStrings` should never appear in any of the sample problems.  Those are arcane incantations that are not needed anymore, and have no place in a PGML problem.

pstaabp:
> In fact `Context()->texStrings` and `Context()->normalStrings` should never appear in any of the sample problems. Those are arcane incantations that are not needed anymore, and have no place in a PGML problem.

I'll clean up the problems for these as well.  I didn't think they were needed anymore.

pstaabp:
Another update to this.  A lot of this is because of openwebwork/webwork2#2066
- I merged pstaabp/pg#19 from @drgrice1 
- split apart the `parse-prob-doc.pl` file and placed many of the subs in a new module.  This allows it to be reused. 
- consolidated all of the sidebar and main template files for the sample problems into a single one.

doc/sample-problems/DiffEq/HeavisideStep.pg
```diff
@@ -0,0 +1,117 @@
...
+## Date(06/01/2015)
+## Institution(Hope College)
+## Author(Paul Pearson)
+## MO(1)
+## KEYWORDS('differential equations', 'adding the Heaviside function to the context')
+
+#:% name = Heaviside or Step Function
+#:% type = Sample
+#:% subject = differential equations
+#:% categories = [heaviside/step]
```
drgrice1, line 17: Since these categories are used for html id's they can not have forward slashes in them.  Those are invalid characters for id's.  So this will need to be `heaviside-step`.

pstaabp: okay

Alex-Jordan:
Note to test, need to install `Pandoc` perl module.

Alex-Jordan:
I clicked "Sample Problems" then clicked "Sample problems by cateegory" and then clicked "difference quotient" and then "Difference Quotient" and it just hangs.

drgrice1:
Yeah, I added the `Pandoc` perl module (and `pandoc` external program needed by that module) to `check_modules.pl` for webwork2 in openwebwork/webwork2#2066.

Alex-Jordan:
OK, now it goes through after I installed `pandoc`.

drgrice1:
If openwebwork/webwork2#2066 is merged before the release (it probably will be?), then the installation instructions will need to be updated.

Alex-Jordan:
So I can get to this page:
Screen Shot 2023-06-30 at 3 51 40 PM

and I have refreshed the javascript dependencies. But clicking a clipboard does nothing (there is nothing new in my clipboard to paste). I have some console errors:

```
ReferenceError: utilsService is not defined
[main.js:1:65](moz-extension://7beb0d23-df3f-4237-be5e-5e794b4d4225/content_scripts/main.js)
Uncaught TypeError: navigator.clipboard is undefined
    <anonymous> http://vmwebworkdevw02.pcc.edu/webwork2_files/js/SampleProblemViewer/sample-problem.a75fd6a6.min.js:1
```

drgrice1:
What browser are you using?  Although it shouldn't matter.  @pstaapb is using `navigator.clipboard` and the clipboard api that is standard in all browsers.  Safari was (as usual) the latest to adopt it.  That was in 2020 for version 13.1.

Alex-Jordan:
Firefox 114.0.2 on Mac.

Alex-Jordan:
Not that it matters but I had the wrong screenshot. I edited that post with what I meant.

drgrice1:
I am not sure if your screenshot is what you meant.  There aren't any clipboard icons directly on the PG editor page.  I assume you have navigated further to a sample problem.  In any case, the issue may be a permissions issue.  The clipboard api requires the user to grant read or write permissions it seems.  I am not sure though.  It seems that the `navigator.clipboard` object would still be defined though.

drgrice1:
Another thing that might be involved is that according to some sources `navigator.clipboard` is only available in secure contexts (via https).

Alex-Jordan:
> I am not sure if your screenshot is what you meant. 

It wasn't. If you view the post in GitHub, I corrected it there.

> Another thing that might be involved is that according to some sources navigator.clipboard is only available in secure contexts (via https).

This sounds like it's my issue. I am testing on a firewalled server, where there is no https.

drgrice1:
@Alex-Jordan: What if you try the clipboard icons at https://openwebwork.github.io/pg-docs/sample-problems/problem-techniques/AdaptiveParameters.html?  Do they work for you on that site?  That could be an indicator that this is probably the https issue.

Alex-Jordan:
> Do they work for you on that site?

Yes if I go there, they work.

Now that I think about it, I think I have seen this issue before with PTX HTML. On a page like [this](https://pretextbook.org/examples/sample-book/noparts/html/section-note-on-proofs.html), to the left of paragraphs, etc there is a faint button/link for copying a permalink to the clipboard. The javascript in use for that does not work if thee page is served http.

drgrice1:
Perhaps we should add a check if the site is secure, and hide the buttons if not.

bin/parse-problem-doc.pl
```diff
@@ -0,0 +1,162 @@
...
+use Mojo::Template;
+use File::Basename qw(fileparse basename);
+use Getopt::Long;
+use File::Copy qw(copy);
+use YAML::XS qw(DumpFile);
+use Pod::Simple::Search;
+
+use SampleProblemParser qw(parseSampleProblem generateMetadata);
+
+my $problem_dir = "$pg_root/doc/sample-problems";
```
drgrice1, line 26: This line needs to be updated for the new location.  I am not sure if you noticed that I added this.  This makes it so that you don't need to pass the `problem_dir` option if you are running this script on the default location.

tutorial/sample-problems/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+All lines following the documentation lines are considered code until the next `#:%` line is encountered.
+
+## Generate the documentation
+
+The documentation is generated with the `parse-prob-doc.pl` script in the `bin`
+directory of pg. There are the following options (and many are required):
+
+- `problem_dir` or `d`:  The directory where the sample problems are.  This is
+`PG_ROOT/doc/sample-problems`, but required to add.
```
drgrice1, line 79: This also needs to be updated for the new location.

drgrice1: Also, this is no longer required.  I didn't notice this, or I would have changed this before.

pstaabp:
Updated with suggestions above and @drgrice1, I pushed in a new exponential problem that uses CSS styling to do it instead of niceTables.pl.

drgrice1:
@pstaabp: I added another pull request to this branch that updates one problem and fixes another.

tutorial/templates/problem-template.mt
```diff
@@ -0,0 +1,110 @@
...
+	</div>
+
+	<script type="module">
+		for (const pre of document.body.querySelectorAll('pre.CodeMirror')) {
+			CodeMirror.runMode(pre.textContent, 'PG', pre);
+		}
+
+		for (const btn of document.querySelectorAll('.clipboard-btn')) {
+			if (typeof navigator.clipboard === 'undefined') btn?.remove();
+			btn.addEventListener('click', () => navigator.clipboard.writeText(btn.dataset.code));
```
drgrice1, line 105: This should probably be
```suggestion
			if (typeof navigator.clipboard === 'undefined') btn?.remove();
			else btn.addEventListener('click', () => navigator.clipboard.writeText(btn.dataset.code));
```
It would not be a good idea to attempt to add an event listener to a btn that was just removed.

drgrice1, line 105: Another thing to note is that you don't need to use `typeof`.  You could just do
```suggestion
			if (navigator.clipboard) btn.addEventListener('click', () => navigator.clipboard.writeText(btn.dataset.code));
			else btn?.remove();
```

tutorial/sample-problems/Misc/Matching.pg
```diff
@@ -0,0 +1,126 @@
...
+#:
+#: If a grader is desired that awards full credit when all questions are correct
+#: and no credit otherwise, use the commented out standard problem grader code
+#: instead.
+#:
+#: Create a list of 6 questions and answers, 2 extra answers, and a
+#: 'None of the above' answer that will be made last with `makeLast`.
+#: So the popup list must have 9 entries A through I.
+#:
+#: As an alternative, see PROBLINK('MatchingAlt.pl') for another way to write
```
drgrice1, line 51: The `PROBLINK` call should be `PROBLINK('MatchingAlt.pg')`.  When you run the `parser-problem-doc.pl` script or view the problem in webwork2 it gives a "use of uninitialized value in concatentation (.) or string" warning from this.

drgrice1 (approved):
I think this pull request is looking pretty good now.  I have not had a chance to look over and test the latest sample problems that were added, but I see that they aren't giving any errors at least when the `parser-problem-doc.pl` script is run.  So I will for now accept them as they are.

PR #812, opened 2023-04-16
Replace bin/run-perltidy with bin/run-perltidy.pl

drgrice1:
The primary advantage of this perl script over the previous shell script is that it used File::Find instead of relying on the extglob, globstar, and nullglob shell options.  Those options are not supported by all shells.

If this script is run with no arguments, then it does the same thing that the previous shell script did, and runs perltidy on all .pm, .pl, and .t files in the PG_ROOT directory.

If file names are passed on the command line, then perltidy is only executed on the passed filenames.  Another advantage of this script over the previous shell script.

The script enforces the use of the .perltidyrc file in the PG_ROOT directory.  It also makes the default for -bext to be '/', which means that backup files are created with the .bak extension, and deleted if no errors occur.  That can be overridden if backups are desired.  Other that those things, all options accepted by perltidy are also accepted by this script.

One other thing that this script does is enforces that the version of perltidy used is version 20220613 or newer.  Older versions have an issue with the -vxl='q' option.

This also fixes the reverse translation of BEGIN_LATEX_IMAGE in the perltidy-pg.pl script.

### DIFF
--- a/bin/perltidy-pg.pl
+++ b/bin/perltidy-pg.pl
@@ -82,7 +82,7 @@ my $postfilter = sub {
 	$evalString =~ s/\h*SOLUTION\(EV3P\(<<'END_SOLUTION'\)\);/BEGIN_SOLUTION/g;
 	$evalString =~ s/\h*HINT\(EV3P\(<<'END_HINT'\)\);/BEGIN_HINT/g;
 	$evalString =~ s/(.*)->tex\(<<END_TIKZ\);/$1->BEGIN_TIKZ/g;
-	$evalString =~ s/(.*)->tex\(<<END_LATEX_IMAGE\);/(.*)->BEGIN_LATEX_IMAGE/g;
+	$evalString =~ s/(.*)->tex\(<<END_LATEX_IMAGE\);/$1->BEGIN_LATEX_IMAGE/g;
 
 	# Care is needed to reverse the preprocessing here.
 	# First in all occurences of an odd number of backslashes the first backslash is replaced with two tildes.
--- a/bin/run-perltidy
+++ b/bin/run-perltidy
@@ -1,7 +1,0 @@
-#!/bin/bash
-
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null 2>&1 && pwd)"
-INPUTDIR=$(dirname $SCRIPT_DIR)
-
-shopt -s extglob globstar nullglob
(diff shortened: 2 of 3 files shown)


pstaabp:
Looks good.  It might be nice to have a verbose flag that turns on the "Tidying file: ..." line instead of always printing that.

pstaabp (approved):

Alex-Jordan:
I can't run this on Oracle:
```
alex.jordan@vmwebworkdevw02:/opt/webwork/pg/bin $ ./run-perltidy.pl 
Version 20220613 or newer of perltidy is required for this script.
The installed version is 20180220.
```

And `yum update perltidy` is not changing things. I suppose I should manually install a recent perltidy, but it could be something to note in installation instructions.

drgrice1:
Can you install a newer version Perl::Tidy from cpan? Versions prior to the one the script asks for won't work right with our .perltidyrc file.

This is a developer script, so I don't think that a note in the installation instructions is merited.

Alex-Jordan:
OK, I did as you suggested and the script works.

However, it actually _worked_. It changed one file:

```
diff --git a/macros/math/draggableSubsets.pl b/macros/math/draggableSubsets.pl
index 95d63360..56082916 100644
--- a/macros/math/draggableSubsets.pl
+++ b/macros/math/draggableSubsets.pl
@@ -354,8 +354,7 @@ sub cmp_preprocess {
                $ans->{preview_latex_string} = join(
                        ',',
                        map {
-                               "\\{\\text{"
-                                       . join(',', map { $self->{shuffledSet}[$_] } grep { $_ >= 0 } @{ $_->{data} }) . "}\\}"
+                               "\\{\\text{" . join(',', map { $self->{shuffledSet}[$_] } grep { $_ >= 0 } @{ $_->{data} }) . "}\\}"
                        } @{ $ans->{student_value}{data} }
                );
        }
```

Should that not be happening?

drgrice1:
I saw that too with the newer versions of Perl::Tidy.  Only the draggableSubsets.pl file is changed out of all of the pg and webwork2 code.  I thought about specifically locking the version of Perl::Tidy to 20220613 because of that.   This is one thing that is a bit annoying with the various revisions of Perl::Tidy.  Prior to version 20220613 qw quotes were not aligned by default.  Then with that version they made it so that they were by default, and so we added the -vxl='q' option.  However, that option didn't exist for prior versions and so it causes an error for them.  Now they have changed what happens with things being turned into one line a little bit.

Alex-Jordan:
OK, I did that and all is clean now. With your implied approval plus mine and @pstaabp's, merging this.

PR #813, opened 2023-04-17
linear relation context

Alex-Jordan:
This is an evolution of `macros/parsers/parserLinearInequality.pl` to `macros/contexts/contextLinearRelation.pl`.

* I'm not sure about this, but I feel that this macro is a context, not a parser. For one thing, you have to load the "LinearRelation" context to be able to make a "LinearRelation" MathObject.
* This handles `!=` which was left out of `parserLinearInequality.pl`.
* It handles ≤, ≥, and ≠ if "alternative" parsing is in the configuration.
* It prints formula inequalities the way they were entered. So `x+1=y+2` stays that way, instead of printing as `x-y=1`. Unless you set a context flag for standard form.

I'd like to add a method where you test a LinearRelation object at a given point, and it returns true/false. But I'm punting for now, out of time for today.

Sample problem:

```
DOCUMENT();
loadMacros(
    "PGstandard.pl",
    "PGML.pl",
    "contextLinearRelation.pl",
);

Context("LinearRelation")->flags->set(reduceConstants=>0);


BEGIN_PGML
# Not Converted to Standard Form
END_PGML

for my $sign ('=', '<', '>', '<=', '=<', '≤', '>=', '=>', '≥', '!=', '<>', '≠') {
$pointvector = LinearRelation([1,1,1], [1,2,3], $sign);
$formula = Compute("pi x + 1/3 y $sign 2+z");

BEGIN_PGML
math: [`[$pointvector]`]  
string: [$pointvector]

math: [`[$formula]`]  
string: [$formula]
----
END_PGML
}

Context()->flags->set(standardForm => 1);

BEGIN_PGML
# Converted to Standard Form
END_PGML

for my $sign ('<', '>', '<=', '=<', '≤', '>=', '=>', '≥', '!=', '<>', '≠') {
$pointvector = LinearRelation([1,1,1], [1,2,3], $sign);
$formula = LinearRelation("pi x + 1/3 y $sign 2+z");

BEGIN_PGML
math: [`[$pointvector]`]  
string: [$pointvector]

math: [`[$formula]`]  
string: [$formula]
----
END_PGML
}

ENDDOCUMENT();
```

### DIFF
--- a/macros/contexts/contextLinearRelation.pl
+++ b/macros/contexts/contextLinearRelation.pl
@@ -1,0 +1,369 @@
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2022 The WeBWorK Project, https://github.com/openwebwork
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+# Artistic License for more details.
+################################################################################
+
+=head1 NAME
+
+contextLinearRelation.pl - Implement linear relations.
+
+=head1 DESCRIPTION
+
+This gives a context C<LinearRelation> that implements an implicit linear relations
+MathObject using =, <, >, <=, >=, or !=. Activate the context with:
+
+    Context("LinearRelation");
+
+Use C<LinearRelation(formula)>, C<Formula(formula)>, or C<Compute(formula)> to 
+to create a LinearRelation object using a string formula. Alternatively, use
+C<LinearRelation(point,vector,sign> where C<point> and C<vector> are array references
+and C<sign> is one of the (in)equality symbols.
+
+Usage examples:
+
+    $R = LinearRelation("x +y + 2z <= 5");
+    $R = Formula("x +y + 2z <= 5");
+    $R = Compute("x +y + 2z <= 5");
+    $R = LinearRelation([1,2,1], [1,1,2], "<=");
+
+Sloppy inequality signs (=<, =>, <>) may be used.
+
+If C<$pg{specialPGEnvironmentVars}{parseAlternatives}> is true in your configuration,
+then you may also work directly with the characters ≤, ≥, and ≠.  
+
+By default, the context has three variables x, y, and z. You should explicitly set
+the variables if your situation should use something different. For example with
+
+    Context()->variables->are(m => 'Real', n =>'Real');
+
+There is one special context flag.
+
+=over 
+
+=item S<C<< standardForm >>>
+
+This determines whether something like C<<LinearRelation("x+2 < y+z")>> will be
+displayed as C<<x+2 < y+z>> or converted to standard form: C<<x-y-z < -2>>.
+It is 0 by default.
+
+=back 
+
+=cut
+
+loadMacros('MathObjects.pl');
+
+sub _contextLinearRelation_init { LinearRelation::Init() };    # don't reload this file
+
+##################################################
+#
+#  Initialize the contexts and make the creator function.
+#
+
+package LinearRelation;
+our @ISA = qw(Value::Formula);
+
+sub Init {
+	my $context = $main::context{LinearRelation} = Parser::Context->getCopy("Vector");
+	$context->{name}                       = "LinearRelation";
+	$context->{precedence}{LinearRelation} = $context->{precedence}{special};
+	$context->{value}{LinearRelation}      = "LinearRelation";
+	$context->{value}{Formula}             = "LinearRelation::formula";
+	$context->parens->remove("<");
+	$context->flags->set(standardForm => 0);
+	$context->operators->add(
+		'=' => {
+			precedence    => .5,
+			associativity => 'left',
+			type          => 'bin',
+			string        => ' = ',
+			kind          => 'eq',
+			class         => 'LinearRelation::inequality',
+			formulaClass  => "LinearRelation"
+		},
+
+		'<' => {
+			precedence    => .5,
+			associativity => 'left',
+			type          => 'bin',
+			string        => ' < ',
+			kind          => 'lt',
+			class         => 'LinearRelation::inequality',
+			formulaClass  => "LinearRelation"
+		},
+		'>' => {
+			precedence    => .5,
+			associativity => 'left',
+			type          => 'bin',
+			string        => ' > ',
+			kind          => 'gt',
+			reverse       => 'lt',
+			class         => 'LinearRelation::inequality',
+			formulaClass  => "LinearRelation"
+		},
+
+		'<=' => {
+			precedence    => .5,
+			associativity => 'left',
+			type          => 'bin',
+			string        => ' <= ',
+			TeX           => '\le ',
+			kind          => "le",
+			class         => 'LinearRelation::inequality',
+			formulaClass  => "LinearRelation",
+			alternatives  => ["\x{2264}"]
+		},
+		'=<' => {
+			precedence    => .5,
+			associativity => 'left',
+			type          => 'bin',
+			string        => ' <= ',
+			TeX           => '\le ',
+			kind          => "le",
+			class         => 'LinearRelation::inequality',
+			formulaClass  => "LinearRelation"
+		},
+
+		'>=' => {
+			precedence    => .5,
+			associativity => 'left',
+			type          => 'bin',
+			string        => ' >= ',
+			TeX           => '\ge ',
+			kind          => 'ge',
+			reverse       => 'le',
+			class         => 'LinearRelation::inequality',
+			formulaClass  => "LinearRelation",
+			alternatives  => ["\x{2265}"]
+		},
+		'=>' => {
+			precedence    => .5,
+			associativity => 'left',
+			type          => 'bin',
+			string        => ' >= ',
+			TeX           => '\ge ',
+			kind          => 'ge',
+			reverse       => 'le',
+			class         => 'LinearRelation::inequality',
+			formulaClass  => "LinearRelation"
+		},
+
+		'!=' => {
+			precedence    => .5,
+			associativity => 'left',
+			type          => 'bin',
+			string        => ' != ',
+			TeX           => '\ne ',
+			kind          => 'ne',
+			reverse       => 'ne',
+			class         => 'LinearRelation::inequality',
+			formulaClass  => "LinearRelation",
+			alternatives  => ["\x{2260}"]
+		},
+		'<>' => {
+			precedence    => .5,
+			associativity => 'left',
+			type          => 'bin',
+			string        => ' != ',
+			TeX           => '\ne ',
+			kind          => 'ne',
+			reverse       => 'ne',
+			class         => 'LinearRelation::inequality',
+			formulaClass  => "LinearRelation"
+		}
+	);
+	main::PG_restricted_eval('sub LinearRelation {Value->Package("LinearRelation()")->new(@_)}');
+}
+
+sub new {
+	my $self    = shift;
+	my $class   = ref($self) || $self;
+	my $context = (Value::isContext($_[0]) ? shift : $self->context);
+	my $formula = "Value::Formula";
+	return shift if scalar(@_) == 1 && ref($_[0]) eq 'LinearRelation';
+	$_[0] = $context->Package("Point")->new($context, $_[0])  if ref($_[0]) eq 'ARRAY';
+	$_[1] = $context->Package("Vector")->new($context, $_[1]) if ref($_[1]) eq 'ARRAY';
+
+	my ($p, $N, $plane, $vars, $d);
+	if (scalar(@_) >= 2 && Value::classMatch($_[0], 'Point', 'Vector') && Value::classMatch($_[1], 'Vector')
+		|| Value::isRealNumber($_[1]))
+	{
+		#
+		# Make a plane from a point and a vector and optionally a
+		# symbol <= or = or =>,
+		# e.g. LinearRelation($point, $vector, '<=');
+		# or from a list of coefficients and the constant
+		# e.g. LinearRelation($dist, [3,5,6], '<=')
+		# one can optionally add new Context
+		# variables ($point, $vector, '<=',[qw(a1,a2,a3)])
+		$p = shift;
+		$N = shift;
+		my $bop = shift || "=";
+		if (Value::classMatch($N, 'Vector')) {
+			$d = $p . $N;
+		} else {
+			$d = $context->Package("Real")->make($context, $N);
+			$N = $context->Package("Vector")->new($context, $p);
+		}
+		$vars = shift || [ $context->variables->names ];
+		$vars = [$vars] unless ref($vars) eq 'ARRAY';
+		my @terms = ();
+		my $i     = 0;
+		foreach my $x (@{$vars}) { push @terms, $N->{data}[ $i++ ]->string . $x }
+		$plane = $formula->new(join(' + ', @terms) . $bop . $d->string)->reduce(@_);
+	} else {
+		#
+		#  Determine the normal vector and d value from the equation
+		#
+		$plane = shift;
+		$plane = $formula->new($context, $plane) unless Value::isValue($plane);
+		$vars  = shift || [ $context->variables->names ];
+		$vars  = [$vars] unless ref($vars) eq 'ARRAY';
+		Value::Error("Your formula doesn't look like a linear inequality")
+			unless $plane->type eq 'Equality';
+		#
+		#  Find the coefficients of the formula
+		#
+		my $f = ($formula->new($context, $plane->{tree}{lop}) - $formula->new($context, $plane->{tree}{rop}))->reduce;
+		my $F = $f->perlFunction(undef, [ @{$vars} ]);
+		my @v = split('', '0' x scalar(@{$vars}));
+		$d = -&$F(@v);
+		my @coeff = (@v);
+		foreach my $i (0 .. scalar(@v) - 1) { $v[$i] = 1; $coeff[$i] = &$F(@v) + $d; $v[$i] = 0 }
+		$N = Value::Vector->new([@coeff]);
+		$plane =
+			$self->new($N, $d, $plane->{tree}{bop}, $vars, '-x=-y' => 0, '-x=n' => 0)
+			->with(original_formula => $plane, original_formula_latex => $plane->TeX);
+		#
+		#  Check that the student's formula really is what we thought
+		#
+		Value::Error("Your formula isn't a linear one")
+			unless ($formula->new($plane->{tree}{lop}) - $formula->new($plane->{tree}{rop})) == $f;
+	}
+	Value::Error("The equation of a linear inequality must be non-zero somewhere")
+		if ($N->norm == 0);
+	$plane->{d} = $d;
+	$plane->{N} = $N;
+	return bless $plane, $class;
+}
+
+#
+#  We already know the vectors are non-zero, so check
+#  if the equations are multiples of each other.
+#
+sub compare {
+	my ($self, $l, $r) = Value::checkOpOrder(@_);
+	$l = new LinearRelation($l) unless ref($l) eq ref($self);
+	$r = new LinearRelation($r) unless ref($r) eq ref($self);
+	my ($lN, $ld, $ltype, $lrev) = ($l->{N}, $l->{d}, $l->{tree}{def}{kind}, $l->{tree}{def}{reverse});
+	my ($rN, $rd, $rtype, $rrev) = ($r->{N}, $r->{d}, $r->{tree}{def}{kind}, $r->{tree}{def}{reverse});
+
+	#
+	#  Reverse inequalities if they face the wrong way
+	#
+	($lN, $ld, $ltype) = (-$lN, -$ld, $lrev) if $lrev;
+	($rN, $rd, $rtype) = (-$rN, -$rd, $rrev) if $rrev;
+
+	#
+	#  First, check if the type of inequality is the same.
+	#  Then check if the dividing (hyper)plane is the right one.
+	#
+	return 1 unless $ltype eq $rtype;
+	# 'samedirection' is the second optional input to isParallel
+	return $lN <=> $rN
+		unless $lN->isParallel($rN, ($ltype ne 'eq' && $ltype ne 'ne'))
+		;    # inequalities require same direction which is the second optional input to isParallel
+	return $ld <=> $rd if $rd == 0 || $ld == 0;
+	return $rd * $lN <=> $ld * $rN;
+}
+
+sub cmp_class {'a Linear Relation'}
+sub showClass { shift->cmp_class }
+
+sub cmp_defaults { (
+	Value::Real::cmp_defaults(shift),
+	ignoreInfinity => 0,    # report infinity as an error
+) }
+
+#
+#  Only compare two equalities
+#
+sub typeMatch {
+	my $self  = shift;
+	my $other = shift;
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I added a `check_at` method. So for example you can do:

```
# In default context with 3 variables x, y, z...
$LR = LinearRelation("x + 2y <= 3");
$LR->check_at([1, 1, 1]);  # returns true
$LR->check_at(Point(1, 2, 3));  # returns false
$LR->check_at(Vector(1, 2));  # error, because the context has three variables, not two
```

macros/contexts/contextLinearRelation.pl
```diff
@@ -0,0 +1,399 @@
...
+			type          => 'bin',
+			string        => ' >= ',
+			TeX           => '\ge ',
+			kind          => 'ge',
+			reverse       => 'le',
+			class         => 'LinearRelation::inequality',
+			formulaClass  => "LinearRelation",
+			alternatives  => ["\x{2265}"]
+		},
+		'=>' => {
```
pstaabp, line 161: I'm not sure I like this.  I know this is an 'extra' way of doing '>=', but just looks like a perl hash.  Do any other language allow greater than or equal like this.   Also, same comment on `=<`.

macros/contexts/contextLinearRelation.pl
```diff
@@ -0,0 +1,399 @@
...
+	{
+		#
+		# Make a plane from a point and a vector and optionally a
+		# symbol <= or = or =>,
+		# e.g. LinearRelation($point, $vector, '<=');
+		# or from a list of coefficients and the constant
+		# e.g. LinearRelation($dist, [3,5,6], '<=')
+		# one can optionally add new Context
+		# variables ($point, $vector, '<=',[qw(a1,a2,a3)])
+		$p = shift;
```
pstaabp, line 221: We're trying to get away from using shift to pull arguments lower in a subroutine.  
On line 203 or 204 ` my ($p, $N, $bop, $vars) = @_;`

macros/contexts/contextLinearRelation.pl
```diff
@@ -0,0 +1,399 @@
...
+sub cmp_defaults { (
+	Value::Real::cmp_defaults(shift),
+	ignoreInfinity => 0,    # report infinity as an error
+) }
+
+#
+#  Only compare two equalities
+#
+sub typeMatch {
+	my $self  = shift;
```
pstaabp, line 314:
```
my ($self, $other, $ans) = @_;
```

pstaabp:
With your example problem, I'm seeing the following error: 
```
ERRORS from evaluating PG file: Unexpected character '≤'; see position 13 of formula at line 2 of (eval 13520) from within LinearRelation::new called at line 2 of (eval 13520) from within main::LinearRelation called at line 17 of setlinear_relation/example.pg
```
and this error occurs for '≥' and '≠' as well.  I assume these are unicode.

macros/contexts/contextLinearRelation.pl
```diff
@@ -0,0 +1,399 @@
...
+			$d = $p . $N;
+		} else {
+			$d = $context->Package("Real")->make($context, $N);
+			$N = $context->Package("Vector")->new($context, $p);
+		}
+		$vars = shift || [ $context->variables->names ];
+		$vars = [$vars] unless ref($vars) eq 'ARRAY';
+		my @terms = ();
+		my $i     = 0;
+		foreach my $x (@{$vars}) { push @terms, $N->{data}[ $i++ ]->string . $x }
```
pstaabp, line 234: foreach -> for

macros/contexts/contextLinearRelation.pl
dpvc: I doubt any student will say "Gee, that looks like a Perl hash" (or more likely, a Javascript function definition). It's not about programming languages, it's about what students are likely to type for the idea they are trying to get across. I agree that it is not pretty, but it is something students have typed in the past.

macros/contexts/contextLinearRelation.pl
```diff
@@ -0,0 +1,399 @@
...
+		'=<' => {
+			precedence    => .5,
+			associativity => 'left',
+			type          => 'bin',
+			string        => ' <= ',
+			TeX           => '\le ',
+			kind          => "le",
+			class         => 'LinearRelation::inequality',
+			formulaClass  => "LinearRelation"
+		},
```
dpvc, line 147: You can just use

``` perl
                '=<' => {
                         alias => '<='
		},
```

here (that wasn't available when the Inequalities content was created).

dpvc (approved):
I think you have done a nice job over all.  I just have a few minor suggestions below.

> I'm not sure about this, but I feel that this macro is a context, not a parser.

The distinction is somewhat arbitrary.  When the primary use is through the creator function (like `LinearRelation()`), then I would call it a parser, even though it has a related context.  When the main use it to define operators and functions that allow a particular syntax to be handled, then I called it a context.  But the distinction is not particularly clear-cut, and since you can use both `LinearRelation()` and `Compute()` to create these, making this a context is certainly reasonable.

Alex-Jordan:
This comes from `parserLinearinequality.pl` which I think Mike developed based off of `parserImplicitPlane.pl`. Some things you are seeing (like the overall good job, and how arguments are called up) are just direct legacy from that file. Given Davide's explanation for parser vs context, maybe this should just be edits to that file instead of a new file. I would need to change the default for the `standardForm` flag in order to not muck up problems that already use `parserLinearinequality.pl`, but that's no problem. I could also update things like how arguments are called up.

So unless someone chimes in that this is better as a new file, I will change it to be edits to `parserLinearinequality.pl`.

@pstaabp Do you have `$pg{specialPGEnvironmentVars}{parseAlternatives}` set to true in the config chain? That would explain `≤` being caught as unexpected.

Sometimes students actually type `≤`, `≥`, and `≠`. It's more common with a Mac keyboard where the option key lets you alter `<`, `>`, and `=`. But the other thing students do is copy-paste. In my experience, that's more common. And of course this is the case even with clear written instructions to type `>=`, with a certain subset of the population.

macros/contexts/contextLinearRelation.pl
Alex-Jordan: Juts to make note here, this is so a student can enter an answer like `x + y ≥ 4`, either by pasting the `≥` character, by using a Mac keyboard `⌥ >`, or some other way of typing `≥` directly. And for that to be allowed, you would have to enable the "parseAlternatives" in localOverrides.conf.

macros/contexts/contextLinearRelation.pl
Alex-Jordan: This is tricky because if the 2nd argument is a `Vector`, the arguments are interpreted one way. Otherwise, a different way. So what is `$N` in one conditional block is `$vars` in the other. It's like `my ($p, $N, $bop, $vars) = @_;` in one case and `my ($plane, $vars) = @_;` in the other. Since `$vars` makes sense as a name for the 4th argument sometimes but the 2nd argument other times, that throws a wrench in naming these all at the head of the subroutine.

macros/contexts/contextLinearRelation.pl
Alex-Jordan: Does it make sense to handle `≤` etc this way too? With `alias` instead of with `alternatives`?

Should both `=<` and `≤` be `alternatives`? Then they only work if `parseAlternatives` is set.

macros/contexts/contextLinearRelation.pl
```diff
@@ -0,0 +1,399 @@
...
+	my $self = shift;
+	my ($l, $r) = @_;
+	{
+		eq => ($l == $r),
+		lt => ($l < $r),
+		le => ($l <= $r),
+		gt => ($l > $r),
+		ge => ($l >= $r),
+		ne => ($l != $r)
+	}->{ $self->{def}{kind} };
```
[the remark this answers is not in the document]
Alex-Jordan: Thanks, I made this change.

macros/contexts/contextLinearRelation.pl
drgrice1: That is actually a problem with much of how PG is designed.  Its design has not followed good programming practices, and that has made things very confusing and unclear for future development.  This is a pain for those that are familiar with the PG design, and is extremely bad for those not already familiar with the design and trying to get started with it.

Optional arguments should always be at the end.  This is true in any language. It is unfortunate that MathObjects in general have the optional context argument as the first argument.

macros/contexts/contextLinearRelation.pl
```diff
@@ -0,0 +1,399 @@
...
+	return $self->eval(map { $variables[$_] => ($point->value)[$_] } (0 .. $#variables));
+}
+
+#
+#  We subclass BOP::equality so that we can give a warning about
+#  things like 1 = 3, and compute the values of inequalities.
+#
+
+package LinearRelation::inequality;
+our @ISA = qw(Parser::BOP::equality);
```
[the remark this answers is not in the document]
Alex-Jordan: Could I just slip `$self->{type} = Value::Type('Relation',1);` into `_check` near the end of `_check`, before the error?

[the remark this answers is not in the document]
dpvc: Yes, you could do that.  There are some MathObjects that don't determine their types until the `_check()` method.

macros/contexts/contextLinearRelation.pl
dpvc: > Does it make sense to handle `≤` etc this way too?

That would work, but would mean it was always defined, not just when `parseAlternatives` is set.  That would make this inconsistent with other alternative notations.

> Should both =< and ≤ be alternatives? 

That would also be possible.  It depends on whether you want that to be based on `parseAlternatives` or not.  I don't feel strongly either way, with perhaps a slight lean toward both as alternatives.

macros/contexts/contextLinearRelation.pl
dpvc: >  Its design has not followed good programming practices

Of course what constitutes "good programming practices" changes as the language and the paradigms that it supports evolve.

> Optional arguments should always be at the end. This is true in any language.

A rather bold statement.  A number of languages allow alternative call signatures for a given function, particularly for an untyped language like Perl.  For example, in this case we have `LinearRelation(point, normal)` and `LinearRelation(formula)`, and the other MathObjects all work similarly (e.g., `Complex($a, $b)`, `Complex([$a, $b])`, `Complex("$a + $bi")`, and so on). So one could think of this as having two additional signatures, `LinearRelation(context, point, normal)` and `LinearRelation(context, formula)`.

This gets used to make MathObjects in a particular context that may not be the current context, and since it is used on generic objects, in order for the context to be at the end of the constructor argument list, MathObjects would need to know the argument lists of all possible MathObjects (an open-ended list), in order to be able to put it in the right position for each kind of MathObject that may need to be created.  Moreover, since some MathObjects have arbitrary length argument lists, like `Point($a, $b, $c, $d)`, there is no optional position at the end of the list, as the length of the list is not specified.  

You might suggest that one looks to see if the last argument is a context and remove that and use it if so.  But, some objects take key-value pairs as the final arguments (as Alex does here with the remaining arguments after `$p`, `$N`, `$bop`, and `$vars` being used as arguments to `reduce()`).  So in that case, the last "argument" could be a pair like `context => $context`, which would make the last two arguments in `@_` be `context` and $context.  One could not easily distinguish this from the last argument being an optional context argument.

While it would be possible to use

``` perl
my ($p, $N, $bop, $vars, %reduce) = @_;
```

to process the arguments (once the context has been shifted off), it would *not* work to do 

``` perl
my ($p, $N, $bop, $vars, %reduce, $context) = @_;
```

to capture the context.  You would have to do processing to remove the context from the argument list whether it is at the end or at the beginning, so I don't think there is anything to be gained by moving it to the end, and in fact some potential complications when doing so.

Perhaps thinking about this as the constructor having additional call signatures will help it make more sense to you.

macros/contexts/contextLinearRelation.pl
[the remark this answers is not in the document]
Alex-Jordan: OK, I did this to my branch, to be pushed soon.

Alex-Jordan:
Question for the group. Right now things like `LinearRelation("1<1")` and `LinearRelation("2<=3")` throw errors. What matters is that you are not allowed to let both sides be constant.

These sorts of thing are either outright false (like the first one) and abstractly represent the empty set. Or the are outright true and abstractly represent the full space (R^3 with default variables x, y, z).

Consider `LinearRelation([0,0,0], [0,0,0], "!=")` and `LinearRelation([0,0,0], [0,0,0], "=")`. These also represent the empty set and the whole space. Note that in all cases, the relation is stored internally as a (point, vector, symbol) like this. Even when a formula is passed.

Would you prefer that I convert things like `LinearRelation("1<1")` and `LinearRelation("2<=3")` to `LinearRelation([0,0,0], [0,0,0], "!=")` and `LinearRelation([0,0,0], [0,0,0], "=")`?

Or leave it as it is and throw an error when both sides are constant?

Alex-Jordan:
There are similar issues with things like `LinearRelation("x + 1 = x + 2")`. It currently throws an error but I could convert that to `LinearRelation([0,0,0], [0,0,0], "!=")`.

macros/contexts/contextLinearRelation.pl
drgrice1: "Perl Best Practices" (implemented by perlcritic at severity level 4) recommends unpacking @_ on the first line of the function to achieve this clarity of what arguments are expected.

The point is to make the arguments for the method clear so that developers know by looking at line 1 of the method (or in the sub definition itself if signatures were enabled) what arguments are expected.  Different call signatures (in C or C++ for instance) does this.  You can look at the function headers and you know what is expected (for C and C++ even types are precise).  Perl doesn't have the concept of alternate call signatures in such a clear and concise way.

I am not saying that the MathObject code should now be rewritten to change the way the optional context argument is written at this point, and in some cases achieving what would be done in other languages with call signatures is not entirely possible in a way that will satisfy perlcritic.  This is where documentation comes in.  Comments or POD before the function would help to clarify these things.

Alex-Jordan:
I pushed another commit.

* `parserLinearRelation.pl`, not `contextLinearRelation.pl`
* I decided I want to change enough things about `parserLinearInequality.pl` that this will be a new file, not edits to `parserLinearInequality.pl`. It will behave differently in a few ways.
* I dropped support for making up new variables at the moment where you create the object. It's just going to use the context variables. I see that feature is in `parserImplicitPlane.pl`...is there a case to be made to keep this feature?
* To me it makes more sense to define the plane's normal vector first like `LinearRelation(vector, point, symbol)` instead of `LinearRelation(point, vector, symbol)`, and since that also helped me in another way, I made it work that way.

I nearly have this how I would like it, where you are allowed to make things like `LinearRelation([0,0,0], point, "<")` which would essentially be "false" and represent the empty set. 

It's working except that a LinearRelation defined with a string formula, like `LinearRelation("0 < 0")`, `LinearRelation("0 <= 1")`, etc. is causing problems. If `$plane` is the string `0 < 0`, then the line:

```
$plane = $formula->new($context, $plane) unless Value::isValue($plane);
```

is reducing to a simple boolean value instead of staying an unreduced formula. This presents a problem for me because once it is reduced, there is no tree and there is no lop, bop, rop for general things that happen downstream. Is there a way to prevent things like `0 < 0` from reducing?

Alex-Jordan:
I almost have this, but I need to fix how comparing one object to another works when the normal vector is 0. I think I know what to do, just waiting til tomorrow. I would not want to take anyone's time looking at it until I push again.

Alex-Jordan:
OK, that took less than I thought it would. It's ready for a test, starting with the attached file. Remember to set `parseAlternatives` to use the sloppy or unicode alternatives for the operators.

[LinearRelation.txt](https://github.com/openwebwork/pg/files/11268634/LinearRelation.txt)

macros/parsers/parserLinearRelation.pl
```diff
@@ -0,0 +1,379 @@
...
+		my @terms;
+		my $i = 0;
+		for my $x ($context->variables->names) { push @terms, $N->{data}[ $i++ ]->string . $x }
+		$bop      = '=' unless defined($bop);
+		$leftside = $formula->new(join(' + ', @terms))->reduce(@_);
+		$plane    = $formula->new($leftside . $bop . $d->string);
+	} else {
+		# Determine the normal vector and d value from the equation
+		$plane = $N;
+		$context->flags->set(reduceConstants => 0);
```
dpvc, line 221: I don't think this is the best way to accomplish your goal of keeping things like `0 < 1` as a formula, as this setting is permanent on the context, and so will prevent any constant reduction in any computation within that context.

Instead, you could make the inequality operators always mark themselves as non-constant, so they are never reduced by adding

``` perl
        $self->{isConstant} = 0;
```

at the end of the `_check` function of `LinearRelation::inequality`.  This will make `LinearRelation("0 < 1")` remain as a formula "0 < 1".  However, `Compute("0 < 1")` will still return `1`, since it uses `$formula->isConstant` to determine whether to reduce the final result, and that function tests whether there are any variables in the formula to tell if it is constant rather than the `isConstant` property of the top node in the tree. 

To handle that case as well, you can add

``` perl
sub isConstant {
  my $self = shift;
  return 0 if $self->type eq "Relation";
  return self->SUPER::isConstant;
}
```

to the end of the `LinearRelation` class.  Then `Compute("0 < 1")` will return a Formula rather than reducing it a constant.

Using this approach will allow `reduceConstants` to act normally on everything other than relations, which are never reduced.

macros/parsers/parserLinearRelation.pl
```diff
@@ -0,0 +1,379 @@
...
+
+sub _check {
+	my $self = shift;
+	$self->SUPER::_check;
+	$self->{type} = Value::Type('Relation', 1);
+}
+
+sub _eval {
+	my $self = shift;
+	return &{ $self->{def}{eval} }(@_);
```
dpvc, line 361: After writing this, I remembered that `_eval` is supposed to return a MathObject, though it probably doesn't matter in this case.  If you want to, you could use

``` perl
	my $context = $self->context;
	return $context->Package("Real")->new($context, &{ $self->{def}{eval} }(@_) ? 1 : 0);
```

or something like that.  But again, it probably doesn't really matter here, so might not be worth it.

dpvc (approved):
Looks good.  Thanks for putting the work into this.

macros/parsers/parserLinearRelation.pl
Alex-Jordan: Thanks for all your feedback. I have applied these changes. I was under the false impression that `$context` was a copy of the current context instead of the actual current context.

pstaabp:
@Alex-Jordan I didn't have the `$pg{specialPGEnvironmentVars}{parseAlternatives}` set.  Perhaps clarify in the POD where to set this.  (localOverrides.conf or PGcourse.pl)?

macros/parsers/parserLinearRelation.pl
```diff
@@ -0,0 +1,386 @@
...
+}
+
+#
+#  If the vectors are zero, check if true or false
+#  If the vectors are non-zero, check if the equations are multiples of each other.
+#
+sub compare {
+	my ($self, $l, $r) = Value::checkOpOrder(@_);
+	$l = new LinearRelation($l) unless ref($l) eq ref($self);
+	$r = new LinearRelation($r) unless ref($r) eq ref($self);
```
drgrice1, line 252: This should be changed to
```suggestion
	$l = LinearRelation->new($l) unless ref($l) eq ref($self);
	$r = LinearRelation->new($r) unless ref($r) eq ref($self);
```
See https://metacpan.org/pod/Perl::Critic::Policy::Objects::ProhibitIndirectSyntax

macros/parsers/parserLinearRelation.pl
```diff
@@ -0,0 +1,386 @@
...
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+# Artistic License for more details.
+################################################################################
+
```
drgrice1, line 15: You need to add `=encoding utf8` here to be able to use `≤`, `≥`, and `≠` as you do on line 52 in the POD.

macros/parsers/parserLinearRelation.pl
```diff
@@ -0,0 +1,386 @@
...
+#    $R = Compute("x + y + 2z <= 5");
+#    $R = LinearRelation([1,2,1], [1,1,2], "<=");
+#    $R = LinearRelation([1,2,1], 5, "<=");
+
+sub new {
+	my ($self, $N, $p, $bop) = @_;
+	my $class   = ref($self) || $self;
+	my $context = (Value::isContext($_[0]) ? shift : $self->context);
+	my $formula = "Value::Formula";
+	return shift if scalar(@_) == 1 && ref($_[0]) eq 'LinearRelation';
```
drgrice1, line 193: Something looks wrong here.  I think you should put this back to the way that you had it before @pstaabp and my comments on unpacking arguments on the first line.  We probably should have kept our mouths shut on this for now.  With the current setup of MathObjects that just doesn't work, and @dpvc should be heeded.  Lines 191 and 193 will not work right with this setup because `$self` has not been shifted out of the input argument array `@_`.

drgrice1 (commented):
I think this will be a nice addition, but there are a couple of minor issues, and one bigger problem.  See the comments in the code.

Alex-Jordan:
OK, I applied those edits. My test file is working the same either way, so I may not be testing what I should be.

drgrice1 (approved):
This looks good.  Thanks for making the changes.

pstaabp (approved):

Alex-Jordan:
There is a problem with this. If I make something like `LinearRelation("x + -2y = z")` then it prints as `x+(-2)y=z` instead of `x-2y=z`. Using `->reduce` has no effect.

@dpvc, do you see how to let this kind of reduction happen, without always rearranging the equation into standard form?

Alex-Jordan:
I added a line after line 220:
```
$plane = $plane->reduce;
```

This calls the formula reduce method and leads to `LinearRelation("x + -2y = z")` displaying `x-2y=z`, so that much is nice. 

However, I can't stop this from also turning `LinearRelation("x+-2/3y=z")` into `x-0.666667y=z`. I've set the flag `reduceConstants=>0` in various places but is has no effect.

dpvc:
> I can't stop this from also turning `LinearRelation("x+-2/3y=z")` into `x-0.666667y=z`.

So here's the deal.  The `reduceConstants` flag only applies to the parsing of the original formula (e.g., `Compute("(1+2)x")` to get `3x`), not to explicit `=>reduce()` calls, where constants are always reduced.

The possible solutions, as I see it, are:

* Add `&& $self->context->flag('reduceConstants')` to
    https://github.com/openwebwork/pg/blob/a29b85f259fac1924688a2a6f8b80b9715c4bc55/lib/Parser/BOP.pm#L71-L72
    so that `reduceConstants` will affect `->reduce` calls as well.  You would probably want to do the same for
    https://github.com/openwebwork/pg/blob/a29b85f259fac1924688a2a6f8b80b9715c4bc55/lib/Parser/Function.pm#L64
    for `reduceConstantFunctions`.  This is a potential breaking change, as it is a significant change to the scope of `reduceConstants`.

* Replace the class for the `/` operator(s) in the `LinearRelation` context to a subclass of `Parser::BOP::divide` that overrides the `reduce` method (not the `_reduce` method) so that it has the extra `reduceConstants` check indicated above (or with a new divide-specific flag if you prefer), so that the change only affects the LinearRelation context.  This is a hack, but the safest and easiest way to handle it.

* Add new reduction rules for the various operations that targets constants, like `a / b`, 'a * b`, `n!`, etc. that would control the reductions of constant operations so that these could be individually controlled.  There could also be a generic `a bop b` rule that could be in place of the `reduceConstants` check in the first bullet point above, so that you could control all the operators at once (rather than having to set *all* of them off individually).  It would be possible to add ones for functions, as well, but there are a *lot* more of those than operators, so probably isn't worth it.

The last option is probably the most flexible one, but would take more work.

PR #816, opened 2023-04-21
Update the CDN version of JSX graph and add the missing CDN version of plotly.

drgrice1:
The version update of JSX graph was forgotten when the npm version was updated, and the plotly version should have been added before.


pstaabp (approved):

drgrice1:
This pull request is preempted by #817.

PR #817, opened 2023-04-21
Rewrite third party asset handling.

drgrice1:
This removes the third-party-assets.json file, and the contents of that file are no longer added to the static-assets.json file.  Instead the frontend (webwork2 or otherwise) should load the package.json file and use that to determine how to serve any file located in node_modules.

Note that a version tag is added to mathquill in the package.json file to lock the specific commit of mathquill to use.  The tag for PG-2.18 is WeBWorK-2.18.  Previously a tag for 2.17 was added.  If a hotfix is needed for mathquill, then we will append a subversion to ensure that browser caching is avoided.

This is paired with https://github.com/openwebwork/webwork2/pull/1950.  That pull request is needed to test this side of things.


taniwallach (commented):
In my testing with morbo using both Chrome and Firefox, in the `GET` request for MathQuill I see
```
GET http://localhost:3000/pg_files/node_modules/mathquill/dist/mathquill.css?version=github:openwebwork/mathquill
```
which does not include the `#WeBWorK-2.18`.
I suspect that something is failing, but maybe I am not looking in the right place.
I do see version numbers in the URLs and `GET` calls for `jsxgraph` when I loaded a problem which uses it.

drgrice1:
If you look at the source in the browser you will see that th `#WeBWorK-2.18` is there.  However, I think that the browser is interpreting the `#` as a hash property instead of as part of the url parameter.  So now I changed it so that the `#` is replaced with an `@` symbol.  That works for me.

drgrice1:
By the way, that is done in webwork2/lib/WeBWorK/Utils.pm, and so not seen in this pull request on the pg side.

taniwallach:
> If you look at the source in the browser you will see that the `#WeBWorK-2.18` is there. However, I think that the browser is interpreting the `#` as a hash property instead of as part of the url parameter. So now I changed it so that the `#` is replaced with an `@` symbol. That works for me.

Thanks. The change on the `webwork2` side gets the `@WeBWorK-2.18` into the URL.

taniwallach (approved):

PR #818, opened 2023-04-21
Run workflows on all pull requests.  Not just those to main or develop.

drgrice1:
Pull requests to the release candidate should run at least.


taniwallach (approved):
This makes sense to me, as we want any code being pulled in via a pull request, even via a different branch to go via these workflows.

pstaabp (approved):
I'll merge in so we can get tests running again.

PR #822, opened 2023-04-27
Add an essay field preview button and restructure the equation editors a bit.

drgrice1:
The essay preview button is added via the htdocs/js/apps/Essay/essay.js javascript.  This javascript is now always added to the problem page (via load_js in PG.pl).  This preview button is added below essay answers regardless of the entry assist method.

The restructuring is to accomodate the usage of these editors for the webwork2 manual problem grader comment boxes (see https://github.com/openwebwork/webwork2/pull/1959), but I think is better in general.  The button attached to the top right corner of the textarea doesn't really look that great.

For the MathQuill equation editor, the button that activates the equation editor is now placed below the essay textarea and to the left of the new preview button.  In addition, the "Insert" and "Clear" buttons of the equation editor are placed in the card footer below the MathQuill input instead of to the right of it.

For the MathView equation editor, the button that activates the equation editor is also placed below the essay textarea, but to the right of the preview button.  Furthermore, those buttons are wrapped in a div that aligns its content to the right.  This makes it so that when the equaiton editor popover opens it does not obscure the textarea.  The fallback placements have been modified so that the popover prefers to go to the right if there is room, but if not it tries to first go down, and then up.  It is not allowed to go to the left.  In some extreme situations (short windows with the button already on the far right) this will cause the page to be extended to the right to make room for the popover.

Also for the MathView equation editor, a "Clear" button is added and the "Insert" and "Clear" buttons are placed right justified below the equation editor input instead of to the right of it.

Note that this is built on top of #807 and is to address requests made by @Alex-Jordan there.

### DIFF
--- a/htdocs/js/apps/Essay/essay.js
+++ b/htdocs/js/apps/Essay/essay.js
@@ -1,0 +1,72 @@
+'use strict';
+
+(() => {
+	const addPreviewButton = (latexEntry) => {
+		if (latexEntry.dataset.previewBtnAdded) return;
+		latexEntry.dataset.previewBtnAdded = 'true';
+
+		const buttonContainer = document.createElement('div');
+		buttonContainer.classList.add('latexentry-button-container', 'mt-1');
+
+		const button = document.createElement('button');
+		button.type = 'button';
+		button.classList.add('latexentry-preview', 'btn', 'btn-secondary', 'btn-sm');
+		button.textContent = 'Preview';
+
+		button.addEventListener('click', () => {
+			button.dataset.bsContent = latexEntry.value
+				.replace(/</g, '< ')
+				.replace(/>/g, ' >')
+				.replace(/&/g, '&amp;')
+				.replace(/\n/g, '<br>');
+			if (button.dataset.bsContent) {
+				if (button.dataset.popoverShown) button.dispatchEvent(new Event('hidden.bs.popover'));
+				button.dataset.popoverShown = 'true';
+				const popover = new bootstrap.Popover(button, {
+					html: true,
+					trigger: 'focus',
+					placement: 'bottom',
+					delay: { show: 0, hide: 200 }
+				});
+				button.addEventListener(
+					'hidden.bs.popover',
+					() => {
+						delete button.dataset.popoverShown;
+						popover.dispose();
+					},
+					{ once: true }
+				);
+				if (window.MathJax) {
+					button.addEventListener(
+						'show.bs.popover',
+						() => {
+							MathJax.startup.promise =
+								MathJax.startup.promise.then(() => MathJax.typesetPromise(['.popover-body']));
+						},
+						{ once: true }
+					);
+				}
+				popover.show();
+			}
+		});
+
+		buttonContainer.append(button);
+		latexEntry.after(buttonContainer);
+	};
+
+	document.querySelectorAll('.latexentryfield').forEach(addPreviewButton);
+
+	const observer = new MutationObserver((mutationsList) => {
+		for (const mutation of mutationsList) {
+			for (const node of mutation.addedNodes) {
+				if (node instanceof Element) {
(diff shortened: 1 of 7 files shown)


htdocs/js/apps/Essay/essay.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+(() => {
+	const addPreviewButton = (latexEntry) => {
+		if (latexEntry.dataset.previewBtnAdded) return;
+		latexEntry.dataset.previewBtnAdded = 'true';
```
pstaabp, line 6: Do you want this as the string 'true' or the boolean `true`?  I think it works either way.

drgrice1: HTML data attributes can not take boolean values.  It must be a string.

pstaabp (approved):

htdocs/js/apps/Essay/essay.js
drgrice1: Just for clarification, it does work to use a boolean `true` here.  However, the actual type of of the data attribute will be `string`.  All that is really needed in this case is that the data attribute is defined and is "truthy".  You could actually set this to the boolean value of `false`, and the type of the data attribute will still be `string`, and its string value will be 'false'.  Note that is "truthy".  So I deem it better to not use a boolean value as it can be confusing.

Alex-Jordan:
I tried testing this this morning, but now something bigger is not working with my dev server. The OS is Oracle 8. I'm having issues getting the webwork2 service to work, as well as something called dprobed.service that I think is some Oracle thing.

Alex-Jordan:
I think it might look better if "Equation Editor" is not boldface. See what you think though.

drgrice1:
I thought it looked better with "Equation Editor" in bold.  That is why I added the class to make it bold.  If you think it looks better not bold, I guess I am fine with that.

Alex-Jordan:
It probably does look better in bold for sure if you focus on the equation editor card only. To me, when the problem itself has no headings or boldface, then the bold "Equation Editor" becomes more more important/prevalent than anything else in the problem, which feels off. I still want my eye's/mind's focus to be on the problem, and the editor should be a secondary thing supporting how I interact with the problem. But if you don't see it that way, it's all good and I'd defer to you.

drgrice1:
I am fine with removing the bold font.  However, it occurs to me that the idea of making the card header the primary color has the same effect.  In addition, after messing with it I really don't like it.  How about using the secondary color instead?  That is the same color as the equation editor (square root) and preview buttons.  To me that is not as intrusive.   One other reason that the primary color is related to the close button color.  To use a darker color (including both the primary and secondary colors) you need to add the 'btn-close-white' class to the close button to make it white.  The problem is that the primary color is not always a dark color, namely if the math4-yellow theme is used.  The secondary color is the same though, so the white button is fine for all themes with that.

drgrice1:
Although, I actually think the default card title background color is the best.

Alex-Jordan:
That sounds good. In the meeting after I stopped screen sharing, I mocked up that color on the header and I thought it was better too.

drgrice1:
I adjusted the appearance as discussed, and fixed the clear button flickering of the toolbar.  I am not entirely sold on the buttons being above the equation editor.  I don't like the separation of the editor and the textarea.  I understand not wanting the buttons to move, but the separation doesn't look right to me.

Alex-Jordan:
OK, thanks for all the work. I think it's good as it is.

Alex-Jordan (approved):

520qzx520:
How can I make mathquill support equations? That is, a system of equations with {

drgrice1:
@520qzx520:  Unfortunately you can't.  MathQuill does not support all of LaTeX.  Matrices are another thing that are not supported.

drgrice1:
For an essay field you can still type the LaTeX for that directly into the essay textarea.  That will work assuming MathJax supports it.

PR #826, opened 2023-05-08
Nice tables fix

Alex-Jordan:
This fixes an issue with vertical rules in hardcopy. When there are vertical rules set in the `texalignment` string argument, and in some cell something was invoking `\multicolumn{}{}{}`, it was failing to use the vertical rules from the `texalignment` in situations when it should.

Also it was needlessly using `\multicolumn{}{}{}` on every cell in a row with a header, when that is only needed on the first cell (and only when `headerrules` is also used).

Test with the table I proposed using here:

https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8266&parent=20277

And I added a warning message if `texalignment` has something unrecognized. For example someone might do `texalignment => 'c!{2pt}c'` when they meant `texalignment => 'c!{\vrule height 2pt}c'` and this alerts them that `!{2pt}` is unrecognized.

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -891,8 +891,14 @@ sub Row {
 				$columntype =~ s/^p/b/ if ($valign eq 'bottom');
 				$columntype =~ s/^p/m/ if ($tableOpts->{valign} eq 'middle');
 				$columntype =~ s/^p/b/ if ($tableOpts->{valign} eq 'bottom');
-				$columntype .= '|'
-					if ($i == 0 && $cellOpts->{colspan} == 1 && $tableOpts->{rowheaders} && $tableOpts->{headerrules});
+				$columntype = getLaTeXcolumnWidth($alignment->[0]{left}) . $columntype
+					if ($i == 0 && $alignment->[0]{left} && !$cellOpts->{halign});
+
+				if ($i == 0 && $cellOpts->{colspan} == 1 && $tableOpts->{rowheaders} && $tableOpts->{headerrules}) {
+					$columntype .= '|';
+				} elsif (!$cellOpts->{halign}) {
+					$columntype .= getLaTeXcolumnWidth($cellAlign->{right});
+				}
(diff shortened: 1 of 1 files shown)


macros/ui/niceTables.pl
```diff
@@ -1432,6 +1440,16 @@ sub getLaTeXthickness {
 	return $output;
 }
 
+sub getLaTeXcolumnWidth {
+	my $input = shift;
+	return '' unless $input;
+	if ($input =~ /^\d+$/) {
```
pstaabp, line 1446: The rest of this sub can be written a little more compactly with:
`return ($input =~ /^\d+$/) ? '|' x $input : "!{\\vrule width $input}";`

pstaabp:
Which table were you specifying in your example above.  I noticed the table:
```
 DataTable(
    [
      ['',['Business', headerrow => 1],'Science','Liberal Arts','Other','Total'],
      ['Female',10,9,9,7,35],
      ['Male',12,11,10,7,40],
      ['Total',22,20,19,14,75],
    ],
    horizontalrules => 1,
    rowheaders => 1,
    booktabs => 0,
    texalignment => '|c|c|c|c|c|c|',
    caption => 'Academic Major by Gender',
    captioncss => 'text-align: center;',
  );
```
was missing the outside vertical lines in develop, but this PR fixes this.  

Also, your other example cited above is throwing errors--in the hash they need to be quoted.  `writing-mode` needs to be `'writing-mode'` and `empty-cells` needs to be `'empty-cells'`, although I don't see these in the niceTables code, so not sure what they do.

Alex-Jordan:
At the time I posted here, the only table I had posted on that forum thread is the one you just posted here.

If you read the post that follows, I was just copy-pasting something that OP posted. It has junk in it that I don't know where it came from. There are no `writing-mode` or `empty-cells` keys, and I don't know where he got that. But anyway in the context of that thread, those latter posts are to point out something else about why it will be hard to implement rowspan.

drgrice1 (approved):
This looks good.  However, I am not seeing the difference for the table in the linked forum post between this pull request and the develop branch.

PR #828, opened 2023-05-11
More macro reorganization.

drgrice1:
Move the PGlateximage.pl macro out of core (it is not a core PG macro) and into the graph directory with the PGtikz.pl macro.  Those two macros certainly should always be kept together.  That directory really should be named `graphics` not `graph` as most of the macros in that directory are for generating more general graphics, not just graphs.

Also move problemRandomize.pl, PeriodicRerandomization.pl, answerDiscussion.pl, PGtextevaluators.pl, and problemPreserveAnswers.pl into deprecated where these should be.

Also move Parser.pl into core.  That is a core macro, and should be kept with Value.pl.

More should be done with this.  The current directory organization seems to have been determined more by what the macro file name was than by what the macro actually does or is used fore.  As a result it is difficult to find the macro file you are looking for.

I think that the parsers and contexts directory should be made a single directory.  All of those are things that derive from math objects, and note that the distinction between a parser and context is not so well defined.  See @dpvc's comment in https://github.com/openwebwork/pg/pull/813#pullrequestreview-1390171518.


pstaabp (approved):

Alex-Jordan:
I wonder if parserGraphTool should go into graph (or graphics) too. A typical problem author would look there for that, and not understand what a parser is.

drgrice1:
I agree.  I think it would fit better there as well.  I moved it there.

PR #829, opened 2023-05-12
update PTX rowheaders

Alex-Jordan:
Earlier PTX did not support row header cells. I made a guess about what the markup might be one day. As of today, PTX supports row header cells, and the markup is different. This aligns the markup with how PTX will actually be going forward.


drgrice1 (approved):
Looks good.

pstaabp (approved):

PR #830, opened 2023-05-17
change how scaffold identifies which section an answer comes from

Alex-Jordan:
This follows a forum thread I started here: https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8275

I have some concerns about the change since I do not fully understand its implications. But it does fix the issue and appears to not break anything.

### DIFF
--- a/macros/core/scaffold.pl
+++ b/macros/core/scaffold.pl
@@ -697,21 +697,21 @@ sub is_open {
 }
 
 #
-#  Return a boolean array where a 1 means that answer blank has
-#  an answer evaluator assigned to it and 0 means not.
+#  Return a boolean array where a 1 means that answer blank
+#  has had its responses initialized and 0 means not.
 #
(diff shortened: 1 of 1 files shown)


somiaj:
@Alex-Jordan I have lots of problems (both scaffolded and not scaffoleded) where previous student's answers are used to access the following answers. There are some subtle things I have with this, but if you would like to send your problem my direction, I could probably get it to work without this patch.  Note I don't spread multianswers across scaffolds, but instead just save and use the student answer later on.

Alex-Jordan:
Thanks, I saw your forum post and wrote back there.

Not sure what to do about this PR. It seems useful to allow multianswers to span multiple sections. Maybe both approaches are really just workarounds though. Maybe scaffold.pl should have a built-in way for you to code a section conditionally based on whatever the user entered as answers in previous sections.

Issue #831, opened 2023-05-17
niceTables.pl has confusing warning if texalignment has more columns than used.


somiaj:
Just noticed that you get a warning from `niceTables.pl` when there are extra columns defined in `texalignment` that are not used. The table outputs just fine, so the warning should probably be fixed or a better warning to help point the problem author on how to fix the warning would help. Example problem:

```
DOCUMENT();
loadMacros(
    'PGstandard.pl',
    'PGML.pl',
    'niceTables.pl',
    'PGcourse.pl'
);

$table = DataTable(
    [
        ['A', 1, 2, 3],
        ['B', 5, 6, 7],
    ],
    align => '|c|c|c|c|c|',
    horizontalrules => 1,
);


BEGIN_PGML
        [$table]*
END_PGML
ENDDOCUMENT();
```

The warning is "Use of uninitialized value $a in pattern match (m//) at line 1406 of [PG]/macros/ui/niceTables.pl."

The warning goes away if the issue with too many columns is fixed, `align => '|c|c|c|c|'.

pstaabp:
Fixed with #832

PR #832, opened 2023-05-18
add cells when texalignment suggest more cells than data

Alex-Jordan:
This addresses #831.

First, the variable that caused the warning is initialized to an empty string if needed.

But also if a problem has more columns in the tex alignment string than actually shows up in a row, this adds empty cells to that row to fill it out. This will make the author see that they have too many things in the alignment string, if that is the case. Or you can use this intentionally if you want a lot of empty cells without actually coding them into the data array.

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -426,6 +426,46 @@ sub DataTable {
 	# $alignment is a 1D array of hash references, with options for each column
 	my $alignment = ParseAlignment($tableOpts->{texalignment});
 
+	# if the user's data implies fewer cells in any row than what is in texalignment
+	# then we add empty data cells, update $colCount, $tableOpts, and $alignment
+	my $needToUpdate;
+	for my $j (0 .. $#$tableArray) {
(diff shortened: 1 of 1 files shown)


macros/ui/niceTables.pl
```diff
@@ -1393,6 +1425,7 @@ sub suffix {
 
 sub css {
 	my ($a, $b) = @_;
+	$a = '' unless defined $a;
```
drgrice1, line 1428: I prefer the syntax `$a //= '';` for this sort of thing.  It is equivalent, and much more concise.

drgrice1 (approved):
Looks good and fixes the issue.

Alex-Jordan:
I made that suggested change.

Also there was a potential problem with this. After more cells are (potentially) added, some global variables for the table should be updated. So I added lines to update them.

If you are wondering, it's not easily possible to add the empty cells *before* those variables are calculated the first time. So it's not easy to save efficiency by only calculating them only once. But I made it so the update only happens when needed.

pstaabp (approved):
Looks good.  Seems to be

PR #834, opened 2023-05-26
In WeBWorK::PG::Environment get `$pg_root` from the PG_ROOT environment variable.

drgrice1:
This is just for correctness as the current code does work.  However, the only reason it works is because of the fact that `pg_root` is no longer set in the webwork2 CourseEnvironment.  So what happens is that the eval statement on lines 64-68 returns undefined and so the `or do` statement is executed.  So in the end the `$pg_root` is correctly set, but from the PG_ROOT environment variable set by the Mojolicious webwork2 app when it starts.  Note in this case it does not use the system environment variable even if that is set since the app overrides it with the value of pg_dir in webwork2.mojolicious.yml.


lib/WeBWorK/PG/Environment.pm
```diff
@@ -58,16 +58,14 @@ use YAML::XS qw/LoadFile/;
...
 
 	eval {
 		require WeBWorK::CourseEnvironment;
-		$ce      = WeBWorK::CourseEnvironment->new({ webwork_dir => $ENV{WEBWORK_ROOT}, courseName => $courseName });
-		$pg_root = $ce->{pg_root};
-	} or do {
-		$pg_root = $ENV{PG_ROOT};
-		die 'The pg directory must be defined in PG_ROOT.' unless -d $pg_root;
+		$ce = WeBWorK::CourseEnvironment->new({ webwork_dir => $ENV{WEBWORK_ROOT}, courseName => $courseName });
 	};
```
pstaabp, line 69: Does this block still need to be in an `eval`?  Is there anyway for this block to fail.  I thought that was generally what `eval` was for?

drgrice1: It certainly needs to be an `eval` still.  The standalone renderer will not have the `WeBWorK::CourseEnvironment` module, and so the code in the eval block will fail for the standalone renderer.  That failure will quietly be ignored, because that is what is supposed to happen.  Without the `eval` block there would be an error that would make the code cease to function.

pstaabp (approved):

PR #835, opened 2023-05-29
Fix moving points that have been graphed.

drgrice1:
Currently if you graph a point, and then move it, and then click one of the problem submit buttons, the point is moved back to its initial position.  This was caused by an obvious mistake.  The `updateText` method needs to be called, not just stated.


pstaabp (approved):
Fixes the issue and minor, so will merge.

PR #836, opened 2023-05-29
Update macro pod

pstaabp:
Working on #811 noticed that some of the macros didn't have POD, others were in bad shape. 

This is intended to be some cleanup to the POD in the macros with no changes to code.  This was just an initial pass through on some of the missing POD and some other cases.  I did fair bit of exchanging leading tabs in POD to four spaces, but not throughout the macros.

### DIFF
--- a/macros/contexts/contextInequalitiesAllowStrings.pl
+++ b/macros/contexts/contextInequalitiesAllowStrings.pl
@@ -1,18 +1,27 @@
-###
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2022 The WeBWorK Project, https://github.com/openwebwork
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+# Artistic License for more details.
+################################################################################
 
 =head1 NAME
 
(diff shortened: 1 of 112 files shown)


drgrice1:
If you want to go further with this, you could run `podchecker` on the macros.  Doing so reveals more issues with the POD.  The most common is a warning about `=head?` without preceding higher level.

macros/contexts/contextInequalitiesAllowStrings.pl
```diff
@@ -14,8 +14,8 @@ =head3 Allows string answers for the set of real numbers and the empty set
 
 =pod
```
drgrice1, line 15: You should also deleted the `=pod` here.  That doesn't do anything after `=head2`.

drgrice1: I also see an `=cut` statement immediately before this on line 11.  That is also unneeded.

drgrice1: I ran `podchecker` on this and it says the `=head3` should be a `=head2` (I am paraphrasing).  Also the parts in `=head1 NAME` and `=head1 Synopsis` should have no indentation.  `podchecker` says "*** WARNING: Verbatim paragraph in NAME section" on this.

drgrice1 (approved):
This looks good.  Clearly there is more room for improvement on this, but this is a start.

pstaabp:
I'll take a look at `podchecker`.  It would be nice to do some more cleanup of some of the macro files--I'm also noticing that a lot should be deprecated.

pstaabp:
More cleanup after running `podchecker`.  This is a good first round of edits.

drgrice1:
I have carefully checked that the only thing that is changed in this pull request is POD, and that no code changes were done.  The POD for all macros (except PG.pl and those in deprecated) passes all `podchecker` tests.  Furthermore, the generated html POD using the webwork2 `bin/dev_scripts/generate-ww-pg-pod.pl` script looks considerably better for many of the macros, and there is now POD for some macros that didn't have it before.

So I will merge this now.

PR #838, opened 2023-05-31
parse out backticks in PGML

Alex-Jordan:
This PR parses each backtick in a PGML block and does things to it before shipping to output.

* HTML: ``<span>`</span>``
* LaTeX: ``\\`{}``
* PTX: `` ` ``

Except it won't do anything to a backtick inside math, code, or verbatim. Those stay as bare `` ` `` in the output. Verbatim is also now wrapped in a `span.tex2jax_ignore`. No math should get processed in a verbatim bit.

Unless I missed something, no pair of backticks you type inside a PGML block will now trigger MathJax AsciiMath. For example, all of these render as math before this commit. But with this commit, they do not:

```
BEGIN_PGML

# `Heading`

[|`verbatim`|]

`plain text wrapped in backticks`

END_PGML
```

If anyone is determined, they can still get AsciiMath processed with backticks like this:

```
$string = "`backticks saved`";

BEGIN_PGML

[$string]

[@"`OK, technically these are typed within the PGML block`"@]

END_PGML
```

But my purpose here is to quash the notion among innocent beginning problem authors that you can legitimately use backticks direclty in PGML to get math. Anyone using these last methods is trying to hack around something and will get what they deserve.

### DIFF
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -45,7 +45,7 @@ my $code      = '```';
 my $pre       = ':   ';
 my $quoted    = '[$@%]q[qr]?|\bq[qr]?\s+(?=.)|\bq[qr]?(?=\W)';
 my $emphasis  = '\*+|_+';
-my $chars     = '\\\\.|[{}[\]()\'"]';
+my $chars     = '\\\\.|[{}[\]()\'"`]';
 my $ansrule   = '\[(?:_+|[ox^])\]\*?';
 my $open      = '\[(?:[!<%@$]|::?:?|``?`?|\|+ ?)';
 my $close     = '(?:[!>%@$]|::?:?|``?`?| ?\|+)\]';
@@ -195,6 +195,7 @@ sub All {
 		/<</           && do { return $self->Center($token) };
 		/>>/           && do { return $self->Align($token) };
 		/```/          && do { return $self->Code($token) };
+		/`/            && do { return $self->Backtick($token) };
 		/:   /         && do { return $self->Preformatted($token) };
 		$self->Text($token);
 	}
@@ -502,6 +503,12 @@ sub Code {
 	$self->Begin($token, "```");
 }
 
+sub Backtick {
+	my $self  = shift;
+	my $token = shift;
+	$self->Item("backtick", $token);
+}
+
 sub Preformatted {
 	my $self  = shift;
 	my $token = shift;
@@ -711,11 +718,12 @@ my $balanceAll = qr/[\{\[\'\"]/;
 		combine     => { align => "align", par => 1 },
 		noIndent    => -1
 	},
-	"#"      => { type => 'heading', parseAll => 1, breakInside => 1, combine => { heading => "n" } },
-	"*"      => { type => 'bold',    parseAll => 1, cancelPar   => 1 },
-	"_"      => { type => 'italic',  parseAll => 1, cancelPar   => 1 },
-	"bullet" => { type => 'bullet',  parseAll => 1 },
-	"list"   => { type => 'list',    parseAll => 1, combine => { list => "bullet", par => 1 }, noIndent => -1 },
+	"#"      => { type => 'heading',  parseAll => 1, breakInside => 1, combine => { heading => "n" } },
+	"`"      => { type => 'backtick', parseAll => 1 },
+	"*"      => { type => 'bold',     parseAll => 1, cancelPar => 1 },
+	"_"      => { type => 'italic',   parseAll => 1, cancelPar => 1 },
+	"bullet" => { type => 'bullet',   parseAll => 1 },
+	"list"   => { type => 'list',     parseAll => 1, combine => { list => "bullet", par => 1 }, noIndent => -1 },
 );
 
 ######################################################################
@@ -1198,6 +1206,7 @@ sub string {
 			/quote/    && do { $string = $self->Quote($item, $strings[-1] || ''); last };
 			/rule/     && do { $string = $self->Rule($item);                      last };
 			/code/     && do { $string = $self->Code($item);                      last };
+			/backtick/ && do { $string = $self->Backtick($item);                  last };
 			/pre/      && do { $string = $self->Pre($item);                       last };
 			/verbatim/ && do { $string = $self->Verbatim($item);                  last };
 			/break/    && do { $string = $self->Break($item);                     last };
@@ -1231,6 +1240,7 @@ sub Heading  { return "" }
 sub Quote    { return "" }
 sub Rule     { return "" }
 sub Code     { return "" }
+sub Backtick { return "" }
 sub Pre      { return "" }
 sub Verbatim { return "" }
 sub Break    { return "" }
@@ -1434,6 +1444,13 @@ sub Code {
 	return $self->nl . '<pre style="margin:0"><code' . $class . '>' . $self->string($item) . "</code></pre>\n";
 }
 
+sub Backtick {
+	my $self = shift;
+	my $item = shift;
+	# prevent MathJax AsciiMath processing
+	return '<span>`</span>';
+}
+
 sub Pre {
 	my $self = shift;
 	my $item = shift;
@@ -1504,7 +1521,7 @@ sub Verbatim {
 	my $item = shift;
 	my $text = $self->Escape($item->{text});
 	$text = "<code>$text</code>" if $item->{hasStar};
-	return $text;
+	return "<span class='tex2jax_ignore'>$text</span>";
 }
 
 sub Math {
@@ -1579,6 +1596,12 @@ sub Code {
 	my $self = shift;
 	my $item = shift;
 	return $self->nl . "{\\pgmlPreformatted\\ttfamily%\n" . $self->string($item) . "\\par}%\n";
+}
+
+sub Backtick {
+	my $self = shift;
+	my $item = shift;
+	return '\\`{}';
 }
 
 sub Pre {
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Alternatively, if we implement the MathJax config approach that Davide posted about at https://groups.google.com/g/mathjax-users/c/zmmEun8rG4I/m/CDpA6SvMDAAJ, that would simplifies things a lot and even help when PGML is not used. I favor that approach, until such time as someone points out a problem with it :)

pstaabp:
Just for documentation, see the conversation in openwebwork/webwork2#1980

dpvc:
@Alex-Jordan, I think this is the wrong approach, and that @somiaj's idea of using the `Escape()` method is the right one.  The back-tic is not a command in PGML, and is just another character that should be treated as text.  The `Escape()` functions are meant to make whatever modifications are needed for that text to be displayed properly in the format being used.  So that is the right place to make the adjustment for back-tics and TeX delimiters that shouldn't be treated as such in HTGML output.  It certainly should *not* be a block-level command in PGML, which are meant to handle chunks of text with delimiters at both ends (unless you want to require open and close back-tics in order to have the special processing).

I note that you want to use `` \`{} `` to render the back-tic in TeX mode, and for that you would want to add it to the list:

https://github.com/openwebwork/pg/blob/2763604901c69faf1beeacbf5413a7ad278e7cce/macros/core/PGML.pl#L1521-L1536

so that back-tics are substituted in the TeX format along with the other special characters.

As somiaj point out, you can handle the `\(` and `\[` delimiters in `Escape()` as well.  If WeBWorK is running with `processEscapes` enabled (the default), then you may need to prevent MathJax from processing `\\` and `\$`, which it will turn into `\` and `$`.  Alternatively,  set `processEscapes: false` in the MathJax config.

drdrew42:
https://github.com/openwebwork/pg/blob/2763604901c69faf1beeacbf5413a7ad278e7cce/macros/core/PGML.pl#L1542

This will also need to be updated to capture the back-tic for escaping in TeX output.

somiaj:
> Unnecessary, but I may come around and concede that they may be harmless.

In both our use cases, the escapes are useful because they are matching what would happen in TeX output by stopping mathjax under some configurations from rendering the math.  This to me is a visual indicator for those who may have an incorrect problem, and being able to see the problem not work in HTML output (which might be the only output a user checks) let's the user know there is an issue with the problem and they fix it. So in the end a properly written problem shouldn't haven't any of these characters to escape, and thus mostly harmless since there shouldn't be any mathJax deliminators to escape in a well written problem, and thus no extra span tags.

The only real use case I can think of having these characters in a problem and escaped with span tags is maybe in describing how to enter formulas into an essay answer or teach someone TeX.  And in that case you do want them escaped so they aren't rendered in any output (though maybe the backtick you are focusing on isn't as useful here).

The only issue I can think of with having these extra spans is if a user a custom css that hits all spans causing the text to render differently in the span than outside it (but in this case there maybe other issues custom css could cause).  Another possible issue I'm unsure about since I don't use a screen reader is if a screen reader has issues with extra spans or not. But since most problems will not have these extra deliminator, not a big issue for a well written problem.

I still think it is worth putting in a patch for this, as a properly written problem shouldn't have these extra characters to be escaped, and it might help users find issues with their problems they may have not been aware of.  I am probably going to apply a similar patch locally unless there is enough consensus that this should be fixed with PGML.

Alex-Jordan:
We can group discuss at tomorrow's meeting. This is (to me at least) not a comfortable thing to put in at the 11th hour before a release. If enough people think it is good, I'll agree. Also, while I know it doesn't help with your `\(` issue, I'd like to look into the MathJax config solution to blocking backticks as I have time. (I'm not super comfortable with javascript stuff so it takes me a bit to wrap my head around merging Davide's config with the current config in the right way.)

I noticed your patch wraps dollar signs in a span. You should check, but that may not be necessary. Our config doesn't use dollar signs to activate MJ. Also I was wondering if you would want to match on `` `+``, not just on `` ` ``, to put adjacent backticks together in one span.

Alex-Jordan:
I haven't explored, but there are things I don't yet understand about repeated backslashes. Like 

```
BEGIN_PGML

\

\\

\\\

\\\\

\\\\\

END_PGML
```

makes HTML output with 1, 1, 1, 1, 2 backslashes on the lines. I wanted to look into that and see first, if things are behaving as they should be. (I don't know if this is intended behavior.) And whether they are or aren't, if you tack on a `(` at the end, and then the patch wraps `\(` in a span, if output is correctly (un)affected. If you are working with your patch, maybe this is something else to look at.

dpvc:
> I don't yet understand about repeated backslashes.

Are you doing the test within an actual PG problem, or through another means?

Dealing with backslashes is always complicated in PG, because PG initially doubles all backslashes so that you can include TeX more easily in your PG problems.  E.g., in regular Perl,

``` perl
$f = "\(x \ne y\)";
```
would produce a string containing `(x ` followed by a newline followed by `e y)`.

But by doubling all the backslashes, that doubles them all inside `BEGIN_PGML/END_PGML` blocks as well.  Because of that, the `BEGIN_PGML/END_PGML` code must first un-doubles the backslashes and then processes the result as PGML.  To accommodate this, there are two entry points to PGML, `PGML::Format()` and `PGML::Format2()`.  The latter simply removes double backslashes and then calls `PGML::Format()` on the result.

It looks like your PGML has not gone through the backslash-doubling process.  When I run your example through PGML in an actual PG problem, I get 1, 1, 2, 2, 3 backslashes, which is what I expected (though the treatment of a backslash followed by a newline should be might be debated, and one might expect this to be 0, 1, 1, 2, 2 instead).

When I run your expression through `PGML::Foramt2()` using a direct call to `PGML::Format()` outside of the doubling done by PG, then I get the numbers you report.

Admittedly, I did my testing on an older version of WeBWorK/PG, so perhaps something has changed about the backslash doubling in the meantime (I confess to not paying too close attention to the changes that have been made over the last few years).  But I am confused by your results and wonder if you can shed light on that in any way.

Alex-Jordan:
Thanks Davide, those backslash counts are what I've been seeing with the
2.18 branch where I was looking at the backtick parsing, Jaimos's patch,
etc. But just a "clean" 2.18 branch.

drdrew42:
Raw HTML output on 2.18 is as @dpvc describes (1, 1, 2, 2, 3 backslashes), but when rendered in the browser, any remaining double-slash is converted to `<span>\</span>`. As a result, what is ultimately displayed is (1,1,1,1,2).

Raw HTML output:
```html
<div class="PGML">
<h2 style="margin:0">Slash Test</h2>
<div style="margin-top:1em"></div>
\
<div style="margin-top:1em"></div>
\
<div style="margin-top:1em"></div>
\\
<div style="margin-top:1em"></div>
\\
<div style="margin-top:1em"></div>
\\\
<div style="margin-top:1em"></div>
</div>
```

Browser DevTools inspect HTML:
```html
<div class="PGML">
<h2 style="margin:0">Slash Test</h2>
<div style="margin-top:1em"></div>
\
<div style="margin-top:1em"></div>
\
<div style="margin-top:1em"></div>
<span>\</span>
<div style="margin-top:1em"></div>
<span>\</span>
<div style="margin-top:1em"></div>
<span>\</span>\
<div style="margin-top:1em"></div>
</div>
```

Alex-Jordan:
With my production 2.17 server, I am putting this in a problem:

```
BEGIN_PGML

a\ b\\ c\\\ d\\\\ e\\\\\

END_PGML
```
I render the problem with Firefox and see;
Screen Shot 2023-06-01 at 9 07 23 AM


I view HTML source with Firefox (note, viewing source, not using inspector) and see:
```
<div id="output_problem_body"><div class="PGML">
a b\ c\ d\\ e\\\
<div style="margin-top:1em"></div>
</div>
```

I also confirm what drew sees with each backslash sequence on its own line, which is a different behavior than above.

drdrew42:
adding `processEscapes: false` produces the expected output w.r.t. multiple backslashes.

dpvc:
> when rendered in the browser, any remaining double-slash is converted to <span>\</span>.

Yes, that explains it, thanks.  It is, as @drdrew42 points out, due to `processEscapes` being true.  I did mention in my first comment above that "you may need to prevent MathJax from processing `\\` and `\$`, which it will turn into `\` and `$`", but had forgotten it already myself.  That means you either want to set `processEscapes` to false (probably a good idea in any case), or replace `\` with `<span>\</span>` (then you could ignore `\(` and `\[`, as they will be prevented from matching by having isolated their backslashes).

dpvc:
> I also confirm what drew sees with each backslash sequence on its own line

Yes, normally in Markdown (on which PGML is based), a backslash prevents the character that follows it from having its normal Markdown effect, and if the next character, so that `\*abc\*` would show `*abc*` literally rather than having abc in bold, as it usually would.  If the following character has no special meaning, then the backslash-character combination is replaced by just the character.  So `\ ` with a space just becomes a space.

I haven't looked closely, but I suspect that the backslash-newline combination is not being treated as backslash-character, but just a backslash at the end of the line, so doesn't match the pattern `/\\./` of a backslash followed by a character that triggers the backslash handling, and that trailing backslash is left in the text unchanged.  That is probably not the desired effect.

I could see two ways to treat backslash-newline:  either just remove the backslash and process the newline as normal, or treat backslash-newline as the hard line-break like ending the line with two spaces.  I could see either one making sense, but I think the first is probably more consistent with Markdown's meaning for backslashes.

In any case, that should probably be considered a separate bug that needs fixing.

Issue #833, opened 2023-05-24
Setting new symbol in Currency context


tbflowers:
I had originally posted this in the forum, but it seems like it may be better suited to be here. Alex Jordan shared a MWE which I will paste below.  (Forum post:  https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8278 )

DOCUMENT();
loadMacros(qw( PGstandard.pl PGML.pl contextCurrency.pl));

Context("Currency")->currency->set(symbol=>'!');   
$a = Currency(1000);
 
ENDDOCUMENT();

An error comes from setting the new symbol. This error does not occur in 2.15, but has been seen in versions 2.17 and 2.18.

drdrew42:
I have my suspicions that this arises from changes made to `Value::Context::Data::set` in #526 

`Currency::set` calls `SUPER::set` (`Value::Context::Data::set`) and follows with `Currency::update`, but by the time we get to `Currency::update`, the old operator has already been removed since `Value::Context::Data::set` now sets `$update = 1`. We then get this error from `Currency::update` attempting to redefine an operator that's already been removed.

drgrice1:
Commenting out line 339 of contextCurrency.pl (the `update` call in the `Currency::Context::currency::set` method) fixes the problem.  Probably not in a proper way though, but the problem works then.  This seems to confirm that @drdrew42 is correct in his assessment.

drgrice1:
Actually, looking closer it seems that this is due to #518, and not #526.  The branch of code that is called when the currency `set` method is called is there.  Since `$update = 1` is set here and thus the `SUPER` update method is called, perhaps the correct fix is to remove line 339 of contextCurrency.pl.

drdrew42:
No, I think the solution should be the removal of these lines from `Currency::update`

```perl
$operators->redefine($data->{symbol}, from => $context, using => $self->{symbol});
$operators->remove($self->{symbol});
```

perhaps even removing that entire conditional, since we can trust the parent to handle the symbol updates.

drgrice1:
Testing this it seems that my suggestion of removing line 339 (which means the entire `set` override method should be removed) works.  Testing the suggestion to remove lines 385 and 386 of the `update` method does not.  Test with the following MWE modified from @Alex-Jordan's previous example:
```perl
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl contextCurrency.pl));

Context('Currency')->currency->set(symbol => '!');
$a = Currency(1000);

[...]

PR #839, opened 2023-06-01
Remove the contextCurrency.pl `set` method override.

drgrice1:
The only thing the override did was call the `update` method to force patterns to be updated.  This is no longer needed since uddate is now always called by the SUPER method when tokens are added.  This was changed in #518 and #526.

This fixes issue #833 in what I believe is the correct way.  @dpvc please advise if this is incorrect, or if there is a better way to do this.


Alex-Jordan (approved):
Tested that this fixes the reported issue.

pstaabp (approved):
Fixes the issue.

drdrew42 (approved):

PR #840, opened 2023-06-04
Add a default checker to the parserGraphTool.pl macro.

drgrice1:
The default checker is designed to grade the graph based on appearance. This means that if a student graphs duplicate objects, then the duplicates are ignored.  Furthermore, if two objects are graphed whose only difference is that one is solid and the other is dashed (in this case the dashed object is covered by the solid object and only the solid object is really visible), then the dashed object is ignored.

Note that fills are only checked by this checker if the other graph objects are correct.  This is because, without the other elements of the correct answer, there is not a valid setting to evaluate if the fill is correct or not.

The variable `$graphToolObjectCmps` is made available for use in custom checkers, and contains a hash whose keys are the types of the objects available (line, circle, parabola, fill, point, cubic, quadratic, and inverval), and whose values are methods that can be called passing one of these objects (typically a correct answer in a checker).  When one of these methods is called it will return two methods.  The first method when called passing a MathObject point will return 0 if the point satisfies the equation of the object, -1 if the equation evaluated at the point is negative, and 1 if the equation evaluated at the point is positive.  The second method when called passing another graph tool object (for example a student answer) will return 1 if the two objects are exactly the same, and 0 otherwise.  A second parameter may be passed and if that parameter is 1, then the method will return 1 if the two objects are the same ignoring if the two objects are solid or dashed, and 0 otherwise.  See the added POD for an example of using this in a custom checker.


drgrice1:
Examples of problems using the default checker are attached. 
[examples.zip](https://github.com/openwebwork/pg/files/11647345/examples.zip)

Alex-Jordan (approved):
Checked out this branch and tested by playing with a few sample files, and it worked as advertised.

pstaabp:
The shaded regions in the plane I don't think are working.  Here's an example for Problem #11

image

When I clicked the point (0,0) for the fill point, I got the correct answer.  

I also had trouble with the same issue on problem #10 in that the fill only gave the correct answer when clicked at the exact point in the solution.

drgrice1:
The behavior you are describing would be the behavior for this problem without this pull request.  I just verified that with that problem and this pull request, I can choose any point in the correct region, and it is marked correct.  Are you sure that you have this pull request checked out?  Also possible (but seemingly unlikely) is that you have a local copy of the parserGraphTool.pl macro in your courses macro directory that is overriding this.

drgrice1:
What is the seed that you are using?

drgrice1:
Also, try adding a warn statement in the default checker added in this pull request to see that the checker is being called.

pstaabp:
Seems to be working now.  Not sure what was going on.

pstaabp (approved):
Looks good.  By the way, huge improvement with default checkers.  I remember writing up some problems and I was guessing most authors would not attempt to write one of these problem needing to write a default answer checker. 

Now off to update the sample problems....

Alex-Jordan:
Two approvals, merging.

PR #841, opened 2023-06-05
Add a method to the parserGraphTool.pl macro to generate more graphs.

drgrice1:
Just call `$gt->generateAnswerGraph` in a problem solution to display the correct answer.

### DIFF
--- a/htdocs/js/apps/GraphTool/graphtool.js
+++ b/htdocs/js/apps/GraphTool/graphtool.js
@@ -9,7 +9,7 @@ window.graphTool = (containerId, options) => {
 	const gt = {};
 
 	gt.graphContainer = document.getElementById(containerId);
-	if (getComputedStyle(gt.graphContainer)?.width === '') {
+	if (gt.graphContainer.offsetWidth === 0) {
 		setTimeout(() => window.graphTool(containerId, options), 100);
 		return;
 	}
--- a/htdocs/js/apps/GraphTool/graphtool.scss
+++ b/htdocs/js/apps/GraphTool/graphtool.scss
@@ -348,3 +348,23 @@
 		}
 	}
 }
+
+.graphtool-solution-container {
+	.graphtool-graph {
+		width: 402px;
+		height: 402px;
+
+		@media only screen and (max-width: 600px) {
+			width: 302px;
+			height: 302px;
+		}
+	}
+
+	.graphtool-number-line {
+		height: 67px;
+
+		@media only screen and (max-width: 600px) {
+			height: 57px;
+		}
+	}
+}
--- a/macros/graph/parserGraphTool.pl
+++ b/macros/graph/parserGraphTool.pl
@@ -341,6 +341,53 @@ the C<IntervalTool> is used.
 
 =back
 
+=head1 METHODS
+
+=over
+
+=item generateAnswerGraph
+
+This method may be called for a GraphTool object to output a static version of the graph into the
+problem. The typical place where this might be desired is in the solution for the problem. For
+example
+
+    BEGIN_PGML_SOLUTION
+    The correct graph is
+
+    [@ $gt->generateAnswerGraph(ariaDescription => 'a better description than the default') @]*
+    END_PGML_SOLUTION
+
+The following options may be passed to this method.
+
+=over
+
+=item C<showCorrect>
+
+Whether to show correct answers in the graph. This is 1 by default.
+
+=item C<cssClass>
+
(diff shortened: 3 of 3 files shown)


drgrice1:
@Alex-Jordan:  You may need to make adjustments for the `PTX` display mode.

Alex-Jordan:
I tested and this works for me for all three output modes.

Would it make sense for the "text" method of a graph tool to call this method? So like
```
BEGIN_PGML_SOLUTION
The correct graph is [$gt].
END_PGML_SOLUTION
```
would print the actual graph instead of something like `{line, solid, (3,0), (0,-6)}`?

drgrice1:
I assume by the "text" method you mean the Value (MathObject) "string" method, which is what would be called in your example.  Unfortunately, that can't be done here.  The reason is that the "string" method is what is called to convert the correct answer into text.

To see this, insert `sub string { return shift->generateAnswerGraph(@_); }` after the added `generatedAnswerGraph` method in `parserGraphTool.pl` in this pull request.  Then you could do
```
BEGIN_PGML
The correct graph is

[gt]*
END_PGML
```
(Note the need of the `*` to prevent PGML from escaping the html output this would give.)

That works both in html and TeX.  However, now check the "Show:  Correct Answers" checkbox and click "Check Answers", and then click on the graph shown in the "Correct Answer" column of the results table.  You will see where the problem is.

Alex-Jordan:
PGML never implemented the image block `[! ... !]` (nor the link block `[< ... >]`). I took a look just now to see what it would take to implement it, but I am stumped. Maybe for the next release, we can make it possible to do like:

```
[! $gt !] 
[! $latex-or-tikz-image !]
[! $wwplot-image !]
[! path-to-local-image-file !]
```

with options to control size.

drgrice1:
I found a way to make `[$gt]*` work.  It may be that this could cause problems in some way by using the string method in this way, but I can't think of anything wrong with it or see any problems in testing.

drgrice1:
The `string` override needed to be reverted.  The MathObject code is calling this method several times even if the problem does not use it via the `[$gt]*` approach.  I am seeing that it is called three times when a graph tool problem loads even if the problem does not use the `[$gt]*` construct.  That is particularly bad in hardcopy as this results in latex being run three additional times (in addition to the graph that is actually displayed), and I can see four images generated when only one should be generated.

Alex-Jordan (approved):

Alex-Jordan:
Here is a discussion about `[! ... !]` from a few years ago:
https://github.com/openwebwork/pg/discussions/562#discussioncomment-577446

If we implement this the way David suggests:
`[! alt text !]{src}`
PGML could inspect the `ref` of `src` and do the right thing when that ref is `parser::GraphTool`. Or the other ways discussed there but the point is to inspect the `ref` of `src` and do different things in different cases.

For writing PTX source that produces PG, there is nothing that directly invokes `[@ ... @]*`. But it would be nice if PTX like:
```
<image source="$g">
    <description>my picture</description>
</image>
```
simply translated to something like `[! my picture !]{$g}` where `ref $g` could be `parser::GraphTool`, `PGlateximage`, `WWPlot`, etc. It would all be written the same way by an author.

I'm just musing, and it's not important for this PR.

Alex-Jordan:
OK, I have `[!alt text!]{source}{width}{height}` working, with the last two optional. It works for static images, WWPlot, PGtikz, and PGlateximage.

And it works for `parser::GraphTool` objects as well. But there is no place to send the width or height. Are these images always supposed to be a hard coded size or would you want width and/or height specified?

drgrice1:
Sizes are determined by width and height settings in a css class.  To set that you need to set a css class for the image, and add that via HEADER_TEXT.  So that probably won't work with this scheme.  Generally, you shouldn't change the width and height.  The graphtool is not really designed to work well with other sizes, and any answer graphs you generate should be the same size.

pstaabp (approved):
I know we got a bit off topic here, but this works.

drgrice1:
Also, try adding a warn statement in the default checker added in this pull request to see that the checker is being called.

Alex-Jordan:
Could there be height and width options (in pixels for consistency with other things) that are only used for sizing the hardcopy?

Alex-Jordan:
If in the future there were width/height arguments you could pass to `generateAnswerGraph()`, do you know what you would call them? And if it will be OK to assume they are plain integer pixel counts? (Even if they end up only affecting tex). 

If I know what to call them, I can go ahead and put that into an edit I have open for `image()` in `PGbasicmacros.pl`.

PR #844, opened 2023-06-07
Fix cubic's in hardcopy for the GraphTool.

drgrice1:
There was an obvious issue with a variable that is not defined in the block.


drgrice1:
Here is a minimal example problem to test with.
[GraphCubic.pg.txt](https://github.com/openwebwork/pg/files/11671255/GraphCubic.pg.txt)

On develop you will see errors when you try to generate a hardcopy.  With this pull request -- no errors!

drgrice1:
Actually, here is the example I meant to attach.  With this one you will see errors on the develop branch.  With the other example there are only warnings.
[GraphCubic.pg.txt](https://github.com/openwebwork/pg/files/11671271/GraphCubic.pg.txt)

pstaabp (approved):
Fixes the issue.

PR #845, opened 2023-06-07
Image markup for PGML

Alex-Jordan:
This lets you use `[! alt text!]{image source}{optional width}{optional height}` to display an image in PGML. It works for static images files relative to the problem file, WWPlot images, PGtikz images, and PGlateximages.

It also works with GraphTool images once #841 is merged.

The attached problem file demonstrates it. If you use it as is, put this image file in the same folder for the demo of the static image.

webwork-logo

Here is the sample problem file (change the extension).
[PGMLimage.txt](https://github.com/openwebwork/pg/files/11672997/PGMLimage.txt)

### DIFF
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -1366,11 +1366,7 @@ sub Image {
 	my $width    = $item->{width}  || 100;
 	my $height   = $item->{height} || '';
 	my $tex_size = $width / 600 * 1000;
-	if (ref $source eq 'parser::GraphTool') {
-		return ($source->generateAnswerGraph(ariaDescription => $text));
-	} else {
-		return (main::image($source, alt => $text, width => $width, height => $height, tex_size => $tex_size));
-	}
+	return (main::image($source, alt => $text, width => $width, height => $height, tex_size => $tex_size));
 }
 
 ######################################################################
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I pushed another commit to by branch at the Alex-Jordan GitHub account, where this PR comes from.  I can see it there:

https://github.com/Alex-Jordan/pg/commits/PGMLimage

But when I visit this PR in GitHub I still only see one commit. I'm not sure what is going on.

somiaj:
I see two commits, one was made yesterday about 11 hours ago, and one was made today about 15 mins ago.

Alex-Jordan:
Weird. I have been hard refreshing this page and still only see one.

somiaj:
Here is what I see, just so you can double check they are the correct commits.

commits_000

And here is the link to the most recent commit (maybe you can see that)

https://github.com/Alex-Jordan/pg/commit/c361c8529c542a53b611754d4d78ab65aece8dd2

Alex-Jordan:
What do you see at the top of this PR page? I do see the two commits over in my branch. But here in this PR I still see:

Screen Shot 2023-06-07 at 10 16 25 AM

somiaj:
Sorry, misunderstood you. I to only see one commit here.

Alex-Jordan:
I changed the base to `develop` and then back to `PG-2.18` and that seems to have reset things.

drgrice1:
I merged #841 with this branch to test this with the parserGraphTool.pl macro as described.  It works with your problem when you uncomment the line for this.  However, you will get a warning about a missing answer blank.  This is because the problem does not add the graphtool answer rule.  The macro is not designed to be a static image tool.  Of course usually you would have the answer rule, and the image would be in the solution.

PR #846, opened 2023-06-07
Fix the TeX fill clipping code for points.

drgrice1:
Points should not have clipping code.  This breaks fills if you have a static object point because it essentially clips out evertything.


drgrice1:
Attached is a problem to test this with.  Generate a hardcopy of that problem.  There should be a circle of radius 5 centered at the origin that is filled in, and a point at (6,8).  With the develop/PG-2.18 branch the circle will not be filled in.  With this pull request it will be.

[GraphLine.txt](https://github.com/openwebwork/pg/files/11681442/GraphLine.txt)

PR #848, opened 2023-06-08
image size defaults

Alex-Jordan:
This tweaks what defaults are used for the size of an image from `image()`.

Before this, `width` always defaults to 100 when unspecified, and `tex_size` always to 800 when unspecified.

Now it's more nuanced. An unspecified `width` will default to 100 _only_ when `height` is also undeclared. If `height` is specified, then `width` is left unspecified, and so you can specify a height and let the browser maintain the image's natural aspect ratio.

Also `tex_size` will default to 800 _only_ when `width` has no value (no value passed to `image()`, and the above has not set it to 100). When `width` does have a value, `tex_size` defaults to something that tries to respect whatever `width` is, translating from a presumed 600px wide reading area to something proportional out of 1000.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -3008,9 +3008,9 @@ sub image {
 	}
 
 	# Get options for width, height, and tex_size, with a sanity check for integer values.
-	my $width    = $out_options{width}    =~ /[1-9]\d*/ ? $out_options{width}    : '';
-	my $height   = $out_options{height}   =~ /[1-9]\d*/ ? $out_options{height}   : '';
-	my $tex_size = $out_options{tex_size} =~ /[1-9]\d*/ ? $out_options{tex_size} : '';
+	my $width    = $out_options{width}    =~ /^[1-9]\d*$/ ? $out_options{width}    : '';
+	my $height   = $out_options{height}   =~ /^[1-9]\d*$/ ? $out_options{height}   : '';
+	my $tex_size = $out_options{tex_size} =~ /^[1-9]\d*$/ ? $out_options{tex_size} : '';
 	$width = 200 unless ($width || $height);
 
 	$tex_size = $width ? int($width / 0.6) : 800 unless $tex_size;
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -2963,7 +2963,9 @@ PGtikz object, or parser::GraphTool object.
 
 C<width> and C<height> are positive integer pixel counts for HTML display. If both
 are missing, C<width> will default to 200 and C<height> will remain undeclared,
-letting the browser display the image with its natural aspect ratio.
+letting the browser display the image with its natural aspect ratio. Except with
+a parser::GraphTool object, if both are missing then nothing will be passed along
+to the GraphTool method for display, and that macro's defaults will be used.
 
 C<tex_size> is also a positive integer, per 1000 applied to the line width in TeX.
 For example 800 leads to 0.8\linewidth. If over 1000, then 1000 will be used.
@@ -3043,8 +3045,7 @@ sub image {
 			push(
 				@output_list,
 				$image_item->generateAnswerGraph(
-					width           => $width,
-					height          => $height,
+					$out_options{width} || $out_options{height} ? (width => $width, height => $height) : (),
 					ariaDescription => shift @alt_list // ''
 				)
 			);
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -3006,18 +3006,16 @@ sub image {
 				display_options2(%known_options);
 		}
 	}
-	my $width    = $out_options{width};
-	my $height   = $out_options{height};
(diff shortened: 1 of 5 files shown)


drgrice1:
I am toying with the possibilities of allowing width and height for the graph tool answer graphs, but one thing I am seeing is that the default width of 100 is always too small to be of any use for the graph tool.

To be honest, I believe it is too small for images in general.  Is it time to consider setting that to a more sane default?  I guess for backwards compatibility with bad past practices we are probably stuck with that?

Alex-Jordan:
Not sure, but at a minimum there could be a different default for a GraphTool object.

Alex-Jordan:
In the 90s, monitors were like 640x480. Maybe 100px is not so small in absolute terms when that's the resolution.  According to the GitHub Blame on `PGbasicmacros.pl`, that default of `100` has been there since 2003 when the commit history begins, and probably earlier.

drgrice1:
I think a sane minimum width for the graph tool is 200.  Working with what you have, I was adding a minimum check in the graph tool macro to ensure a minimum of 200.

To answer your question on #841 (https://github.com/openwebwork/pg/pull/841#issuecomment-1583463486), I think I will just use `width` and `height` for the parameters for html, and `texSize` for the TeX size.

The plan is to make the default value of `texSize` the value of the object's `texSize` parameter (which defaults to 400).  So you will also need to pass in your computed `tex_size` for that parameter in the `image` method.

Alex-Jordan:
I think I am fine with changing the default width to 200. It's the problem author's responsibility to either write a width, or be happy with the macro's default, and I'm not sure that never changing macro defaults is a good policy. Screen resolution in general has more than doubled (in one dimension) since that 100 was set, so now 100 looks small. And since 100 has been small in absolute terms for a while now, I wonder if that has actually motivated authors to write in a width.

somiaj:
For what it is worth, `PGgraphmacros.pl` uses a PG configuration setting `onTheFlyImageSize` (set in `conf/pg_config.dist.yml`) for its default, which is set to 400.   You could use that environment variable `$main::envir{onTheFlyImageSize}` for the default width, which is then configurable.

Alex-Jordan:
Several `100` are now `200`. Also I fixed a mistake with the width for a video in PTX output.

Here's an example of a problem that uses the default and the images are way too small: `Library/Dartmouth/setStewartCh10S2/problem_1/problem_1.pg`

macros/core/PGbasicmacros.pl
```diff
@@ -3023,7 +3040,14 @@ sub image {
 	while (@image_list) {
 		my $image_item = shift @image_list;
 		if (ref $image_item eq 'parser::GraphTool') {
-			push(@output_list, $image_item->generateAnswerGraph(ariaDescription => shift @alt_list // ''));
+			push(
+				@output_list,
+				$image_item->generateAnswerGraph(
+					width           => $width,
+					height          => $height,
+					ariaDescription => shift @alt_list // ''
```
drgrice1, line 3048: I said earlier a sane default for the graph tool was a width of 200, but what I really should have said is a sane minimum for the width is 200.  Really, that is generally still to small.  The sane default is really what the graph tool uses by default to begin with.  That is why it uses that.  That also has the added advantage of being a responsive size.  On narrow screens it is smaller than on wide screens.  So it would be better here if the width or height were only given in the case that one of them is explicitly given.  Then if neither are given, the graph tool will use the true sane defaults.

drgrice1, line 3048: Here is one way to do what I am suggesting.
```suggestion
				$image_item->generateAnswerGraph(
					$out_options{width} || $out_options{height} ? (width => $width, height => $height) : (),
					ariaDescription => shift @alt_list // ''
```

drgrice1:
@Alex-Jordan:  I added a pull request to this branch that adds the `height` and `width` options to the graph tool `generateAnswerGraph` method.   I thought it might be good to put that with this for testing purposes, and to work out details.

macros/core/PGbasicmacros.pl
```diff
@@ -2995,18 +3006,24 @@ sub image {
...
 		}
 	}
-	my $width       = $out_options{width};
-	my $height      = $out_options{height};
-	my $tex_size    = $out_options{tex_size};
+
+	# Get options for width, height, and tex_size, with a sanity check for integer values.
+	my $width    = $out_options{width}    =~ /[1-9]\d*/ ? $out_options{width}    : '';
+	my $height   = $out_options{height}   =~ /[1-9]\d*/ ? $out_options{height}   : '';
+	my $tex_size = $out_options{tex_size} =~ /[1-9]\d*/ ? $out_options{tex_size} : '';
```
drgrice1, line 3013: I realized that these sanity checks aren't quite right.  They should be `/^[1-9]\d*$/`.  If the regex is not anchored, then something like 'asdf500' would match and cause a warning.

Alex-Jordan: Good catch, updated now.

pstaabp (approved):
Looks good.

drgrice1 (approved):

Issue #849, opened 2023-06-10
GraphTool correct answer graph of line when one point is just outside viewing window


Alex-Jordan:
If you render the following problem, and click to show Correct Answers, it is clearly wrong.

```
DOCUMENT();

loadMacros(qw(PGstandard.pl PGML.pl parserGraphTool.pl));

$gt = GraphTool("{line, solid, (9,3), (11, 5)}");

BEGIN_PGML

[_]{$gt}

END_PGML

ENDDOCUMENT();
```

The line should have slope 1, but I see
<img width="264" alt="Screen Shot 2023-06-09 at 7 17 57 PM" src="https://github.com/openwebwork/pg/assets/4732672/c433aefb-8807-428c-938d-299572349102">

It seems that the point (11,5) has been snapped to (10,5).

Note that (11,5) is outside the viewing window, but it would not be uncommon for this to happen when generating a problem randomly. The student can still make the correct line using two points within the viewing window. I did not bother including a checker for this MWE.

I haven't searched exhaustively but I have only seen this happen when one point is only one unit outside of the viewing window.

pstaabp:
Fixed with #850

PR #850, opened 2023-06-10
Implement workaround for JSXGraph issue with points created off the board.

drgrice1:
If a point is created off the board with `snapSizeGrid` true, then the coordinate that is outside of the board's bounding box range is moved one snap size unit toward the bounding box range.  See https://github.com/jsxgraph/jsxgraph/issues/545 for issues on this bug in JSXGraph.

To fix this points need to be created with `snapToGrid` false (the default value), and then that attribute set to true afterward.  In order for this to work the coordinates of the point need to be manually rounded to the snap size.

This is only done for initial creation of objects (answers or static objects).  It does not need to be done for objects that are created dynamically (by student interaction) since those points are forced to be on the board anyway.  This is also not done for 'point' graph tool objects, since the correct answer should not be a point off the board anyway.

This workaround fixes issue #849.

### DIFF
--- a/htdocs/js/GraphTool/cubictool.js
+++ b/htdocs/js/GraphTool/cubictool.js
@@ -129,9 +129,12 @@
 				},
 
 				createPoint(gt, x, y, grouped_points) {
-					const point = gt.board.create('point', [x, y], {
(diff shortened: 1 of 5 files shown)


pstaabp (approved):
Fixes the issue.

pstaabp:
@drgrice1 Will need to resolve conflicts.  Just merged #851.

PR #852, opened 2023-06-12
padding option for niceTables

Alex-Jordan:
This allows you to have a `padding` option for a table from `niceTables.pl`, as described in the POD.

It slightly changes the default padding so that (1) now it is based on the (present-day) default font sizes, (2) the default for top/down padding in hardcopy is unchanged, and (3) the left-right padding is measured on the same scale as top-down padding, which is the font size.

This makes the left-right padding a little bit smaller for a DataTable, and a little bit larger for a LayoutTable. And it changes the default HMTL top-down padding in a way I did not examine They could be adjusted to exactly match either what they currently are for HTML, or what they currently are for PDF, but not both. What I chose here is a happy medium, with "nice" numbers 0, 0.5, and 1.

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -154,6 +154,18 @@ rule to the right of a column of row headers. Default is 1.
 Can be C<'top'>, C<'middle'>, or C<'bottom'>. Applies to all rows.
 See below to override for an individual row.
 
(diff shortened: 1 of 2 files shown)


drgrice1 (approved):
Thanks for adding this.  This seems to work very well for what I wanted.

pstaabp (approved):
👍

PR #855, opened 2023-06-19
only assign type when needed

Alex-Jordan:
This should fix openwebwork/webwork2#2065.


drgrice1 (approved):
This fixes the problem in my test.

drgrice1:
Two approvals here.  So merging.

PR #856, opened 2023-06-21
remove remnant from original WeBWorK-MBX bridge

Alex-Jordan:
These are old artifacts from the original 2016 WeBWorK-PTX bridge, that tried to use WeBWorK tex output directly. Now just clutter in the tex output.

Probably the special line of PTX output can be taken out and folded into `conf/pg_config.dist.yml`, but I don't know what that file is for so leaving it be.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -2491,18 +2491,7 @@ A wide variety of google widgets, youtube videos, and other online resources can
 =cut
 
 sub beginproblem {
-	my $out = "";
-	$out .= MODES(
-		TeX => "\n%%% BEGIN PROBLEM PREAMBLE\n"
-		,    #Marker used in PreTeXt LaTeX extraction; contact alex.jordan@pcc.edu before modifying
-		HTML => ""
-	);
-	$out .= MODES(%{ main::PG_restricted_eval(q!$main::problemPreamble!) });
-	$out .= MODES(
-		TeX => "\n%%% END PROBLEM PREAMBLE\n"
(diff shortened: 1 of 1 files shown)


drgrice1:
The `conf/pg_config.dist.yml` file is used for default values for the WeBWorK::PG::Environment.  For webwork2, those values are overridden by the course environment.  For the standalone renderer those values are overridden by a copy of that file to `conf/pg_config.yml`.

As to the `problemPreamble` and `problemPostamble` values, those are set to avoid errors if those environment variables are undefined.  The `problemPreamble` is only used in the `beginproblem` method, so if we remove it there then that variable is no longer needed.  The `problemPostamble` is only used in the `ENDDOCUMENT` method of `PG.pl`, so if line 551 of that file is deleted, then that variable is also not needed.  Note that these are still set in `defaults.config` and commented out code is present in `localOverrides.conf` for webwork2.

We just need to decide if there is any valid reason that these variables should be used.  If not (and I think not), then we should go all the way with them, and delete them.  The commented out examples of their usage in webwork2's `defaults.config` and `localOverrides.conf` are certainly not valid things that these should be used for.

drgrice1 (approved):
Looks good to me.

pstaabp (approved):
This seems to keep WeBWorK problems the same--haven't tested PTX.

Alex-Jordan:
This was not even PTX really. I know the comment says "PreTeXt" but that was from some point where all instances of "MathBook XML" aka MBX were relabeled. These comments were for MBX, which would cut up WW TeX output into chunks, and used this comment to identify where the statement of a problem in tex output actually began. But it's been a long time since WW just started making PTX output instead, and PreTeXt no longer cares what WW's TeX output has in it.

PR #857, opened 2023-06-21
Fix a mistake in WWPlot.pm that makes gif image generation fail.

drgrice1:
This was a mistake that I made when I added the `ext` method for TikZ image generation.  Fortunately, I doubt anyone uses `gif` images much (if at all), since `png` is better.


Alex-Jordan (approved):
It's good, of course. I'll let a second person merge though.

pstaabp (approved):

PR #858, opened 2023-06-21
replace MathObjects negative space with \mathopen{}

Alex-Jordan:
See https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8303 for an explanation. I think this would be a safe way to drop the negative space without introducing a big gap. The cases I know of where this use of `\mathopen{}` could cause things to look bad are not situations that will come up in these two places.


pstaabp:
This looks good.  I tested the problem posted in the forum as well as a few other cases involving ( ).  I wonder if this might affect other cases.

dpvc (approved):
This should be OK, and better than the negative space.

drgrice1 (approved):
This looks good to me.

PR #859, opened 2023-06-21
a few additional units

Alex-Jordan:
In upgrading my production server, I had to take notice of all the customizations we've added. These are some that are worth contributing. I was especially surprised to see that gallons are missing.

### DIFF
--- a/lib/Units.pm
+++ b/lib/Units.pm
@@ -230,10 +230,21 @@ our %known_units = (
 		factor => 0.0001,
 		m      => 3
 	},
+	# U.S./English volume units
 	cup => {
 		factor  => 0.000236588,
 		m       => 3,
 		aliases => ['cups']
+	},
+	pt => {
+		factor  => 0.000473176473,
+		m       => 3,
+		aliases => [ 'pint', 'pints' ]
+	},
+	qt => {
+		factor  => 0.000946352946,
+		m       => 3,
+		aliases => [ 'quart', 'quarts' ]
 	},
 	gallon => {
 		factor  => 0.00378541,
--- a/lib/Units.pm
+++ b/lib/Units.pm
@@ -122,11 +122,12 @@ our %known_units = (
 	hour => {
 		factor  => 3600,
 		s       => 1,
-		aliases => [ 'hours', 'hr', 'h' ]
+		aliases => [ 'hours', 'hrs', 'hr', 'h' ]
 	},
 	day => {
-		factor => 86400,
-		s      => 1
+		factor  => 86400,
+		s       => 1,
+		aliases => ['days']
 	},
 	month => {    # 60 * 60 * 24 * 30
 		factor  => 2592000,
@@ -179,6 +180,11 @@ our %known_units = (
 	fm => {    # femtometer
 		factor => 1E-15,
 		m      => 1
+	},
+	meter => {
+		factor  => 1,
+		m       => 1,
+		aliases => ['meters']
 	},
 
 	# ENGLISH LENGTHS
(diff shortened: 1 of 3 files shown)


lib/Units.pm
```diff
@@ -180,6 +181,11 @@ our %known_units = (
 		factor => 1E-15,
 		m      => 1
 	},
+	meter => {
+		factor  => 1,
+		m       => 1,
+		aliases => ['meters']
+	},
```
drgrice1, line 188: This should not be added as a new unit.  Instead "meter" and "meters" should be added directly as aliases of "m" (defined on line 42 of this file).

Alex-Jordan:
Oh I didn't realize the fundamental units had this style of definition too.

drgrice1:
Yeah, all of the fundamental units are defined first in the`%known_units` hash.

pstaabp (approved):

lib/Units.pm
```diff
@@ -233,14 +235,19 @@ our %known_units = (
 		m       => 3,
 		aliases => ['cups']
 	},
+	gallon => {
+		factor  => 0.00378541,
+		m       => 3,
+		aliases => [ 'gallons', 'gal' ]
+	},
```
drgrice1, line 242: I think there are some other related units that could be added here.  Pint and quart go with these volume measurements, and are equally standard with cup and gallon.  There are also teaspoon and tablespoon, that are rather common.  If you really want to go the distance there are also pottle, barrel, shot, gill, hogshead, etc.  I haven't heard of some of those before.  Some fun names too!  See [customary US units](https://en.wikipedia.org/wiki/United_States_customary_units).

drgrice1 (approved):
Looks good.  But how far do we go with this?

lib/Units.pm
pstaabp: I'll have a hogsheads of mead with lunch today. 😄

Alex-Jordan:
I'm fine with going further. But here are some possible principles that
could restrain us.

Any reasonably common unit for stem problems should be in there. Gallons
counts, hogshead not.

For any unit in there, all reasonable aliases in the context of what a
student might type should be available. Like "meter".

Individual problems can always add new units like hogshead.

Questions I have though:
* What about metre and litre?
* We have L and ml but not mL.
* Students can want to type periods after some units, should we cater to
that?

As far as this pr goes these are currently just the units we've added at
PCC.

pstaabp:
I'm prepping a PR to this branch for pints and quarts and adding some tests for these.  Should I add 'litre(s)', 'metre(s)' etc. to this as well?

pstaabp:
@Alex-Jordan I put a PR into your branch.  Also added the new units to the help file.

lib/Units.pm
duffee: For those that like extreme edge cases, the Gal is the acceleration of 1 cm/s^2, but was deprecated by the Int. Ast. Union in 1989.  Probably an idea to include a comment that the gallon is US, not Imperial.

https://en.wikipedia.org/wiki/Gal_(unit)
https://www.iau.org/publications/proceedings_rules/units/

Alex-Jordan:
I accidentally rebased on 2.18 instead of merging 2.18 here. The GitHub diff looks weird now. I'll see if I can fix it.

Alex-Jordan:
I think I straightened this out.

drgrice1:
A rebase should work fine.

Alex-Jordan:
It works better when you see there was a conflict and fix it, instead of just pushing it half-rebased.

lib/Units.pm
Alex-Jordan: @duffee In the help file, I put "US " in front of the title for each of gallon, quart, pint, and cup.

PR #863, opened 2023-06-23
pg.sty

Alex-Jordan:
This introduces a latex sty file for PG macros. For getting started, it has one macro, `\answerRule{}` which takes a numerical argument and makes a rule so many ex wide. I sunk it 3pt lower from what answer rules have always been in `PGbasicmacros.pl`.

This pairs with https://github.com/openwebwork/webwork2/pull/2081.

To test, get on both branches. Have a problem with answer rules, preferably that includes an answer rule array. First visit the problem editor. Test that you can build the PDF (and see if the sunken underline is in effect.) Then test that you can download the tex and build it locally.

Then put the problem in a set. See if you can generate the PDF hardcopy. And see if you download the tex bundle if you can build the set.

### DIFF
--- a/assets/tex/pg.sty
+++ b/assets/tex/pg.sty
@@ -1,0 +1,4 @@
+\ProvidesFile{pg}
+
+\newcommand{\answerRule}[1]{\raisebox{-3pt}{\parbox[t]{#1ex}{\hrulefill}}}
+
--- a/assets/tex/pg.sty
+++ b/assets/tex/pg.sty
@@ -1,4 +1,4 @@
 \ProvidesFile{pg}
 
-\newcommand{\answerRule}[1]{\raisebox{-3pt}{\parbox[t]{#1ex}{\hrulefill}}}
+\newcommand{\answerRule}[2][]{\raisebox{-3pt}{\parbox[t]{#2ex}{\hrulefill}}}
 
--- a/assets/tex/pg.sty
+++ b/assets/tex/pg.sty
@@ -1,4 +1,5 @@
-\ProvidesFile{pg}
+\NeedsTeXFormat{LaTeX2e}
+\ProvidesPackage{pg}[2023/06/26 version 2.18]
 
 \newcommand{\answerRule}[2][]{\raisebox{-3pt}{\parbox[t]{#2ex}{\hrulefill}}}
 
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -316,7 +316,7 @@ sub NAMED_ANS_RULE {
 	$tcol = $tcol < 40 ? $tcol : 40;           ## get min
 
 	MODES(
-		TeX        => ($name ? "\\answerRule[$name]{$tcol}" : "\\answerRule{$tcol}"),
+		TeX        => "\\answerRule[$name]{$tcol}",
 		Latex2HTML => qq!\\begin{rawhtml}<input type=text size=$col name="$name" value="">\\end{rawhtml}!,
 
 		# Note: codeshard is used in the css to identify input elements that come from pg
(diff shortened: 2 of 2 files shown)


pstaabp:
This all works well.  As mentioned earlier, I would recommend having the default be an empty box instead of a line.  If we do that, we should switch the essay_box over to a box as well.

Alex-Jordan:
I don't think an outlined box is good for print when you consider someone
trying to hand write something like a rational expression.

drdrew42:
Is there any way to label these answer rules without affecting the output? I'm thinking of ways to post-process the output (once the answers are processed) to insert the correct answers... in HTML the inputs are labeled with an ID, can we do something similar here?

Alex-Jordan:
If it's within the scope of a single problem then you could tag it with the
answer name. Like \answerRule[AnSwEr0001]{10} or something. And the macro
could give you whatever marker you need in the .tex for that.

Alex-Jordan:
I changed it so that the .tex file has like `\answerRule[AnSwEr0001]{10}`. This is a start to what I think @drdrew42 commented on.

macros/core/PGbasicmacros.pl
```diff
@@ -316,7 +316,7 @@ sub NAMED_ANS_RULE {
 	$tcol = $tcol < 40 ? $tcol : 40;           ## get min
 
 	MODES(
-		TeX        => "\\answerRule{$tcol}",
+		TeX        => ($name ? "\\answerRule[$name]{$tcol}" : "\\answerRule{$tcol}"),
```
drgrice1, line 319: The `$name ? ...` checking is pointless.  `$name` will always be set.  All answers are actually labeled answers internally for PG.  So this should be changed to
`TeX => "\\answerRule[$name]{$tcol}"`.  That is what will always happen anyway.  The same for the rest of the instances in this file.

Alex-Jordan: OK, I changed those.

assets/tex/pg.sty
```diff
@@ -0,0 +1,4 @@
+\ProvidesFile{pg}
```
drgrice1, line 1: Shouldn't this be `\ProvidesPackage{pg}`?  It is my understanding that this should be used for a `.sty` file, `\ProvidesClass` for a `.cls` file, and `\ProvidesFile` for anything else.

The same is true of the `webwork2.sty` file now in the webwork2 code base.  I missed this when you added that file.

The `\ProvidesFile` command should work fine for these, but the `\ProvidesPackage` is probably more correct, and it does do some checking specifically for packages as I understand it.  It also has an optional package version argument.

drgrice1 (approved):
This looks good.  Just a question on the `\ProvidesFile` usage.

drgrice1:
By the way, [here](https://latexref.xyz/Class-and-package-commands.html#index-_005cProvidesClass) is more information on the `ProvidesPackage` and `ProvidesFile` commands.

assets/tex/pg.sty
Alex-Jordan: I'm not an expert with this, but I think because it is not a true package, and it's rather more like locally defined macros, that `ProvidesFile` was the recommendation. But I was just reading a stackexchange post, not authoritative.

drgrice1: The stack exchange post that I found says the opposite of the one that you see.

I don't believe that LaTeX distinguishes between local and system packages, or that this is not a true package.

pstaabp:
After doing some more testing, I included a problem with a `PGchoicemacros.pl` and it seems to be if the `match_questions_list` is used for this, then I get an error when generating the hardcopy.  Here's an example:
`Library/Indiana/Indiana_setSeries5IntegralTest/ur_sr_5_11.pg`.  (This problem has some error, but runs on the current PG-2.18 branch).

Here's one of the tex errors:
```
! Argument of \\answerRule has an extra }.
<inserted text> 
                \par 
l.40 \item[\answerRule[AnSwEr0002]
                                  {5}2.] \(\displaystyle \sum_{n=1}^\infty \...
I've run across a `}' that doesn't seem to match anything.
For example, `\def\a#1{...}' and `\a}' would produce
this error. If you simply proceed now, the `\par' that
I've just inserted will cause me to report a runaway
argument that might be the root of the problem. But if
your `}' was spurious, just type `2' and it will go away.
```

Alex-Jordan:
I made that change and pushed, but I still get a compilation error. I'm looking into it.

Alex-Jordan:
Eh, I just didn't have the corresponding webwork2 branch going. Working for me now.

pstaabp (approved):
Looks good now.  @drgrice1 do you want to look at this anymore or else we can merge.

drgrice1:
This is good.

PR #864, opened 2023-06-25
Fix invalid `L<...>` usage in POD.

drgrice1:
With this change the current pod parsing script will still work. Without this change the updated pod parsing script in https://github.com/openwebwork/webwork2/pull/2083 will not correctly resolve these links, because the incorrect parsing that is currently being done for these invalid links is removed.

The most prevalent invalid `L<...>` usage comes in the form of `L<macros/macroName.pl>`.  Following the correct usage rules this means a link to the header "macroName.pl" in the "macros" package.  Clearly that is not correct.

See https://perldoc.perl.org/perlpod#L-a-hyperlink for the `L<...>` usage rules.  The updated script will honor those rules with the additional rules for local files that fit into that scheme:

* For a link to a PG module, use the package name (exactly as the rules for `L<...>` dictate).
* For a link to a `.pod` file, just use the basename without the `.pod` extension.  So for example for a link to doc/MathObjects/UsingMathObjects.pod, use L<UsingMathObjects>.  (See below)
* For a link to a macro file (or script), use the basename of the macro file with the `.pl` extension and no path.  This is actually how it already is in most files, except those fixed here.

This also fixes the `L<MathObjects>` links.  Those don't currently work, and won't start to work with the updated script.  There is no `MathObjects` module or pod file.  (There is a MathObjects.pl file, but that doesn't have good POD.) So this changes those to go to the `doc/MathObjects/UsingMathObjects.pod` file.

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -13,40 +13,40 @@ MathObject Matrix methods: L<http://webwork.maa.org/wiki/Matrix_(MathObject_Clas
 MathObject Contexts: L<http://webwork.maa.org/wiki/Common_Contexts>
 CPAN RealMatrix docs: L<http://search.cpan.org/~leto/Math-MatrixReal-2.09/lib/Math/MatrixReal.pm>
 
-Allowing Matrices in Fractions: 
-L<http://webwork.maa.org/moodle/mod/forum/discuss.php?d=2978>	
+Allowing Matrices in Fractions:
+L<http://webwork.maa.org/moodle/mod/forum/discuss.php?d=2978>
 
      Context()->parens->set("[" => {formMatrix => 1});
 
 Files interacting with Matrices:
 
-L<lib/MatrixReal1.pm>
+L<MatrixReal1>
 
-L<macros/MatrixReduce.pl>
+L<MatrixReduce.pl>
 
-L<lib/Matrix.pm> 
+L<Matrix>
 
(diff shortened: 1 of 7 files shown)


pstaabp (approved):
Looks good.

lib/Value/Matrix.pm
```diff
@@ -13,40 +13,40 @@ MathObject Matrix methods: L<http://webwork.maa.org/wiki/Matrix_(MathObject_Clas
...
+L<PGmorematrixmacros.pl>
 
-L<macros/PGnumericalmacros.pl>
+L<PGnumericalmacros.pl>
 
-L<macros/tableau.pl>
+L<tableau.pl>
 
-L<macros/quickMatrixEntry.pl>
+quickMatrixEntry.pl
```
Alex-Jordan, line 43: Is this one supposed to have the `L<...>`?

drgrice1: That macro does not have POD.  So there is nothing to link to in the generated POD files.  So no.

Alex-Jordan (approved):

drgrice1:
Oops.  I was wrong in my statement that this will work with the existing script.  The macro links still work, but the module links don't.  Open the link `http://your.server.address/webwork2/pod/lib/Value/Matrix.pm`, and you will see that the `MatrixReal1` and `Matrix` links don't work, and that the macro links do.  Oh well, there are only those two and one more in tableau.pl, and they will be fixed once openwebwork/webwork2#2083 is merged.

PR #866, opened 2023-06-28
Remove some obsolete javascript files that are causing a security alert in Github.

drgrice1:
See https://github.com/openwebwork/pg/security/dependabot/3.

The related css and html files are also removed.  These are unmaintained, and I think only @mgage knows anything about them.


pstaabp (approved):
I'm guessing this was some kind of test for @mgage 's quick matrix entry.

dlglin (approved):

PR #868, opened 2023-06-30
Perltidy improvements

drgrice1:
The `bin/run-perltidy.pl` script is set up the same as the corresponding webwork2 script to autodetect the pg root directory and no longer depends on the PG_ROOT environment variable.

A new perltidyrc file (`bin/perltidy-pg.rc`) is added.  This is used for perl tidying pg code.  The only differences are that spaces are used for indentation, and a line limit of 80 characters is used instead of 120.

The `WeBWorK::PG::Tidy` module is added.  This module is a wrapper around the `Perl::Tidy` module that basically does everything the `bin/perltidy-pg.pl` script did before.  That script now just calls the exported `pgtidy` method.  This module will also be used by the webwork2 PG problem editor to tidy PG problem code.

### DIFF
--- a/bin/perltidy-pg.pl
+++ b/bin/perltidy-pg.pl
@@ -20,7 +20,7 @@ pg-perltidy.pl -- Run perltidy on pg problem files.
 
 =head1 SYNOPSIS
 
-  pg-perltidy.pl [options] file1 file2 ...
+    pg-perltidy.pl [options] file1 file2 ...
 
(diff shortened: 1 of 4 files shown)


Alex-Jordan (approved):
I just tested by messing up some perl and running the script. And it is working as far as that test went.

pstaabp (approved):
Looks good.

PR #869, opened 2023-06-30
Update GeoGebra location for the deployggb.js file.

drgrice1:
The file is no longer available at the location in the code.  So GeoGebra problems are not working.

### DIFF
--- a/macros/graph/AppletObjects.pl
+++ b/macros/graph/AppletObjects.pl
@@ -38,7 +38,7 @@ Usage:    C<$applet = GeogebraWebApplet(...);>
 =cut
 
 sub GeogebraWebApplet {
-	ADD_JS_FILE("https://www.geogebra.org/deployggb.js", 1);
+	ADD_JS_FILE("https://cdn.geogebra.org/apps/deployggb.js", 1);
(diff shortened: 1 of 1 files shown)


drgrice1:
You can test this with Contrib/CUNY/CityTech/Calculus/setExplore_-_Graphing_the_Derivative/randomized-derivative-graph-with-solution.pg or Library/FortLewis/Calc1/06-01-Antiderivatives-graphically/AF2/AF2.pg, for example.  Of course there are many others in Contrib and the OPL.

Alex-Jordan:
This raises the question for me: is there a good way to just directly enter `Library/FortLewis/Calc1/06-01-Antiderivatives-graphically/AF2/AF2.pg` somewhere and then I'm editing it? All I know of is workarounds where I put it in a set first then edit it, or navigate there with the Library Browser. Or I put it in a `render_rpc` URL that I have bookmarked.

drgrice1:
I don't know if an easier way at this point.

Alex-Jordan (approved):
This works. I saw the broken problem before, and the fixed problem after.

pstaabp (approved):

PR #870, opened 2023-07-01
refactor hardcopy themes

Alex-Jordan:
This pairs with openwebwork/webwork2#2099 that refactors hardcopy themes.

Part of that refactor is cleanly separating:

- tex packags and macros that supports PG directly
- tex packags and macros that `webwork2` uses for hardcopy production
- tex packags and macros that a particular theme needs

This PR moves everything of the first flavor over to the `pg` repo. Now in theory something other than `webwork2` can fully process individual problem files via tex to PDF, without needing something from the `webwork2` repo.

### DIFF
--- a/assets/tex/CAPA.tex
+++ b/assets/tex/CAPA.tex
@@ -1,0 +1,15 @@
+% capa tex macros
+
+\newcommand{\capa}{{\sl C\kern-.10em\raise-.00ex\hbox{\rm A}\kern-.22em%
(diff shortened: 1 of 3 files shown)


Alex-Jordan:
I was wondering if it makes sense to move the contents of `PGML.tex` and `CAPA.tex` into `pg.sty` to make sure that file is closer to self-contained.

drgrice1 (approved):
This looks good (paired with https://github.com/openwebwork/webwork2/pull/2099 of course).

PR #874, opened 2023-07-18
use the hidden #enter_key_submit button for enter key hits

Alex-Jordan:
This pairs with https://github.com/openwebwork/webwork2/pull/2119, as suggested by @drgrice1, so that the feature works as described when MathQuill is used.

### DIFF
--- a/htdocs/js/MathQuill/mqeditor.js
+++ b/htdocs/js/MathQuill/mqeditor.js
@@ -477,7 +477,9 @@
 				delete answerQuill.toolbar;
 
 				// For ww2 homework, depends on $pg{options}{enterKey}
-				document.getElementById('enter_key_submit')?.click();
+				const enterKeySubmit = document.getElementById('enter_key_submit');
+				if (enterKeySubmit) enterKeySubmit.click();
+				else document.getElementById('previewAnswers_id')?.click();
 				// For gateway quizzes, always the preview button
 				document.querySelector('input[name=previewAnswers]')?.click();
 				// For ww3
--- a/htdocs/js/MathQuill/mqeditor.js
+++ b/htdocs/js/MathQuill/mqeditor.js
@@ -468,7 +468,7 @@
 
 		if (latexEntryMode) return;
 
-		// Trigger an answer preview when the enter key is pressed in an answer box.
+		// Trigger a button press when the enter key is pressed in an answer box.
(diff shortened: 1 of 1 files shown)


drgrice1 (commented):
You should also update the comment on line 471.  It should probably say something like "Trigger the appropriate submit button when the enter key is pressed in an answer box."

Alex-Jordan:
I made it generic there and got specific in the comments below.

drgrice1 (commented):
This looks good.

Although, I have mixed feelings about relying on a front end setting for this.  For the standalone renderer this will be a bit odd.  @drdrew42, what do you think on this?

Alex-Jordan:
What is the javascript way to make it fall back to `'previewAnswers_id'` if there is no `'enter_key_submit'`?

htdocs/js/MathQuill/mqeditor.js
```diff
@@ -468,17 +468,17 @@
...
 				// Ensure that the toolbar and any open tooltips are removed.
 				answerQuill.toolbar?.tooltips.forEach((tooltip) => tooltip.dispose());
 				answerQuill.toolbar?.remove();
 				delete answerQuill.toolbar;
 
-				// For ww2 homework
-				document.getElementById('previewAnswers_id')?.click();
-				// For gateway quizzes
+				// For ww2 homework, depends on $pg{options}{enterKey}
+				document.getElementById('enter_key_submit')?.click();
```
drgrice1, line 480: Perhaps this should be changed to:
```suggestion
				const enterKeySubmit = document.getElementById('enter_key_submit');
				if (enterKeySubmit) enterKeySubmit.click();
				else document.getElementById('previewAnswers_id')?.click();
```
That way the standalone renderer can continue to function without any changes.  This will still work with the webwork2 changes in https://github.com/openwebwork/webwork2/pull/2119 (and without those changes).

drgrice1:
> What is the javascript way to make it fall back to `'previewAnswers_id'` if there is no `'enter_key_submit'`?

I didn't see this comment before.  Although I answered it.  With that change I don't have any reservations about this pull request anymore.  This will work for the new webwork2 option, and will work for the standalone renderer (or other front end) that doesn't provide this hidden submit button.

pstaabp (approved):

Issue #871, opened 2023-07-15
unecessary warning for RPC rendered problem using base64 encoding


Alex-Jordan:
The following are the same problem rendered via rpc, one using htm2xml, the other using render_rpc. In both cases, there is a perl warning from PGalias.pl. I suspect this was already there all along, but in 2.18, the warning is being exposed. There seems to be an assumption in PGalias.pl that there should be a sourceFilePath, but these examples use base64 encoded problems as the problemSource param. There is no sourceFilePath and that leads to this warning.

This is an issue for PTX because older books (like [here](https://spot.pcc.edu/math/orcca/ed2/html/section-variables-and-evaluating-expressions.html)) use an iframe embedding with the html2xml flavor as the src, so if they rely on a server that upgrades to 2.18, these warnings are now showing up. 

[...]

PR #875, opened 2023-07-19
don't check for pgFileName when problem has problemSource

Alex-Jordan:
This is a minimal approach to addressing #871. What @drgrice1 suggested there may be better, but this is targeting the situation that led to the report.

### DIFF
--- a/lib/PGalias.pm
+++ b/lib/PGalias.pm
@@ -112,7 +112,7 @@ sub check_parameters {
 
 	# Problem specific data
 	$self->warning_message('The path to the current problem file template probFileName is not defined.')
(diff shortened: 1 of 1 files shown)


drgrice1 (commented):
I don't think that this is a good way to go.  This covers one case where `probFileName` might be undefined in the environment, but there are others.  There are also `rawProblemSource` and `uriEncodedProblemSource`.  Furthermore, it doesn't actually deal with the issues that would be caused if a problem source attempted to use something that would call the PGalias.pm code that actually uses the value of `pgFileName`.

Alex-Jordan:
Closing in favor of #877.

PR #876, opened 2023-07-19
typo

Alex-Jordan:
Only related to https://github.com/openwebwork/webwork2/pull/2120 on the surface. This is just a plain old typo. But to get all of them to not show up on a grep search, I submit this as well.


drgrice1 (approved):

drgrice1:
I think this doesn't need a lot of review.  I am merging.

PR #878, opened 2023-07-20
Escape MathJax characters in PGML.

somiaj:
Wrap `\`, `` ` ``, and `$` in span tags to prevent MathJax from processing the characters using the `tex2jax_ignore` class.  For more information see https://github.com/openwebwork/webwork2/issues/1980.

### DIFF
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -1382,6 +1382,9 @@ sub Escape {
 	$string =~ s/</&lt;/g;
 	$string =~ s/>/&gt;/g;
 	$string =~ s/"/&quot;/g;
+
+	# Wrap the characters \, `, and $ in span tags to prevent MathJax from processing them.
+	$string =~ s/([\\`\$]+)/<span class="tex2jax_ignore">$1<\/span>/;
 	return $string;
 }
 


drgrice1:
Unfortunately this is to late for the 2.18 release.  We are releasing as we speak.  You will need to retarget this to develop, and either wait for the next release or lobby for a hot fix.

macros/core/PGML.pl
```diff
@@ -1382,6 +1382,9 @@ sub Escape {
 	$string =~ s/</&lt;/g;
 	$string =~ s/>/&gt;/g;
 	$string =~ s/"/&quot;/g;
+
+	# Wrap the characters \, `, and $ in span tags to prevent MathJax from processing them.
+	$string =~ s/([\\`\$]+)/<span class="tex2jax_ignore">$1<\/span>/;
```
dpvc, line 1387: You don't need the `class="tex2jax_ignore"`, here; just using a plain `span` will do.

somiaj: @dpvc Since I'm matching multiple characters at once, this could escape multiple backslashes, which I thought you mentioned would then be escaped by MathJax, `<span>\\\\\\</span>` for instance.  I also thought adding the class would help anyone who saw these spans in the html better understand their purpose.

dpvc: Sorry, didn't look carefully enough.  You are right.  I retract my comment.

Alex-Jordan:
Sorry Jaimos, I know I was reluctant about this earlier and it led to a
delay of this PR getting going. I still feel like something about this I
can't put my finger on could have an unwanted side effect, but I'm OK with
just seeing if that ever happens.

Alex-Jordan (approved):
We should consider merging this. If it has any unforeseen consequences, the earlier we merge it, the more likely we will catch them before the next release.

I am working on an update to the PGML lab, and I reached this example about verbatim:

```
[|This is not math [`x+1`] in here.|]
```

But it does not comes out verbatim, of course. It comes out like:

`This is not math [x+1] in here.`

with the x+1 rendered by MathJax. So it's the same issue. I tested, and with this change I see

``This is not math [`x+1`] in here.``

with no rendered math, as expected.

drgrice1 (approved):

drgrice1:
I will go ahead and merge this as suggested by @Alex-Jordan.

PR #883, opened 2023-07-24
Add `nonstopmode` when LaTeXImage.pm runs latex or pdflatex.

drgrice1:
Currently if a LaTeX error is encountered when latex or pdflatex is run, user interaction is required for the execution to continue.  Since this is run by the server no such interaction can occur.  Thus the request waits until the timeout, and then dies.  This causes several problems.

See issues [#2041](https://github.com/openwebwork/webwork2/issues/2041) and [#2148](https://github.com/openwebwork/webwork2/issues/2148) for details.  This may not entirely fix those issues, but it is part of what is happening.

This fix is certainly needed independent of if it does fix those issues. Create a problem that uses LaTeXImage.pm and make the TeX have an error. For example, in TikZ don't add a semicolon at the end of a line that should have it.  Without this pull request the problem will hang until the timeout (typically 60 seconds), and with this pull request it will return immediately and show the TeX error that occurred.

This also adds some clean up to the `insertGraph` method of PGcore.pm. Basically, instead of just warning when errors occur and then continuing on to code that is guaranteed to fail, the warnings are given and the later code is skipped.  In adition if the `draw` method does not return any content, the temporary image file is not written.  Previously an empty file  was written.

### DIFF
--- a/lib/LaTeXImage.pm
+++ b/lib/LaTeXImage.pm
@@ -203,7 +203,7 @@ sub draw {
 		close $fh;
 		system "cd $working_dir && "
 			. WeBWorK::PG::IO::externalCommand('latex')
-			. " image-dvisvgm.tex > latex.stdout 2> /dev/null && "
+			. " --interaction=nonstopmode image-dvisvgm.tex > latex.stdout 2> /dev/null && "
 			. WeBWorK::PG::IO::externalCommand('mv')
 			. " image-dvisvgm.dvi image.dvi";
 		chmod(0777, "$working_dir/image.dvi");
@@ -218,7 +218,7 @@ sub draw {
 		close $fh;
 		system "cd $working_dir && "
 			. WeBWorK::PG::IO::externalCommand('pdflatex')
-			. " image.tex > pdflatex.stdout 2> /dev/null";
+			. " --interaction=nonstopmode image.tex > pdflatex.stdout 2> /dev/null";
 		chmod(0777, "$working_dir/image.pdf");
 	}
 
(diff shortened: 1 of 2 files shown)


drgrice1:
This should be considered for a hotfix as well.

Alex-Jordan:
I wonder if this might fix another issue I have been meaning to investigate. I have `xelatex` in use instead of `pdflatex`. Every now and then I check on the server and find there are several `xelatex` processes that have been running for hours, or even days. I kill them and carry on. But I haven't been able to trigger these. Running a problem with broken tikz code does not trigger on of these for instance. At least not reliably. So I kicked the can for looking into it. Soon I will st up nightly reboots and that may sweep that issue under the rug.

Alex-Jordan:
Is there a reason to prefer `nonstopmode` over `batchmode` for this? I think in other places, `nonstopmode` is appropriate because I think you are capturing terminal output for a log. But is that the case here? If not, is `batchmode` appropriate?

Alex-Jordan:
I'm sure this is a good change, but it did not resolve my issue. I applied this, restarted webwork2, unassigned and reassigned the test to my fake student, cleared all image files, and took the test. Again when I graded, the same thing happened as with #2148. Again it reached number 25 with scoring, and then all 0s.

drgrice1:
In this case the output is captured, and so `nonstopmode` is needed.  That output is "warned" if there are errors which results in that output being shown in the page response.

You closed #2148, but there still is some issue there that I want to look into.  Even if there is some error in pg, it shouldn't hang, and fail to continue on to other problems.  I will look at the stuff you posted.

Alex-Jordan:
I'm in favor of making this a hotfix too. The scenario where some random seed invalidates the latex could be common enough.

pstaabp (approved):
Works great.  I love having the errors from the latex..

PR #887, opened 2023-07-26
parse derivatives instead of just relying on tree

drdrew42:
Taking derivatives using the `D` method ends up creating formulas that refuse to reduce (e.g. constant products appearing in the derivative of a polynomial).

This PR modifies the behavior of the `D` method -- preserving the existing tree structure of the derivative, but now processing the result of that process (respecting the existing context flags for free).

### DIFF
--- a/lib/Parser/Differentiation.pm
+++ b/lib/Parser/Differentiation.pm
@@ -39,7 +39,11 @@ sub Parser::D {
 		return (0 * $self)->reduce('0*x' => 1) unless defined $self->{variables}{$x};
 		$f = $f->D($x);
 	}
-	return $self->new($f);
+	my $deriv = $self->new($f);
+	$deriv->{string} = $deriv->{tree}->string;
+	$deriv->tokenize;
+	$deriv->parse;
+	return $deriv;
 }
 
 #


lib/Parser/Differentiation.pm
```diff
@@ -39,7 +39,11 @@ sub Parser::D {
 		return (0 * $self)->reduce('0*x' => 1) unless defined $self->{variables}{$x};
 		$f = $f->D($x);
 	}
-	return $self->new($f);
+	my $deriv = $self->new($f);
+	$deriv->{string} = $deriv->{tree}->string;
```
dpvc, line 43: This line means that any constants are clipped to 6 significant digits (the default number format), so the result may have different from the original.

You may also want to add a new reduction rule that controls doing this doing this, so it can be turned off, if desired.

Alex-Jordan:
I think we can close this. The specific instance cited (e.g. constant products appearing in the derivative of a polynomial) has already been addressed by #1025. Is there an example of something different that this would still help with?

In case this should be kept, replying to @dpvc's comment:  could the `string` method at the new line 43 take an argument that briefly changes the number format to use more precision? Say 17 significant digits?

dpvc:
The `string` methods already take optional parameters, so you can't use that approach to adjust the precision.  But there is a context flag for the number format, and an expression can override that using its `format` property, so you could use

``` perl
	my $deriv = $self->new($f);
	$deriv->{format} = '%.17g';
	$deriv->{string} = $deriv->{tree}->string;
	delete $deriv->{format};
```

in order to get the higher precision.  I doubt it is necessary, but you could save any old `format` value and replace it, if there is one.

PR #890, opened 2023-07-27
fix _macro_init definitions (hotfix)

Alex-Jordan:
These were not quite right in the move from the OPL.

### DIFF
--- a/macros/math/algebraMacros.pl
+++ b/macros/math/algebraMacros.pl
@@ -3,7 +3,7 @@
 # define any custom subroutines you want and then use them in
 # problems by including the file in loadMacros() calls.
 
-sub _init_macro_template {
+sub _algebraMacros_init {
 
 	#Possibly add initialization code here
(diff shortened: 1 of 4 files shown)


drgrice1 (approved):
Good find on this.  I should have caught this earlier.

drgrice1 (approved):

macros/math/algebraMacros.pl
```diff
@@ -3,7 +3,7 @@
 # define any custom subroutines you want and then use them in
 # problems by including the file in loadMacros() calls.
 
-sub _init_macro_template {
```
pstaabp, line 6: Doesn't this need to have the name of the macro? So `_algebra_macros_init` ?

drgrice1:
Actually, it needs to be `_algebraMacros_init`.

Alex-Jordan:
Changed on both PRs and force pushed.

PR #893, opened 2023-07-27
Fix line numbers in warnings from problems and macros.

drgrice1:
If you have edited problem files and received warnings, you are probably familiar with the fact that the line numbers are off.  The reason for this is new lines added to the evaluated strings.  This just removes those new lines.  See the comment on line 23 of WWSafe.pm for the motivation for this fix.


drdrew42 (approved):
Awesome!

Issue #894, opened 2023-07-28
q within [@...@] PGML blocks


Alex-Jordan:
PR https://github.com/openwebwork/pg/pull/760 brought support for `q` within `[@...@]` PGML blocks, and I think it was working on a 2.17 installation with those commits applied.  Now on 2.18, it is not working in at least one case. Here is a MWE:

```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl));
BEGIN_PGML
[@ q(foo) @]
END_PGML
ENDDOCUMENT();
```

This gives me a warning/error message about a substring index being outside of the string length. If I change the `q(foo)` to `q!foo!` then it goes away, so it seems to have to do with the paren matching. (And there is at least a workaround.)

drgrice1:
I have noticed this as well. I never tested on 2.17, but saw this at the beginning on 2.18.  I believe the same happens with `q{ }`.

Alex-Jordan:
I can't say for sure at this point, but I believe I took the PGML.pl from #760 and put it in our macros folder back around when #760 was merged. We have problems that use `[@ q(...) @]` and with no one reporting anything, I assume they were working in winter and spring without warnings. But now in summer on 2.18, they are making these warnings.

drgrice1:
I just tested with 2.17 using the PGML.pl file from #760, and I get the same error messages that I get with 2.18.

PR #895, opened 2023-07-28
Fix q( ), q{ }, and q[ ] quotes inside PGML [@ @] blocks.

drgrice1:
What was happening is that the
`my $quoted = '[$@%]q[qr]?|\bq[qr]?\s+(?=.)|\bq[qr]?(?=\W)';` pattern combined with the later
`my $chars = '\\\\.|[{}[\]()\'"]';`
pattern in the PGML::Parse $splitPattern combined such that an empty string was injected between the `q` and the beginning `(`, `{`, or `[` delimiter.  For example the PGML block `[@ q(string) @]` was split into `['[@', ' ', 'q', '', '(', 'string', ')', ' ', '@]']`.  The `Quoted` method expected the delimiter to be in the next position after `q`.  So this skips that empty string and moves on to the next position in the split in that case.

The next part I do not fully understand, and hopefully @dpvc can shed some light on this.  When this is parsed until the matching end delimiter is reached it ends by gobbling the end parenthesis (the ending ')' in the list above), and leaves the empty string.  If the parser is left with that to continue then a later PGML::Eval is called, and this results in the error `Use of uninitialized value $code in concatenation (.) or string at [PG]/lib/WeBWorK/PG/Translator.pm line 1246` when `PG_restricted_eval` is called.  So this also skips to the next split in this case.

Furthermore, if the PGML block is `[@ q(string q) @]`, then this is split into `['[@', ' ', 'q', '', '(', 'string ', 'q) ', '', '@]'`.  Then at the end of parsing the `q` expression and gobbling the ending parenthesis the split that is left is the space at the end of `'q) '`.  This results in the same PGML::Eval error later.  So this also skips to the next split in this case.  Both this case and the previous are handled on line 546 of PGML.pl in this pull request.

This fixes issue #894.

### DIFF
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -522,7 +522,7 @@ sub Preformatted {
 sub Quoted {
 	my $self  = shift;
 	my $token = shift;
-	my $next  = $self->{split}[ $self->{i} ];
+	my $next  = $self->{split}[ $self->{i} ] || $self->{split}[ ++$self->{i} ];
 	my $quote = substr($next, 0, 1);
 	$self->{split}[ $self->{i} ] = substr($next, 1);
 	my $pcount = 0;
@@ -542,7 +542,8 @@ sub Quoted {
 			my $i = index($text, $close);
 			if ($i > -1) {
 				$self->Text(substr($text, 0, $i + 1));
-				$text = $self->{split}[ $self->{i} ] = substr($text, $i + 1);
+				$self->{split}[ $self->{i} ] = substr($text, $i + 1);
+				++$self->{i} if $open && $self->{split}[ $self->{i} ] =~ /\s*/;
 				return;
 			}
 		}
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -538,12 +538,12 @@ sub Quoted {
 			$pcount++;
 		} elsif ($open && $text eq $close && $pcount > 0) {
 			$pcount--;
-		} elsif (!$open || ($text ne $qclose && $pcount == 0)) {
+		} elsif ($pcount == 0 && $text ne $qclose) {
 			my $i = index($text, $close);
 			if ($i > -1) {
 				$self->Text(substr($text, 0, $i + 1));
 				$self->{split}[ $self->{i} ] = substr($text, $i + 1);
-				++$self->{i} if $open && $self->{split}[ $self->{i} ] =~ /\s*/;
+				$self->Text($self->{split}[ $self->{i}++ ]) if $self->{i} % 2;
 				return;
 			}
 		}


drgrice1:
An example problem for testing this is attached.
[quotes.pg.txt](https://github.com/openwebwork/pg/files/12205331/quotes.pg.txt)

dpvc:
Well, it looks like I didn't do a very good job with handling the case where you have different open and close delimiters (like `(` and `)` rather than a single character for both).  As I recall, that was the last thing I added, and I totally blew it!

You're right about getting the open ing delimiter when it is a parenthesis:  that has to come from the element two down from the opening `q`, not the next one.  So your correction to line 525 is good.  But the final `elsif` clause is messed up (my fault) and your solution doesn't really fix it.  Here is a corrected one:

``` perl
		} elsif ($pcount == 0 && $text ne $qclose) {
			my $i = index($text, $close);
			if ($i > -1) {
				$self->Text(substr($text, 0, $i + 1));
				$self->{split}[ $self->{i} ] = substr($text, $i + 1);
				if ($self->{i} % 2) {
					$self->Text($self->{split}[ $self->{i}++ ]);
				}
				return;
			}
		}
```

The test for the `elsif` was incorrect, so wasn't always being run when it should, and the issue that you were getting with the blank entries really wan't because they were blank, but because the loop was ending on the wrong index.  This is because the `$self->{split}` is made up of the things matching any of the split patterns separated by the text that falls between them, the parity of `$self->{i}` tells whether the entry is plain text or a token we need to process.  The main loop in the `Parse()` method assumes the parity of `$self->{i}` is properly maintained in the functions it calls, so that `$self->{i}` is even at the top of the loop.  It adds the i-th entry as text (even ones are text) and then processes the next as a token that has some meaning in PGML.

The problem is that the processing in `Quoted()` can end either on a text entry of `$self->{split}` (when the close character is fond within the text string) or on a token entry (when it is `)` for example, which is one of the token patterns).  So we need to check the parity and adjust if we are in the wrong position at the end of the loop.  The code above does that with the `% 2` check, so gets the main processing loop back in synch.

Your example with `q(string q)` involves an extra complication, because the close parenthesis is actually part of a different token pattern:  the pattern for lists:

``` perl
my $list      = '(?:^|(?<=[\t ]))(?:[-+o*]|(?:\d+|[ivxl]+|[IVXL]+|[a-zA-Z])[.)]) +';
```

which matches `q) ` as a potential list item rather than producing `q`, `)`, ` ` as `Quoted()` had expected (the addition of `()` to the `$chars` pattern was supposed to guarantee that the braces and parentheses would be individual tokens, but I hand't thought about the list pattern interfering).  Your proposed fix did handle this situation by incrementing the index, but I think it is better to target the real problem, which is the parity of the index, as whether it has spaces or not is not the issue.  And I think you really wanted `/^\s+$/` as the pattern in any case.

So I'd recommend the change above instead.

drgrice1:
Thanks @dpvc.  Your explanation helps a lot.  I was noticing that the `[@` block was not being properly terminated, and you cleared that up as to why that was happening.  I adjusted the code on line 546 as you suggested.

drgrice1:
I also adjusted the final `elsif` conditional as you suggested.

pstaabp (approved):
Looks like it fixes the issue.

PR #897, opened 2023-08-03
PG Text post processing

drgrice1:
Macros or problems can add a hook by calling `add_content_post_processor` with a subroutine that will be called after the translator has processed answers.  For all display modes except TeX, the subroutine will be passed the problem text parsed into a Mojo::DOM object and a reference to the header text.  For the TeX display mode a reference to the problem text is passed.  The subroutine can then modify the problem DOM or text, and header text.  The resulting Mojo::DOM object will be converted back to a string and the problem text reference returned by the translator set to point to that string.
    
The hidden MathQuill inputs for the latex string is added in this post processing stage (see the ENDDOCUMENT method in PG.pl).
    
The scaffold.pl macro is updated to use this post processing.  This means that answers are no longer evaluated by the macro.  Instead the scaffold sections are finalized after the answers are processed normally by the translator.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -29,6 +29,7 @@ jobs:
             libjson-perl \
             libjson-xs-perl \
             liblocale-maketext-lexicon-perl \
+            libmojolicious-perl \
             libtest2-suite-perl \
             libtie-ixhash-perl \
             libuuid-tiny-perl \
--- a/cpanfile
+++ b/cpanfile
@@ -18,6 +18,7 @@ on runtime => sub {
 	requires 'JSON::XS';
 	requires 'Locale::Maketext';
 	requires 'Locale::Maketext::Lexicon';
+	requires 'Mojolicious';
 	requires 'Tie::IxHash';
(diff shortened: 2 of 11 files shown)


drgrice1:
That doesn't exactly make sense.  The submit buttons are not added by PG at all.  They are added by webwork2 after PG finishes rendering the problem.

drgrice1:
@pstaabp:  I am guessing that you have something else going on unrelated to this pull request.  Perhaps you need to run `generateAssets.js` in the webwork2 htdocs location.  I suspect that the `visually-hidden-unfocusable` class added in openwebwork/webwork2#2119 is not in your math4 css and the new hidden submit button is showing up.

pstaabp:
You're right.  It was doing this in develop and it was a problem with my custom theme--I had another issue with the fact that some of the files were copies and not links to the math4/ versions.  All fixed now.

Alex-Jordan:
Probably all is well now, but there is an extra button like the preview button here that is usually hidden. It is the button that is "clicked" if you hit enter from an input field. You can configure so that button might be check/submit instead. I'm guessing this is the button you saw.

PR #898, opened 2023-08-03
Remove the pass through createDirectory method in PGcore.pm.

drgrice1:
This method is not used anywhere in PG except directly in that file, and PGcore.pm can direcly use the WeBWorK::PG::IO::createDirectory method.  This method existing means that any problem can call `$PG->createDirectory('/path/to/dir', 0770, -1)` to create a directory anywhere that the server has write permissions for.
    
Also the WeBWorK::PG::IO module should not export methods by default (this is done by setting `our @EXPORT`).  Instead it should only export methods on request using `our @EXPORT_OK`.  Furthermore, since that array doubles as the list of methods that are shared in the safe compartment, care should be taken as to what is included in that list.  In particular the `createDirectory` method should not be. Otherwise problems can directly call `createDirectory('/path/to/dir', 0770, -1)`.
    
Any other `lib` module that wants to use the WeBWorK::PG::IO methods can still do so either by importing the desired exported method by name (if exported), or by calling the method with its full namespace (as is done everywhere currently).  Note that it is not necessary to `use WeBWorK::PG::IO` to use the methods therein when calling a method with its full namespace.  It is necessary that at some point the module is "required" for this to work.  This technically can be done in any file, but if a file uses a module, then for clarity and to be sure the module is loaded that file should call `require module`.
    
Also correct the list of modules in `pg_config.dist.yml` and the comments on how that works.  This is all about the difference between "use Module", "require Module", and "Module->import" (note `import` is a method and should not be called as `import Module`).  Calling `use Module LIST` is exactly equivalent to `BEGIN { require Module; Module->import(LIST); }`. Note that to call `require` on a module it must be contained in a file by the same name with the `.pm` extension. If a module is listed first in the modules list, then essentially that is equivalent to calling `use Module` (which both requires and imports the module and the module must be contained in its own file).  If a module is listed second, then that is roughly equivalent to calling `Module->import`.  For PG and translation, this means that if a module has its own file (i.e., Module is defined in Module.pm), then it should be first in the list, and if Module2 is defined in Module.pm, then Module2 should be listed in the same list after Module.
    
Note that these changes will also need to made for webwork2 in defaults.config.

### DIFF
--- a/conf/pg_config.dist.yml
+++ b/conf/pg_config.dist.yml
@@ -52,7 +52,6 @@ directories:
     - $pg_root/macros/ui
     - $pg_root/macros/deprecated
 
-
 URLs:
   # The public URL of the html directory above.
   html: /pg_files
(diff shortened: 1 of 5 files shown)


pstaabp (approved):
This looks good. 

Since this closes a possible security hole, should we make a hot fix as well?

PR #899, opened 2023-08-06
input field correctness indicators

Alex-Jordan:
This adds indicators that a text or MathQuill input field is (in)correct, beyond the current colored glow. It also colors the background. This is what PTX does for embedded WW exercises.

Screen Shot 2023-08-06 at 3 05 54 AM

Screen Shot 2023-08-06 at 3 05 17 AM

This could bee a way to address the second item at https://github.com/openwebwork/webwork2/issues/2157.

Similar things can be done with select, radio, and checkbox inputs. I'll see what people think about this first.

### DIFF
--- a/htdocs/js/InputColor/color.js
+++ b/htdocs/js/InputColor/color.js
@@ -17,7 +17,10 @@
 					if (!radioGroups[input.name]) radioGroups[input.name] = [];
 					radioGroups[input.name].push(input);
 				} else {
-					input.classList.add(type);
+					if (input.type.toLowerCase() === 'text') {
+						if (type === 'correct' || input.value !== '') input.classList.add(type);
+					} else
+						input.classList.add(type);
 				}
 			});
 
--- a/htdocs/js/MathQuill/mqeditor.js
+++ b/htdocs/js/MathQuill/mqeditor.js
@@ -501,7 +501,9 @@
 			if (answerLabel.includes(tableLink.dataset.answerId)) {
 				if (tableLink.parentNode.classList.contains('ResultsWithoutError'))
 					answerQuill.classList.add('correct');
-				else answerQuill.classList.add('incorrect');
+				else {
+					if (answerQuill.input.value !== '') answerQuill.classList.add('incorrect');
+				}
 			}
 
 			// Make a click on the results table link give focus to the mathquill answer box.
--- a/htdocs/js/Problem/problem.scss
+++ b/htdocs/js/Problem/problem.scss
@@ -88,7 +88,7 @@
 
 		&.incorrect {
 			padding-right: 29px;
-			background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 12 12' width='12' height='12' fill='none' stroke='%23dc3545'%3e%3ccircle cx='6' cy='6' r='4.5'/%3e%3cpath stroke-linejoin='round' d='M5.8 3.6h.4L6 6.5z'/%3e%3ccircle cx='6' cy='8.2' r='.6' fill='%23dc3545' stroke='none'/%3e%3c/svg%3e");
+			background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 12 12' width='12' height='12' fill='none' stroke='%23AE5757'%3e%3ccircle cx='6' cy='6' r='4.5'/%3e%3cpath stroke-linejoin='round' d='M5.8 3.6h.4L6 6.5z'/%3e%3ccircle cx='6' cy='8.2' r='.6' fill='%23AE5757' stroke='none'/%3e%3c/svg%3e");
 		}
 	}
 
--- a/htdocs/js/Problem/problem.scss
+++ b/htdocs/js/Problem/problem.scss
@@ -77,18 +77,18 @@
 	}
 
 	input[type=text], span.mq-editable-field {
-		background-size: auto 80%;
+		background-size: 20px auto;
 		background-position: right;
 		background-repeat: no-repeat;
 
 		&.correct {
-			padding-right: 30px;
-			background-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' version='1.1' height='20px' width='30px'><text x='24' y='16' fill='%23060' text-anchor='end'>✓</text></svg>");
+			padding-right: 27px;
+			background-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' version='1.1' height='20px' width='27px'><text x='18' y='16' fill='%23060' text-anchor='end'>✓</text></svg>");
 		}
 
 		&.incorrect {
-			padding-right: 30px;
-			background-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' version='1.1' height='22px' width='30px'><text x='28' y='18' fill='%23943D3D' text-anchor='end'>⚠</text></svg>");
+			padding-right: 27px;
+			background-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' version='1.1' height='22px' width='27px'><text x='24' y='18' fill='%23AE5757' text-anchor='end'>⚠</text></svg>");
 		}
 	}
 
--- a/htdocs/js/Problem/problem.scss
+++ b/htdocs/js/Problem/problem.scss
@@ -76,6 +76,24 @@
 		}
 	}
 
+	input[type=text], span.mq-editable-field {
+		background-size: auto 80%;
+		background-position: right;
+		background-repeat: no-repeat;
+
+		&.correct {
+			padding-right: 30px;
+			background-color: #8F8; /* green */
+			background-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' version='1.1' height='20px' width='30px'><text x='24' y='16' fill='%23060' text-anchor='end'>✓</text></svg>");
+		}
+
+		&.incorrect {
+			padding-right: 30px;
+			background-color: #DAA; /* red */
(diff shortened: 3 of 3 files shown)


drgrice1:
I don't like setting the background color.  I find that intrusive.  I think that having the previous edge glow together with the icon is sufficient to address your accessibility concerns.

This needs some more work for MathQuill.  When the input becomes taller to accommodate things like fractions and such, the icon starts to not work so well.  Here is a screen shot.
mq-fraction

Alex-Jordan:
I removed the background color, made the height of the indicators fixed, and made a few other small tweaks.

drgrice1:
That looks better.  There is one minor tweak that could be made.  The correct indicator looks well positioned, but the incorrect indicator seems like it is to close to the right edge.  This is for both with MathQuill or without.

Just for play I tried Bootstrap's form validation styles for this.  To see that replace the style you added with
```
		background-repeat: no-repeat;
		background-position: right calc(.375rem + .1875rem) center;
		background-size: calc(.75em + .375rem) calc(.75em + .375rem);

		&.correct {
			padding-right: calc(1.5em + .75rem);
			background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 8 8'%3e%3cpath fill='%23198754' d='M2.3 6.73.6 4.53c-.4-1.04.46-1.4 1.1-.8l1.1 1.4 3.4-3.8c.6-.63 1.6-.27 1.2.7l-4 4.6c-.43.5-.8.4-1.1.1z'/%3e%3c/svg%3e");
		}

		&.incorrect {
			padding-right: calc(1.5em + .75rem);
			background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 12 12' width='12' height='12' fill='none' stroke='%23dc3545'%3e%3ccircle cx='6' cy='6' r='4.5'/%3e%3cpath stroke-linejoin='round' d='M5.8 3.6h.4L6 6.5z'/%3e%3ccircle cx='6' cy='8.2' r='.6' fill='%23dc3545' stroke='none'/%3e%3c/svg%3e");
		}
```
That is essentially copied from their form validation rules for inputs that are valid or not.  Their `calc`'s could probably be simplified.  I am sure those are generated from variables in sass.

drgrice1:
The bootstrap sizing doesn't quite work right for our inputs though.  They add more padding to inputs, and their sizing works with that.

drgrice1:
Going a little further using the Bootstrap icons and fiddling with the size and position I found the following works pretty well.  One advantage of the svg not using unicode characters is a more uniform sizing between the two icons.
```
		background-size: 20px auto;
		background-position: right 5px center;
		background-repeat: no-repeat;

		&.correct {
			padding-right: 30px;
			background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 8 8'%3e%3cpath fill='%23198754' d='M2.3 6.73.6 4.53c-.4-1.04.46-1.4 1.1-.8l1.1 1.4 3.4-3.8c.6-.63 1.6-.27 1.2.7l-4 4.6c-.43.5-.8.4-1.1.1z'/%3e%3c/svg%3e");
		}

		&.incorrect {
			padding-right: 30px;
			background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 12 12' width='12' height='12' fill='none' stroke='%23dc3545'%3e%3ccircle cx='6' cy='6' r='4.5'/%3e%3cpath stroke-linejoin='round' d='M5.8 3.6h.4L6 6.5z'/%3e%3ccircle cx='6' cy='8.2' r='.6' fill='%23dc3545' stroke='none'/%3e%3c/svg%3e");
		}
```

Alex-Jordan:
The unicode characters might also cause problems with not being available in whatever font. I think I see that more often on a mobile device.

Is it easiest for you to commit these changes? For what it's worth I approve of all that you posted here.

drgrice1:
I pushed the change.  I tweaked the padding an position by one pixel from what I had posted.

drgrice1:
If you want to you could switch the colors in the svg's to our usual red and green.  I noticed that the Bootstrap colors are slightly different.  They are pretty close though.

Alex-Jordan:
I could not find that wee have a green that wee use for text (only background, glow, etc). So I left the bootstrap green alone.

I changed the red to the red I had for the unicode character, which is a 10% lighter version (or was it 20% lighter?) of a red we use.

I think this is ready. The only thing is if we want similar indicators for other input field types. These can be done, but the mechanism is different enough that I'm not sure if it should be part of this PR.

drgrice1:
Yeah, I really meant that you could switch to the colors you had before my commit. I just assumed those were the colors used for the border glow.

drgrice1 (approved):

somiaj (approved):
Tested it out and looks good.

Some thoughts (though maybe for future changes).  I personally found the incorrect mark on a blank answer a bit too much, and wonder if blank answer parts be left alone (to me a blank answer is not the same as an incorrect answer). I was also thinking what about indications for different types of incorrect or partial credit.  Maybe a different color/symbol for a problem part which has a syntax error and isn't parsed correctly, and also maybe another symbol for a question that is partially correct.

Alex-Jordan:
The PTX counterpart to this recognizes blank answers and given them a friendly light blue background. And recognizes partial credit and gives yellow background (but I can't recall right now what symbol it gives, maybe the same warning as for incorrect). But right now I think webwork2 does not supply classes for these situations, so that would have to be added.

pstaabp:
This looks nice, but would agree that it would be nice to exclude empty answers from the incorrect marks.  

Last week we talked about #897, which includes some post processing and how we might get rid of the attempts table.  So this might be the start of some new UI.

pstaabp:
I was just looking at a RadioMultiAnswer problem and here's what this PR does with this:
image

Although the checkmarks on the inputs which are not radio selected are (I guess) technically correct, I would argue, it would be better if they are not checked.

drgrice1:
It wouldn't be to hard only add the incorrect class to text inputs (or MathQuill inputs) that have a value set.  The coloring matches what is shown in the attempts table which is really the same as it was before.  The icons just make this more noticeable than before.

It would take a bit more to deal with the partial credit case.  Currently that information is not passed to the javascript.

I don't really think the case of a problem part which has a syntax error and isn't parsed correctly should be dealt with separately though.  It is still incorrect, and I am not sure that you can truly distinguish in a reliable way between that and incorrect for other reasons.

This will all be much easier when this isn't handled by javascript anymore.  Pull request #897 makes that possible, as well as removing the attempts table and implementing a better way of giving feedback.

drgrice1:
@Alex-Jordan: I put in a pull request to this branch that skips empty text or MathQuill inputs for the incorrect class.  Empty correct answers will still be marked correct.  So it won't fix the parserRadioMultiAnswer.pl issue that @pstaabp noted.  That actually will be challenging, and may not even be possible to do without giving something away in the DOM about what the correct answer is.  I haven't looked closely though or thought about it too much.

drgrice1:
I am not entirely certain that not marking the empty incorrect answers is the right thing to do.  In fact I was just looking at a parserRadioMultiAnswer.pl problem where part A has a sub answer and part A is incorrect.  If you select A, but don't enter the sub answer, then the radio gets the incorrect glow, but the empty text input is unmarked.  In that case it seems the blank input should still be marked as incorrect.  But in general, I think that empty answers should actually be marked as incorrect if they are incorrect.  If there are several answers that are all visible, and the student only answers one part, the empty answers are still incorrect, and should be marked as such.

The case of scaffolding where a new scaffold has just opened is the one exception to this.  The answers in the newly opened scaffold should not be marked correct.  The attempts table has this same issue.

So in short I disagree with @somiaj that blank answers should be left alone.  Although, I suspect the scaffolding may be where he brings this up from, and there I agree.  Scaffolds need work.

Alex-Jordan:
Here is another issue with this. For an array input, does it mislead a student to have all the entries marked incorrect when really only on entry is incorrect? 

Screen Shot 2023-08-07 at 9 53 09 PM

Alex-Jordan:
And there is also this kind of thing for a multianswer with `singleResult => 1`. As with my previous post, it's not that this is wrong behavior, but I wonder if it is misleading to a student. Also as with the previous post, I don't have any ideas to do something better.

Screen Shot 2023-08-07 at 10 00 37 PM

Alex-Jordan:
Here is a another thing about @pstaabp's example. After getting the unwanted green checkmarks, you can change your selection and it is misleading.

Screen Shot 2023-08-09 at 9 58 20 PM

drgrice1:
Yeah, that is a bit confusing.  Should the javascript be adapted to just not add indicators for any empty answer?  This would only be an intermediate solution.  Potentially an empty input could be correct, and as I said before I think that incorrect empty answers should be marked incorrect (unless in a newly opened scaffold).

I think the things that are being discussed here need to be discussed at detail in our meeting.  The attempts table needs to go, and it will take some work to properly replace it.

drgrice1:
On that issue with the parserRadioMultiAnswer.pl macro and switching answers after checking/submitting answers, I think that also applies much more generally.  Even if you enter a single answer, and checking marks it incorrect.  If you then type the correct answer in that input, of course it still shows incorrect.  I think that in general, javascript should remove the correct/incorrect class as soon as a student starts modifying an answer since the correct/incorrect status is no longer valid.

Also, the javascript detects incorrect/correct status by looking at the class added to the "Result" column in the attempts table.  If an answer is completely correct (meaning the score is greater than or equal to 1), then the "ResultsWithoutError" class is added.  If an answer is completely incorrect (meaning the score is equal to 0), then the "ResultsWithError" class is added.  For any score strictly between 0 and 1, no class is added.  Currently the javascript only checks to see if the "ResultsWithoutError" class is there, and if so marks the answer correct and otherwise marks it incorrect.  So to mark partial credit, the javascript only needs to also check if the "ResultsWithError" class is there.  In that case of course it would be marked incorrect, but if there is no class then it can be marked as partial credit.  We just needs to create a css class and styles for that.

PR #900, opened 2023-08-09
Clean up the PG.pl POD.

drgrice1:
All POD in the file is moved up to where each method is defined.  This ensures that the order of the generated html is consistent with the order in the file, and makes it easier to maintain the POD.  Methods are now `head2`'s instead of items so that they are in the index.

The POD in this file was not cleaned up in #836, so this finishes that task.

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -1,7 +1,88 @@
-#use AnswerEvaluator;
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2023 The WeBWorK Project, https://github.com/openwebwork
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):

macros/PG.pl
```diff
@@ -185,18 +360,19 @@ ()
...
 
-If external is 1, it is assumed the full url is provided.  If external is 0 or
-not given, then file name will be prefixed with the webwork2/htdocs/ directory.
+If external is 1, it is assumed the full URL is provided. If external is 0 or
+not given, then file name will be served from the C<pg/htdocs> directory (if
+found).
 
 Additional attributes can be passed as a hash reference in the optional third
-argument.  These attributes will be added as attributes to the script tag.
+argument. These attributes will be added as attributes to the script tag.
```
pstaabp, line 375: This existed before, but I think 5 lines down should have a `http:` or `https:` in front of the URL.

pstaabp:
I also see an error in the POD generation: 

```
Around line 818:
Expected '=item *'
```

macros/PG.pl
drgrice1: I changed this to the link currently used by AppletObjects.pj which does include `https:`.  Thanks for pointing this out.

drgrice1:
I also fixed the POD error.  That one was a last minute addition, and I was sloppy.

pstaabp (approved):
Looks good now.

Issue #901, opened 2023-08-09
radio buttons with only one option


Alex-Jordan:
If a RadioButtons has only one option, something goes wrong. Here is a MWE:

```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl parserRadioButtons.pl));

$radio = RadioButtons(["Select me"], 0);

BEGIN_PGML
[_]{$radio}
END_PGML

ENDDOCUMENT();
```

And here is a screenshot:

<img width="656" alt="Screen Shot 2023-08-09 at 12 29 04 PM" src="https://github.com/openwebwork/pg/assets/4732672/213262c1-dc82-4264-b268-1e38c279588d">

This arose in an exercise where the author has a scaffolded problem, and one section has a one-option radio button as a method to get the student to confirm that they read a certain paragraph and understand it. So I think it is legitimate to use a on-option radio buttons.

drgrice1:
In my opinion it is not valid to use a radio button with one option. It is not what radio buttons are meant for. A checkbox would be better suited for this.

drdrew42:
I concur regarding the appropriateness of checkbox for this use-case. However, I am also of the opinion that a one-option multiple choice object should still "work" (maybe include a warning for privileged accounts that an alternative should be considered?)

drgrice1:
I agree that it should still work, and I am in fact looking into it.

drgrice1:
This also happens with check boxes with one option.  So that is the real bug.

Alex-Jordan:
No longer an issue, at least not in develop.

PR #902, opened 2023-08-09
Fix the response group for single option radio or checkbox answers.

drgrice1:
The new handling in the response group when a radio or checkbox answer is extended that was added in #778 is also needed when a response is initially added.

This fixes issue #901.

This should be considered for a hotfix I believe.


Alex-Jordan (approved):
Tested this with the problem file that led to the reported issue, and it works.

pstaabp (approved):
Fixes the issue.

pstaabp:
Hotfix seems reasonable.

PR #905, opened 2023-08-12
fix hrule thickness argument

Alex-Jordan:
With horizontal rules of any kind in a `niceTables.pl` table, you can declare their thickness. You can be explicit with a unit like `2pt` or you can just use a number. If you just use a number, it is supposed to come out as that many pixels.

In hardcopy, if you are not using booktabs, this is ignored and you just get the standard hline thickness. 

If you are using booktabs, we take advantage of the ability to declare thickness on the booktabs horizontal rule commands. But it turns out that px is not a legal unit of measure. So this change converts that number of pixels to pt, using the convention 1px is 0.75pt.

To test, try

```
[@ DataTable([[[1, rowbottom => 2]]]) @]*
```

and make a hardcopy. It should fail without this commit, and work with it.

Side note: When it fails, it takes a long time, and then reports `Hardcopy generation error: The operation was aborted.` I do not understand why it happens this way, instead of the usual thing when there is bad latex: it more quickly downloads the bundle with the tex file, log, etc.

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -1492,7 +1492,7 @@ sub getLaTeXthickness {
 	my $input  = shift;
 	my $output = '';
 	if ($input =~ /^\s*(\.\d+|\d+\.?\d*)\s*$/) {
-		$output = "$1px" if $1;
+		$output = $1 * 0.75 . 'pt' if $1;
 	} elsif ($input) {
 		$output = "$input";
 	}


pstaabp:
I didn't get an error. Were you running in xelatex? From here: https://tex.stackexchange.com/questions/41370/what-are-the-possible-dimensions-sizes-units-latex-understands

it appears that px is defined in luatex and pdftex.

Alex-Jordan:
Yeah, you are right. It is a XeLaTeX thing. Try:

```
\documentclass{article}
\usepackage{booktabs}
\begin{document}
\begin{tabular}{c}
a\\\midrule[2px]
c
\end{tabular}
\end{document}
```

with both `pdflatex` and `xelatex`.

I still think this should go in. Supporting xelatex should be a long term goal, and maybe lualatex too.

pstaabp:
I think it should go also.  Just commenting.  

What I've read is that lualatex is the way to go for unicode and extra font support these days.

drgrice1 (approved):
I see the issue when `$externalPrograms{pdflatex}` is set to use xelatex.  So this is needed for that case.  It still works correctly for pdflatex too.

Note that if I download the TeX source generated by webwork2 with the develop or main branch for the given MWE, and then run xelatex in the terminal the same thing happens.  It takes forever to fail.  Something to do with xelatex in batchmode with this.

Alex-Jordan:
IIRC if you comment out the batchmode, you will find that it is prompting you for an acceptable dimension to replace the px count. I'm thinking that with batchmode, that prompt is still happening but you don't see it. And it takes forever because actually it's timing out. That's my hunch anyway.

drgrice1:
Perhaps we also need the `-halt-on-error` option.  Then when an error occurs it will just stop.

drgrice1:
I will merge this since the hotfix of this was merged.

PR #906, opened 2023-08-12
keep column styling when using multicolumn

Alex-Jordan:
With a niceTables table in hardcopy, certain situations lead to cells using multicolumn as a way to change the column type (for example making a cell left-aligned instead when the column as a whole is centered). Before this PR, that obliterates other styling that may be in play for the column, such as a background color. This brings that styling back.

You can test with:
```
[@ DataTable([
            [1, 2],
            [3, 4],
        ],
        rowheaders => 1,
        align => '>{\columncolor{lightgray}}lr',
) @]*
```

Before this commit, the mechanism that makes the row headers all be in boldface overrides the lightgray background in the first column. With this commit, the background will still be light gray.

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -942,6 +942,7 @@ sub Row {
 				$columntype =~ s/^p/b/ if ($valign eq 'bottom');
 				$columntype =~ s/^p/m/ if ($tableOpts->{valign} eq 'middle');
 				$columntype =~ s/^p/b/ if ($tableOpts->{valign} eq 'bottom');
+				$columntype = ">{$cellAlign->{tex}}" . $columntype if $cellAlign->{tex};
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Looks good.  I see the problem with the develop brancyh, and this fixes the problem.

Alex-Jordan:
I need group wisdom if things like these should also have hotfix peers. There are already 3 PRs open right now to fix some small thing with niceTables. Should they patch main too? Or go through develop first, then open one hotfix PR with all of them? And of course there will be more. There are one or two bugs I know about that I have yet to work on.

drgrice1:
I think I will defer to you on the `niceTables.pl` changes.  I say hotfix'em!

drgrice1:
I will merge this since the hotfix of this was merged.

PR #910, opened 2023-08-15
Remove the usage of the system `cp`, `rm`, and `mv` external programs.

drgrice1:
`rm` and `mv` were used in the `LaTeXImage.pm` module.  Removing the `mv` usage was easy.  The `move` method from the `File::Copy` module can be used without being shared to the safe compartment.  The `rm` method is trickier since its usage was a recursive directory removal.  You can't use the `remove_tree` method of the `File::Path` module as that requires sharing the `File::Spec` module to the safe compartment.  That is not a good idea.  So instead a `remove_tree` method is implemented in `WeBWorK::PG::IO` using `File::Find` (which works without being shared to the safe compartment).

The unused `copyCommand` method has been removed.  The last usage of the `curlCommand` method was replaced with the `externalCommand` method, and the `curlCommand` method removed.  Those commands can still be used via the more general `externalCommand` method, although the `cp` external command shouldn't be used anymore.
    
There is also some general clean up of the `WeBWorK::PG::IO` module.  The no-op `convertPath` method has been removed.  The POD has been cleaned up.  All methods that can be used by other modules are now document (in some cases poorly though -- like the AskSage method).  There were some syntax errors and perlcritic issues fixed in the code.

The `LaTeXImage.pm` has some related clean up as well.

Note that this includes pull request #898.

### DIFF
--- a/conf/pg_config.dist.yml
+++ b/conf/pg_config.dist.yml
@@ -52,7 +52,6 @@ directories:
     - $pg_root/macros/ui
     - $pg_root/macros/deprecated
 
-
 URLs:
   # The public URL of the html directory above.
   html: /pg_files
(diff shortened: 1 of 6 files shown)


pstaabp (approved):
Looks good.  Tested on a tikz image.

drdrew42 (approved):

PR #913, opened 2023-08-20
PGML tables

Alex-Jordan:
This brings table-making to PGML, starting with work by @dpvc responding to #909, with some amendments I made. It is not ready yet, because options on cells do not work. Options on a cell should work like `[. cell content .]{b => 1}` for example to make the cell bold. But instead the presence of option braces is somehow preventing the cell from even becoming a part of its row. @dpvc , do you see what the issue is there?

```
DOCUMENT();
loadMacros(
    "PGstandard.pl",
    "PGML.pl",
);

BEGIN_PGML
[# [- [. [`x`] .] [. [`x^2`].] -] [- [. [`1`] .] [. [_]{1} .] -] #]

[# [- [. [`x`] .] [. [`x^2`].] -] [- [. [`1`] .] [. [_]{1} .] -] #]{center => 0}

[#
    [-
        [. [`x`] .]
        [. [`x^2`].]
     -]
    [-
        [. [`1`] .]
        [. [_]{1} .]
     -]
#]{align => '|c|c|', horizontalrules => 1}

[#
    [-
        [. [`x`] .]
        [. [`x^2`].]
     -]
    [-
        [. [`1`] .]
        [. [_]{1} .]
     -]
#]{center => 0, align => '|c|c|', horizontalrules => 1}
END_PGML
ENDDOCUMENT();
```

I also need to look into the codemirror syntax highlighting, since a line opening with `#]` is being treated as a heading.

### DIFF
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -47,8 +47,8 @@ my $quoted    = '[$@%]q[qr]?|\bq[qr]?\s+(?=.)|\bq[qr]?(?=\W)';
 my $emphasis  = '\*+|_+';
 my $chars     = '\\\\.|[{}[\]()\'"]';
 my $ansrule   = '\[(?:_+|[ox^])\]\*?';
-my $open      = '\[(?:[!<%@$]|::?:?|``?`?|\|+ ?)';
-my $close     = '(?:[!>%@$]|::?:?|``?`?| ?\|+)\]';
+my $open      = '\[(?:[!<%@$#.-]|::?:?|``?`?|\|+ ?)';
+my $close     = '(?:[!>%@$#.-]|::?:?|``?`?| ?\|+)\]';
 my $noop      = '\[\]';
 
 my $splitPattern =
@@ -151,7 +151,8 @@ sub Parse {
 		my $token = $self->{split}[ ($self->{i})++ ];
 		next unless defined $token && $token ne '';
 		for ($token) {
-			$block->{terminator} && /^$block->{terminator}\z/ && do { $self->Terminate($token); last };
+			$block->{terminator}   && /^$block->{terminator}\z/   && do { $self->Terminate($token);    last };
+			$block->{containerEnd} && /^$block->{containerEnd}\z/ && do { $self->EndContainer($token); last };
 			/^\[[@\$]/ && ($block->{parseAll} || $block->{parseSubstitutions}) && do { $self->Begin($token); last };
 			/^\[%/     && ($block->{parseAll} || $block->{parseComments})      && do { $self->Begin($token); last };
 			/^\\./     && ($block->{parseAll} || $block->{parseSlashes})       && do { $self->Slash($token); last };
@@ -208,8 +209,17 @@ sub Begin {
 	my $options = shift || {};
 	my $def     = { %{ $BlockDefs{$id} }, %$options, token => $token };
 	my $type    = $def->{type};
+	my $class   = $def->{class} || 'PGML::Block';
 	delete $def->{type};
-	my $block = PGML::Block->new($type, $def);
+	delete $def->{class};
+	my $block = $class->new($type, $def);
+	my $end   = $self->{block}{isContainer} ? $self->{block}{terminator} : $self->{block}{containerEnd};
+	$block->{containerEnd} = $end if $end;
+
+	if ($block->{container} && $self->{block}{type} ne $block->{container}) {
+		PGML::Warning "A $type must appear in a $block->{container}";
+		$block->{hasWarning} = 1;
+	}
 	$self->{block}->pushItem($block);
 	$block->{prev}        = $self->{block};
 	$self->{block}        = $block;
@@ -222,6 +232,7 @@ sub End {
 	my $action = shift || "paragraph ends";
 	my $endAt  = shift;
 	my $block  = $self->{block};
+	return          if $block->{isContainer};
 	$block->popItem if $block->topItem->{type} eq 'break' && $block->{type} ne 'align';
 	while ($block->{type} ne 'root') {
 		if (ref($block->{terminator}) eq 'Regexp' || $block->{cancelPar}) {
@@ -247,7 +258,7 @@ sub Terminate {
 	foreach my $field (
 		"prev",     "parseComments",    "parseSubstitutions", "parseSlashes",
 		"parseAll", "cancelUnbalanced", "cancelNL",           "cancelPar",
-		"balance",  "terminateMethod",  "noIndent"
+		"balance",  "terminateMethod",  "noIndent",           "ignoreIndent"
 		)
 	{
 		delete $block->{$field};
@@ -257,6 +268,15 @@ sub Terminate {
 		if    (scalar(@{ $block->{stack} }) == 0) { $prev->popItem }
 		elsif ($block->{combine})                 { $prev->combineTopItems }
 	}
+}
+
+sub EndContainer {
+	my $self  = shift;
+	my $token = shift;
+	while (!$self->{block}{isContainer}) {
+		$self->Terminate($token);
+	}
+	$self->Terminate($token);
 }
 
 sub Unbalanced {
@@ -334,6 +354,7 @@ sub Par {
 sub Indent {
 	my $self  = shift;
 	my $token = shift;
+	return if $self->{block}{ignoreIndent};
 	if ($self->{atLineStart}) {
 		my $tabs = $token;
 		$tabs =~ s/    /\t/g;    # turn spaces into tabs
@@ -568,6 +589,31 @@ sub NOOP {
 my $balanceAll = qr/[\{\[\'\"]/;
 
 %BlockDefs = (
+	"[#" => {
+		type         => 'table',
+		class        => 'PGML::Block::Table',
+		parseAll     => 1,
+		ignoreIndent => 1,
+		terminator   => qr/#\]/,
+		allowStar    => 1,
+		options      => [ "align", "midrules", "center" ]
+	},
+	"[-" => {
+		type         => 'table-row',
+		class        => 'PGML::Block::TableRow',
+		parseAll     => 1,
+		ignoreIndent => 1,
+		container    => 'table',
+		terminator   => qr/-\]/
+	},
+	"[." => {
+		type        => 'table-cell',
+		parseAll    => 1,
+		isContainer => 1,
+		container   => 'table-row',
+		terminator  => qr/.\]/,
+		options     => ["headerrow"]
+	},
 	"[:" => {
 		type               => 'math',
 		parseComments      => 1,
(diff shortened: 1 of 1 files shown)


pstaabp:
@Alex-Jordan I don't think it's worth looking into codemirror right now.  We may switch to codemirror 6 this release which would mean a complete rewrite.

macros/core/PGML.pl
```diff
@@ -1129,6 +1181,46 @@ sub pushItem {
...
+our @ISA = ('PGML::Block');
+
+sub pushItem {
+	my $self = shift;
+	my $item;
+	while ($item = shift) {
+		if ($item->{type} eq 'text') {
+			my $text = join('', @{ $item->{stack} });
+			PGML::Warning 'Table row text must be in cells' unless $text =~ m/^\s*$/;
+		} elsif ($item->{type} eq 'table-cell') {
```
dpvc, line 1215: It turns out that we need to allow an `options` block to be used on the row stack (it will come after a `table-cell` block`).  This was leading to the options not being ignored (and an error being flagged).  It turns out that PGML doesn't show the error messages, but it probably should.  This should do the trick:

```suggestion
		} elsif ($item->{type} eq 'table-cell' || $item->{type} eq 'options') {
```

dpvc (changes requested):
The issue was my fault:  I was not allowing options to be processed in table rows.  (As I think I mentioned, I don't remember all the details of how PGML works, as it is something like 20 years old, now.)  I've included a suggestion that should fix the issue.

Alex-Jordan:
Thanks @dpvc, that did the trick and things I've tested have been successful.

Here is on of the `niceTables` wiki doc tables, redone this way:

```
DOCUMENT();
loadMacros(
    "PGstandard.pl",
    "PGML.pl",
);
BEGIN_PGML
[#
    [-
        [.Name.]{headerrow => 1}
        [.Biography.]
     -]
    [-
        [._James Kirk_.]
        [.James Tiberius Kirk was born on March 22, 2233, in Riverside, Iowa..]
     -]
    [-
        [._Jean-Luc Picard_.]
        [.Jean-Luc Picard was born to Maurice and Yvette Picard in La Barre, France, on July 13, [_]{2305}, and dreamed of joining Starfleet..]
     -]
#]{caption => 'Captains of the Enterprise', horizontalrules => 1, align => 'lX'}
END_PGML
ENDDOCUMENT();
```

I think where this will be most useful is tables with answer blanks in them. Otherwise using the `[@ DataTable(...) @]*` construction is about the same effort. But with answere blanks and the ability to use PGML answer blank markup, this will be better. So like with:

```
DOCUMENT();
loadMacros(
    "PGstandard.pl",
    "PGML.pl",
);

BEGIN_PGML
[#
    [-
        [. [`x`] .]{headerrow => 1}
        [. [`x^2`].]
     -]
    [-
        [. [`1`] .]
        [. [_]{1} .]
     -]
    [-
        [. [`2`] .]
        [. [_]{4} .]
     -]
    [-
        [. [`3`] .]
        [. [_]{9} .]
     -]
#]{center => 0, align => '|c|c|'}
END_PGML
ENDDOCUMENT();
```

Alex-Jordan:
One thing about this that could be simplified is to cut out the `[-   -]` delimiters for rows. Instead, a double line break could indicate moving to a new row.  So instead of:

```
[#
    [-
        [. [`x`] .]{headerrow => 1}
        [. [`x^2`].]
    -]
    [-  [.[`1`].]  [.[_]{1}.]  -]
    [-  [.[`2`].]  [.[_]{4}.]  -]
    [-  [.[`3`].]  [.[_]{9}.]  -]
#]{center => 0, align => '|c|c|'}
```

it could be:

```
[#
    [. [`x`] .]{headerrow => 1}
    [. [`x^2`].]

    [.[`1`].]  [.[_]{1}.]

    [.[`2`].]  [.[_]{4}.]

    [.[`3`].]  [.[_]{9}.]
#]{center => 0, align => '|c|c|'}
```

I can't tell which is worse for authors, having to manage `[-   -]` or double line breaks.

pstaabp:
@Alex-Jordan I had tested this last week.  Not sure where my comments went.  I'll have to reproduce.  However, with your last comment, would putting a * on the last cell be a good way to indicate the end of a row?  

Your example above, could be:
```
[#
    [. [`x`] .]{headerrow => 1}
    [. [`x^2`].]*
    [.[`1`].]  [.[_]{1}.]*
    [.[`2`].]  [.[_]{4}.]*
    [.[`3`].]  [.[_]{9}.]*
#]{center => 0, align => '|c|c|'}
```
Seems a little simpler. Not sure how easy to code though.

Alex-Jordan:
> putting a * on the last cell 

I like this idea, but I realized recently it will not be so trivial to implement. The niceTables argument syntax effectively marks where a row begins *and* ends using brackets:

```
DataTable(
    [
        [cellA ,cellB, cellC],
        [cellD ,cellE, cellF]
    ]
)
```

The proposal to use stars means that in PGML, when processing cellC (which would have a star), we have to do something outside the scope of cellC, grouping the previous cells together into a row. This is an inherent challenge with the star idea that effectively only marks up the *end* of a row.

A further challenge would come from cellF. Should authors be required to star that cell? If not, while processing cellF, we will have to again look outside the scope of cellF to understand it is the end of a row, along with the grouping of cells needed at cellC.

I'll give it a try. I am unsure if I should approach this entirely within PGML.pl, or if I should grow niceTables.pl to accommodate an alternate syntax that would make the PGML.pl code straightforward. Like:

```
DataTable(
    [cellA ,cellB, {data => cellC, terminateRow => 1}, cellD ,cellE, cellF]
)
```

That would make things easy on the PGML.pl side. And niceTables.pl could recognize when the first argument is this way and immediately convert to the first form. I think that will be my approach unless someone (@dpvc ?) proposes something better.

dpvc:
> I like this idea, but I realized recently it will not be so trivial to implement. 

It is not that hard.  I have attached a version of PGML.pl that implements it.  The changes are that the table row syntax and implementation is removed, and the table block now just contains a collection of cells.  The `Table()` subroutine that is the one that actually calls the `DataTable()` function is modified to break up the list of cells into the rows based on the location of the starred cells.  That is not all that difficult to do.  Here's the complete diff against the current PGML.pl:

``` diff
50,51c50,51
< my $open      = '\[(?:[!<%@$#.-]|::?:?|``?`?|\|+ ?)';
< my $close     = '(?:[!>%@$#.-]|::?:?|``?`?| ?\|+)\]';
---
> my $open      = '\[(?:[!<%@$#.]|::?:?|``?`?|\|+ ?)';
> my $close     = '(?:[!>%@$#.]|::?:?|``?`?| ?\|+)\]';
604,611d603
< 	"[-" => {
< 		type         => 'table-row',
< 		class        => 'PGML::Block::TableRow',
< 		parseAll     => 1,
< 		ignoreIndent => 1,
< 		container    => 'table',
< 		terminator   => qr/-\]/
< 	},
616c608
< 		container   => 'table-row',
---
> 		container   => 'table',
617a610
> 		allowStar   => 1,
1196,1214d1188
< 		} elsif ($item->{type} eq 'table-row') {
< 			$self->SUPER::pushItem($item);
< 		} elsif ($item->{type} eq 'comment') {
< 		} else {
< 			PGML::Warning 'Tables can contain only table rows';
< 		}
< 	}
< }
< 
< package PGML::Block::TableRow;
< our @ISA = ('PGML::Block');
< 
< sub pushItem {
< 	my $self = shift;
< 	my $item;
< 	while ($item = shift) {
< 		if ($item->{type} eq 'text') {
< 			my $text = join('', @{ $item->{stack} });
< 			PGML::Warning 'Table row text must be in cells' unless $text =~ m/^\s*$/;
1219c1193
< 			PGML::Warning 'Table rows can contain only table cells';
---
> 			PGML::Warning 'Tables can contain only table cells';
1348,1354c1322,1330
< 	if ($item->{type} eq 'table-cell') {
< 		return [ $self->string($item), @options ];
< 	} elsif ($item->{type} eq 'table-row') {
< 		return [ map { $self->Table($_) } @{ $item->{stack} } ];
< 	} elsif ($item->{type} eq 'table') {
< 		my $def = [ map { $self->Table($_) } @{ $item->{stack} } ];
< 		return ($item->{hasStar} ? main::LayoutTable($def, @options) : main::DataTable($def, @options));
---
> 	return [ $self->string($item), @options ] if $item->{type} eq 'table-cell';
> 	my $table = [];
> 	my $row = [];
> 	for my $cell (@{ $item->{stack} }) {
> 		push(@$row, $self->Table($cell));
> 		if ($cell->{hasStar}) {
> 			push(@$table, $row);
> 			$row = [];
> 		}
1355a1332,1333
> 	push(@$table, $row) if @$row;
> 	return ($item->{hasStar} ? main::LayoutTable($table, @options) : main::DataTable($table, @options));
```

I also attach the complete file here:  [PGML.pl.zip](https://github.com/openwebwork/pg/files/12629287/PGML.pl.zip)

I think that should do what you want.  One issue that might be a concern is that the star notation doesn't provide a means of indicating an empty row.  I'm not sure that really matters, as you can always use `[. .]*` for a row with a single cell that is blank, but it's not technically an empty row.  On the other hand, the current implementation doesn't allow empty rows either, since empty blocks are removed (so empty rows and empty cells are removed, though `[. .]` is not empty, since it contains a space as text).  But it would be possible to extend the current notation to allow empty rows, while that is not so easy for the star notation.

Alex-Jordan:
Thanks @dpvc , I committed your edits and it works well with this test file. Which has a mix of using stars on the table itself for DataTable versus LayoutTable, and a mix of using stars on the very last cell or not.

```
DOCUMENT();
loadMacros(
    "PGstandard.pl",
    "PGML.pl",
);

BEGIN_PGML
[# [. [`x`] .] [. [`x^2`].]* [. [`1`] .] [. [_]{1} .]* #]

[# [. [`x`] .] [. [`x^2`].]* [. [`1`] .] [. [_]{1} .] #]{center => 0}

[#
        [. [`x`] .]
        [. [`x^2`].]*
        [. [`1`] .]
        [. [_]{1} .]*
#]{align => '|c|c|', horizontalrules => 1}

[#
        [. [`x`] .]
        [. [`x^2`].]*
        [. [`1`] .]
        [. [_]{1} .]
#]{center => 0, align => '|c|c|', horizontalrules => 1}
END_PGML
ENDDOCUMENT();
```

From my testing, there is one last thing I find that would make this better, which I couldn't figure out on my own. Right now if there is a blank line (or blank line with white space) within a table, in between cells, it causes an error. For example:

```
[#
        [. [`x`] .]
        [. [`x^2`].]*

        [. [`1`] .]
        [. [_]{1} .]*
#]{align => '|c|c|', horizontalrules => 1}
```

or 

```
[#
        [. [`x`] .]
    
        [. [`x^2`].]*
        [. [`1`] .]
        [. [_]{1} .]*
#]{align => '|c|c|', horizontalrules => 1}
```

I tried a few things but don't understand the issue. These new lines aren't themselves triggering the error, but they lead to the subsequent cells triggering error messages like "A table-cell must appear in a table".

Alex-Jordan:
Nevermind @dpvc, I think I found what I needed. I added `allowPar => 1` to the table block. Now things are working as far as my testing goes. For anyone who would like to test, as a starting point try:

```
DOCUMENT();
loadMacros(
    "PGstandard.pl",
    "PGML.pl",
);

BEGIN_PGML
## Layout Table, one line, terminates last row

[# [. [`x`] .] [. [`x^2`].]* [. [`1`] .] [. [_]{1} .]* #]

## Data Table, one line, does not terminate last row

[# [. [`x`] .] [. [`x^2`].]* [. [`1`] .] [. [_]{1} .] #]{center => 0}

## Layout Table, multi line, has blank line, bad indentation, terminates last row

[#
        [. [`x`] .]{bgcolor => 'yellow'}
        
[. [`x^2`].]*{bgcolor => 'yellow'}

        [. [`1`] .]
        [. [_]{1} .]*
#]{align => '|c|c|', horizontalrules => 1}

## Data Table, multi line, does not terminate last row

[#
        [. [`x`] .]
        [. [`x^2`].]*
        [. [`1`] .]
        [. [_]{1} .]
#]{center => 0, align => '|c|c|', horizontalrules => 1}
END_PGML
ENDDOCUMENT();

```


One thing I would like @pstaabp s help with. In the PG editor, codemirror is treating lines like `#]{align => '|c|c|', horizontalrules => 1}` as heading lines, because they begin with a `#`. Do you have a recommendation for preventing that? Also does something need to be done to recognize `[# ... #]` and `[. ... .]` as blocks for codemirror? Or would we just leave that be?

dpvc:
> I think I found what I needed. I added `allowPar => 1` to the table block. 

Yes, that is the correct addition.  Normally, a paragraph break ends an open block (e.g., a blank line ends a list, for example), so the blank lines were ending the table blocks, and then the cells where outside a table, causing the warning messages.  The `allowPar` option prevents a paragraph from ending the block, which is what you need.

So good work in figuring that out.

pstaabp:
> One thing I would like @pstaabp s help with. In the PG editor, codemirror is treating lines like #]{align => '|c|c|', horizontalrules => 1} as heading lines, because they begin with a #. Do you have a recommendation for preventing that? Also does something need to be done to recognize [# ... #] and [. ... .] as blocks for codemirror? Or would we just leave that be?

This will be a good nudge to start looking into rewriting a PG/PGML mode for code mirror 6, which can hopefully take care of these issues.  I'll start looking into this.  For now, let's not worry about this.

pstaabp:
This is looking good.  Seems pretty easy to write tables with this syntax.  Haven't thoroughly tested it though.

drgrice1 (approved):
This looks good to me.  @dpvc: Have your change requests been addressed?

dpvc (approved):
Looks good, and is a nice addition to PGML.

Issue #912, opened 2023-08-18
Variable adjacent to code substitution in TEXT causes hardcopy error.


Alex-Jordan:
The following could be old, and could be expected behavior to WW developers. But it is unexpected behavior for problem authors. 

Try this MWE:

```
DOCUMENT();
loadMacros("PGstandard.pl");
BEGIN_TEXT

$DOLLAR\{ ans_rule(1) \}

END_TEXT
ANS(Real(1)->cmp);
ENDDOCUMENT();
```

and build hardcopy, and it fails. The tex file it creates is:

```
\batchmode
\documentclass{article}
\usepackage{webwork2}
\pagestyle{empty}
\parindent=0pt
\begin{document}

\newpage
{\bf (Error: syntax error  in '
$DOLLAR{\\answerRule[AnSwEr0001]{3}}

')}

\vspace{\baselineskip}\par{\small{\it Correct Answers:}
\begin{itemize}
\item
$\displaystyle 1$
\end{itemize}}\par

\end{document}
```

Note this is not about dollar signs in tex. You get similar errors with any variable that is immediately adjacent to a code block inside a `TEXT` block.

apizer:
This is not an old issue. It is new to WW2.18.  It breaks one of the oldest WW problems (Problem 2 in Set 0) but I thought the problem was corrupted and fixed it by accident.  I could not get the minimal problem above to work, but the following (please forgive my old technique) works in WW2.17 but fails in WW2.18.

DOCUMENT();
loadMacros("PGstandard.pl");
BEGIN_TEXT
Enter 1:
$BR\{ ans_rule(1) \}

END_TEXT
$ans =1;
ANS(std_num_cmp($ans));

ENDDOCUMENT();

drgrice1:
This seems to be due to #863.  If you add a space in the TeX string on line 319 of PGbasicmacros.pl, then hardcopy generation succeeds.  However, it also adds a space in the output which is not desired.  This seems to be another case of EV3P parsing something in an odd way.

Alex-Jordan:
Quick not that my MWEE was missing `MathObjects.pl`. 

```
DOCUMENT();
loadMacros("PGstandard.pl", "MathObjects.pl");
BEGIN_TEXT

$DOLLAR\{ ans_rule(1) \}

END_TEXT
ANS(Real(1)->cmp);
ENDDOCUMENT();
```

drgrice1:
[...]

PR #914, opened 2023-08-21
Wrap the `\\answerRule`` in TeX output in an `\\mbox{...}`.

drgrice1:
This fixes issue #912.  See the discussion there.

@Alex-Jordan: If you have a better idea for this, then please feel free to put in another pull request, and I will close this one.


Alex-Jordan:
Do I recall right that the outer brace in `{\\answerRule[$name]{$tcol}}` is something we added to protteect against issues with LaTeX and bracket `[` matching?

drgrice1:
Yes.  The problem was that LaTeX doesn't like `\item[\answerRule[$name][$tcol]]`.  The nested brackets don't work.

drgrice1:
While `\item[{\answerRule[$name][$tcol]}]` does work.

Alex-Jordan:
I'm also having success against the issue by putting `\\null` in front of that brace. We want something before that brace to prevent perl from seeing a hash access. I'm thinking `\\null` is less of a change than `\\mbox`. But still not sure what is best.

Alex-Jordan:
I'm trying to think through if there is a correct way to escape that brace too, so that it comees out as a plain brace for LaTeX, but is escaped at that point for perl. Can you see that working out?

drgrice1:
Yeah, `\\null` should work. In fact that is more like what I was looking for, but I couldn't remember or find a latex noop command. That is it.

Alex-Jordan:
I guess we can't distinguish a brace in an author's `${DOLLAR}` from the brace in `{\\answerRule[$name]{$tcol}}`. One should be understood as a brace by perl, the other should not. So there is no hope of escaping it in an automated way.

drgrice1:
So it seems that `\null` is the equivalent of `\hbox{}`.  So that isn't that different than wrapping the answer rule in `\mbox{...}`.  The output looks the same.  However, wrapping in the `\mbox` might mess with the `\raisebox` of the `\answerRule` definition.  So I will switch to `\null` for now.

I can't think of anyway to escape the beginning brace so that it is not interpreted as a hash accessor by perl, but yet is a regular brace for latex.  That is unless we add some special parsing to the EV3P method.

Alex-Jordan:
> That is unless we add some special parsing to the EV3P method.

That's the line I was thinking on. Because adding these `\\null`, `\\mbox` or whatever feels like a hack and there's not much protection against a future developer changing something there, and not realizing the job these things would be doing. But in EV3P, if something could handle this in all cases (like how actual dollar signs in the author's problem are handled there) that would be better.  

But it seems impossible to distinguish the kind of brace that we would want escaped from the kind of brace we would not want escaped.

Alex-Jordan:
Is this proposition reasonable?

> There will never be a hash accessor construction of the form `$hashname{\\string}`

Based on that, I added:

```
$eval_string =~ s/\{(?=\\\\[^\\])/\\\{/g if $options{escapeBraces};
```
to `EV3P` and that works in my testing, without any changes to `PGbasicmacros.pl`. "escapeBraces" is a misnomer and this line could either be unconditional, or I can find a better name.

I'm seeking a solution like this so that wee can be free to write/edit macros for tex without having to be aware of what could happen when the macro starts with a brace.

drgrice1:
I think that should be okay.  Perhaps rather than using `if $option{escapeBraces}`, use `if $displayMode eq 'TeX'`.  I don't think this should ever be needed for any other display modes.

Also, I think that for a hotfix we should use `\\null`, and then we can add your suggestion to develop.  I am cautious of this, and think that adding it to develop for now gives us time to test it, and see if it causes any other problems.  Adding `\\null` as in this pull request shouldn't break anything.

Alex-Jordan:
That all sounds good. I'll amend the PR I opened in a bit.

dpvc:
Would it be better to use `\relax` than `\null`?  The former is a no-op, but the latter inserts an empty hbox.  Not quite the same thing, and probably trivial, but making an Xbox does have consequences (like initiating horizontal mode if you are in vertical mode).

drgrice1:
Yes, I agree that would be better.

drgrice1:
@dpvc:  What do you think of the idea in https://github.com/openwebwork/pg/pull/914#issuecomment-1686434286 and #916?

drgrice1:
I am going to close this in favor of #917.

@Alex-Jordan: This doesn't need to be merged.  We don't have to merge things that aren't needed into develop.

dpvc:
My thought is that these solutions seem to target the answer-rule output specifically, but that the problem is more general than that.  It turns out that other output from `\{...\}` can interact badly with variables that precede it, and variables within `\{...\}` can interact badly with text that follows.  For example, consider

``` perl
$x = "one"; $x = "two"; $x = "three";
$y = '$x';

BEGIN_TEXT
$x\{"x"\} and $x\{"x"\}x  and \{'$y'\}x
END_TEXT
```

which will produce `two and three and two` with the current `EV3P()`, because after evaluating the command substitutions, the string is `$xx and $xxx and $xx`, which is then evaluated for variable substitution.

This shows that the interaction of variables before and within command substitution can interact with the text produced by and following the command.  It's not just a preceding variable when braces at the beginning of the command output.

To resolve this, one would want a way to prevent variables on the left of a command substitution from "bleeding into" the command output, and from variables within the command output from bleeding into the following text, without affecting the output.   We want something that we can wrap around the command output to isolate it from the surrounding text.  For TeX output, adding `\relax` at the beginning and end of command output would do, but what about other output modes?

It turns out that we can do something that is independent of the output mode.  Because the issue is only when there is variable substitution, we can isolate the command output by using a variable reference on both ends, with a variable that inserts nothing.  E.g., if we add `${__blank__}` on both ends of the command output, (and have `$__blank__ = ""`), then that will keep variable substitution from bleeding into or out of the output.

So if you define

``` perl
sub safe_evp {
  my @result = &safe_ev;
  $result[0] = '${__blank__}' . $result[0] . '${__blank__}';
  return @result;
}
```

after the definition of `save_ev` and then replace

https://github.com/openwebwork/pg/blob/e1313567a81dc3b9b6a70b8a45817b3804171c09/macros/core/PGbasicmacros.pl#L2297

by

``` perl
        our $__blank__ = '';
	$string = ev_substring($string, "\\\\{", "\\\\}",
                               $options{processVariables} ? \&safe_evp : \&safe_ev) if $options{processCommands};
```

that would fix the issue with the answer rules, and also the problems I indicate above.

The only issue I can think of that this could potentially cause would be if someone was using `\{...\}` in a way that they actually *wanted* to bleed through.  For example

``` perl
@x = ("zero", "one", "two", "three");

BEGIN_TEXT
$x[\{2\}]
END_TEXT
```

though I think this probably isn't being done, and it could be fixed by using

``` perl
@x = ("zero", "one", "two", "three");

BEGIN_TEXT
$x["\{2\}"]
END_TEXT
```

if you really wanted to do that.  In any case, I doubt this is a real issue.

dpvc:
Sorry, I see you just closed this.  But those are still my comments on the issue.

drgrice1:
@dpvc:  Your comments are still valid, and you certainly make a good point.  We can continue this discussion in #917.

PR #915, opened 2023-08-21
Prefix the `{\\answerRule...}`` in TeX output with `\\relax` (hotfix).

drgrice1:
This fixes issue #912.  See the discussion there.

PR #916, opened 2023-08-21
escape pattern for EV3P for perl/latex brace conflict (hotfix)

Alex-Jordan:
This is an alternative to #915, as mentioned there. This approach escapes a brace that we are pretty sure is not intended to be part of a perl hash value accessor. I lean toward this approach since it doesn't require macro authors to know about the underlying issue with braces. And if it needs adjusting, it's just in one place.

This is a conservative regex pattern that only escapes the brace if followed by backslash,backslash,not-backslash. It could be a little less conservative. Or a lot less conservative like `s/\{(?!\s*[\$\w])/\\\{/g`, _always_ escaping a brace unless it is followed by a word character or a dollar sign, which could indicate the brace was indeed meant to be part of a hash value accessor.

I'll wait for discussion to see if this goes forward ro needs changing, before opening a `develop` version.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -2286,11 +2286,12 @@ sub EV3P {
 	my $option_ref = {};
 	$option_ref = shift if ref($_[0]) eq 'HASH';
 	my %options = (
-		processCommands  => 1,
-		processVariables => 1,
-		processParser    => 1,
-		processMath      => 1,
-		fixDollars       => 1,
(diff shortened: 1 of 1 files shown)


drgrice1:
I think you have not yet seen my comment at https://github.com/openwebwork/pg/pull/914#issuecomment-1687201200.  As per that comment, this should be targeted for develop, and adjusted a bit.

Alex-Jordan:
Closing in favor of #914, #915, and #917.

PR #917, opened 2023-08-22
escape braces that are not for perl hash access

Alex-Jordan:
This is a divergence from the hotfix being applied in #915, to address the issue in #912.


Alex-Jordan:
Closing this in favor of #918.

PR #918, opened 2023-08-22
Protect EV3P command output from variable substitution.

drgrice1:
This an alternate to #917 (and #914) to fix issue #912.  This uses the technique described by @dpvc in
https://github.com/openwebwork/pg/pull/914#issuecomment-1688929625 with a minor tweak to limit the scope of `$__blank__`.

Here is a MWE with all of the issues that have been brought up:
```perl
DOCUMENT();

loadMacros('PGstandard.pl', 'MathObjects.pl');

$x = "one"; $xx = "two"; $xxx = "three";
$y = '$x';

BEGIN_TEXT
$DOLLAR\{ ans_rule(1) \}
$PAR
$x\{"x"\} and $x\{"x"\}x and \{$y\}x
END_TEXT

ANS(Real(1)->cmp);

ENDDOCUMENT();
```

With the develop branch the above example will work in html, but the output on the second line will be "two and three and two".  With this pull request the output on the  second line will be "onex and onexx and onex".

With the develop branch the above example will not work in hardcopy, but with this branch it will.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -2070,6 +2070,12 @@ sub safe_ev {
 	($out, $PG_eval_errors, $PG_full_error_report);
 }
 
+sub safe_evp {
+	my @result = &safe_ev;
+	$result[0] = '${__blank__}' . $result[0] . '${__blank__}';
+	return @result;
+}
+
 sub old_safe_ev {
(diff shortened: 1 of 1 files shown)


pstaabp (approved):

drdrew42 (approved):
Tested 👍

PR #925, opened 2023-09-15
enforce MA checkTypes for each evaluator

drdrew42:
## MultiAnswer and Type Checking

When using MultiAnswer, the `checkTypes` setting is effectively ignored. See attached file for MWE.

What happens is that when the user response is parsed, MathObjects do type checking by default and add a feedback message. The presence of this feedback message stops the evaluator from calling the custom checker in the MultiAnswer object. Culprit: L417 below.

https://github.com/openwebwork/pg/blob/e1313567a81dc3b9b6a70b8a45817b3804171c09/macros/parsers/parserMultiAnswer.pl#L414-L422

## Fix(es?)

This PR stifles the type warnings by setting `showTypeWarnings` to false (only when `checkTypes` is false, of course) within each individual evaluator/AnswerHash that the MultiAnswer contains. 

Another possible solution would be to modify L417 as follows:
``` perl
return if ($self->{checkTypes} && $ans->{ans_message}) || !defined($ans->{student_value});
```

This second solution has the benefit of retaining the default type-checking message, but I wonder about situations where `ans_message` might be defined for reasons other than type-checking failures. Does anyone know of potential edge-cases here?

This PR takes (what I believe to be) the safer route.

### DIFF
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -234,6 +234,13 @@ sub cmp {
 			});
 		}
 	}
+	# showing type warnings will prevent the custom checker
+	# from being called, regardless of the checkTypes setting
+	if (!$self->{checkTypes}) {
+		foreach my $cmp (@{ $self->{cmp} }) {
+			$cmp->ans_hash(showTypeWarnings => 0);
+		}
+	}
 	my @cmp = ();
 	if ($self->{singleResult}) {
 		push(@cmp, $self->ANS_NAME(0)) if $self->{namedRules};
@@ -372,7 +379,7 @@ sub entry_cmp {
 }
 
 #
-#  Call the correct answser's checker to check for syntax and type errors.
+#  Call the correct answer's checker to check for syntax and type errors.
 #  If this is the last one, perform the user's checker routine as well
 #  Return the individual answer (our answer hash is discarded).
 #
@@ -424,7 +431,7 @@ sub perform_check {
 	$rh_ans->{isPreview} = $inputs->{previewAnswers}
 		|| ($inputs_{action} && $inputs->{action} =~ m/^Preview/);
 
-	Parser::Context->current(undef, $context);                                 # change to multi-answser's context
+	Parser::Context->current(undef, $context);                                 # change to multi-answer's context
 	my $flags = Value::contextSet($context, $self->cmp_contextFlags($ans));    # save old context flags
 	$context->{answerHash} = $rh_ans;                                          # attach the answerHash
 	my @result = Value::cmp_compare([@correct], [@student], $self, $rh_ans);


drdrew42:
Looks like I forgot to attach the MWE. 

[multianswer-mwe.pg.txt](https://github.com/openwebwork/pg/files/12624018/multianswer-mwe.pg.txt)

Before this PR, if a user responds with "x", the error message will be "Your answer is not a number..." and the checker obviously does not run because no credit is given.

With this PR, the response of "x" receives the feedback message "Your answer has the wrong type" (from the custom checker) and receives full credit because the custom checker has been called.

dpvc:
The `checkTypes` option actually works as documented.  The description for this option is

> whether the types of the student and professor's answers must match exactly or just pass the usual type-match error  checking (in which case, you should check the types before you use the data).

That is, the answers are always subject to the type-match checking; `checkTypes` only determines if the types must match *exactly* or not.  For example, if an answer is a Real, then a String answer will not produce a type mismatch error (since a string like `DNE` is allowed for a potential numeric limit answer, for instance), but a String answer would prevent the checker from being called if `checkTypes` is true.

The reasoning for this is that the checker often assumes the student answer is the same type as the correct answer, and most people will not know to take the needed precautions to use the student answer correctly when it isn't.  For example, they might try to take the square root of the student answer when it is expected to be a number, which will fail when the student entered `DNE`.  So the checker is not called in that situation by default, and the author must specifically elect to do so, and ones hopes that that would help them realize that they must take precautions about that.

The option wasn't intended to disable the type matching as a whole.  One could use a post-filter to adjust the message when a type-match fails, for example.  I admit that the name `checkTypes` was a bad choice, and it might have been better called `exactTypes` or something.  Perhaps it would be useful to have a new option that does what your proposal suggests, or to change `checkTypes` to be the new feature and use `exactTypes` for the existing one, but one would want to check its usage in the OPL to decide if changing its meaning would cause problems.  If you keep `checkTypes` for what not currently does, then perhaps `checkTypeMatch` might be the new option.

If y9ou go with a new option, then your first approach (the one actually in this PR) is the better of the two.  The answer message can be set for a variety of reasons during the check, such as syntax errors (though in that case the `student_value` will be undefined), so using that as an indicator of a type mismatch is probably not a good idea.

drdrew42:
Thanks @dpvc for weighing in. You are correct regarding the documentation -- I think I was reading what I wanted `checkTypes` to mean rather than what the documentation actually said. 

The 'global' change to the behavior of `checkTypes` in this PR isn't strictly necessary if we consider the convenience method added in drdrew42/pg#1 ([06a0aea](https://github.com/drdrew42/pg/pull/1/commits/06a0aeaea2a6c835a41f41b2948fdcb7dbe87d29)) -- which can be used to set comparison flags (such as `showTypeWarnings => 0`) on any individual answer checker in a MultiAnswer. The added benefit of this approach is its granularity and flexibility across all MathObject answer checkers.

drdrew42:
Closed in favor of #926

PR #926, opened 2023-09-18
Convenience Method for MultiAnswer

drdrew42:
## MultiAnswer and AnswerEvaluator Attributes

After some discussion about #925, I think the clearer solution is to offer a convenience method for passing settings for individual answer rules.

Thanks to @dpvc for feedback regarding maintaining existing functionality.

### setCmpFlags

* The first argument to `setCmpFlags` is the answer rule to be targeted (starting with 1). 
* Following this first argument any number of settings may be provided (as a list, not a HASH ref).

#### Type Checking:

The original problem I faced involved customizing the error messages when type matching failed. I wrote a post-filter to replace the default error messages, but I also wanted to provide partial credit. However, the presence of the type-match warning prevented my checker from doing so (see #925 for details). Now, with this method I can avoid the default type-match warning and provide my own instead with partial credit in my checker: 

``` perl
$multianswer_obj->setCmpFlags(1, showTypeWarnings => 0);
```

#### Custom Context Flags:

Thanks to @drgrice1 for this example: 

When working with `Context("Fraction")`, the `studentsMustReduceFractions => 0` setting will automatically reduce student responses _before_ they reach our `MultiAnswer->{checker}` subroutine. If one wishes to provide partial credit for un-reduced fraction answers, it is necessary to ensure that this flag is true. Before the introduction of this convenience method, code contortions were necessary to configure this setting in a MultiAnswer object.

### DIFF
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -381,7 +381,7 @@ sub entry_cmp {
 }
 
 #
-#  Call the correct answser's checker to check for syntax and type errors.
+#  Call the correct answer's checker to check for syntax and type errors.
 #  If this is the last one, perform the user's checker routine as well
 #  Return the individual answer (our answer hash is discarded).
 #
@@ -433,7 +433,7 @@ sub perform_check {
 	$rh_ans->{isPreview} = $inputs->{previewAnswers}
 		|| ($inputs_{action} && $inputs->{action} =~ m/^Preview/);
 
-	Parser::Context->current(undef, $context);                                 # change to multi-answser's context
+	Parser::Context->current(undef, $context);                                 # change to multi-answer's context
 	my $flags = Value::contextSet($context, $self->cmp_contextFlags($ans));    # save old context flags
 	$context->{answerHash} = $rh_ans;                                          # attach the answerHash
 	my @result = Value::cmp_compare([@correct], [@student], $self, $rh_ans);
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -209,6 +209,15 @@ sub new {
 }
 
 #
+#  Set flags to be passed to individual answer checkers
+#
+sub setCmpFlags {
+	my ($self, $cmp_number, %flags) = @_;
+	die "Answer $cmp_number is not defined." unless defined($self->{cmp}[ $cmp_number - 1 ]);
+	$self->{cmp}[ $cmp_number - 1 ]->ans_hash(%flags);
+}
+
+#
 #  Creates an answer checker (or array of same) to be passed
 #  to ANS() or NAMED_ANS().  Any parameters are passed to
 #  the individual answer checkers.
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -79,6 +79,7 @@ sub setCmpFlags {
 	my ($self, $cmp_number, %flags) = @_;
 	die "Answer $cmp_number is not defined." unless defined($self->{cmp}[ $cmp_number - 1 ]);
 	$self->{cmp}[ $cmp_number - 1 ]->ans_hash(%flags);
+	return $self;
 }
 
 #
@@ -435,19 +436,28 @@ or a separate result for each rule.
 
 =head1 ATTRIBUTES
 
-Create a new C<MultiAnswer> item by passing a list of answers to the constructor. The answers
-are converted into C<MathObjects> if they aren't already.
+Create a new C<MultiAnswer> item by passing a list of answers to the constructor.
+
+The answers may be provided as C<MathObjects>, C<AnswerEvaluators>, or as strings (which will be
+converted into C<MathObjects>).
 
 C<MultiAnswer> objects have the following attributes:
 
-=head2 checker
+=head2 checker (required)
 
 A coderef to be called to check student answers. This is the only required attribute.
 
-The C<checker> routine is passed four parameters: a reference to the array of correct answers,
+The C<checker> routine receives four parameters: a reference to the array of correct answers,
 a reference to the array of student answers, a reference to the C<MultiAnswer> object itself,
 and a reference to the checker's answer hash. The routine should return either a score or a
 reference to an array of scores (one for each answer).
+
+    # this checker will give full credit for any answers
+    sub always_right {
+		my ($correct,$student,$multi_ans,$ans_hash) = @_;  # get the parameters
+		return [ (1) x scalar(@$correct) ];                # return an array of scores
+	}
+	$multianswer_obj = $multianswer_obj->with(checker=>~~&always_right);
 
 =head2 singleResult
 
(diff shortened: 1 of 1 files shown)


dpvc:
> I think the clearer solution is to offer a convenience method for passing settings for individual answer rules.

There is another alternative, which would be to allow you to pass a MathObjects answer checker directly to `MultAnswer()` itself.  Currently, you can pass either a MathObject or a string that parses to one, but you could allow an answer checker as well, provided it came from a MathObject.  The presence of, say, the `correct_value` property should be sufficient to identify a MathObject checker from other ones.  

The lines that need changing are

https://github.com/openwebwork/pg/blob/e1313567a81dc3b9b6a70b8a45817b3804171c09/macros/parsers/parserMultiAnswer.pl#L189-L192

and making them something like

``` perl
	foreach my $x (@data) {
		if (ref($x) eq 'AnswerEvaluator') {
			Value::Error('Only MathObject answer checkers can be passed to MultiAnswer()')
				unless defined($x->{correct_value});
			push(@cmp, $x);
			$x = $x->{correct_value};
		} else {
			$x = Value::makeValue($x, context => $context) unless Value::isValue($x);
			push(@cmp, $x->cmp(@ans_defaults));
		}
	}
```

should do the trick (though I haven't tested this).

Then you can use the individual answer checkers to set their options.  E.g.,

``` perl
$ma = MultiAnswer(
	Fraction(1,2)->cmp(studentsMustReduceFractions => 1),
	Real(3)->cmp(showTypeWarnings => 0)
)->with(
	checker  => sub {
		...
	}
);
```

drdrew42:
I actually attempted to do this before, supporting AnswerEvaluators passed into the MultiAnswer constructor, but I couldn't get around the retrieval of the correct answer from the evaluator. 

In your code suggestion, `correct_value` is not a property of the AnswerEvaluator object. The AnswerEvaluator does, however, have a ref to the AnswerHash, so this can be resolved (in a way) -- except it seems that the AnswerHash only stores `correct_ans` as a string. So, in this approach, the author would create a MathObject and store its `cmp` method. Then, when passed to MultiAnswer, we would pull the stringified MathObject to create a new MathObject which is then stored in the `data` array of the MultiAnswer. I think the reparsing that this solution requires could/would cause unexpected things to happen with the display of the correct answer.

It would be nice to supplement to the convenience method by supporting this approach as well, but do you know of a better way for us to recover the original MathObject from the AnswerEvaluator?

dpvc:
> In your code suggestion, `correct_value` is not a property of the AnswerEvaluator object. 

Sorry, you are right, it is on the AnswerEvaluator's answer hash, not the AnswerEvaluator itself (as I mentioned, I hadn't tested the code; I was writing from memory).  

> except it seems that the AnswerHash only stores `correct_ans` as a string.

True, but you don't want `correct_ans`, you want `correct_value`, which is put there by the MathObject answer checkers, and is a reference to the MathObject that created the AnswerEvaluator.  That is what you want.  

You don't want to re-parse the `correct_ans` string, as you will lose any properties that are set on the original MathObject (e.g., a Function object might have limits set, or test points or other values that are not carried by the `correct_ans` string).  So you want to use the `correct_value` that is on the answer hash.

macros/parsers/parserMultiAnswer.pl
```diff
@@ -549,3 +421,163 @@ sub ans_array {
...
+and their corresponding values.
+
+C<$which_rule> begins counting at 1.
+
+If the specified C<$which_rule> does not correspond to an existing comparison object within
+the C<MultiAnswer> instance, this method will throw an error with the message
+"Answer $which_rule is not defined."
+
+    $success = MultiAnswer($fraction_obj)->setCmpFlags(1, studentsMustReduceFractions => 1);
+    $failure = MultiAnswer($fraction_obj)->setCmpFlags(2, studentsMustReduceFractions => 1);
```
drgrice1, line 510: This is a little confusing.  If you add code like this to the POD, authors may think that this might work.  I think you mean that the first example will throw an error message, and the second will not.  But that is not entirely clear.  Also, it seems to insinuate that the `$success` or `$failure` variables are something other than what they are.  The value of the `$success` variable would actually be the answer hash for the `$fraction_obj` answer.  Of course `$failure` would not be set because the `die` would occur.  In any case, the result of the `MultiAnswer` constructor call is never stored in a variable, and so the call would not be particularly useful.

macros/parsers/parserMultiAnswer.pl
```diff
@@ -549,3 +421,163 @@ sub ans_array {
...
+
+This method sets the provided message and does B<not> return early -- allowing an answer checker
+to return a non-zero value for partial credit.
+
+C<$which_rule> begins counting at 1.
+
+If the specified C<$which_rule> does not correspond to an existing answer rule, this method
+will throw an error with the message "Answer $which_rule is not defined."
+
+    $failure = MultiAnswer($math_obj)->setMessage(2, "It's like a jungle sometimes...");
```
drgrice1, line 529: This has the same problem as the example code in the `setCmpFlags` section.

drgrice1 (commented):
I think that this generally looks good.  A couple of suggestions though.

I think it would be good to add the new capability of passing an `AnswerEvaluator` directly to the `MultiAnswer` constructor to the POD documentation.

Also, you removed the section in the POD that describes how the checker works, and that states that a checker is required.  I think that should be added back.  The examples don't show everything that was previously in the explanation, and are not clear on the return values of the checker.

I think a comment about the width of answer rules should still be in the POD as well.  That should be updated for how to do that with PGML answers.

macros/parsers/parserMultiAnswer.pl
```diff
@@ -208,6 +72,15 @@ sub new {
...
 }
 
+#
+#  Set flags to be passed to individual answer checkers
+#
+sub setCmpFlags {
+	my ($self, $cmp_number, %flags) = @_;
+	die "Answer $cmp_number is not defined." unless defined($self->{cmp}[ $cmp_number - 1 ]);
+	$self->{cmp}[ $cmp_number - 1 ]->ans_hash(%flags);
+}
```
dpvc, line 82: In light of the comments by @drgrice1, perhaps this method should return `$self` so that you can chain the calls together.  E.g., 

``` perl
$ma = MultiAnswser("1/2", "sqrt(x-10)")->
         setCmpFlags(1, studentsMustReduceFractions => 1)->
         setCmpFlags(2, limits => [10, 15]);
```

and still have `$ma` be the MultiAnswer object.

drgrice1 (approved):
I think this looks good now.

dpvc (approved):
Looks good.

macros/parsers/parserMultiAnswer.pl
```diff
@@ -549,3 +422,179 @@ sub ans_array {
...
+=head1 DESCRIPTION
+
+The C<MultiAnswer> class is designed to represent MathObjects with interrelated answers.
+It provides functionality to tie several answer rules to a single answer checker, allowing one
+answer to influence another. You can choose to produce either a single result in the answer table
+or a separate result for each rule.
+
+=head1 ATTRIBUTES
+
+Create a new C<MultiAnswer> item by passing a list of answers to the constructor.
```
drgrice1, line 439: I use editorconfig with the setting to remove spaces at ends of lines on save, and just noticed that this line has one.  Perltidy removes spaces at the end of a line in the code, but not in the POD.  Please remove it!

macros/parsers/parserMultiAnswer.pl
```diff
@@ -549,3 +422,179 @@ sub ans_array {
...
+            my ($correct,$student,$multi_ans,$ans_hash) = @_;  # get the parameters
+            my ($f,$x1,$x2) = @{$student};                     # extract the student answers
+            return $f->eval(x=>$x1) == $f->eval(x=>$x2);
+        },
+    );
+    ANS($mp->cmp);
+
+We can also make use of named subroutines. If using C<with> after assigning the C<MultiAnswer> to a
+variable, note that the C<with> method returns a shallow copy of the C<MultiAnswer> object. If you
+do not store the result when calling C<with>, your parameters will not be applied.
```
drgrice1, line 582: This line also has a space at the end.

macros/parsers/parserMultiAnswer.pl
drdrew42: Thanks for the pointer towards .editorconfig 👍

drgrice1:
I will merge this since it has two approvals.

PR #927, opened 2023-09-20
add they're/(s)he's and they've/(s)he's contractions

Alex-Jordan:
I wanted to write something like "how long until they've reached their goal?" using variable pronouns, without unpacking the contraction, so I added these.


pstaabp (approved):
Good addition.

drgrice1 (approved):
Looks good to me.

PR #929, opened 2023-09-27
Fix an issue that makes unanswered CheckboxList answers in a MultiAnswer answered.

drgrice1:
When a CheckboxList is one of the answers for a MultiAnswer object and `allowBlankAnswers` is 1 for the MultiAnswer object, then for some not quite determined reason the cmp_preprocess method of parserCheckboxList.pl is called at an earlier stage than otherwise, and additionally at that point the answer hash `student_value` is a Value::String whose data array contains one empty element.  As such the the answer strings end up begin set to the last checkbox answer label due to the way that the `labelText` method works.

Note that the checker still works correctly, but what is displayed in the results table is misleading.

Essentially the CheckboxList cmp_preprocess is not correctly detecting when no checkboxes are checked.  This skips setting answer strings in the case that no checkboxes are checked.

A minimal working example is attached.  Test this by checking or submitting answers without answering anything.  With the develop branch it will show that the first answer is not answered, but the second is.  With this pull request it will show both parts unanswered.
[ma-checkbox.pg.txt](https://github.com/openwebwork/pg/files/12739084/ma-checkbox.pg.txt)

### DIFF
--- a/macros/parsers/parserCheckboxList.pl
+++ b/macros/parsers/parserCheckboxList.pl
@@ -470,7 +470,7 @@ sub cmp_defaults {
 # Adjust student preview and answer strings to be the actual choice strings rather than the value strings.
 sub cmp_preprocess {
 	my ($self, $ans) = @_;
-	if (defined $ans->{student_value} && @{ $ans->{student_value}->data }) {
+	if (defined $ans->{student_value} && (grep { defined && /\S/ } @{ $ans->{student_value}->data })) {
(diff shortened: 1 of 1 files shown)


drdrew42:
Something is still off here.  An initially empty answer is recognized as such, however:
1. submit no answers, get an empty results table
2. select a checkbox (correct/incorrect doesn't matter) and submit, results table still accurate
3. unselect the checkbox and submit, results table shows selection from step two and scores accordingly.

drdrew42 (approved):
works in ww2, I will work on the renderer

pstaabp (approved):
fixed.

PR #930, opened 2023-09-27
Fix CheckboxList part answers of a RadioMultiAnswer.

drgrice1:
If a CheckboxList is one of the answers in a RadioMultiAnswer part, then the checkboxes are not enabled/disabled correctly as the radio buttons are checked.  This fixes that.

This was not implemented previously because there was no MathObject checkbox answer.  Now there is so this is needed.

An example problem is attached.  In that problem with the develop branch if you select various radio buttons in the problem you will see that only the first checkbox is enabled or disabled when the radio button for its part is selected or unselected.  With this pull request all checkboxes are enabled and disabled appropriately.
[rma-with-checkbox-sub.pg.txt](https://github.com/openwebwork/pg/files/12743017/rma-with-checkbox-sub.pg.txt)


drdrew42 (approved):
Tested 👍

pstaabp (approved):

PR #931, opened 2023-09-29
restore state for contextForm answers

Alex-Jordan:
This is based on:

https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8226


pstaabp:
This PR seems to fix the first issue (with contextForm.pl), but not the Ordering one you later posted.

drgrice1 (approved):
This at least fixes the evaluator state for the changes made here.  Have you tested some of the other things mentioned in the forum post since the PG post processing has been added?  Some of that may be resolved with that because scaffolds no longer process answers in the way they did before.

pstaabp (approved):

PR #932, opened 2023-09-29
add a convenience subroutine for getting a simplified radical expression

Alex-Jordan:
This macro was previously in the PCC section of the OPL macros. This addition is a tool we have used in some problem files. It's simplest on my end to leave it here, but I acknowledge it may be a better fit somewhere else...

### DIFF
--- a/macros/contexts/contextLimitedRadical.pl
+++ b/macros/contexts/contextLimitedRadical.pl
@@ -130,6 +130,25 @@ sub _contextLimitedRadical_init {
 
 ###########################
 #
+# Convenience
+#
+# Pass $a,$b, get Formula("$a sqrt($b)") but simplified
+
+sub preprad {
+	my $a = shift;
+	my $b = shift;
+	if ($a == 0 or $b == 0) { return Formula("0"); }
+	my $sign        = ($a > 0)       ? ''    : '-';
+	my $simplifieda = (abs($a) == 1) ? $sign : $a;
+	if ($b == 1) {
+		return Formula("$a");
+	} else {
+		return Formula("$simplifieda sqrt($b)");
+	}
+}
+
+###########################
+#
 #  Create root(n, x)
 #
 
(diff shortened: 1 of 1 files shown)


macros/contexts/contextLimitedRadical.pl
```diff
@@ -128,6 +128,25 @@ sub _contextLimitedRadical_init {
...
+	my $a = shift;
+	my $b = shift;
+	if ($a == 0 or $b == 0) { return Formula("0"); }
+	my $sign        = ($a > 0)       ? ''    : '-';
+	my $simplifieda = (abs($a) == 1) ? $sign : $a;
+	if ($b == 1) {
+		return Formula("$a");
+	} else {
+		return Formula("$simplifieda sqrt($b)");
+	}
```
drgrice1, line 147: It would be better to unpack `$a` and `$b` on the first line together.

Also, `$sign` and `$simplifieda` are only used in the `else` clause, and so they should be in its scope only.

I don't really see the need for the `$sign` variable that is used only once.  Just use its value where `$sign` is used that one time.

Finally, instead of an `if ... else ...` with a return statement in each one, this could just do `return ... if ...`.

So this could be condensed to
```suggestion
	my ($a, $b) = @_;
	return Formula("0")  if $a == 0 or $b == 0;
	return Formula("$a") if $b == 1;
	my $simplifieda = abs($a) == 1 ? ($a > 0 ? '' : '-') : $a;
	return Formula("$simplifieda sqrt($b)");
```

drgrice1 (commented):
I agree that this would probably fit better somewhere else.  This method could be useful in many situations where the `contextLimitedRadical.pl` is not needed.  I am not sure where exactly to advise you to put this though.

Alex-Jordan:
This is old code that I just submitted as it was. I should have reviewed it for inefficiencies or other ugliness. I'll make changes, probably exactly as you suggest, soon.

Alex-Jordan:
I committed your suggestion plus one minor cleanup. Still not sure where to put this other than here. It returns a `Formula()` so it relies on MathObjects.

pstaabp (approved):
Let's merge this.  I don't think anyone has found a better place for it.

drdrew42 (approved):
Yeah, let's put it in.

PR #933, opened 2023-10-01
Add uncaught exceptions when processings answers to the PG warnings messages.

drgrice1:
Currently these uncaught exceptions are added to a local `%errorTable` hash, and then nothing is ever done with that hash.  This results in the error messages from these exceptions being lost.  Instead you get useless message "The evaluated answer is not an answer hash".

Now the message from such an uncaught exception is added to the the PG warning messages, and the backtrace is added to the PG debug messages if the `view_problem_debugging_info` environment variable is set (this is set for instructors by webwork2).

The `%errorTable` hash has been removed.  The backtrace is cached in a scalar variable instead.  There can only be one uncaught exception, and there is no need to cache the backtrace for exceptions that are caught. Those are handled after all.

On such uncaught exception occurs if if the checker for a MultiAnswer object returns the wrong number of results as in the following example:

```perl
DOCUMENT();
loadMacros('PGstandard.pl', 'PGML.pl', 'parserMultiAnswer.pl');

$a = random(0,  10);
$b = random(11, 20);
$c = random(21, 30);

$ma = MultiAnswer($a, $b, $c)->with(allowBlankAnswers => 1, checker => sub { return (1, 1) });

BEGIN_PGML
Enter [`[$a]`]: [_____]{$ma}

Enter [`[$b]`]: [_____]{$ma}

Enter [`[$c]`]: [_____]{$ma}
END_PGML

ENDDOCUMENT();
```

With the develop branch if you check or submit answers you will get the message "The evaluated answer is not an answer hash", but with this pull request you will get the message "Checker subroutine returned the wrong number of results".  If you are the instructor the debug messages will also be shown with the backtrace.

### DIFF
--- a/lib/WeBWorK/PG/Translator.pm
+++ b/lib/WeBWorK/PG/Translator.pm
@@ -915,38 +915,21 @@ sub process_answers {
 		? @{ $self->{PG_FLAGS_REF}->{ANSWER_ENTRY_ORDER} }
 		: keys %{$rh_correct_answers};
 
-	# Define custom warn/die handlers for answer evaluation. These used to be inside the for loop around the conditional
-	# involving $new_rf_fun, but for efficiency we've moved it out here. This means that the handlers will be active
-	# during the code before and after the actual answer evaluation.
-
+	# Define custom warn/die handlers for answer evaluation.
 	my $outer_sig_warn = $SIG{__WARN__};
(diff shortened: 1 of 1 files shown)


drdrew42 (approved):
Tested 👍 
Glad to get something more useful than "not an Answer Hash"!

pstaabp (approved):

PR #934, opened 2023-10-02
Forgotten amendment to #921.

drgrice1:
This was mentioned by @drdrew42 a couple of weeks ago, and I forgot to do it.  The argument name of the `pretty_print` method was incorrect. Although it doesn't affect the usage it is misleading as to what it does.

Also remove the warning I added in PGUtil.pm that is no longer pertinent with the addition of the two arguments to the method in PG.pl.


drdrew42 (approved):
Good think _you_ remembered, I sure didn't!

pstaabp (approved):

Issue #655, opened 2022-02-11
Compute doesn't handle small real numbers


dlglin:
When creating a MathObject Real from a perl value that is small enough to be displayed in scientific notation, Compute gives the wrong result.

e.g. `$a = Compute(0.0000128);`

The value of $a is then stored as -1.5206.  Looking at the latex string, it is displaying `1.28e-5`, which leads me to believe that Compute is converting 0.0000128 to a string, and then parsing that string and interpreting the e as exp(1).

dpvc:
That is exactly what it does.  Compute is for turning strings into MathObjects, and converts any other perl object to a string first before computing.  You probably want to use

```
$a = Real(0.0000128);
```

Unfortunately it is not easy to tell a number from a string in Perl, and you are not supposed to have to do so in general.  The ways to do it are not straight-forward, and most seem to be introduced in perl 5 (MathObjects predate that).

I suppose one could add check for a numeric value, but that was never the intent for `Compute()`, which was for parsing strings.

pstaabp:
fixed in #935

PR #935, opened 2023-10-03
Check for numbers in Compute

drdrew42:
## Background

Compute is a convenience for parsing MathObjects. A problem arises when Compute is asked to parse a value that rises above or below the threshold for perl to represent in scientific notation. The string for such values contains 'e', which Compute then parses as Euler's number.

**Recent Examples:**
* #655 
* [normal_prob Forum post](https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8361)

### "Fix"

With this PR, passing a valid scientific notation value to Compute() will (by default) create a `Real` MathObject instead of a `Formula`. However, this new default behavior will also come with a warning about the ambiguous input. The warning instructs authors to decide between Real() and Formula() in choosing an explicit MathObject.

In the code, I used Value::matchNumber to identify arguments to `Compute()` as valid strictly-numerical strings. In such cases, the Formula and Real MathObjects are compared and a warning is triggered whenever there is ambiguity. For all arguments identified as numbers, even if ambiguous, the Real version is returned.

This change to the default behavior will only negatively impact problems that have attempted to parse a linear expression in 'e' (with no whitespace) using Compute(). These problems, if they exist, will now see warnings along with the change in the default behavior.

### Testing

```
DOCUMENT();

$x = Compute(6e+20);

BEGIN_PGML

The answer is [$x]: [__]{$x}

END_PGML

ENDDOCUMENT();
```

Before this PR, the correct answer is '6*e+20'.

With this PR, the correct answer is '6*10^20' and you will see the correct answer is also displayed in scientific notation. You will also receive a warning about the ambiguous value. Changing from `Compute()` to `Real()` or `Formula()` will remove the error message and parse the correct answer accordingly.

### Further Consideration

Note that `[__]{6e+20}` will neither create a warning nor have the correct answer value. I don't think we should consider supporting this use case. In fact, I think we should consider triggering a warning any time PGML encounters an answer that is not a MathObject/AnswerEvaluator.

### DIFF
--- a/macros/core/Parser.pl
+++ b/macros/core/Parser.pl
@@ -95,6 +95,14 @@ original can be obtained from the C<original_formula> property.
 sub Compute {
 	my $string  = shift;
 	my $formula = Formula($string);
+	if (Value::matchNumber($string)) {
+		my $real = Real($string);
+		warn "Compute() called with ambiguous value: $string\n"
+			. '-- use Real() for scientific notation'
+			. " ($real) or use Formula() for $formula \n"
+			unless ($real == $formula);
+		return $real;
+	}
 	$formula = $formula->{tree}->Compute if $formula->{tree}{canCompute};
 	my $context = $formula->context;
 	my $flags   = Value::contextSet($context, reduceConstants => 0, reduceConstantFunctions => 0, showExtraParens => 0);


pstaabp (approved):
Looks good.  I like the warning.

Alex-Jordan:
It is curious that the test is based purely on $real not equal to $formula, and yet the message more specifically brings up scientific notation. So there is an assumption that the only way $real would not equal $formula is if it is this issue we have with "e" and scientific notation.

I've never seen it come up, but there are other ways in which some valid number format for perl will be interpreted differently for a MathObject Formula. For example, `Compute(010)` is the same as `Real(8)`, but `Compute("010")` is the same as `Real("10")`. And `Compute(0x10)` is the same as `Real(16)`, but `Compute("0x10")` is the same as `Formula("0x * 10")`.

Would it be right to tweak the message to be less specific, but still appropriate to mention scientific notation as a common culprit? Or maybe these other kinds of things are just never going to happen in practice. I'm just unsure what other things like this are possible.

drdrew42:
The issue is not with the difference in how Compute handles perl values vs. strings -- rather with how Compute parses the _stringification_ of the incoming value.

In your example `010` stringifies to "8", and as such, the Formula and the Real version will match.
The string `"010"` also produces identical results with Formula and Real.

In your example `0x10` stringifies to "16", and again, the Formula and Real version will match.
Here, the string `"0x10"` will not match the `matchNumber` regex, and thus would not be parsed as a Real.

Alex-Jordan:
OK. So in other words, are you sure that any possible difference between $real and $formula stems from this "e" ambiguity? Are stringified perl numbers always either "normal" or in scientific notation, and there's nothing else?

PR #936, opened 2023-10-05
Add PG's own localization.

drgrice1:
This separates the webwork2 and PG localizations.  PG now uses its own translation po and pot files and provides its own language handle for maketext.  Only the language is passed in by the caller.

The translations in the PG code have been transferred from webwork2.

Note that this includes the fixes for using `maketext` inside the safe compartment that were in the now closed https://github.com/openwebwork/webwork2/pull/2213.

That is the `_compile` method of `Locale::Maketext` is overridden to remove the `use strict` usage inside of its code eval call. That is the only essential difference between the override method added here and the original method in `Locale::Maketext`.  With this you can use `maketext` interpolation in PG. For example, you can call `maketext('[quant,_1,dog] ate my homework', $num_dogs)`.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -28,6 +28,7 @@ jobs:
             libhtml-parser-perl \
             libjson-perl \
             libjson-xs-perl \
+            liblocale-maketext-lexicon-perl \
             libtest2-suite-perl \
             libtie-ixhash-perl \
             libuuid-tiny-perl \
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
-conf/pg_config.yml
+*~
+*.swp
 *.bak
+
+conf/pg_config.yml
 cover_db/
 
 htdocs/node_modules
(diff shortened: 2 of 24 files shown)


pstaabp:
I hadn't test the `update-localization-files` before, on this, but testing this now for #940, I'm getting the `getopt` error:
```
./bin/update-localization-files -p
getopt: illegal option -- o
```

I'm wondering if this is some strange MACOS error of `getopt`, even though it appears that is was supposed to be standardized, wikipedia says it can be system dependent. :(

drgrice1:
I assume that you also get that error on the webwork2 develop or main branch then.  So the version of `getopt` on OSX is not fully compatible with the linux version it seems.  I am not sure what to do about that.  In any case, this is not really an issue introduced by this pull request it seems.

pstaabp:
> I assume that you also get that error on the webwork2 develop or main branch then. So the version of `getopt` on OSX is not fully compatible with the linux version it seems. I am not sure what to do about that. In any case, this is not really an issue introduced by this pull request it seems.

Yes.  Same issue there, so not introduced here.  The script still works.

PR #937, opened 2023-10-06
Remove Unnecessary niceTables Warning

drdrew42:
Right now, if the texalignment string for niceTables includes any whitespace, warnings are thrown. Furthermore, the warning message is unclear since the token is whitespace: "Token in texalignment could not be parsed" (seemingly with no specific token referenced).

I see no reason to warn about whitespace in these strings, so this PR prevents the unnecessary warning.

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -1365,7 +1365,7 @@ sub ParseAlignment {
 
 			# could parse these further for color identification, etc
 		} else {
-			main::WARN_MESSAGE("Token $token in texalignment could not be parsed");
+			main::WARN_MESSAGE("Token $token in texalignment could not be parsed") unless ($token =~ /^\s*$/);
 		}
 	}
 


pstaabp:
I'm not seeing warnings.  Here's a simple example:
```
DOCUMENT();

loadMacros("PGstandard.pl","PGML.pl","niceTables.pl");

$tab = DataTable([[1,2],[3,4]], texalignment => ' |c |c ');

BEGIN_PGML
[$tab]*
END_PGML
ENDDOCUMENT();
```

drgrice1:
I think @drdrew42's comments on this are a bit misleading.  The warnings occur when the value of `texalignment` consists entirely of white space (not just includes white space).  You will see the warnings with the following MWE:
```perl
DOCUMENT();

loadMacros('PGstandard.pl', 'PGML.pl', 'niceTables.pl');

BEGIN_PGML
[@ DataTable([ [ 1, 2 ], [ 3, 4 ] ], texalignment => '  ') @]*
END_PGML

ENDDOCUMENT();
```

However, I am not entirely certain that this is a very important issue.  Why are you giving the `texalignment` option consisting entirely of whitespace?

drdrew42:
I was seeing this error with leading whitespace in the align string.

I updated this PR to simply strip whitespace while stripping newlines at the start of `ParseAlignment`

pstaabp:
This looks good.  I'm noticing that someone adds
```
texalignment => ' |c|  |  | c|'
```
this is the same as `|c|||c|`, and the middle bar is thick as documented.

Alex-Jordan:
I don't think we can just kill all spaces. Part of the alignment string could be something like `!{\vrule width 3pt}` and then `!{\vrulewidth3pt}` will not be understood. Perhaps at line 1329, it should not push a token if the token is only whitespace.

pstaabp (approved):
This looks good to me now.  Haven't been able to generate the warning with any whitespace.

PR #940, opened 2023-10-09
Clean up translation strings

pstaabp:
This removes some trailing spaces in maketext strings, so there isn't two additional translated strings. 

Also removed the Latex2HTML from the two MODES calls related.  

This is built on #936 with the `pg.pot` files.


drgrice1 (approved):
This looks good (pending #936).

PR #941, opened 2023-10-10
move Hint and Solution knowls to details/summary disclosure

Alex-Jordan:
This is certainly not ready for a merge. But read on.

This moves "Hint" and "Solution" away from knowls that need the knowl js, and instead uses the HTML5 native disclosure widget that is a `details` element with a `summary` child. The advantage of the move is that something fancy with js is now supported mostly with just HTML and CSS.

For now, there are a bunch of default bootstrap classes that get the styling started.

At minimum, it needs much better styling, and I think @drgrice1's expertise is needed here. The styling may need a level of bootstrap understanding that you have better than I do. Also since the current knows are styled using `pg/htdocs/js/Knowls/knowl.css`, I am wondering if some CSS file inside `pg` is where styling for this should happen (as opposed to in `webwork2/.../math4.scss`). And I'm unsure where that should be and if it would be a new (s)css file, how to introduce that.

There is a chevron next to each "Hint" or "Solution" and it is possible to animate that to rotate it as you open/close the item. And it is possible to have a little js to animate the expansion/retraction of the hidden content. Here is one page demonstrating an approach for both of these: https://christianoliff.com/blog/building-a-bootstrap-accordion-with-the-details/summary-html-tags/

Again since it may involve a new CSS file, and especially if it involves new js, I am wondering if @drgrice1 is interested in handling the cosmetic touches for this. PTX has moved to using details/summary for many of its (former) knowls, but does not use bootstrap.


Alex-Jordan:
I didn't look into it, but perhaps the `knowlLink` subroutine could be rebuilt to use `details/summary` and then maybe we could drop `knowl.js`.

pstaabp:
I think is a start.  A couple of things.  
1. I think it would be nice to animate the opening/closing.  I looked at a few examples using CSS3 animations, so should be possible. 
2. I assume we will extend to other knowls items like the helpLink

Alex-Jordan:
I tweaked some CSS and now in math4, it looks like:

Screenshot 2023-10-25 at 10 34 04 PM

I think it's OK, but there are two problems.
* To do this, the color of the "Hint" and "Solution" relies on the theme. In particular, it uses `var(--bs-primary)` which is set over in the theme. Is it a concern for this PG thing to rely on a webwork2 style?
* While this looks fine for blue, red, and green, it's bad for yellow. I'm not sure how to set the text color differently when the theme is the yellow theme.

macros/core/PGbasicmacros.pl
```diff
@@ -1182,8 +1182,24 @@ sub SOLUTION {
...
-			knowlLink(SOLUTION_HEADING(), value => $solution_body, type => 'solution'), '</div>');
+		TEXT(tag(
+			'div',
+			class => 'solution accordion my-3',
+			tag(
+				'details',
+				class => 'accordion-item',
+				tag(
+					'summary',
+					class => 'accordion-button collapsed p-2',
```
drgrice1, line 1193: I think this looks a lot better if the horizontal margin is not changed.  That makes the alignment of the header text and the body text look off.  I think your primary goal was probably to reduce the vertical padding anyway.  You can do this by using `py-2` instead.  The same for hints as well of course.

htdocs/js/Problem/problem.scss
```diff
@@ -123,6 +123,12 @@
 			border: 1px solid black;
 		}
 	}
+
+	/* Hints and Solutions */
+	.hint summary, .solution summary {
+		font-weight: bold;
+		color: var(--bs-primary);
```
drgrice1, line 130: Using the `--bs-primary` color has contrast issues for the `math4-yellow` theme.  Instead you should use the primary text color.  That will be the same for all of the other themes, but for `math4-yellow` is a darker color.  Unfortunately, there isn't a css variable that contains that.  However, this can still be attained by deleting the `color` setting here, and instead adding the class `text-primary` to either the `summary.accordion-button` element or the `div.accordion-header` element inside that.

You could also add the `fw-bold` class to the same element and then you don't need the addition here, but either way is the same for that.

drgrice1 (commented):
This looks mostly good.  One contrast issue needs to be fixed though.

Alex-Jordan:
I applied the changes from this morning about the font color, font weight, and padding.

drgrice1 (approved):
Looks good.

pstaabp (approved):
Looks great.

PR #947, opened 2023-10-26
Alternate to #946 and probably better (slightly) approach.

drgrice1:
Investigating things closer what I see is that the `Real` call added in #935 is not honoring the current context.  In the problem where the issue is being seen, that context happens to be a variant of the `Fraction-NoDecimals` context.  The `Real` call that was added in #935 is not creating a fraction in that context when it should.

This alternate approach only returns the result of the `Real` call in the case that an ambiguous value is detected.  Otherwise it continues on the the rest of the compute method the way that it did before #935 was added.

### DIFF
--- a/macros/core/Parser.pl
+++ b/macros/core/Parser.pl
@@ -97,11 +97,12 @@ sub Compute {
 	my $formula = Formula($string);
 	if (Value::matchNumber($string)) {
 		my $real = Real($string);
-		warn "Compute() called with ambiguous value: $string\n"
-			. '-- use Real() for scientific notation'
-			. " ($real) or use Formula() for $formula \n"
-			unless ($real == $formula);
-		return $real;
+		unless ($real == $formula) {
+			warn "Compute() called with ambiguous value: $string\n"
+				. '-- use Real() for scientific notation'
+				. " ($real) or use Formula() for $formula \n";
+			return $real;
+		}
 	}
 	$formula = $formula->{tree}->Compute if $formula->{tree}{canCompute};
 	my $context = $formula->context;
--- a/macros/core/Parser.pl
+++ b/macros/core/Parser.pl
@@ -93,17 +93,16 @@ original can be obtained from the C<original_formula> property.
 # ^uses Formula
 # ^uses Value::contextSet
 sub Compute {
-	my $string  = shift;
-	my $formula = Formula($string);
-	if (Value::matchNumber($string)) {
-		my $real = Real($string);
-		unless ($real == $formula) {
-			warn "Compute() called with ambiguous value: $string\n"
-				. '-- use Real() for scientific notation'
-				. " ($real) or use Formula() for $formula \n";
-			return $real;
-		}
+	my $string = shift;
+	if ($string =~ m/^\s*-?(?:\d+(?:\.\d*)?|\.\d+)(?:e[-+]\d+)\s*$/
+		&& Value::matchNumber($string)
+		&& ($string ^ $string) eq '0')
+	{
+		warn "Compute() called with ambiguous value: $string\n"
+			. "-- use Real() for scientific notation or Formula() for Euler's number e\n";
+		$string = uc($string);
 	}
+	my $formula = Formula($string);
 	$formula = $formula->{tree}->Compute if $formula->{tree}{canCompute};
 	my $context = $formula->context;
 	my $flags   = Value::contextSet($context, reduceConstants => 0, reduceConstantFunctions => 0, showExtraParens => 0);


drgrice1:
Here is a minimal working example that shows the issue caused by #935.
```perl
DOCUMENT();

loadMacros('PGstandard.pl', 'PGML.pl', 'contextFraction.pl');

Context('Fraction-NoDecimals');

$a = Compute('5/3');
$b = Compute('3');
$c = $a + $b;

BEGIN_PGML
[`[$a] + [$b] = [$c]`]
END_PGML

ENDDOCUMENT();
```
With the current develop branch (which includes #935) or #946 (which will soon be closed) this problem will show `1.66667 + 3 = 4.66667`.  With this pull request (or the develop branch prior to #935) this problem will show `\frac{5}{3}+3 = \frac{14}{3}` typeset with MathJax.

drgrice1:
I still don't think that this is the full fix.  I think that #935 was not correct, and needs further examination.

@dpvc:  Could you look at what was done there, and see if you can offer a better solution for the issue @drdrew42 was trying to solve?

drgrice1:
You can add the lines
```
[@ '$a is a ' . ref($a) @]*

[@ '$b is a ' . ref($b) @]*

[@ '$c is a ' . ref($c) @]*
```
to the `BEGIN_PGML/END_PGML` section of the above example to see the real problem that is occurring.  `$a` and `$c` should be `context::Fraction::Fraction`'s, and instead they are `context::Fraction::Real`'s with #935.

Alex-Jordan (approved):

dpvc:
To begin with, I can tell you what is happening in your MWE.  Because the new code in `Compute()` uses `Value::matchNumber()`, the result is dependent on the context, because `Value::matchNumber` uses the context's `signedNumber` pattern to determine what matches a number.  In the `Fraction-NoDecimals` context (and the Fraction contexts in general), the signed number includes things that looks like fractions.  So `Compute("5/3")` with have `Value::matchNumber()` being true, and the new code runs.  In the #935 code, `$real` will be returned in this case regardless of whether `$real == $formula`, and in the Fraction contexts, `Real()` will produce a `context::Fraction::Real` due to this line:

https://github.com/openwebwork/pg/blob/daff8afd342f06d25feb87d4c00280980492e318/macros/contexts/contextFraction.pl#L267

A call to `Real()` will never return a `context::Fraction::Fraction`.

In your modification, `$real` is only returned when the error message is given.  Because `$real == $formula` is true in these contexts, the message isn't given and `$real` is not returned, so `Compute()` does what it usually does.

In both the original and this new version, I think the use of `Value::matchNumber()` is ill-advised.  Since you are really only concerned about exponential notation, I'd just test for that directly, rather than looking for the context-specific number format (or you could do the exponential notation test first, and then also check if it passes `Value::matchNumber()`), so something like

```perl
  if ($string =~ m/$\s*-?(?:\d+(?:\.\d*)?|\.\d+)(?:e[-+]\d+)\s*$/) {
    ...
  }
```

or

``` perl
  if ($string =~ m/$\s*-?(?:\d+(?:\.\d*)?|\.\d+)(?:e[-+]\d+)\s*$/ && Value::matchNumber($string)) {
    ...
  }
```

would avoid running the extra code on all the numbers that *don't* look like exponentiation.  

There are some other potential issues with both code blocks.

First, it is possible that `Real($string)` will throw an error (for example, a context might require reals to be integer valued, or to not allow them at all), so it might be prudent to run this code within an error-trapping call (`Parser::Eval()` could be used for that).

Second, the equality check`$real == $formula` could cause an error.  For example, if `e` in this context happened to be a strange object that could not be compared to a real, say some kind of vector where you could add a real, but not compare to a real.  I know, not very likely, but because `Compute()` must be able to be used in **any** context, no matter how strange, we don't want to do things that could produce errors (other than the initial `Formula($string)` call, since we do want to know about parsing errors during that).

There are also a couple of alternative approaches.  For example, one could suggest that if `$r` is a perl real, then one should use

``` perl
Compute(uc($r));
```

rather than just `Compute($r)`.  Of course, that requires the author to know to do that.  On the other hand, something like

``` perl
$a = .000012;
$b = 10;
Compute("${a}x+$b");
```
would end up with the expression `1.2e-5x+10` being parsed, and I don't see how you will be able to detect that and warn the author, so authors do need to be aware of this issue at least at some level.

Another approach might be to use a test to see if the value passed to `Compute()` is actually a number rather than a string.  In general, this is difficult to do, but you could use something like the following:

``` perl
  my $string = shift;
  if ($string =~ m/$\s*-?(?:\d+(?:\.\d*)?|\.\d+)(?:e[-+]\d+)\s*$/ && Value::matchNumber($string)) {
    my $test = ($string ^ $string);
    $test =~ s/./x/g;
    $string = uc($string) if $test eq 'x';
  }
  my $formula = Formula($string);
```

as the beginning of the code for `Compute()`.  Here's what this does:  when $string is actually a number, then `$srting ^ $string` does an xor on that number, and that means all the bits will cancel out, giving a result of 0; but if $string is actually a string, then this does an xor on each character in the string, meaning it a string of null bytes of the same length as the original string.  The `$test =~ s/./x/g;` changes every character to an 'x', so in the case of a number, where `$test` is now 0, that will make it `x`, but in the case of a string, each null byte will become an 'x'.  Since we know the string matches the pattern for exponential notation, we know that for a string value, it will have more than one character (at least a digit followed by `e`, then a plus or minus, and digits following that).  So `$test` will be `x` when a number and several `x` when a string.

(I tried to use `($string ^ $string) == 0` as the test, but the string of nulls seemed to match that as well, so that didn't work as hoped.)

This means `Compute(.000012)` will produce the correct value, as will `Compute(3e+10)`.  But if you have a formula as an actual string, like `Compute("3e+10")`, then this will be treated as `3*e + 10`.  That is, when something that might be considered exponential notation is in a string, it will be parsed as a string version of a formula, but when it is an actual Perl real, it will be treated as a real.

The problem with this is that many people do `Compute("$r")` when `$r` is a perl real.  I don't understand why they do this, as I can see no point in stringifying the real other than to reduce the precision to the default precision for printed output, but that seems unlikely.  But again, one needs to be aware of how reals stringify when inserted into strings, as indicated by the polynomial above, so I can see an argument for this approach.

dpvc:
OK, I figured out how to distinguish the number from the string better:

``` perl
my $string = shift;
  if ($string =~ m/$\s*-?(?:\d+(?:\.\d*)?|\.\d+)(?:e[-+]\d+)\s*$/ && Value::matchNumber($string)) {
    $string = uc($string) if ($string ^ $string) eq "0";
  }
  my $formula = Formula($string);
```

This uses `eq` rather than `==` to force the result of a number 0 to be `"0"` which will not equal a string of null bytes.

drgrice1:
I switched to using the regular expression number check as suggested by @dpvc in this pull request.

The warning is left for now.  Do we really want that?

As noted by @dpvc this will still not detect things like
```perl
$a = .000012;
Compute("${a}x + $b");
```
so authors still need to be aware of this issue for those sort of things.

drdrew42 (approved):

PR #948, opened 2023-10-26
Update the parserWordCompletion.pl macro.

drgrice1:
Use html native data list element for word completion instead of a jquery-ui autocomplete.

The message that was previously shown in a native html dialog if a student entered an answer that is not among the given choices is instead shown as a feedback message in the attempts table.

With these changes no javascript or css is even needed for the macro anymore.  However, there may be accessibility concerns.  See the [accessibility concernts](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist#accessibility_concerns) on the MDN web docs on this.  Although, I doubt that the jquery-ui autocomplete was much better.  In any case, we could try to find another javascript library for this if we want to try to do better.

Attached is a set definition file containing all OPL and Contrib problems that use this macro.  There are 58 problems.  Note there are a few other problems that load the macro, but don't use them.
[setOPL_Word_Completion.def.txt](https://github.com/openwebwork/pg/files/13183641/setOPL_Word_Completion.def.txt)

### DIFF
--- a/macros/parsers/parserWordCompletion.pl
+++ b/macros/parsers/parserWordCompletion.pl
@@ -13,215 +13,193 @@
 # Artistic License for more details.
 ################################################################################
 
-=head1 NAME
+loadMacros('MathObjects.pl');
 
-parserWordCompletion.pl
+sub _parserWordCompletion_init {
+	parser::WordCompletion::Init();
+	return;
+}
 
-=head1 DESCRIPTION
-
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
When I checked this before Safari was rendering these very strangely.  There was an Safari update I installed this week and it looks good now.  Hopefully that fix will stay. 🤷

PR #949, opened 2023-10-27
Update the quickMatrixEntry.pl macro

drgrice1:
This removes the dependence on jQuery and jquery-ui.

This also adds an ans_rule method for PGML usage.  The POD documentation that was added explains only how to use the macro in this way.

There is deprecated code for backwards compatibility, although I would gladly remove that.  The only known problem that uses this macro is in the pg repository.  It is the file `t/matrix_tableau_tests/tableau_javascript.pg`.  I would also like to delete that file.

### DIFF
--- a/htdocs/js/QuickMatrixEntry/quickmatrixentry.js
+++ b/htdocs/js/QuickMatrixEntry/quickmatrixentry.js
@@ -1,0 +1,132 @@
+/* global bootstrap */
+
+'use strict';
+
+(() => {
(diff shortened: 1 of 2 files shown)


drgrice1:
With this pull request and #948, I believe that all jquery-ui usage has now been removed from the webwork2 and pg code.  Perhaps all jquery usage in general?  I am not entirely certain of that though.

drgrice1:
The `liveGraphics.js` file uses jquery.

pstaabp (approved):
works.

PR #950, opened 2023-10-27
allow empty set symbol for empty solution sets

Alex-Jordan:
This lets a student enter ∅ directly if they are trying to say that there are no solutions. For now this is just if they happen to know how to get that symbol into the input field.


drgrice1 (approved):
This looks fine.

We could make some changes to MathQuill so that when this context is the context for an answer then the `\emtpy` latex command is made an autocommand, and a button is added to the toolbar to insert that into the MathQuill input.  MathQuill would need some modification for this, but that is trivial.  Currently the text that is output by MathQuill for `\emtpy` is "varnothing", but that could be changed to be the unicode symbol used here perhaps.

pstaabp (approved):

drdrew42 (approved):

Issue #944, opened 2023-10-24
inserting an answer between two multiAnswer inputs


Alex-Jordan:
Not sure if this is a bug or just bad problem-coding (which it certainly is). But if I put an answer blank in between two answer blanks from a multiAnswer object, the order of the answer blanks gets mixed up. For example:

```
OCUMENT();

loadMacros(
    "PGstandard.pl",
    "PGML.pl",
    "parserMultiAnswer.pl",
);

$m = MultiAnswer("x^2", "x")->with(
    checker => sub {
        my ($c,$s,$self) = @_;
        return ($c->[0] == $s->[0]) && ($c->[1] == $s->[1]);
    }
);

BEGIN_PGML
[_]{$m} should be [`x^2`]

[_]{1} should be [`1`]

[_]{$m} should be [`x`]
END_PGML

ENDDOCUMENT();
```

actually expects you to answer `x` in the second blank, and `1` in the third.

I can think of legitimate reasons to do this, as long as you use `singleResult => 0`. But also one could say that multiAnswer inputs are not supposed to be split up like this, and if you really needed this you could include the `1` among the multiAnswer inputs and just score it separately in the checker.

Should this count as a bug to try and fix?

dpvc:
To do this, you need to use named answer rules.  The MultiAnswer object has a `namedRules` option for that:

https://github.com/openwebwork/pg/blob/daff8afd342f06d25feb87d4c00280980492e318/macros/parsers/parserMultiAnswer.pl#L133-L138

but unfortunately, PGML doesn't deal with that properly.  So you have to use a named rule for the other answer, as in

``` perl
BEGIN_PGML
[_]{$m} should be [`x^2`]

[_]{1}{name => "one"} should be [`1`]

[_]{$m} should be [`x`]
END_PGML
```

Unfortunately, the MultiAnswer results will be shown together in the results table, with the 1 following them, so the order in the results table is confusing.  It might be better to use `singleResult => 1`, which at least makes it clear that the two answers are linked.

[...]

PR #952, opened 2023-10-30
Restructure how answer names are associated with answer evaluators.

drgrice1:
Before describing the restructuring some terminology needs to be clarified.

Henceforth we will no longer refer to any answer as being "unlabeled". All answers are labeled (or better "named" -- more on that later), and always have been.  Every answer rule method calls a named answer rule method, and `ANS` calls `NAMED_ANS` (previously it called `LABELED_ANS`).  Instead answers are either explicitly named (meaning an author chose the name -- hopefully by calling `NEW_ANS_NAME`), or implicitly named (meaning that the association of the name and evaluator is handled by PG by order of ANS and ans_rule calls).

Also, the term "label" should not be used for this. In HTML this is the "name" attribute on the input, and a "label" is a different thing entirely. As such the deprecation flag on the `NAMED_ANS` method has been removed.  In fact, `NAMED_ANS` is no longer an alias for `LABELED_ANS`.  Instead it is the other way around.  So we should encourage using `named_ans_rule` methods together with `NAMED_ANS`.

Now for the restructuring.  Instead of using a fragile pairing of answer name and answer evaluator counts, two FIFO stacks and a name answer hash are used.  The stacks are basically as described by @dpvc in https://github.com/openwebwork/pg/issues/944#issuecomment-1781881663.

When an `ans_rule` method is called is calls `NEW_ANS_NAME` as before. However, `NEW_ANS_NAME` only increments a counter that is only used to make the answer names unique, and does not record the answer anymore. The `ans_rule` method is now responsible for recording the answer after calling `NEW_ANS_NAME`.  It does this by calling
`RECORD_IMPLICIT_ANS_NAME`.  That method pushes the answer name onto the implicit answer name stack.  Then the named answer rule method (which must be called by the answer rule method) records the answer by calling `RECORD_ANS_NAME`.  That actually adds the answer name to the `PG_ANSWERS_HASH`.  Then when `ANS` is called it shifts the first answer name from the stack and associates that answer name to the given answer evaluator by calling `NAMED_ANS`.

The previous paragraph is assuming that the answer rule method is called before the `ANS` method.  If the `ANS` method is called first, then it pushes the answer evaluator onto the other stack which is the implicit evaluator stack.  Then when the `ans_rule` method is called it shifts the first evaluator from the stack and associates it with the given answer name by calling `NAMED_ANS` when the answer is recorded by the named answer rule method.

In the case that an author provides a name (which should be obtained by calling `NEW_ANS_NAME` to obtain a name that will work in all scenarios in which problems are rendered), then these explicit names and associated evaluators are not added to the above stacks.  Of course this is done as before by calling a name answer rule method and `NAMED_ANS` (or its alias `LABELED_ANS`).  This basically circumvents the stacks and makes the association directly.  If the `NAMED_ANS` method is called first, then the given name and evaluator are added to the explicit named evaluator hash.  Then when the named answer rule method is called with the same answer name it removes the evaluator from the hash and adds it to the PG_ANSWERS_HASH.  This ensures that the order of PG_ANSWERS_HASH is the same as the order that the answer rule methods are called.  Thus the order in the (soon to be removed) attempts table will always be correct.  This order will still matter somewhat without the attempts table in some cases as it is the order that the answer evaluators are called in answer processing.

With this pull request I think that all of the examples described in work).  With this it is very easy to intersperse implicitly and explicitly names answers.  If a `MultiAnswer` object is used, its answers can be mixed with other answers easier.  If PGML is used (tweaked by this pull request as discussed in #944), then you don't even need to use the `namedRules` option to do so and PGML handles the details of getting the order correct.  If PGML is not used, you still need to use `namedRules`, but you don't need to name the other answers that are mixed in.

Of course this is completely compatible with existing problems and there is nothing changed for problem authors in terms of the PG answer rule API.

Now it is truly valid to tell authors to use `NEW_ANS_NAME` to get an answer name to use, and it won't interfere with the association of answer names to answer evaluators.  However, care should be taken to call the `NEW_ANS_NAME` method with the answer rule method it will be used for.  This is because the answer names are used to generate the aria labels for answers.  If the `NEW_ANS_NAME` is called before or after another answer rule method then the answer names will be out of order.  You will have something like AnSwEr0002, AnSwEr0001, AnSwEr0003 for order of the names used, and thus the aria labels will be 'answer 2', 'answer 1', 'answer 3' in the order of the page.

There was also POD clean up in PGcore.pm, and some reordering of the methods therein to make the POD headers work better.

There were also several answer rule methods that are not used anywhere in PG or in the OPL that were removed.  For example all of the `tex` answer rules.

### DIFF
--- a/lib/PGcore.pm
+++ b/lib/PGcore.pm
@@ -71,38 +71,35 @@ sub new {
 		OUTPUT_ARRAY      => [],    # holds output body text
 		HEADER_ARRAY      => [],    # holds output for the header text
 		POST_HEADER_ARRAY => [],
-		#		PG_ANSWERS                => [],  # holds answers with labels # deprecated
-		#		PG_UNLABELED_ANSWERS      => [],  # holds unlabeled ans. #deprecated -replaced by PG_ANSWERS_HASH
-		PG_ANSWERS_HASH => {},    # holds label=>answer pairs
+		PG_ANSWERS_HASH   => {},    # holds label=>answer pairs
 
 		# Holds other data, besides answers, which persists during a session and beyond.
-		PERSISTENCE_HASH         => $envir->{PERSISTENCE_HASH} // {},    # Main data, received from DB
-		PERSISTENCE_HASH_UPDATED => {},    # Keys whose updated values should be saved by the DB
-
-		answer_eval_count            => 0,
-		answer_blank_count           => 0,
-		unlabeled_answer_blank_count => 0,
-		unlabeled_answer_eval_count  => 0,
-		KEPT_EXTRA_ANSWERS           => [],
-		ANSWER_PREFIX                => 'AnSwEr',
-		ARRAY_PREFIX                 => 'ArRaY',
-		vec_num                      => 0,                       # for distinguishing matrices
-		QUIZ_PREFIX                  => $envir->{QUIZ_PREFIX},
-		PG_VERSION                   => $ENV{PG_VERSION},
-		PG_ACTIVE                    => 1,                       # toggle to zero to stop processing
-		submittedAnswers             => 0,       # have any answers been submitted? is this the first time this session?
-		PG_session_persistence_hash  => {},      # stores data from one invoction of the session to the next.
-		PG_original_problem_seed     => 0,
-		PG_random_generator          => undef,
-		PG_alias                     => undef,
-		PG_problem_grader            => undef,
-		displayMode                  => undef,
-		content_post_processors      => [],
-		envir                        => $envir,
-		WARNING_messages             => [],
-		DEBUG_messages               => [],
-		names_created                => 0,
-		external_refs                => {},      # record of external references
+		PERSISTENCE_HASH            => $envir->{PERSISTENCE_HASH} // {},    # Main data, received from DB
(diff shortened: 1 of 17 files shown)


lib/PGcore.pm
```diff
@@ -812,67 +744,70 @@ sub tempDirectory {
 	return $self->{tempDirectory};
 }
 
-=head2 surePathToTmpFile
+=head1 Message channels
 
-	$path = surePathToTmpFile($path);
+There are three message channels
```
pstaabp, line 749: Is there are 3rd? or just debug and warning?

pstaabp: Or maybe the 3rd is internal_debug?  Just clarifying the comment.

drgrice1: That is a good question.  That is how that was before.  I just moved it, and that is why it shows it changed.

drgrice1: I changed it to `two` because it seems clear that is correct.

pstaabp (approved):
Fixes the issue with the problem in #944 and works with many problems found in the OPL.

PR #954, opened 2023-10-30
Add display "inline" to the feedback popover.

drgrice1:
When a feedback button is clicked the feedback popover is injected into the DOM in its parent element.  A bootstrap popover is constructed with `div`s.  As such if the input the feedback button is associated with has content after it in the same line, the injected display block `div` causes the line to break.

This adds `display:inline` to the popover to prevent the line break.

Unfortunately, this does result in invalid html, but it works.  The only fix that I can think of is to change the `span`s added around the inputs into `div`s that also have `display:inline` set.  To fix problems that do not use PGML and use `$PAR` for line breaks, the `$PAR` definition would need to be changed from being a `p` tag to being a `div` with a margin (same as PGML does).

I also added a minimum width to the popover.  Since the popover is in the parent element of the feedback button, bootstrap tries to make the width fit into the width of the parent element.  So if that is narrow, then so is the popover.

### DIFF
--- a/htdocs/js/Problem/problem.scss
+++ b/htdocs/js/Problem/problem.scss
@@ -214,7 +214,7 @@
 	--bs-popover-body-padding-y: 0;
 	--bs-popover-max-width: 600px;
 	--bs-popover-zindex: 18;
-	display: inline;
+	position: absolute;
 	min-width: 200px;
 
 	.popover-header {
--- a/htdocs/js/Problem/problem.scss
+++ b/htdocs/js/Problem/problem.scss
@@ -214,6 +214,8 @@
 	--bs-popover-body-padding-y: 0;
 	--bs-popover-max-width: 600px;
 	--bs-popover-zindex: 18;
+	display: inline;
+	min-width: 200px;
 
 	.popover-header {
 		text-align: center;
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -298,8 +298,8 @@ sub NAMED_ANS_RULE {
 		TeX => "{\\answerRule[$name]{$tcol}}",
 		# Note: codeshard is used in the css to identify input elements that come from pg.
 		HTML => tag(
-			'span',
-			class => 'text-nowrap',
+			'div',
+			class => 'text-nowrap d-inline',
 			tag(
 				'input',
 				type           => 'text',
@@ -414,8 +414,8 @@ sub NAMED_ANS_BOX {
 	return MODES(
 		TeX  => qq!\\vskip $height in \\hrulefill\\quad !,
 		HTML => tag(
-			'span',
-			class => 'text-nowrap',
+			'div',
+			class => 'text-nowrap d-inline',
 			tag(
 				'textarea',
 				name       => $name,
@@ -855,8 +855,8 @@ sub NAMED_POP_UP_LIST {
 		|| $displayMode eq 'HTML_tth')
 	{
 		return tag(
-			'span',
-			class => 'text-nowrap',
+			'div',
+			class => 'text-nowrap d-inline',
 			tag(
 				'select',
 				class => 'pg-select',
@@ -1359,7 +1359,7 @@ sub PAR {
 	MODES(
 		TeX        => '\\vskip\\baselineskip ',
 		Latex2HTML => '\\begin{rawhtml}<P>\\end{rawhtml}',
-		HTML       => '<P>',
+		HTML       => '<div style="margin-top:1em"></div>',
 		PTX        => "\n\n"
 	);
 }
--- a/t/pg_problems/problem_file.t
+++ b/t/pg_problems/problem_file.t
@@ -23,11 +23,11 @@ is(
 	qq{<div class="PGML">\n}
 		. qq{Enter a value for <script type="math/tex">\\pi</script>.\n}
 		. qq{<div style="margin-top:1em"></div>\n}
-		. qq{<span class="text-nowrap">}
+		. qq{<div class="text-nowrap d-inline">}
 		. qq{<input aria-label="answer 1 " autocapitalize="off" autocomplete="off" class="codeshard" }
 		. qq{dir="auto" id="AnSwEr0001" name="AnSwEr0001" size="5" spellcheck="false" type="text" value="3.14159">}
 		. qq{<input id="MaThQuIlL_AnSwEr0001" name="MaThQuIlL_AnSwEr0001" type="hidden" value="">}
-		. qq{</span>}
+		. qq{</div>}
 		. qq{<input name="previous_AnSwEr0001" type="hidden" value="3.14159">\n}
 		. qq{</div>\n},
 	'body_text has correct content'


pstaabp:
This doesn't fix it in Safari. :(  Works in chrome. 

Here's what it looks like: 
image

I toggled the first two buttons. I've been playing with html/css to see if I can get something to work.  So far no luck.

drgrice1:
What happens when you open the mathquill tool bar for those inputs?

drgrice1:
I went ahead and made the change from `span` and `p` tags to `div`'s as described in my initial comment on this pull request.  The change is in the last commit and can easily be reverted if needed.  @pstaabp: This should fix the issue for all browsers.

pstaabp:
> What happens when you open the mathquill tool bar for those inputs?

I don't see anything unusual with the toolbar.

pstaabp:
That last commit didn't fix it.   I'll keep playing with it to see if I can figure out what is causing the line break.

drgrice1:
That doesn't make any sense that the MathQuill toolbar does not make line breaks, and particularly that the latest changes do not fix it.

Can you share the code for the problem that you are testing this with?

Are you sure you have all of the javascript updated via `generate-assets.js` (just checking)?

Alex-Jordan (approved):
This worked for me, including with Safari.

Alex-Jordan (commented):
I see the Safari issue too, sorry. So discount my "approval". I'll see if I can remove it.

Alex-Jordan:
Here is a jsfiddle to start with:

https://jsfiddle.net/h29uo6cf/

For me, this works in Firefox and Safari. So it is not a MWE reproducing the issue. How should I do the example differently to make it more like what we have in WeBWorK?

Alex-Jordan:
And sorry to rehash, but I want to circle back to make sure I understand something. The goal is to put the new content from the popover closely following the button in the DOM, so that it reads logically for screenreader users.

So the strategy is to wrap with the new outer `div.text-nowrap.d-inline` so that you can append the popover div as its last child. If I have that right so far, why would you not instead use the `.after` method to insert the popover div as an immediate following-sibling to the button?

drgrice1:
This is closer to what we have: https://jsfiddle.net/drgrice1/y8kctLo9/3/

We don't have access to what bootstrap uses to insert the popover div.  We are only given the `container` option to work with, and bootstrap appends the popover as the last child of the `container`.

drgrice1:
In the fiddle I posted, try commenting out the second to last line, and uncommenting the last line of the javascript click handler.

Alex-Jordan:
In the fiddle as given, no issue with running it in Safari.

> commenting out the second to last line,
Then I see the issue happen in Safari

> and uncommenting the last line
and then it's back to not having the issue

All this is the same in Firefox though as it is in Safari.

On Mon, Nov 6, 2023 at 4:20 PM Glenn Rice ***@***.***> wrote:

> In the fiddle I posted, try commenting out the second to last line, and
> uncommenting the last line of the javascript click handler.
>
> —
> Reply to this email directly, view it on GitHub
> <https://protect2.fireeye.com/v1/url?k=31323334-501d2dca-3132feb7-454455534531-4a698b15894ac67f&q=1&e=bddee190-b562-4de9-8398-332997a7c434&u=https%3A%2F%2Fgithub.com%2Fopenwebwork%2Fpg%2Fpull%2F954%23issuecomment-1797071471>,
> or unsubscribe
> <https://protect2.fireeye.com/v1/url?k=31323334-501d2dca-3132feb7-454455534531-f6c1ed2113b63137&q=1&e=bddee190-b562-4de9-8398-332997a7c434&u=https%3A%2F%2Fgithub.com%2Fnotifications%2Funsubscribe-auth%2FABEDOADKDP32SL42XEPACTTYDF5BBAVCNFSM6AAAAAA6W2LX2KVHI2DSMVQWIX3LMV43OSLTON2WKQ3PNVWWK3TUHMYTOOJXGA3TCNBXGE>
> .
> You are receiving this because you commented.Message ID:
> ***@***.***>
>


-- 
Alex Jordan
Mathematics Instructor
Portland Community College

drgrice1:
In Firefox when I comment out the second to last line and uncomment the last line, there is only a flicker when the button is clicked where things go onto the next lines, but then it all pops back up onto the same line.  Admittedly even that flicker doesn't happen with this pull request in Firefox or Chrome, so it is still probably not the same.

This is really frustrating I must say.  Safari is doing something odd.

drgrice1:
Here is an update fiddle to experiment with: https://jsfiddle.net/drgrice1/y8kctLo9/35.  What does this do in Safari?

That fiddle is much more like the bootstrap popover.  It adds a div that is not display inline, but is position absolute (which the popover is by default).  The mqeditor toolbar also is set up this way, and as I understand it, the toolbar does not cause this line breaking issue.  So the question is, what is being done differently with the bootstrap popover that is causing this line breaking issue on safari, even though the mqeditor toolbar which is setup very similarly does not?

drgrice1:
Test the latest change to this pull request.  I suspect that it fixes the issue on Safari.

Alex-Jordan:
I tried that jsfiddle, and I get no undesirable behavior in Safari.

Will try the latest on the branch now.

On Tue, Nov 7, 2023 at 4:42 AM Glenn Rice ***@***.***> wrote:

> Here is an update fiddle to experiment with:
> https://jsfiddle.net/drgrice1/y8kctLo9/35. What does this do in Safari?
>
> That fiddle is much more like the bootstrap popover. It adds a div that is
> not display inline, but is position absolute (which the popover is by
> default). The mqeditor toolbar also is set up this way, and as I understand
> it, the toolbar does not cause this line breaking issue. So the question
> is, what is being done differently with the bootstrap popover that is
> causing this line breaking issue on safari, even though the mqeditor
> toolbar which is setup very similarly does not?
>
> —
> Reply to this email directly, view it on GitHub
> <https://protect2.fireeye.com/v1/url?k=31323334-501cfaeb-3132feb7-454455535732-7ae3e89a87b7f345&q=1&e=581fbcc5-974a-4d84-80db-51bc39e0270c&u=https%3A%2F%2Fgithub.com%2Fopenwebwork%2Fpg%2Fpull%2F954%23issuecomment-1798425406>,
> or unsubscribe
> <https://protect2.fireeye.com/v1/url?k=31323334-501cfaeb-3132feb7-454455535732-3d9c3da5b9e5ca49&q=1&e=581fbcc5-974a-4d84-80db-51bc39e0270c&u=https%3A%2F%2Fgithub.com%2Fnotifications%2Funsubscribe-auth%2FABEDOAGYYCE3TFWZFZUGQITYDIUBZAVCNFSM6AAAAAA6W2LX2KVHI2DSMVQWIX3LMV43OSLTON2WKQ3PNVWWK3TUHMYTOOJYGQZDKNBQGY>
> .
> You are receiving this because you commented.Message ID:
> ***@***.***>
>


-- 
Alex Jordan
Mathematics Instructor
Portland Community College

Alex-Jordan:
It looks like you solved it. I'm not seeing the undesirable behavior in Safari now. I suggest letting @pstaabp test this too and then we can merge it.

pstaabp (approved):
Works great now.

PR #955, opened 2023-11-01
Update the LiveGraphics3D.pl macro with modern JavaScript.

drgrice1:
This removes the dependence on jQuery and updates the JavaScript and macro.

Unfortunately, the npm installed version of x3dom does not include the x3dom-full.js file, and the live graphics macro uses x3dom elements (like Polyline2D) which are not supported by the basic x3dom.js file. So that library can not be installed via npm. So instead the x3dom files are served from the www.x3dom.org site.  Another option would be to go back to the old school approach of including the files in the repository directly.  Note that the npm package that we were using is made available by a third party. The main thing that you will notice from switching to using the x3dom-full.js file is that now the cube containing the 3D graph will be visible again (that uses the Polyline2D).

Also the version of x3dom is upgraded to 1.8.3.

Also, the underlying html injected by Perl in macro is different. Instead of using a table with a single row and cell, a div is used. That table actually worked quite poorly in many cases, and certainly was not the right thing to use.  Furthermore, there is no injected script element.  Instead the options are passed via data attributes on that div.

### DIFF
--- a/htdocs/js/LiveGraphics/liveGraphics.js
+++ b/htdocs/js/LiveGraphics/liveGraphics.js
@@ -1,913 +1,913 @@
-// liveGraphics.js
-// This is a javascript based replacement for the LiveGraphics3D java library
-// 
-// This program is free software; you can redistribute it and/or modify it under
-// the terms of either: (a) the GNU General Public License as published by the
-// Free Software Foundation; either version 2, or (at your option) any later
-// version, or (b) the "Artistic License" which comes with this package.
-// 
-// This program is distributed in the hope that it will be useful, but WITHOUT
-// ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
-// FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
-// Artistic License for more details.
+'use strict';
 
-var LiveGraphics3D = function (container, options) {
-    var my = this;    
-    
-    // define x3d container and scene
-    var x3d = $("<x3d/>").appendTo(container)
-	.css('width',options.width+'px')
-	.css('height',options.height+'px')
-	.css('border','none')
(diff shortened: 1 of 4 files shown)


drgrice1:
Attached is a set definition file with all problems from the OPL and Contrib that use this macro (some use it indirectly via the other related macros).
[setLive_Graphics_3D.def.txt](https://github.com/openwebwork/pg/files/13232352/setLive_Graphics_3D.def.txt)

pstaabp (approved):
This is working as well as before.  I noticed a few errors in the problems posted, but they are the same problems as in develop.  

It would be nice to replace this with something that looked a little nicer, but it seems to work.

drgrice1:
Yeah, the errors you see are from things that are not correct in the problems themselves.

We already have a nicer thing that replaces this.  The Plotly3D.pl macro added by @somiaj.  It may even be possible to rework this macro to use plotly.js in the future.  Although, at this point I think there are some things that library does not do yet.  For example vector fields which the LiveGraphics3D.pl macro does support. If I recall correctly @somiaj said that plotly does not support those yet.  The main point of this pull request is to remove jquery support and update the macro for things that are not working as it used to.  When we had the x3dom javascript in the wework2 repository before we were using the x3dom-full.js file, and I switched it because I couldn't find that via npm (because it doesn't exist).  I thought that the x3dom.js file I switched to was equivalent, but it isn't.

somiaj:
Although plotly doesn't have a vector field method built in, it seems you can draw arrows by using a cone as an arrow head, and we could implement a vector field plot with plotly.

somiaj (approved):
Confirmed the problems work with the macro updates.

pstaabp:
I'll merge and we should continue investigating a replacement for these.

PR #956, opened 2023-11-05
update PGML horizontal rule for HTML5

Alex-Jordan:
I noticed the horizontal rules in PGML were not behaving according to documentation in two ways. In some cases a `width` argument was not respected. And in all cases a `size` (thickness/height) argument was not respected. The `size` attribute for an `hr` is not valid HTML5, so I changed that to use CSS. I added a `height` key since `size` no longer sounds right, but left `size` there too in case problems declared size with an explicit key.

And for width, if the argument is just a number, that is the case that was not respected. Now in that case it appends `pt` so the width is a valid value for the CSS width.

### DIFF
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -438,7 +438,7 @@ sub Rule {
 	my $token = shift;
 	if ($self->{atLineStart}) {
 ### check for line end or braces
-		$self->Item("rule", $token, { options => [ "width", "size" ] });
+		$self->Item("rule", $token, { options => [ "width", "height", "size" ] });
 		$self->{ignoreNL} = 1;
 	} else {
 		$self->Text($token);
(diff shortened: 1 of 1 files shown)


macros/core/PGML.pl
```diff
@@ -1589,20 +1589,31 @@ sub Quote {
...
+		'div',
+		main::tag(
+			'span',
+			style => "width:$width; display:inline-block; margin:0.3em auto",
+			main::tag(
+				'hr', style => "width:$width; height:$height; background-color:currentColor; margin:0.3em auto;"
+			)
+		)
+	);
 	return $self->nl . $html . "\n";
```
drgrice1, line 1617: Since the `$html` variable is not subsequently modified, there is no need for it.  This could just be:
```suggestion
	return $self->nl
		. main::tag(
			'div',
			main::tag(
				'span',
				style => "width:$width; display:inline-block; margin:0.3em auto",
				main::tag(
					'hr', style => "width:$width; height:$height; background-color:currentColor; margin:0.3em auto;"
				)
			)
		);
```
Note that I removed the `\n` at the end of that.  Historically there has been an attempt to make the generated html human readable.  That is bad practice for modern html.  Instead you should strive for minimized html which decreases download size and renders faster in the browser.  You can always use tools like prettier to make the html nicely formatted for human readability.  I have been removing these newlines from code whenever I make changes to related code.  (Unfortunately Mojolicious templates do not generate anything close to minimized html for webwork2.)

drgrice1 (approved):
Looks good.  I gave one minor code suggestion, but it is not essential.

Alex-Jordan:
I committed that suggested change.

pstaabp:
I just grabbed the examples from https://webwork.maa.org/wiki/Rules_-_PGML

and am getting errors.  Here's some of them:

```
Bareword found where operator expected at line 1 of (eval 3502), near "200px"
from within PGML::Eval called at line 847 of [PG]/macros/core/PGML.pl
from within PGML::Parse::terminateOptions called at line 256 of [PG]/macros/core/PGML.pl
from within PGML::Parse::Terminate called at line 154 of [PG]/macros/core/PGML.pl
from within PGML::Parse::Parse called at line 70 of [PG]/macros/core/PGML.pl
from within PGML::Parse::new called at line 1931 of [PG]/macros/core/PGML.pl
from within PGML::Format called at line 1949 of [PG]/macros/core/PGML.pl
from within PGML::Format2 called at line 4 of undefined

(Missing operator before px?)
```

Seem to get the same errors on develop or this branch. The rules do seem to be taking the parameters now though.

drdrew42:
Confirming @pstaabp's feedback -- bareword where operator expected, (missing operator before px?), unquoted string "px" warnings all occur when using `---{200px}`. This happens in *main* already and was not introduced by this PR.

Alex-Jordan:
I take back what I recently posted. I can do things like:
```
[:<1,2>:]{Vector}
```
without getting a bareword warning. But 
```
----{200px}
```
causes that warning. So I guess there is a PGML issue here, but it is independent from this PR.

dpvc:
PGML evaluates the arguments for its various options at

https://github.com/openwebwork/pg/blob/daff8afd342f06d25feb87d4c00280980492e318/macros/core/PGML.pl#L800

(or at line 790 when the `=>` form is used).  For something like `Vector`, that just returns the string `'Vector'`, but for something like `200px`, I think perl tries to interpret this as a numeric expression because of the leading `200`, then `px` becomes a bare word that causes the error message.

I'm pretty sure that the evaluation of options was added after the initial development of PGML, but the GitHub history doesn't go back far enough (only until July of 2011, but PGML.pl dates back to some time before 2008 (probably around 2006 or 2007).  I suspect the documentation worked initially, but when the evaluation was added, the documentation wasn't updated.

There are a couple of options.  One could make some options not be evaluated, but I don't like the inconsistence of that.  Alternatively, if an error occurs during evaluation, PGML could simply use the origin string and trap the error, but this makes it harder to identify errors that are actually legitimate ones.  Or finally, you could just change the documentation.  I suspect that is the best choice.

Issue #957, opened 2023-11-05
PGML bold within a word


Alex-Jordan:
Documentation in a few places says that in PGML, for emphasis (bold or italics):

```
Stars can be used in*side* a word, but underlines_don't_work_that_way.
```

But in both 2.17 and develop (haven't checked 2.18) the output is `...used in*side* a word...` without boldface. I looked into what is going on but I couldn't figure out how to fix that without simultaneously messing up stars that are used in other ways.

dpvc:
Changing

https://github.com/openwebwork/pg/blob/daff8afd342f06d25feb87d4c00280980492e318/macros/core/PGML.pl#L400

to 

``` perl
       if ($self->nextChar(' ') !~ m/\s/ && ($type eq 'bold' || $self->prevChar(' ') !~ m/[a-z0-9]/i )) {
```

should do the trick.

Alex-Jordan:
That works, thanks.

pstaabp:
Fixed in #958

PR #958, opened 2023-11-05
make PGML bold work mid-word

Alex-Jordan:
This addresses #957.


somiaj (approved):
Confirmed that this allows bold midword, which doesn't work in develop.

pstaabp (approved):
Fixes it.

PR #960, opened 2023-11-08
Add line breaks back to answer messages.

drgrice1:
I missed this in the original conversion from the previous attempts table.  It used to convert new lines to `<br>` tags.  This adds that to the new feedback mechanism.


somiaj (approved):
Confirmed that this replaces newlines with `<br>` so they format properly in message output.

pstaabp (approved):

Issue #961, opened 2023-11-09
Macros whose first line starts with =head1 fail to load.


somiaj:
Loading a macro with `loadMacros` whose first line starts POD, such as `=head1 NAME`, fail to load with errors as the POD delimiter is ignored, and the text is read as code which fails. If a blank line or comment is added to the start of the macro, it loads just fine. This wasn't an issue in 2.18. For example, saving the following as `testMacro.pl` and loading it fails.

```perl
=head1 NAME
This is a test
=cut
$tau = Real("2pi");
```

But putting a comment on the first line, the macro loads just fine.

```perl
# Comment
=head1 NAME
This is a test
=cut
$tau = Real("2pi");
```

drgrice1:
In my opinion this is not an issue.  That is actually what is expected by Perl POD.  Any of the POD directives (lines that start with `=`) should have and empty line before and after them.

This is caused because I removed the newline from the eval call.  You will need to take your pick of what is more important.  Do you want errors and warning reported at the correct line number in the macro file, or do you want this to work?  In my opinion, this behavior is not important (particularly since it is invalid POD in any case).

drgrice1:
Note also that if you run `perltidy` on a Perl file that starts with a POD directive with no empty line before it (like `=head`), then it will add an empty line.  You should run `perltidy` on your macro files anyway.

somiaj:
I'm okay with this not being a real issue due to I'm using invalid POD. I just had multiple macros stop working for me that did work in 2.18 because I didn't realize there had to be a new line and it took me a while to figure out what the actual problem was with the macro files (the errors were strange but didn't actually point me to the issue with the POD).  I may not be the only person who has custom macro files like this that may break in the upgrade from 2.18.

dlglin:
I added a note to the [release notes](https://webwork.maa.org/wiki/Release_notes_for_WeBWorK_2.19).

drgrice1:
Note that with pull request #972 the macro will compile, but a warning will be emitted.

somiaj:
Fixed with #962.

PR #962, opened 2023-11-09
Warn if a macro file has a POD directive on the first line but don't fail.

drgrice1:
This adds a new line to the beginning of the code for a macro file that starts with `=` which will usually mean an invalid POD directive on the first line of the file (and is invalid in any case for the first character of a Perl file).

In that case a warning is also shown.  This warning will only be shown if the view_problem_debugging_info flag is set (this corresponds to the webwork2 permission that ta's and above have).

This is to address issue #961.

### DIFF
--- a/lib/WeBWorK/PG/Translator.pm
+++ b/lib/WeBWorK/PG/Translator.pm
@@ -1266,6 +1266,13 @@ sub PG_macro_file_eval {
 
 	local $SIG{__DIE__} = 'DEFAULT';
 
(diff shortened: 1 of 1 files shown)


somiaj (approved):
This fixes the issue and provides a much better warning of the invalid POD.

This is a nice protection for others who many have invalid POD in macros.

pstaabp (approved):
fixes it and nice warning message.

PR #963, opened 2023-11-13
Round 2 of updates to the LiveGraphics3D macros

drgrice1:
This does three primary things.

First, this switches from using `x3dom` for graphing to using `plotly`.

Second, the parsing of the LiveGraphics3D data has been re-imagined.  The new parsing is more like how the documentation (that I could find) seems to intend it to be done.  At least as I interpret the documentation.   This allows for implementing a few more things from the LiveGraphics3D data that are used by the other LiveGraphics3D macros that were previously not implemented.  More could be done with this as well.

Third, the other LiveGraphics3D macros have been cleaned up, and the POD improved.

Note that the `LiveGraphicsVectorField2D.pl` macro has been updated, but it really doesn't work very well.  It works better than it did before though.  I don't know of any problems that use this macro, so it probably isn't to big of a deal.  To be honest, I don't understand why this macro was created.  It really doesn't make sense to use a 3 dimensional tool such as  LiveGraphics3D to graph 2 dimensional objects.  However, plotly also supports 2 dimensional graphs, and so (with a little javascript ugliness) the macro could be made to work and actually produce a 2 dimensional graph for this macro.

### DIFF
--- a/htdocs/js/LiveGraphics/liveGraphics.js
+++ b/htdocs/js/LiveGraphics/liveGraphics.js
@@ -1,18 +1,6 @@
 'use strict';
 
 (() => {
-	// Convenience method for easily constructing nested x3dom elements with code that clearly reveals the structure.
-	const createX3DElement = (type, attributes = null, ...children) => {
-		const element = document.createElement(type);
-		if (attributes) {
-			for (const [attribute, value] of Object.entries(attributes)) {
-				element.setAttribute(attribute, value);
-			}
-		}
-		element.append(...children);
-		return element;
-	};
-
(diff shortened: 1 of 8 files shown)


drgrice1:
The set definition file with all problems from the OPL that use these macros posted in https://github.com/openwebwork/pg/pull/955#issuecomment-1789785369 is useful for testing this again.

pstaabp (approved):
These look good. 

I also think that a 2D vector field plot doesn't need to be interactive (as well as most 2D plots).  Using plotly makes sense with such plots.  We also have a sample problem to generate a 2D vector field using tikz.

PR #966, opened 2023-11-25
Add PGplot, a method for generating graphs.

somiaj:
Here is something I've been working on for a while, but probably best to start getting some feedback.

First, PGplot is a new method for generating dynamic graphs and is meant to be a replacement of WWPlot and PGgraphmacros.pl.  The first main difference is PGplot is just an API to store data that can then be used to generate a graph using multiple outputs. Right now only TikZ pgfplots and GD (for testing mostly) is supported. I started trying to fully support both, but as I added more features, I focused only on TikZ pgfplots features (so GD only supports simple things and should probably be removed, but leaving it here for now).

In addition there is a new macro PGplotmacros.pl that is meant to be a drop in replacement of PGgraphmacros.pl. This works in most cases provided the problem doesn't call any GD features directly and doesn't use filling (since I couldn't figure out how to get the GD fill method to work with TikZ).  My hope was this could just replace PGgraphmacros.pl, but there are a non trivial number of problems in the OPL that call GD features directly, that this only worked on most of the OPL problems I tested.

Here are a set of examples and testing problems I have been working with to try out. This contains the problems and a few set definition files you can import. [PGplot-examples.tar.gz](https://github.com/openwebwork/pg/files/13462854/PGplot-examples.tar.gz)

I am starting this as a draft, as I suspect there are going to be lots of changes as we think about the PGplot API being used and what sort of features it should support.

### DIFF
--- a/macros/graph/PGplot.pl
+++ b/macros/graph/PGplot.pl
@@ -1,0 +1,766 @@
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2023 The WeBWorK Project, https://github.com/openwebwork
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+# Artistic License for more details.
+################################################################################
+
+=head1 NAME
+
+PGplot.pl - An object to create dynamic graphs to include in PG problems.
+
+=head1 DESCRIPTION
+
+This macro creates a PGplot object that is used to add data of different
+elements of a 2D plot, then draw the plot. The plots can be drawn using different
+formats. Currently the legacy GD graphics format and TikZ (using pgfplots)
+are available.
+
+=head1 USAGE
+
+First create a PGplot object:
+
+    loadMacros('PGplot.pl');
+    $plot = PGplot();
+
+Configure the L<Axes|Axes.pl>:
+
+    $plot->axes->xaxis(
+        min   => 0,
+        max   => 10,
+        ticks => [0, 2, 4, 6, 8, 10],
+        label => '\(t\)',
+    );
+    $plot->axes->yaxis(
+        min   => 0,
+        max   => 500,
+        ticks => [0, 50, 100, 150, 200, 250, 300, 350, 400, 450, 500],
+        label => '\(h(t)\)'
+    );
+    $plot->axes->style(title => 'Height of an object as a function of time.');
+
+Add a function and other objects to the plot.
+
+    $plot->add_function('-16t^2 + 80t + 384', 't', 0, 8, color => blue, width => 3);
+
+Insert the graph into the problem.
+
+    BEGIN_PGML
+    [@ image(insertGraph($plot), width => 500) @]*
+    END_PGML
+
+=head1 PLOT ELEMENTS
+
+A plot consists of multiple L<Data|Data.pl> objects, which define datasets, functions,
+and labels to add to the graph. Data objects should be created though the PGplot object,
+but can be access directly if needed
+
+=head2 DATASETS
+
+The core plot element is a dataset, which is a collection of points and options
+to plot the data. Datasets are added to a plot via C<$plot-E<gt>add_dataset>, and
+can be added individually, or multiple at once as shown:
+
+    # Add a single dataset
+    $plot->add_dataset([$x1, $y1], [$x2, $y2], ..., [$xn, $yn], @options)>
+    # Add multiple datasets with single call
+    $plot->add_dataset(
+        [[$x11, $y11], [$x12, $y12], ..., [$x1n, $y1n], @options1],
+        [[$x21, $y21], [$x22, $y22], ..., [$x2m, $y2m], @options2],
+        ...
+    );
+
+For example, add a red line segment from (2,3) to (5,7):
+
+    $plot->add_dataset([2, 3], [5, 7], color => 'red', width => 2);
+
+Add multiple arrows by setting the C<end_mark> (or C<start_mark>) of the dataset.
+
+    $plot->add_dataset(
+        [[0, 0], [2,3], color => 'green', end_mark => 'arrow'],
+        [[2, 3], [4,-1], color => 'blue', end_mark => 'arrow'],
+        [[0, 0], [4, -1], color => 'red', end_mark => 'arrow'],
+    );
+
+If needed, the C<$plot-E<gt>add_dataset> method returns the L<Data|Data.pl> object
+(or array of Data objects) which can be manipulated directly.
+
+    $data = $plot->add_dataset(...);
+
+=head2 PLOT FUNCTIONS
+
+Functions can be used to generate a dataset to plot. Similar to datasets
+functions can be added individually or multiple at once:
+
+    # Add a single function
+    $plot->add_function($function, $variable, $min, $max, @options)
+    # Add multiple functions
+    $plot->add_function(
+        [$function1, $variable1, $min1, $max1, @options1],
+        [$function2, $variable2, $min2, $max2, @options2],
+        ...
+     );
+
+This method can be used to add both single variable functions and
+parametric functions (an array of two functions) to the graph.
+
+    # Add the function y = x^2 to the plot.
+    $plot->add_function('x^2', 'x', -5, 5);
+    # Add a parametric circle of radius 5 to the plot.
+    $plot->add_function(['5cos(t)', '5sin(t)'], 't', 0, 2*pi);
+
+Functions can be defined using strings (which are turned into MathObjects),
+MathObjects, or perl subroutines:
+
+    # Add a function from a predefined MathObject.
+    $f = Compute("$a x^2 + $b x + $c");
+    $plot->add_function($f, 'x', -5, 5, width => 3);
+    # Define a function using a perl subroutine.
+    # The variable is undefined since it is not used.
+    $plot->add_function(
+        [ sub { return $_[0]**2; }, sub { return $_[0]; } ],
+        undef,
+        -5,
+        5,
+        color => 'green',
+        width => 2
+    );
+
+Functions can also be added using function strings. Function strings are of the form:
+
+    "$function for $variable in <$min,$max> using option1:value1 and option2:value2"
+
+This can be used to add either single variable functions or parametric functions:
+
+    'x^2 for x in [-5,5) using color:red, weight:3 and steps:15'
+    '(5cos(t), 5sin(t)) for t in <2,2pi> using color:blue, weight:2 and steps:20'
+
+The interval end points configure if an open_circle, C<(> or C<)>, closed_circle, C<[> or C<]>,
+arrow, C<{> or C<}>, or no marker, C<E<lt>> or C<E<gt>>, are added to the ends of the plot. Options are
+listed in the form C<option:value> and can be separated by either commas or the word C<and>.
+Multiple functions can be added at once using a list of function strings, which can be useful
+for creating piecewise functions.
+
+    # Add two single variable functions and a parametric function to the graph.
+    $plot->add_function(
+        'x + 2 for x in [-4, 4] using color:blue and weight:3',
+        'x^2 for x in {-4, 4} using color:red and weight:3',
+        '(2cos(t), 2sin(t)) for t in <0, 2pi> using color:green and weight:2'
+    );
+    # Add a piecewise function to the graph.
+    $plot->add_function(
+        '-3-x for x in {-5,-2.5)',
+        'x^2-4 for x in [-2.5,2.5)',
+        '8-2x for x in [2.5,5}'
+    );
+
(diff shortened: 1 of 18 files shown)


somiaj:
Well looks like my method of splittings up in the `macros/` directory is causing a lot of tests to fail, since it cannot find the objects that are needed. I will have to look into that.

drgrice1 (commented):
Here are some initial thoughts on this.

The `PGplotmacros.pl` needs to be renamed to something that does not contain the word "macros".

macros/graph/unionImage.pl
```diff
@@ -66,12 +66,12 @@ sub Image {
...
 	$ilink = $ilink // '';
-	$image = alias(insertGraph($image)) if (ref($image) eq "WWPlot");
-	$image = alias($image) unless ($image =~ m!^(/|https?:)!i);         # see note
+	$image = alias(insertGraph($image)) if (ref($image) eq "WWPlot" || ref($image) eq 'PGplot');
+	$image = alias($image) unless ($image =~ m!^(/|https?:)!i);                                    # see note
 
 	if ($ilink) {
 		$ilink = alias(insertGraph($ilink)) if (ref($ilink) eq "WWPlot");
-		$ilink = alias($ilink) unless ($ilink =~ m!^(/|https?:)!i);         # see note
+		$ilink = alias($ilink) unless ($ilink =~ m!^(/|https?:)!i);                                # see note
```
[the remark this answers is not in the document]
somiaj: Thanks, I missed that location.

t/macros/load_macros.t
```diff
@@ -34,7 +34,17 @@ my %baseMacros = (
...
+# PGplot helper macros Axes.pl, GD.pl, Tikz.pl, and Data.pl are not really broken, but shouldn't be loaded directly.
+# PGplotmacros.pl overrides some default WWplot objects. Loading this causes other test to fail.
+my %brokenMacros = (
+	'answerDiscussion.pl' => 1,
+	'PG_CAPAmacros.pl'    => 1,
+	'Axes.pl'             => 1,
+	'GD.pl'               => 1,
+	'Tikz.pl'             => 1,
+	'Data.pl'             => 1,
+	'PGplotmacros.pl'     => 1,
```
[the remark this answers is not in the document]
somiaj: Would modules in `lib` be preferable over combining `Axes.pl`, etc in a single file? I just feel it would be better to have everything loaded from a macro as needed, but it probably doesn't matter.

I'll push `PGplotmacros.pl` to a second PR. The current way it is setup, it modifies the WWplot methods, causing other macros to fail their tests, and PGplot can be tested without this wrapper.

somiaj:
I removed the `PGplotmacros.pl` wrapper, and will add that to a future PR. Also combined everything into a single macro file, so now this is just a new macro for creating graphs. The primary output at the time is PGFPlots via a LatexImage, though it is meant to be able to support different outputs.

I'm also noticing a difference in perltidy here and in the tests. In the failed perltidy test, the test wants to add an additional tab to the list of or statements I modified in `PGbasicmacros.pl`, but perltidy doesn't want that tab here. I'm running perltidy version `20220613` in Debian bookworm, could there be a change there?

drgrice1:
Rebase onto develop now, and it will fix the perltidy issue.

somiaj:
I still have some difference between my local perltidy and the test results. My guess is the older version, so I'll look into updating that.

drgrice1:
Make sure you specifically use Perl::Tidy version 20220613.  Both newer and older versions will give different results.  This is one of my annoyances with Perl::Tidy.  When they release new versions, the break compatibility with older versions.

somiaj:
That is strange, I am using 20220613. I did perltidy `PGbasicmacros.pl`, and last night when I tried to perltidy that file again, I had no changes. I just tried, and now I have the same changes as the workflow test. Unsure what is going on, but the issue seems to have gone away.

macros/graph/PGplot.pl
```diff
@@ -0,0 +1,1914 @@
...
+	my $dx   = $x2 - $x1;
+	my $dy   = $y2 - $y1;
+	my $len  = sqrt($dx * $dx + $dy * $dy);
+	my $ux   = $dx / $len;                    # Unit vector in direction of arrow.
+	my $uy   = $dy / $len;
+	my $px   = -1 * $uy;                      # Unit vector perpendicular to arrow.
+	my $py   = $ux;
+	my $hbx  = $x2 - 7 * $w * $ux;
+	my $hby  = $y2 - 7 * $w * $uy;
+	my $head = new GD::Polygon;
```
drgrice1, line 1221: This should be
```suggestion
	my $head = GD::Polygon->new;
```

macros/graph/PGplot.pl
```diff
@@ -0,0 +1,1914 @@
...
+
+=item styles
+
+Hash of data for options for the general axis.
+
+=back
+
+=head1 USAGE
+
+The axes object should be accessed through a PGplot object using C<$plot-E<gt>axes>.
```
drgrice1, line 1417: Instead of using the `E<gt>` syntax, make this readable directly with `C<< $plot->axes >>`.  Not the spaces in that are required.   I saw this several other places in this file where this should be fixed as well.

macros/graph/PGplot.pl
```diff
@@ -0,0 +1,1914 @@
...
+			title      => '',
+			grid_color => 'gray',
+			grid_style => 'solid',
+			grid_alpha => 40,
+			show_grid  => 1,
+		},
+		@_
+	};
+
+	bless $self, $class;
```
drgrice1, line 1586: Here again, bless `$self` directly instead of on a separate line.

macros/graph/PGplot.pl
```diff
@@ -0,0 +1,1914 @@
...
+    $data->style(color => 'blue', width => 3);
+
+=back
+
+=cut
+
+package PGplot::Data;
+
+sub new {
+	my $class = shift;
```
drgrice1, line 1805: Unpack the arguments here by adding `%options`, and bless `$self` below directly.

macros/graph/PGplot.pl
```diff
@@ -0,0 +1,1914 @@
...
+	my $style = $styles[0];
+	if (ref($style) eq 'HASH') {
+		map { $self->{styles}{$_} = $style->{$_}; } (keys %$style);
+		return;
+	}
+	return $self->{styles}{$style};
+}
+
+sub set_function {
+	my $self = shift;
```
drgrice1, line 1860: Unpack arguments.

macros/graph/PGplot.pl
```diff
@@ -0,0 +1,1914 @@
...
+sub _add {
+	my ($self, $x, $y) = @_;
+	return unless defined($x) && defined($y);
+	push(@{ $self->{x} }, $x);
+	push(@{ $self->{y} }, $y);
+	return;
+}
+
+sub add {
+	my $self = shift;
```
drgrice1, line 1905: One last place where arugments need to be unpacked.

drgrice1 (commented):
I think this is a good start.  I noted some code issues.

You have several examples in the tar file that you provided that reference a `PGplotmacros.pl` macro which does not exist.  I would like to see that actually work (but not with a `PGplotmacros.pl` file).  That should be the main goal of this.  That is to make it so that the code in `PGgraphmacros.pl` can be eventually replaced with calls to methods here and it will work for existing problems that use that macro.  There is the issue of problems using `GD` objects directly, but I am sure we can find a way to work around that.

I don't like the merge of the previous separate files all into this one file, making it almost 2000 lines long.  Instead the `PGplot`, `PGplot::Tikz`, `PGplot::GD`, and `PGPlot::Axes` should be made actual modules (each a separate file).  Then this macro should have just the basics need to instantiate the object (much like how the `PGtikz.pl` and `PGlateximage.pl` macros work with respect to the `LaTeXImage` package).  The documentation on how to use the macro should stay here for problem authors (again like the above mentioned macros).  Modular code is much better than a single massive file.

somiaj:
@drgrice1 Thanks for the review. @pstaabp has been doing a bit of work on this and has split everything back out and moved this to `lib/`. He is focused on getting this to work with jsxgraph for html, and also have a way to just export JSON from to plot data. I was informed he may add a PR to my branch with the changes, when those come in I'll implement any of these which are still relevant.

somiaj:
I removed the `PGplotmacros.pl` for now because I thought it would be easier to focus on getting this to an acceptable state before dealing with trying to make existing problems work with a wrapper. I was hoping initially to have a drop in replacement, but as I kept running into how many problems were calling GD directly, that I saw this was going to take a lot more work and should probably be dealt with separately. I also think some of my approach, which was working for some problems, was overriding things in an incorrect way, which is what caused it to break all the auto tests.

I do agree it should be a goal that we can just have a drop in replacement and all current problems are updated to using something newer.

somiaj:
I updated this, I removed the change of putting everything in a single file, then added @pstaabp change to move everything to `lib/`, and some code clean up. I also updated all of @drgrice1 suggestions as well.

This move changed the package to just `Plots`, and the command to create a `Plots` object to `Plot()`. Since the `Plots` object is now apart of `lib/`, the following patch https://github.com/somiaj/webwork2/tree/pgplot-libs is needed. I have updated all the examples naming.

[Plots_examples.tar.gz](https://github.com/user-attachments/files/16651201/Plots_examples.tar.gz)

In updating this I did notice an issue with `pdf2svg` and `xelatex` set as the default. It seems that `xelatex` isn't fully compatible with `pgfplots`, though switching this to use `dvisvgm` by default works just fine (since that calls `latex` directly). The issue I noticed was with the marks test (Problem 8 of the Test set), the rest worked just fine. I think one of the marks uses a symbol that conflicts with `xelatex`. I am just documenting this issue for now, the default setup using `dvisvgm` works properly.

somiaj:
Add the very basics of JSXGraph and Plotly.js output support. Examples updated.

[Plots_examples.tar.gz](https://github.com/user-attachments/files/17691181/Plots_examples.tar.gz)

somiaj:
Thinking about adding JSXGraph and Plotly.js output method a bit more, I think this approach is not the best. Plots is mostly designed around creating non dynamic graphs, so at best adding javascript output could do is allow some moving around/zooming in/out of the graph in some situations.

I think a better approach would create a JSXGraph and Plotly.js plots object that is a subclass, built on the same `Plots::Data` objects, so it can fall back to using TikZ (or some other non dynamic output) for hard copy, instead of try to make Plots do both non-dynamic and dynamic graphs.

drgrice1:
JSXGraph does a great job of rendering static graphs.  I use it for that purpose in problems via a local macro that I have.

somiaj:
I was more thinking about also wanting to be able to also use it for dynamic graphs from the same macro. And unsure if adding the dynamic side of things should be done in Plots, or better to just have a separate macro for that. But so far it is working well with the setup I have here.

drgrice1:
Yeah, I think that this is really for static graphs (dynamically generated but not interactive).  I mean you could have dynamic in the sense of pan and zoom I suppose (both Plotly and JSXGraph support that).

somiaj:
Agreed, so my thought was Plotly and JSXGraph could be separate macros that build on top of this, so they can create static and dynamic graphs (as opposed to what I'm working on now which is just making it an optional output method). Though I guess they can split off later.

somiaj:
Updated everything and rebased against `PG-2.20`. I ditched the idea of support poltly at the time, and have mostly worked on JSXgraph, with the idea that JSXgraph should be the default HTML output and TikZ the hardcopy (though I do like TikZ graphs, I think they generate much slower).

Updated sample problems, think I have JSXgraph mostly in line with TikZ. TikZ does support some things JSXgraph doesn't though the only real thing that I didn't find a good solution was filling area between curves, seems this isn't default in JSXgraph, also if the time comes to try to replace PGplotmacros with this, being able to fill a region given a point is also needed.

[Plots_examples.tar.gz](https://github.com/user-attachments/files/20560271/Plots_examples.tar.gz)

macros/graph/plots.pl
```diff
@@ -0,0 +1,419 @@
...
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See either the GNU General Public License or the
+# Artistic License for more details.
+################################################################################
+
```
pstaabp, line 15: We've gotten rid of this header material.

somiaj: Thanks. It is now gone.

somiaj:
Vector field support added. Examples updated.

[Plots_examples.tar.gz](https://github.com/user-attachments/files/20649653/Plots_examples.tar.gz)

drgrice1:
@somiaj: I added a pull request with the background/border for TikZ images that I mentioned.

somiaj:
@drgrice1 thanks. If you get a chance, can you set it up so a modal can be used to zoom in on the JSXGraph output if the option `$plot->axes->style('jsx_navigation')` is not set?

drgrice1:
I added a pull request that cleans up the `addToPreamble` call.  That was quite unreadable.

drgrice1:
I added another pull request with the `imageview` dialog when the `jsx_navigation` option is false.  There are also some other changes that are needed.

drgrice1:
I added another pull request to this branch.  I figured out what was really going on with the image in the dialog and fixed it.

somiaj:
Grr, I thought I was on my branch with your new merge....should I revert it or do a force push to revert?

PR #967, opened 2023-11-26
avoid an error from misuse of 'AnSwEr'

Alex-Jordan:
Line 557 assumes that if `AnSwEr` is a substring of the answer name, that it is followed by some digits. But the test at line 546 does not look for digits following `AnSwEr`. So if for some reason a problem author manually creates an answer name like `MyAnSwEr`, then there will be an error message.

This just changes the check at line 546 to look for some trailing digits. If they are not there, the subroutine exits early.

No one should be manually creating answers that have `AnSwEr`  as a substring. But sometimes problem authors who are newbies and looking at code somewhere to mimic might end up doing this. This PR addresses one issue if that happens. I note there would be more issues, since other things here assume that the presence of `AnSwEr` means it was an auto-generated answer name. But that's a different matter.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -543,7 +543,7 @@ sub generate_aria_label {
 	my $label = '';
 
 	# if we dont have an AnSwEr type name then we do the best we can
-	if ($name !~ /AnSwEr/) {
+	if ($name !~ /AnSwEr\d+/) {
 		return maketext('answer') . ' ' . $name;
 	}
 


drgrice1 (approved):
Clearly line 546 conflicts with line 557, and improves things some.

The check fro the quiz prefix on line 551 is also not reliable.  The quiz prefix is not always the first prefix on the answer name, so that check will miss many cases.

I think a more robust method for generating aria labels is needed.  A method that is not based solely on the answer name.

Alex-Jordan:
Yes, I think I typed a comment in your recent PR that redid answer names about this, but did not hit send. We are inferring this and that from the answer name, but it would be better if it were just explicitly made available as data when the answer name is created. I started in that direction but reverted back to this little change for now.

drdrew42 (approved):

PR #968, opened 2023-11-28
Remove `be_strict`.

drgrice1:
All modules should call `use strict` and certainly have no need of this method.  They are not restricted from using `use` as they are outside of the safe zone.  They should also all call `use warnings`.  More work is needed because many of the modules actually have problems that cause warnings.

Macros can (and should) call `BEGIN { strict->import }`. This is actually entirely equivalent to previously calling `BEGIN { be_strict() }` (due to an incorrect `import` call -- see below) and can be done in the safe compartment. Note that `use strict` is (essentially) the equivalent of `BEGIN { require strict; strict->import; }`.  For macros there is no need to `require strict` as `strict` will already have been loaded. So using `BEGIN { strict->import }` does essentially what `use strict` does for a macro.

This does not go as far as is really needed.  All macros should call `BEGIN { strict->import; }`.  This should be a requirement for macros in PG going forward.  This is added in `parserGraphTool.pl` in this pull request, and the one issue with that macro is fixed. This is deferred for other macros.

Also remove the unused `ww_strict.pm` module.  Note that module was not actually used even by the `be_strict` method.  The `be_strict` method was defined by

```perl
sub be_strict {
	require ww_strict;
	strict::import();
	return;
}

```

This means that the `ww_strict` module was required, but the line `strict::import()` did not import that module.  Instead that imports the real perl `strict` module. In order to import the `ww_strict` module it would have needed to call `ww_strict->import`. Notice that the `be_strict` method also used an incorrect way of calling the `import` method of `strict`. The `import` method is a package method that should be called with an invocant (via `strict->import`, and not a static method which would be called as `strict::import()` as was done in the `be_strict` method.  It worked though because the `import` method doesn't use the invocant, and the `be_strict` method did not accept any arguments.


drdrew42 (approved):
I looked through a few problems to test this out -- things look good 👍

pstaabp (approved):
This looks good.  Tested a number of problems and work as expected.  Requiring `use strict` (and `use warnings`) for modules as well as `strict->import;` for macros will be good moving forward.

Alex-Jordan:
It seems to come from the `BEGIN { strict->import; }` in the macro files.

drgrice1:
I assume that you have restarted the webwork2 app, right?

Is this happening on the problem page?

What version of perl are you using?

Alex-Jordan:
I have restarted the app. I see this on the problem's page as well as the editor page.

This is on an Oracle 8.7 server, with perl 5.26.3. My production server is the same OS, but perl is 5.36.1. Seems like the perl version is the issue. If it's for sure that we don't want to support this old of perl, I will try to upgrade perl on my dev server.

PR #970, opened 2023-11-29
Update third party JavaScript dependencies.

drgrice1:
The following dependencies were updated.

* plotly.js-dist-min from 2.18.2 to 2.27.0
* autoprefixer from 10.4.13 to 10.4.16
* cssnano from 6.0.0 to 6.0.1
* rtlcss from 4.0.0 to 4.1.1
* sass from 1.57.1 to 1.69.5
* terser from 5.16.1 to 5.24.0
* yargs from 17.6.2 to 17.7.2
* jsxgraph from 1.5.0 to 1.6.2

Most versions of these libraries are compatible with pg's current usage and require no changes to the code.

The exception is jsxgraph.  Most of the the changes are compatible with the current usage.  However, the fullscreen mode has changed in a way that breaks the graphtool usage.  So now the graphtool javsacript handles fullscreen mode using its own code.  This is better than it was before in any case. There were several issues that are resolved with this.  Particularly resize issues related to screen rotation on mobile devices, and enabling or disabling help after switching to fullscreen.

Now all JavaScript dependencies are at their newest versions.


pstaabp (approved):
looks good.  Checked a few graph tool problems in full screen mode.

PR #972, opened 2023-12-01
Change the color of the message dot on the feedback button to yellow.

drgrice1:
This is as discussed in #939 (post merge).


somiaj (approved):
I agree with the yellow choice, as matching the background of the messages text gives a color relation between them.

Alex-Jordan (approved):

Issue #872, opened 2023-07-16
Bug with AnswerHash.pm


Alex-Jordan:
I first posted about this [in the forums](https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8309).

The following MWE illustrates a bug. If you enter the correct answer, it is not accepted and there is a perl warning "The evaluated answer is not an answer hash : ||."

This came to my attention from an instructor who previously used certain problems in 2.16, and they worked. But now they do not work on 2.17. And this MWE does not work on 2.18. So it is possibly a bug introduced between 2.16 and 2.17.

Note that if you change $ans to a Real, there is no issue. I also tried making $ans an Interval and there was no issue. It seems to have to do with $ans being a NumberWithUnits. Also there is no issue if I remove the withPostFilter or pass an empty hash to AnswerHints. So it also seems to have to do with AnswerHints.

```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl parserNumberWithUnits.pl answerHints.pl));
$ans = NumberWithUnits("1 m");
#$ans = Real("1");   # This works.
$cmp = $ans->cmp()->withPostFilter(AnswerHints(2 => "no"));

BEGIN_PGML
Enter [`[$ans]`].

[_]{$cmp}
END_PGML
ENDDOCUMENT();
```

I have done some sleuthing. First, it does not help to revert the two macro libraries here to earlier (say 2.16) versions. The change in behavior is something deeper.  Things go bad inside `lib/AnswerHash.pm`. There is this loop, where I have inserted `warn` statements while debugging
```
        foreach my $i (@post_filters) {
                last if defined($rh_ans->{done}) and $rh_ans->{done} == 1;    # no further action needed
                my @array  = @$i;
                my $filter = shift(@array);    # the array now contains the options for the filter
                warn('A');
                $rh_ans = &$filter($rh_ans, @array);
                warn('B');
                $self->print_result_if_debug('post_filter', $rh_ans, @array);
        }
```

Running the MWE, I get an "A" without a "B". Something about trying to access `&$filter` (in the particular case when it is an AnswerHints filter involving a NumberWithUnits as in the MWE) terminates not only this loop, but also the ambient subroutine. That's what leads to the error message; the ambient subroutine then returns nothing at all, which is "not an answer hash".

I can even change `warn('A');` to `warn('A' . &$filter);`, and then I don't even get an "A". This suggests it's not so much something about executing that filter code, but just accessing it at all.

[...]

Issue #964, opened 2023-11-16
answer hints in multiple choice


Alex-Jordan:
In the MWE below, there a dropdown question. Each answer looks like a number to the student, but one is actually a string because of a space character. It is coded so that each answer should invoke a feedback message from AnswerHints. But only the one answer that is not actually a number gets the feedback message to show. (Same behavior in 2.18 and develop.)

```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl answerHints.pl parserPopUp.pl));

$mc = DropDown(['1.5', '2 ',3], 2); # correct answer is 3

BEGIN_PGML

[_]{$mc->cmp->withPostFilter(AnswerHints(
    '1.5' => "You chose 1.5",
    '2 ' => "You chose 2 with a space",
    3 => "You chose 3"
))}

END_PGML
ENDDOCUMENT();
```

I additionally tried the following, but only tried it in develop:
* I tried a version of this with RadioButtons, and I can't get any of the feedback messages to appear, not even the one for '2 '.
* I tried a version of this with CheckboxList, and I can't get any of the feedback messages to appear, not even the one for '2 ', but I'm not actually sure what the key should be
* I tried a version of this with a Real() instead of a DropDown(), and I did not get the corresponding issue to happen. I can trigger each of the feedback messages in that situation. So it seems this may be specific to the multiple choice parsers.

dpvc:
There are several things that are contributing to your experience.  First, by default, AnswerHints only checks answers when they match the type of the correct answer.  In your case, since that is 3, it only checks those hints that are numbers, so the first two checks are skipped regardless of the student answer.  But again by default, AnswerHints also doesn't check answers when the student answer is marked as correct, so it never checks when the student answer is 3.  That is why you never get an answer hint in this case.

You would need to use

``` perl
[_]{$mc->cmp->withPostFilter(AnswerHints(
    '1.5' => ["You chose 1.5", checkTypes => 0],
    '2 ' => ["You chose 2 with a space", checkTypes => 0]
    3 => ["You chose 3", checkCorrect => 1]
))}
```

in order to get the answers hints to work.

On the other hand, if you used 

``` perl
$mc = DropDown(['1.5', '2 ', '3'], 2);
```

so that the correct answer is a string rather than a number, then you only need to do

``` perl
[_]{$mc->cmp->withPostFilter(AnswerHints(
    '1.5' => "You chose 1.5",
    '2 ' => "You chose 2 with a space",
[...]

PR #974, opened 2023-12-01
fix bugs with answerHints from issue #964

Alex-Jordan:
This fixes the issues discussed in #964.

### DIFF
--- a/lib/Value.pm
+++ b/lib/Value.pm
@@ -268,12 +268,11 @@ sub inContext { my $self = shift; $self->context(@_); $self }
 #############################################################
 
 #
-#
 #  The address of a Value object (actually ANY perl value).
 #  Use this to compare two objects to see of they are
 #  the same object (avoids automatic stringification).
 #
-sub address { oct(sprintf("0x%p", shift)) }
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
Thanks @dpvc. This now also addresses #872.

Alex-Jordan:
Each issue has a MWE problem to test. For me, they both work following these changes.

pstaabp:
This fixes #872, however I'm not seeing answer hints show up at all for the various MWEs.  Reading through #964 we need to implement the translation between the answer and the internal value like @dpvc pointed out, but that doesn't work for me either.

Alex-Jordan:
Do I understand right that this can be merged, and #964 can remain closed, but #872 should be reopened but with extended comments? I think we ultimately want something where you do not have to insert that postFilter, and so #872's original MWE remains an issue although not quite for the same reasons. Or maybe #872 remains closed but I open a new issue.

drgrice1:
This can be merged.  If you want to re-open #872 or create another issue regarding creating an easier way to make answer hints work with radio buttons or checkboxes, that would be good.

pstaabp (approved):

PR #975, opened 2023-12-02
move feedback message to top of feedback card

Alex-Jordan:
This moves the feedback message to the top of the card, as discussed in the #939 thread.

I'm not sure about it yet. Additionally styling adjustments might be in order.

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -1022,7 +1022,7 @@ sub ENDDOCUMENT {
 
 				my %options = (
 					resultTitle      => maketext('Preview'),
-					resultClass      => 'preview',
+					resultClass      => '',
 					btnClass         => 'btn-info',
 					btnAddClass      => 'ms-1',
 					feedbackElements => Mojo::Collection->new,
@@ -1126,10 +1126,11 @@ sub ENDDOCUMENT {
 						if $options{insertElement} && $options{insertElement}->attr->{'data-feedback-insert-method'};
 				}
 
-				# Add the preview/correct/incorrect/partially-correct class and
+				# Add the correct/incorrect/partially-correct class and
 				# aria-described by attribute to the feedback elements.
 				for (@{ $options{feedbackElements} }) {
-					$_->attr(class              => join(' ', $options{resultClass}, $_->attr->{class} || ()));
+					$_->attr(class => join(' ', $options{resultClass}, $_->attr->{class} || ()))
+						if $options{resultClass};
 					$_->attr('aria-describedby' => "ww-feedback-$answerLabel");
 				}
 
@@ -1195,7 +1196,7 @@ sub ENDDOCUMENT {
 							bs_trigger             => 'click',
 							bs_placement           => 'bottom',
 							bs_html                => 'true',
-							bs_custom_class        => join(' ', 'ww-feedback-popover', $options{resultClass}),
+							bs_custom_class        => join(' ', 'ww-feedback-popover', $options{resultClass} || ()),
 							bs_fallback_placements => '[]',
 							bs_content             => Mojo::DOM->new_tag(
 								'div',
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -827,8 +827,8 @@ changes this to "Ungraded" for essay answers.
 =item *
 
 C<resultClass>: This is the CSS class that is added to each answer input in the
-response group. By default it is set to "preview", "correct", "incorrect",
-or "partially-correct" depending on the status of the answer and
+response group. By default it is set to the empty string, "correct",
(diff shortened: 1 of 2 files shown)


drgrice1 (approved):
This looks good to me.  What additional styling adjustments do you thing are needed?  I don't see anything that looks wrong to me.

Alex-Jordan:
I can't put my finger on it. The transition from the message to the h4 below it sometimes seems off to me, depending on the size of the message. And maybe it matters that the headers below (h4) feel heavier than the h3 at the top of the card.

Maybe it's how the yellow message background color has 1:1 contrast ratio with the gray color of the h4 below it. That might explain what I see with the transition from the message to the header below it.

Anyway, I'm not really sure. I'm fine with just leaving it, while still wondering if others see what I see and have ideas.

drgrice1:
None of it looks that off to me, but would it help if there were a black border on the bottom of the feedback message?  Perhaps also on the top of the feedback message?  Or perhaps just a border all around the feedback message?

Alex-Jordan:
Yeah, a border on the bottom of the feedback really helps.

Is it better implemented as a CSS border or an `hr`? I'm leaning toward an `hr` because I can simply exclude it when there is no message, or when there is a message but there is nothing further below the message.

drgrice1:
Also, it isn't hard to implement a border on the bottom with css that is only there if the message is not the last child.

drgrice1:
Looks good.

I noticed another issue though (not caused by this pull request).  I wanted to test the case that there is nothing below the message.  So I used the following problem:
```
DOCUMENT();
loadMacros('PGstandard.pl', 'PGML.pl', 'parserMultiAnswer.pl',);

$ma = MultiAnswer(0)->with(
	singleResult      => 1,
	allowBlankAnswers => 1,
	checker           => sub {
		my ($cor, $stu, $self, $ansHash) = @_;
		$self->setMessage(1, 'message that is always shown');
		return 1;
	}
);

BEGIN_PGML
[__________]{$ma}
END_PGML
ENDDOCUMENT();
```

I noticed that the feedback message card body does not get the border radius of the popover, and so the background color doesn't look the best.

drgrice1:
No, it is just the background color, and it is not just in this pull request.  It is how it already is.  If you add

```
					&:last-child {
						border-bottom-left-radius: 0.5rem;
						border-bottom-right-radius: 0.5rem;
					}
```

To the feedback message it is a little better.  That is the border radius of the popover, but it still doesn't quite match up right.

drgrice1:
To see this check answers with that problem without entering an answer.

Alex-Jordan:
I had to look real hard to see that. I think my red/green color blindness blurs the yellow and gray together a bit and I have to get pretty close to the screen to see the radius issue. That blurring of yellow and gray is probably what was bothering me before the black border too.

drgrice1:
Hmm, I am not sure.  That is a problem introduced with this pull request.  Maybe we just need to make sure there is always a header.  So for a preview, perhaps make the title "Preview Only".

Alex-Jordan:
I fixed the radius, I think. It seemed that I needed to use `calc(0.375rem - 1px)` instead of `0.5rem`.

I added a header for when the popover is just preview.

And I noticed the cursor over the main header was only turning to a pointer when the feedback was "correct". I changed it to always be a pointer on the feedback card header, regardless of feedback type.

Alex-Jordan:
> I fixed the radius, I think. It seemed that I needed to use calc(0.375rem - 1px) instead of 0.5rem.

OK, I think I figured out it should be `var(--bs-card-inner-border-radius)` and I changed it to that, with a force push.

Alex-Jordan:
OK, I think I addressed that now.

This is unrelated to this PR, but ideally an essay answer title would not always be "Ungraded" and the message would not always be "This answer will be graded at a later time." Ideally it would recognize that it's been assigned a score since the last time an answer was submitted, and it would give responses based on that.

pstaabp (approved):
This looks nice.   I like the message at the top.  Also, tested an essay question and I think the last commit fixed the error @drgrice1 pointed out.

drdrew42 (approved):
This is looking really nice -- I much prefer having the feedback message at the top 👍

macros/PG.pl
```diff
@@ -827,8 +827,8 @@ =head2 ENDDOCUMENT
 =item *
 
 C<resultClass>: This is the CSS class that is added to each answer input in the
-response group. By default it is set to the empty string, "correct",
-"incorrect", or "partially-correct" depending on the status of the answer and
+response group. By default it is set to "preview", "correct", "incorrect",
```
drgrice1, line 830: This is now set to the empty string by default again.

Issue #886, opened 2023-07-26
Currency errors with hardcopy


pstaabp:
The following:
```
DOCUMENT();
loadMacros("PGstandard.pl","PGML.pl",'contextCurrency.pl');
Context('Currency');

$c = Currency('1000'); 

BEGIN_PGML
The value is [$c].
END_PGML

ENDDOCUMENT();
```

Generates `The value is $1,000.00` as expected in HTML, however in hardcopy format, there is a backslash in front. 

<img width="236" alt="image" src="https://github.com/openwebwork/pg/assets/879338/ece970a6-acdb-482d-8f27-08aa1bfdfa8a">

drgrice1:
Use ```[`[$c]`]``` instead.  This works for both HTML and hardcopy.

Alex-Jordan:
It's not intuitive, but I think this is expected behavior.

`[$c]` gives the string interpolation for `$c`, which is `\$1,000.00`. Then PGML has `The value is \$1,000.00.` and escapes the characters it needs to escape for the output mode. So for hardcopy, PGML turns the `\` into `{\ttfamily\char92}` (a backslash) and the `$` into `\$`. So you get `The value is {\ttfamily\char92}\$1,000.00.` Which you see in the PDF as `\$1,000.00`.

You could ask, should the string interpolation for `$c` be `$1,000.00` instead? But then when you are not using PGML, that would break hardcopy production.

I think the "right" thing to do is use `[$c]*` to tell PGML "don't escape anything". Unless you think it's good to use math mode in that context, and then use @drgrice1 's suggestion ``[`[$c]`]``. Of course neither is intuitively necessary to an author, but it seems like it is correct behavior.

Meanwhile with `[$c]` for HTML output, the backslash in `\$1,000.00` is not "pre-escaped" to something like `\char92`. And it escapes the dollar sign to just a dollar sign. But then we escape that now and the output is `<span class="tex2jax_ignore">$</span>1,000.00.` which comes out looking fine. 

Or with `[$c]*` for HTML output, you still get `$1,000.00` but now you are saying not to wrap the dollar sign in a span. So that comes out fine too, since we do not use `$` as a MathJax delimiter. (This raises the question for me though why are we wrapping `$` in that span?)

My point is `[$c]*` works in both output forms.

I think this is not actually an issue, unless you have a way for something's string method to be different in a PGML context than it is in a non-PGML context. But it is a messy in-the-weeds PGML thing.

[...]

PR #977, opened 2023-12-05
don't escape dollar for string method in PTX mode

Alex-Jordan:
This is a minor thing I noticed. For `PTX` output, as long as we are not in math mode, the dollar sign should not be escaped. So the `string` method here should not escape the dollar sign for PTX output.

### DIFF
--- a/macros/contexts/contextCurrency.pl
+++ b/macros/contexts/contextCurrency.pl
@@ -250,7 +250,7 @@ sub new {
 			precedence    => 10,
 			associativity => $associativity,
 			type          => "unary",
-			string        => (($main::displayMode eq 'TeX') ? Currency::quoteTeX($symbol) : $symbol),
+			string        => ($main::displayMode eq 'TeX' ? Currency::quoteTeX($symbol) : $symbol),
 			TeX           => Currency::quoteTeX($symbol),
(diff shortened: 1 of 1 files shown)


macros/contexts/contextCurrency.pl
```diff
@@ -250,13 +250,9 @@ sub new {
...
 			associativity => $associativity,
 			type          => "unary",
-			string        => (
-				($main::displayMode eq 'TeX' or $main::displayMode eq 'PTX')
-				? Currency::quoteTeX($symbol)
-				: $symbol
-			),
-			TeX   => Currency::quoteTeX($symbol),
-			class => 'Currency::UOP::currency'
+			string        => (($main::displayMode eq 'TeX') ? Currency::quoteTeX($symbol) : $symbol),
```
drgrice1, line 253: Could you remove the extra parentheses on this line?  Now that this is all on one line those are excessive.  The only parentheses that are needed are those around `$symbol`.

Alex-Jordan:
Let's not merge this until issue #886 is resolved. I'm testing Davide's solution to that issue, and it would clash with this one. Also Davide's solution there already removes the reference to `PTX` here.

drdrew42:
Closed in favor of #981

Issue #654, opened 2022-02-10
PGML unmatching backticks in math mode causes hard to decipher warnings.


somiaj:
Found this issue when trying to debug a problem. When using PGML, if the number of starting and stopping ticks don't match, the warning makes it hard to determine what the issue is.

For example, ```[`\pi``]``` gives the following warning:

![image](https://user-images.githubusercontent.com/18082412/153331884-5b5ea735-4173-43f1-b402-2f419fba79d2.png)

While, ```[``\pi`]``` gives the following warning:

![image](https://user-images.githubusercontent.com/18082412/153331966-bce6860c-f5b6-4b8f-810e-9373ecb7d3f6.png)

It would be helpful if more useful warning messages were produced.

Alex-Jordan:
And if you have unmatched colons like `[: pi ::]` you get no error (or output) at all.

You also get an error from unmatched pipes for comments: `[| pi ||]`.

dpvc:
The intended result for an unmatched delimiter is that PGML issues a warning and treats the unmatched context as literal text.  That is what happens for me when I use my command-line tool for testing, so I was confused when that didn't happen in actual WeBWorK problems.

It took me a while to track it down, but the issue actually turns out to be fairly simple to fix, and a bit surprising.

Terminators come in two forms, literal strings an regex patterns, and there are a couple of places where `ref($block->{terminator})` is used to distinguish between the two.  PGML check if this `ref()` call equals `'Regexp'`, but somewhere along the line, the Safe compartment changed and, in the Safe compartment now, the `ref()` is `Safe::Root23::Regexp`, or something similar with different numbers.  That means that the test fails, and PGML goes down the wrong path for handling the missing ed-delimiter.  In this case, it leads to no PGML warning, but causes the log message you indicate (that turns out to be from passing an undefined math string to `EV3P()`).

The fix is to properly account for the new class type used in the Safe compartment, so that the right path will be taken, and the warning message about unmatched delimiters will generated via `WARN_MESSAGE()`.

Here is a diff that does that:

``` diff
diff --git a/macros/PGML.pl b/macros/PGML.pl
index d0a8756f..8610a58f 100644
--- a/macros/PGML.pl
+++ b/macros/PGML.pl
@@ -26,6 +26,11 @@ sub Eval {main::PG_restricted_eval(@_)}
 
 sub Sort {return main::lex_sort(@_)}
 
+sub isRegexp {
+  my $ref = shift;
[...]

PR #978, opened 2023-12-05
PGML close math with any number of backtick or colon

Alex-Jordan:
This addresses #654, but @dpvc should review this. I might be oversimplifying and introducing a new bug.

Also this does not address the PGML verbatim issue mentioned there.


Alex-Jordan:
I'll note that this commit makes it work to have unbalanced math delimiters, like ````[`\pi``]````. The open delimiter determines which type of math it is.

But maybe we don't want that. Maybe it should be more like what @somiaj says in #654, that these situations should be recognized and there could be a better error message.

somiaj:
If it is possible to detect and warn users if they aren't balanced, maybe do both, and issue a warning that only instructors can see, but render the problem.

somiaj:
I just tested `````[`\pi````]````` (having four ticks when closing), it still closes, so I guess is working as it should, there are just extra "primes" shown, which I think is fine. I was almost going to suggest just changing the regex to ``qr/\`+\]/`` to allow for any number of closing ticks. But since it doesn't error out and visually someone can see the error probably not needed.

Basically #654 came out of spending way to long trying to figure out what was wrong with a long scaffold problem, when it was just due to not matching the number of ticks. I'm fine with this (provided there isn't unforeseen consequences) of allowing unmatching number of ticks, as this will help avoid a simple error that might otherwise be hard to find. The worst issue I could think of is something like ```[`\frac{\pi}{2}``]``` in which someone may have meant to open with two backticks, but had a typo. With this change, this would be easily seen when rendering the problem allowing them to fix it, and not error out.

Alex-Jordan:
Closing this in favor of #980.

PR #979, opened 2023-12-05
Don't always show "Ungraded" for essay problems.

drgrice1:
This requires a change to webwork2 (or any other front end).  Basically, the front end needs to let PG know if the problem has been graded or not via the environment "needs_grading" flag.  This is done for webwork2 in https://github.com/openwebwork/webwork2/pull/2270.

The feedback processing in the ENDDOCUMENT method of PG.pl and PGessaymacros.pl use this flag to show the correct messages for essay answers.

Note that the methods in PG.pl now has more generic code for handling answers like those from the PGessaymacros.pl macro instead of being specifically for that macro.  This done with the `manuallyGraded` and `needsGrading` options that can be set by a `feedback_options` method for a custom answer type.

I think that the answer messages and feedback popover titles need further examination.  I am not sold on using "Graded" and "Ungraded" for manually graded problems.  I also don't like the messages that state things like "All of the gradeable answers are correct".  In reality all questions are "gradeable", it is just a matter of who or what does the grading (and perhaps when that grading is done). I am open to suggestions for improvements to these messages.

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -1325,7 +1325,7 @@ sub ENDDOCUMENT {
 								Mojo::DOM->new_tag(
 									'div',
 									class => 'alert alert-success mb-2 p-1',
-									maketext('All of the gradeable answers are correct.')
+									maketext('All of the computer gradable answers are correct.')
 								)
 							);
 						} else {
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -916,6 +916,19 @@ has not answered a question (for example, this occurs for answers to questions
 created with the L<draggableProof.pl> macro) . So macros that create answers
 with responses like that should override this.
 
+=item *
+
+C<manuallyGraded>: This is a boolean value. This should be true if the answer is
+not graded by the PG problem grader, but is graded manually at a later time, and
+should be false if the PG problem grader sets the grade for this answer. For
+example, essay answers created by the PGessaymacros.pl macro set this to true.
+
+=item *
+
+C<needsGrading>: This is a boolean value. This should be true if the answer is
+not graded by the PG problem grader, but is graded manually at a later time, and
+the answer has changed.
+
 =back
 
 =cut
@@ -1005,9 +1018,10 @@ sub ENDDOCUMENT {
 		add_content_post_processor(sub {
 			my $problemContents = shift;
 
-			my $numCorrect = 0;
-			my $numBlank   = 0;
-			my $numEssay   = 0;
+			my $numCorrect        = 0;
+			my $numBlank          = 0;
+			my $numManuallyGraded = 0;
+			my $needsGrading      = $rh_envir->{needs_grading};
 
 			my @answerNames = keys %{ $PG->{PG_ANSWERS_HASH} };
 
@@ -1016,9 +1030,6 @@ sub ENDDOCUMENT {
 				my $ansHash      = $PG->{PG_ANSWERS_HASH}{$answerLabel}{ans_eval}{rh_ans};
 
 				my $answerScore = $ansHash->{score} // 0;
-				my $isEssay     = ($ansHash->{type} // '') eq 'essay';
-				++$numCorrect if $answerScore >= 1;
-				++$numEssay   if $isEssay;
 
 				my %options = (
 					resultTitle      => maketext('Preview'),
@@ -1032,7 +1043,9 @@ sub ENDDOCUMENT {
 					showEntered      => 1,
 					showPreview      => 1,
 					showCorrect      => 1,
-					answerGiven      => 0
+					answerGiven      => 0,
+					manuallyGraded   => 0,
+					needsGrading     => 0
 				);
 
 				# Determine if the student gave an answer to any of the questions in this response group and find the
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I tested this and some things seem amiss. Comments here might apply here, or might apply to https://github.com/openwebwork/webwork2/pull/2270. Or maybe none of the above if I'm seeing an issue that is not from this PR.

To test, I put the [demo essay problem](https://webwork.maa.org/wiki/EssayAnswer1) into an assignment. OK, so I have a dropdown and an essay, and when I click to submit, the dropdown is left blue, not marked correct/incorrect. This is because `$showPartialCorrectAnswers = 0;`. But in that case, should I be seeing the "All of the gradeable answers are correct." message? It's indirectly revealing that I have that part correct.

I changed to `$showPartialCorrectAnswers = 1;`. Moving on...

When I go to the grader page for the problem, I see:

Screenshot 2023-12-16 at 3 01 01 PM

I get it that at the bottom, it is a legend telling me that names with the background color are the ones where I need to edit an essay answer. But it wasn't clear at first. And the "=" sign rendered through MathJax seems wrong. Maybe this is something that was already there from before this PR. If I were designing this, I would put an indicator on the `div.past-answer` that in the case of my screenshot, has "My test answer" in it. The indicator could include coloring the background, but something else might be better. Like an icon with a title that explains its purpose.

`All of the gradeable answers are correct` could be `All of the autoscorable answers are correct`. Can the messages be reworked to use `autoscorable` (or something like that, `automatically gradable`, etc)? And then refer to "the other answers"? Or "non-automatically gradable", if necessary. BTW GitHub spellchecker says "gradeable" is not valid, but "gradable" is.

>  I am not sold on using "Graded" and "Ungraded" for manually graded problems.

How about "Awaiting Instructor Grading" for "Ungraded", and when it has been graded, use "Incorrect", "Partially Correct", or "Correct".

drgrice1:
> To test, I put the [demo essay problem](https://webwork.maa.org/wiki/EssayAnswer1) into an assignment. OK, so I have a dropdown and an essay, and when I click to submit, the dropdown is left blue, not marked correct/incorrect. This is because `$showPartialCorrectAnswers = 0;`. But in that case, should I be seeing the "All of the gradeable answers are correct." message? It's indirectly revealing that I have that part correct.

This is not from this pull request, and is the way that this has always worked.  In fact see the comment that I added on line 1258 of `macros/PG.pl` (that is the line number in the develop branch), as I point out exactly what you you are getting at here.  I also brought this up in our meeting last week.

> I get it that at the bottom, it is a legend telling me that names with the background color are the ones where I need to edit an essay answer. But it wasn't clear at first. And the "=" sign rendered through MathJax seems wrong. Maybe this is something that was already there from before this PR. If I were designing this, I would put an indicator on the `div.past-answer` that in the case of my screenshot, has "My test answer" in it. The indicator could include coloring the background, but something else might be better. Like an icon with a title that explains its purpose.

The coloring on the student's name rather than the essay answer itself is also how this page has always done it.  No changes have been made to the problem grader in this pull request (or really the webwork2 one).  I added the legend at some point (in the  to help clarify what the coloring meant.  Note that there is also another possible key that may be shown in the legend in the case that there are different source files for different students.  In that case the student answers are colored for all students that have a different source file than the problem that is currently rendered at the top of the page, and the additional legend key is shown explaining that color.  So there would be a conflict of colors, and that would have to be reworked as well.

I don't think that an icon would be a good idea in this case.  I don't think there is a good place to put it.  Perhaps a tooltip would work though.

> `All of the gradeable answers are correct` could be `All of the autoscorable answers are correct`. Can the messages be reworked to use `autoscorable` (or something like that, `automatically gradable`, etc)? And then refer to "the other answers"? Or "non-automatically gradable", if necessary. BTW GitHub spellchecker says "gradeable" is not valid, but "gradable" is.

I was thinking of using the terminology "automatically gradable".  I don't really like "autoscorable" since I am not even sure that "autoscore" is a word.

The "gradeable" misspelling has been in the attempts table since the messages were added in 2015.  I cut and pasted that message over from there.  Good catch though.  That should be fixed.

> How about "Awaiting Instructor Grading" for "Ungraded", and when it has been graded, use "Incorrect", "Partially Correct", or "Correct".

Using "Awaiting Instructor Grading" is perhaps better for answers that have not been graded (although a bit long).  However, the problem is that even when an answer is graded, the grade that has been assigned is not known.  So you can't use the "Incorrect", "Partially Correct", or "Correct" headers.

dlglin:
> I was thinking of using the terminology "automatically gradable". I don't really like "autoscorable" since I am not even sure that "autoscore" is a word.

How about "computer-gradable"?

drgrice1:
I changed the wording of "All of the gradable answers are correct." to "All of the computer gradable answers are correct."  This is only shown if there are manually graded problems on the page (essay problems) , and gives a better distinction between the manually graded problems and the rest.

pstaabp:
> One thing that needs to be changed is the usage of "macro" in a macro name. I find it annoying to speak of the "PGessaymacros.pl macro". That should have been named something like "PGessayanswers.pl". It is too late for that macro due to backwards compatibility, but I think a rule should be that no new macros can be created that have "macro" in the file name.

I agree.  Perhaps we can rename things with "pointers" for problems using old name. I also wonder about using "PG" in a macro name.  Aren't all macros for pg?

pstaabp (approved):
I think this is ready to go unless we need more discussion.

drdrew42 (approved):

PR #980, opened 2023-12-06
Fix warnings for improper PGML block termination (alternate to #978)

drgrice1:
This is basically @dpvc's suggestion at https://github.com/openwebwork/pg/issues/654#issuecomment-1843553494.

There were two lines of code changed in his suggestion that seem to be unrelated.

The first adds a second return value which is the "inline" math mode to the list returned on line 1358.  I also made that change, since it seems correct.  The other return from the same method also returns a list with two values, the second of which is a math mode.

The second just changed the parentheses on line 1364.  I just removed the parentheses because neither the original parentheses or the way they were changed in the suggestion are needed.

### DIFF
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -28,6 +28,11 @@ sub Eval { main::PG_restricted_eval(@_) }
 
 sub Sort { return main::lex_sort(@_) }
 
+sub isRegexp {
+	my $ref = shift;
+	return $ref && ref($ref) =~ m/Regexp$/;
+}
+
 ######################################################################
 
 package PGML::Parse;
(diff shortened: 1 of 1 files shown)


drgrice1:
I personally like this better than attempting to allow any number of backticks, colons, or pipes to terminate a block, but if the general consensus is to go that way instead, then that is okay.

drgrice1:
Note that if we do decide to go with the approach in #987, something will still need to be done about the `Regexp` `ref` matching.  That code should not be left dis-functional.

Alex-Jordan:
I closed #978. (For posterity, it was accidentally referenced here as `#987`.)

drdrew42 (approved):

PR #981, opened 2023-12-07
use stringify in context Currency

Alex-Jordan:
This has Davide's edits from #886. It works to fix the reported issue. Two side effects have been mentioned:

* Something like `$message = "The cost is $c.";` and then in PGML using `[$message]` would not work without a star.
* Something like `[$c]*` currently works, but would no longer work in hardcopy.

### DIFF
--- a/macros/contexts/contextCurrency.pl
+++ b/macros/contexts/contextCurrency.pl
@@ -250,13 +250,9 @@ sub new {
 			precedence    => 10,
 			associativity => $associativity,
 			type          => "unary",
-			string        => (
-				($main::displayMode eq 'TeX' or $main::displayMode eq 'PTX')
-				? Currency::quoteTeX($symbol)
-				: $symbol
-			),
-			TeX   => Currency::quoteTeX($symbol),
-			class => 'Currency::UOP::currency'
+			string        => $symbol,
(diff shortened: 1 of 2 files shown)


somiaj:
I am unsure if this is beneficial in the long run, seems we are just replacing one issue with another. Currently if someone doesn't tell PGML to not escape things, an extra slash appears in hardcopy, annoying but things still work, but after this, if someone is correctly using `[$c]*` to not escape things, hardcopy won't even generate. Just my thoughts, I don't have any suggestions for a good solution.

dpvc:
I see Alex suggest the same thing while I was writing.  Sorry for the extra noise.

somiaj:
Yea, I guess it would matter how many people were trying to work around the work around in PGML, I was just downloading the updated OPL to see if it is actually used there.

somiaj:
In grepping through the OPL, it seems the common solution was to use math mode with currency, ``[`[$c]`]``, and I'm not seeing any uses of `[$c]*` either, so I guess we shouldn't worry about changing old behavior.

drgrice1 (approved):
Looks good to me.

drdrew42 (approved):

PR #983, opened 2023-12-11
Make it possible for the frontend to set $refreshCachedImages.

drgrice1:
The frontend can now pass a value for `refreshCachedImages` in the environment to force dynamically generated images to be regenerated. The PG problem editor of webwork2 will use this.  See https://github.com/openwebwork/webwork2/pull/2277.


drgrice1:
Closing as this is not needed anymore.  A better approach was found in https://github.com/openwebwork/webwork2/pull/2283 .

PR #985, opened 2023-12-12
Minor clean up and issue fixing for scaffold.pl.

drgrice1:
The code for hiding (or really dimming) the results for unnopened sections in the attempts table is no longer relevant.  So that is removed.

Also add `BEGIN { strict->import }` and fix the issues it reveals. This issue on line 317 is my fault.  In a previous pull request I moved that line up before `$self` was defined to group the error calls together at the beginning, but missed that `$self` needed to be changed to `$scaffold` to do that (which is what `$self` is later set to be).

### DIFF
--- a/macros/core/scaffold.pl
+++ b/macros/core/scaffold.pl
@@ -640,7 +640,8 @@ sub can_open {
 	return 1 if $Scaffold::forceOpen;
 	my $method = ($Scaffold::isInstructor ? $self->{instructor_can_open} : $self->{can_open});
 	$method = $self->{after_AnswerDate_can_open} if $Scaffold::afterAnswerDate;
-	$method = "Section::can_open::" . $method unless ref($method) eq 'CODE';
+	return $method->($self) if ref($method) eq 'CODE';
+	$method = "Section::can_open::" . $method;
 	return $self->$method;
 }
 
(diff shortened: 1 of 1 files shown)


macros/core/scaffold.pl
```diff
@@ -676,7 +641,7 @@ sub can_open {
 	my $method = ($Scaffold::isInstructor ? $self->{instructor_can_open} : $self->{can_open});
 	$method = $self->{after_AnswerDate_can_open} if $Scaffold::afterAnswerDate;
 	$method = "Section::can_open::" . $method unless ref($method) eq 'CODE';
-	return &{$method}($self);
+	return $self->$method;
```
dpvc, line 644: Note that `$method` can be either a method name in the `Section::can_open` package (the usual case), or a CODE literal.  (See the POD documentation for `can_open`, and note the `unless ref($method) eq 'CODE'` in the line above).

You probably need to do something like

``` js
if (ref($method) eq 'CODE') {
  return &{$method}($self);
}
$method = "Section::can_open::";
return $self->$method;
```

instead in order to handle both cases.  Similarly for the `is_open` code below.

dpvc (changes requested):
One situation needs additional work, as described below.

macros/core/scaffold.pl
drgrice1: Yeah, I forgot that possibility.  I have fixed that now.  Thanks.

dpvc (approved):
Looks good now.

PR #987, opened 2024-01-02
remove workaround for MathView double star operator

Alex-Jordan:
In MathView js, there is this place (which this PR removes) where `**` gets replaced with `^`. This is unnecessary since the AsciiMath portion of the MathJax config is already handling `**` as if it were a `^`.

Furthermore, this is actually harmful, since it prevents a user from using the AsciiMath `***` to make a `⋆`.

So this PR removes that replacement.


drgrice1 (approved):
Looks good.

pstaabp (approved):

Issue #988, opened 2024-01-16
MathQuill toolbar placement in a relatively positioned parent - bug when in an RTL course.


taniwallach:
When an input box appears in a relatively placed element in an RTL course, the MathQuill toolbar placement is problematic.

See the sample problem file [chain01.txt](https://github.com/openwebwork/pg/files/13950308/chain01.txt) and the image below. All is OK when the course language is English, but with the course language as Hebrew the location blocks the input box.

![mq_menu_placement_issue](https://github.com/openwebwork/pg/assets/39089094/781e0813-b100-4db8-a87f-ce1d4e21218b)

Related to the code at https://github.com/openwebwork/pg/blob/daff8afd342f06d25feb87d4c00280980492e318/htdocs/js/MathQuill/mqeditor.js#L362C9-L362C23

@drgrice1 Could you help with this? It seems the JavaScript code will need to condition how it calculates placement for these cases depending on whether the primary document direction is LTR (default) or RTL.

From `simple.conf`:
```
$language = 'he-IL';
$perProblemLangAndDirSettingMode = 'auto:en:ltr';
```
where dropping the language line gets placement to work.

---

Note: The problem triggers a warning. `Use of uninitialized value $html_attribs in concatenation (.) or string at line 1794 of [PG]/macros/core/PGbasicmacros.pl` A fix is below. Eventually I'll get a PR in.

---

```
diff --git a/macros/core/PGbasicmacros.pl b/macros/core/PGbasicmacros.pl
index 793e7b0c..3aeb3b55 100644
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -1773,7 +1773,7 @@ sub openDivSpan {
                return ();
        }
        my $option_ref = {};
-       my $html_attribs;
+       my $html_attribs = '';
        if (ref($_[0]) eq 'HASH') {
                $option_ref   = shift;
                $html_attribs = processDivSpanOptions($option_ref);


```

drgrice1:
The quick and easy solution is to add `/*rtl:ignore*/` above line 104 of `mqeditor.scss` (before `right: 10px`).  That will make the toolbar appear on the right side of the page just as it does for left-to-right languages.

taniwallach:
Will be resolved by https://github.com/openwebwork/pg/pull/989

PR #989, opened 2024-01-16
Make the MathQuill editor toolbar appear on the right even for right-to-left languages.

drgrice1:
Currenlty if the course language is set to he-IL and the `$perProblemLangAndDirSettingMode` is set to `auto:en:ltr`, then the computations for the toolbar position make the toolbar on the right but way to wide.  This just makes the toolbar appear the same as for left-to-right languages.  That is not strictly correct, but since the toolbar is not currently translated, I am not sure much better can be done at this point.

This addresses issue #988.

### DIFF
--- a/htdocs/js/MathQuill/mqeditor.scss
+++ b/htdocs/js/MathQuill/mqeditor.scss
@@ -100,6 +100,7 @@ input[type='text'].codeshard.mq-edit {
 	border-radius: 4px;
 	border: 2px solid darkgray;
 	background-color: white;
+	/*rtl:ignore*/
 	right: 10px;
 	z-index: 1001;
 	overflow-y: auto;


taniwallach (approved):
Certainly good enough. I think students can deal with the MQ toolbar on the far right on top of the menu.

pstaabp (approved):

pstaabp:
@taniwallach does this need a hot fix?  I didn't check in 2.18.

drgrice1:
@taniwallach: Do you have an answer to @pstaabp's question?  Is this needed as a hotfix for 2.18?

taniwallach:
Sorry... No - I do not think a hot-fix is needed. The three servers I manage directly have the patch already, and the institution where I provide advice and support has queued it for installation. If anyone else is running an RTL language webwork courses, hopefully they will find the patch here or ask in the forums.

PR #990, opened 2024-01-18
new context flag to require constant entries

Alex-Jordan:
This makes it so that you can set a context flag `requireConstantEntries`, and if so, vectors cannot have variables in their entries. But for example something like `t <1, 2, 3>` is still legal. This allows me to write a problem that requires students to "factor out" a parameter from vector answers.

I want to check with @dpvc that this is an OK way to do this. If so I will extend this PR to do the same thing for Matrix objects. And I'll add a default value of `0` to `Default.pm`.

### DIFF
--- a/lib/Parser/List/Vector.pm
+++ b/lib/Parser/List/Vector.pm
@@ -22,6 +22,9 @@ sub _check {
 			$type = (($type =~ m/^[aeiou]/i) ? "an " : "a ") . $type;
 			$self->{equation}->Error([ "Coordinates of Vectors must be Numbers, not %s", $type ]);
 		}
+		if ($self->context->flag("requireConstantEntries") && !($x->{isConstant})) {
+			$self->{equation}->Error("Coordinates of Vectors must be constant");
+		}
 	}
 }
 
--- a/lib/Parser/List/Vector.pm
+++ b/lib/Parser/List/Vector.pm
@@ -22,10 +22,9 @@ sub _check {
 			$type = (($type =~ m/^[aeiou]/i) ? "an " : "a ") . $type;
 			$self->{equation}->Error([ "Coordinates of Vectors must be Numbers, not %s", $type ]);
 		}
-		if ($self->context->flag("requireConstantEntries") && !($x->{isConstant})) {
-			$self->{equation}->Error("Coordinates of Vectors must be constant");
-		}
 	}
+	$self->{equation}->Error("Coordinates of a Vector must be constant")
+		if ($self->context->flag("requireConstantVectors") && !($self->{isConstant}));
 }
 
 sub ijk {
--- a/lib/Parser/List/Matrix.pm
+++ b/lib/Parser/List/Matrix.pm
@@ -24,6 +24,8 @@ sub _check {
 		$self->{equation}->Error("Entries in a Matrix must be Numbers or Lists of Numbers")
 			unless ($x->type =~ m/Number|Matrix/);
 	}
+	$self->{equation}->Error("Entries of a Matrix must be constant")
+		if ($self->context->flag("requireConstantMatrices") && !($self->{isConstant}));
 }
 
 #
--- a/lib/Parser/List/Point.pm
+++ b/lib/Parser/List/Point.pm
@@ -20,6 +20,8 @@ sub _check {
 			$self->{equation}->Error([ "Coordinates of Points must be Numbers, not %s", $type ]);
 		}
 	}
+	$self->{equation}->Error("Coordinates of a Point must be constant")
+		if ($self->context->flag("requireConstantPoints") && !($self->{isConstant}));
 }
 
 #########################################################################


pstaabp:
@Alex-Jordan Can you show an example problem that would use this?

Alex-Jordan:
Here is a problem that uses this. For example, the answer might be `r<3,1,1,0> + s<1,-3,0,1>`. It will not accept `<3r+s,r-3s,r,s>`. If you try that, it returns "Coordinates of Vectors must be constant".


```
# WeBWorK problem written by Alex Jordan 
DOCUMENT();

loadMacros(qw(
    PGstandard.pl PGML.pl PCCmacros.pl
));

Context("Matrix")->flags->set(requireConstantEntries => 1);
Context()->variables->are(
    r => 'Real',
    s => 'Real',
    t => 'Real',
    a => 'Real',
    b => 'Real',
    c => 'Real',
);

$a1 = ColumnVector(random(1,3), 0);
$a2 = ColumnVector(random(-7,-3), random(-9,-2));

$h = non_zero_random(-4,5);
$k = non_zero_random(-4,5);
$a3 = $h*$a1 + $k*$a2;
$g3 = gcd($a3->value);
$a3 = $a3/$g3;

$i = non_zero_random(-3,3);
$j = non_zero_random(-3,3);
$a4 = $i*$a1 + $j*$a2;
$g4 = gcd($a4->value);
$a4 = $a4/$g4;

$b = ColumnVector(0,0);

$x1 = ColumnVector(-$h,-$k,$g3,0);
$x2 = ColumnVector(-$i,-$j,0,$g4);

$A = Matrix($a1, $a2, $a3, $a4)->transpose;

$sol = Formula("r $x1 + s $x2");
$ans = $sol->cmp(checker => sub {
    my ($c,$s,$self) = @_;
    $s = Formula("$s")->reduce;
    Value::Error("Your answer does not solve the system.") unless ($A*$s == Matrix($b)->transpose);
    # How many parameters did student use?
    $spar = $s->usesOneOf('r') + $s->usesOneOf('s') + $s->usesOneOf('t') + $s->usesOneOf('a') + $s->usesOneOf('b') + $s->usesOneOf('c');
    Value::Error("Your answer uses more parameters than is necessary.") if ($spar > 2);
    Value::Error("Your answer does not use enough parameters.") if ($spar < 2);
    # Get dimension of span of student answer space
    @svec = map {$s->eval(r => random(1,9), s => random(1,9), t => random(1,9), a => random(1,9), b => random(1,9), c => random(1,9))} (0..3);
    $sdim = Matrix(@svec)->order_LR;
    Value::Error("Your answer does not parametrize the entire solution set.") if ($sdim < 2);
    # Now we know student answer uses exactly the right number of parameters
    # and also not in some trivial way since the dimension is right
    return 1;
});

# remove the check for equivalency with previous submitted answer
pop(@{$ans->{post_filters}});

BEGIN_PGML
Describe all solutions to the matrix equation [`A\vec{x}=\vec{0}`] where [`A`] is row equivalent to [```[$A]```]
Write the solution set in parametric vector form.
[@ KeyboardInstructions('You may use [`r`], [`s`], [`t`], [`a`], [`b`], [`c`] as parameters, as needed.
Enter vectors using angle brackets. For example, to enter [`\begin{bmatrix}1\\\\2\end{bmatrix}`], type [|<1,2>|]*.') @]**

[`\vec{x}={}`][_]{$ans}

END_PGML

ENDDOCUMENT();
```

lib/Parser/List/Vector.pm
```diff
@@ -22,6 +22,9 @@ sub _check {
 			$type = (($type =~ m/^[aeiou]/i) ? "an " : "a ") . $type;
 			$self->{equation}->Error([ "Coordinates of Vectors must be Numbers, not %s", $type ]);
 		}
+		if ($self->context->flag("requireConstantEntries") && !($x->{isConstant})) {
+			$self->{equation}->Error("Coordinates of Vectors must be constant");
+		}
```
dpvc, line 27: You don't need to check each coordinate separately, as the vector's `isConstant` flag will already be set by the time the `_check` subroutine is called, so I think you can just do

``` perl
	$self->{equation}->Error("Coordinates of Vectors must be constant");
	  if $self->context->flag("requireConstantEntries") && !($self->{isConstant};
```

after the `foreach` loop.

dpvc (changes requested):
I don't have any problem with this, though I left a comment that should allow you to make it more efficient.

You may also want to consider a more specific flag name, like `requireConstantVectorEntries` if you are also planning to do this for Matrices, so that you can control them individually.

Alex-Jordan:
I also see the issue that @pstaabp found. And I see it after applying @dpvc 's suggestion too.

This calls `$self->{equation}->Error()` if there are non-constant entries in a vector. You can also trigger `$self->{equation}->Error()` a few lines above if the entries are not of type number, and when I do that, I do not see anything similar happen. That is, I don't see a portion of the parsed expression disappear. I'm not sure what to look into next.

dpvc:
Does the feedback mechanism properly escape HTML special characters like `<` and `>`?  The `<r,-r,3r,0>` part of the answer might look like a custom tag, whereas `<2,-3,0,3>` does not because a tag name must start with a letter.  I'm not able to run the current develop version of WW, so can't check this myself.  But that's what it looks like might be happening to me.

Alex-Jordan:
So now I'm not able to reproduce the issue. I am wondering if what happened was Davide's suggested change actually did incidentally fix that issue, but while I was testing I forgot to restart webwork2. So it was still running off the old .pm. @pstaabp can you try this now?

If the same issue arises, can you look into Davide's idea? In the HTML source for the page, do you see the angle brackets?

drgrice1:
This is an issue with HTML characters.  The DOM shows that the entire answer is there, but it is not rendered because of the starting `<` tag.  This is only in the entered field, and not the preview.  This is tricky because this is all an HTML encoded attribute value on a div.  The decoding when the popover opens decodes everything into the DOM, and not just what is meant to be HTML.  I will see what I can do about it.

drgrice1:
It is odd that this does not happen with the develop branch though.  It only happens with this pull request.  I haven't figured out why that is the case.

Alex-Jordan:
Well, I'm thoroughly confused and wondering if the behavior is not consistent. I know for sure that prior to my Feb 3 comment here, I tested this on both 2.18 (where I made a local uncommitted edit) and on develop. And I saw the issue both places (in the results table in 2.18 and in the popover with develop). I specifically confirmed that before I made that comment because I was wondering if the results table was a piece of the puzzle.

Earlier today on develop I could not repeat the issue, which made me think that in earlier testing I hadn't restarted webwork2. But now with what you (@drgrice1 ) say, I'm just really confused.

Alex-Jordan:
Oh! I get it...

In my testing today, the answer I tried was `<-r,r,r,0>+ whatever`.  The dash character after the `<` probably saves me from the tag interpretation. But when it is `<r,r,r,0>` or like @pstaabp 's example, the `<r ... >` looks like a tag to the web browser.

So is this issue actually independent of this PR? Would _any_ vector that starts with a letter after the `<` have this presentation issue?

drgrice1:
I am able to get this to happen on develop as well.   I just had to get an error to occur in the student answer (but not the one from this pull request).  Testing this with 2.18 and the attempts table, I see that this also happens there.  So this is not an issue with this pull request, and is also not an issue with the new feedback mechanism.

The issue is that `student_ans` is coming in already escaped when errors do not occur.  However, when an error occurs, then it is not escaped.  I am working on tracking down why that is, but the issue is an inconsistency in how PG is setting the student_ans in the answer hash.

Alex-Jordan:
A note to anyone who might look at this PR: I will extend this to Matrices and Points, and do as Davide suggested and use different flag names for each. So this can wait for a new review until I add those. And the bug @pstaabp found is an independent issue that Glenn has a handle on.

drgrice1:
Testing this in a script from the command line, if the input for the answer has the value `<1,0,0>` then the value of `student_ans` in the answer hash is `&lt;1,0,0&gt;`, and if the input has the value `<s(1,0),0,0>` then `student_ans` is `<s(1,0),0,0>`.

I also tested this with webwork and pg version 2.17, and this issue already occurred with that version.  So this isn't an issue with the Mojolicious change either.

dpvc:
One of the sources of this problem is that it is not clear what `student_ans` should contain; is it HTML, LaTeX, text, or different things under different circumstances?  Because the `student_ans` shows in the results table (or used to), it was used to highlight the error location in the student answer, and so included HTML to do that, so it had to be considered escaped HTML for that purpose.  But that means there is no plain-text version of the student answer.  I've always thought there should be `student_ans_text`, `student_ans_html`, `student_ans_latex`, and similarly `correct_ans_text`, `preview_ans_text`, and so on.  But straightening that all out in all the macros that set one or another of these would be quite an undertaking, I would think.

Alex-Jordan:
This now has three new flags: `requireConstantPoints`, `requireConstantVectors`, `requireConstantMatrices`.

Note that a Point, Vector, or Matrix object would be required to have constant entries, including when used in a Formula. But a Formula that returns Points, Vectors, or Matrices is not necessarily required to be constant. For example, `Formula("x <1,2>")` is a Formula that returns vectors, and is fine to use when `requireConstantVectors` is set.

So I considered having more clear names for these flags, to make it more clear that the entries need to be constant but nothing is required outside of the object: `requireConstantPointEntries`, `requireConstantVectorEntries`, `requireConstantMatrixEntries`. Is it worth the added verbosity?

Whenever this is resolved and merged, I need to update wiki documentation or I'm pretty sure documentation for this will be forgotten about later.

Alex-Jordan:
@dpvc This still has a block on it that you set. I think that the issue that floated up here has been determined to be an independent issue. Do you think what is there now looks good, at least as far as the scope of this PR?

pstaabp:
This looks good now.  Tested both constant vectors and constant points.  Although it make sense to use constantMatrix as well, I'm trying to figure out how to test.   Since the standard is to use a matrix input, there wouldn't be a way to have a variable times that.  @Alex-Jordan do you have an idea about using this?

dpvc (approved):

drgrice1:
@pstaabp:  Matrices can also be asked to be entered in a single text input in the format `[[1,0],[0,1]]` (for example).  In that case you can give scalar multiples.  For example, one could enter `s[[1,0],[0,0]]+t[[0,0],[0,1]]`.

drgrice1:
By the way, here is a very simple example of how to ask for a matrix in the format I mentioned (not utilizing the new flag):
```perl
DOCUMENT();

loadMacros('PGstandard.pl', 'MathObjects.pl', 'PGML.pl');

Context('Matrix');
$matrix1 = Matrix('[[1,0],[0,1]]');

BEGIN_PGML
What is the [`2 \times 2`] identity matrix?

[_]{$matrix1}{10}
END_PGML

ENDDOCUMENT();
```

drgrice1 (approved):

drgrice1:
Actually, something doesn't seem to be working with the matrix flag.  With the following example:
```
DOCUMENT();

loadMacros('PGstandard.pl', 'MathObjects.pl', 'PGML.pl');

Context('Matrix')->flags->set(requireConstantMatrices => 1);
Context()->variables->are(s => 'Real', t => 'Real');
$matrix = Matrix('s[[1,0],[0,0]] + t[[0,0],[0,1]]');

BEGIN_PGML
Enter the matrix [`[$matrix]`]:

[_]{$matrix}{10}
END_PGML

ENDDOCUMENT();
```
I can enter `[[s,0],[0,t]]` and it is counted correct.  Am I missing something?

drgrice1:
Nevermind. It turns out you need to check out this pill request for the flag to work.

PR #992, opened 2024-01-20
Ensure a false score is saved as 0 in AnswerHash.

somiaj:
Somewhere in the Rendering.pm path, a boolean false score gets
  converted into an empty string, which can no longer be used in some
  numerical comparisons, as discussed in openwebwork/webwork2#2300.

  This fixes the case mentioned in the issue, and sets false values to 0
  when setting the score in the AnswerHash to catch other possible cases.

### DIFF
--- a/lib/AnswerHash.pm
+++ b/lib/AnswerHash.pm
@@ -252,7 +252,7 @@ sub input {    #$rh_ans->input('foo') is a synonym for $rh_ans->{student_ans}='f
 sub score {
 	my $self  = shift;
 	my $score = shift;
-	$self->{score} = $score if defined($score);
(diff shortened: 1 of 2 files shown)


drgrice1 (approved):
Looks good.

To be precise the boolean value becomes a string when Mojolicious JSON encodes the subprocess return value to pass back to the parent process, and then JSON decodes the return value in the parent process.  The same thing happens with a boolean false value if you use Cpanel::JSON::XS, JSON, or JSON::PP (Mojolicious uses one of those depending what is available) to encode and decode a hash directly in a Perl script.

pstaabp (approved):

PR #995, opened 2024-01-25
Add `strict` to modules shared to the PG safe zone in pg_config.dist.yml

drgrice1:
Versions of Perl prior to 5.34 need this.  Otherwise warnings are issue by the `BEGIN { strict->import }` approach now used in place of the old `BEGIN { be_strict() }` approach.

Note that this is not used by webwork2, but the standalone renderer will need this (if anyone runs that with an older perl version).  So the standalone renderer will also need to update its distributed pg_config.yml file.

Also change `Main` to `main` in the translator.  I can't resist fixing this any longer.  The proper namespace is `main`.  This is also used inconsistently in `Translator.pm`.  Other than these two instances `main` is used everywhere else.

This is related to (but not needed to fix) the issue brought up in https://github.com/openwebwork/pg/pull/968#issuecomment-1910760107.

### DIFF
--- a/conf/pg_config.dist.yml
+++ b/conf/pg_config.dist.yml
@@ -254,3 +254,4 @@ modules:
   - ['Rserve']
   - [DragNDrop]
   - ['Types::Serialiser']
+  - [strict]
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
Didn't test this, but looks good.

drdrew42 (approved):

PR #996, opened 2024-01-27
handle nested braces in alignment specification

Alex-Jordan:
This fixes a bug with niceTables. Without this commit, if you specify an alignment string like `*{3}{p{2in}}` somewhere in a niceTables table, it will break. The regex treats this: `{p{2in}` as a matched brace pair with `p{2in` as the contents.

This commit changes the regex to balance the braces.

Try this example:

```
DOCUMENT();
BEGIN_PGML
[# [. Cell .]  [. Cell .]  [. Cell .] #]*{align => '*{3}{p{2in}}'}
END_PGML
ENDDOCUMENT();
```

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -1282,10 +1282,10 @@ sub ParseAlignment {
 	$alignment =~ s/\R//g;
 
 	# first we parse things like *{20}{...} to expand them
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Impulsive approval!

PR #997, opened 2024-01-27
constrain images to 100% width

Alex-Jordan:
Consider this example problem:

```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl PGtikz.pl));

$image = createTikZImage();
$image->BEGIN_TIKZ
\fill[color=blue] (0,0) -| (1,1) -| cycle;
END_TIKZ

BEGIN_PGML
[#
    [. Text Cell .]
    [. [@image(insertGraph($image), width=>600)@]* .]*
#]{align => 'p{2.5in}p{2.5in}',}

END_PGML
ENDDOCUMENT();
```

The table tries to constrain the second column to be 2.5 in wide. But the image command wants to make the image 600px wide, more than 2.5in. The `600` wins, and the image is presented badly overflowing.

The author could reduce the `600`, but should they have to? Should they have to play with that number until the image just fits? Even if they cut the `width=>600` that still does leads to a default of `200` being used, which is not good either. I think the author should be able to declare the column to be 2.5in wide, and then pass the image a clearly larger size like `600` and it should just shrink down to fit.

So this PR adds inline styling to the `img` element, with `max-width:100%;`. Normally I would not want to use inline styling, but it seems appropriate here. The surrounding table certainly has inline styling for the "2.5in" width, so you'd want the cap on the image width to be styled inline as well. In case the overall HTML gets separated from the CSS files.

I cannot think of a situation where `max-width:100%;` would cause harm, but I'm aware that I could be overlooking something.

### DIFF
--- a/htdocs/js/ImageView/imageview.scss
+++ b/htdocs/js/ImageView/imageview.scss
@@ -1,4 +1,6 @@
 .image-view-elt {
+	max-width: 100%;
+
 	&:hover {
 		cursor: pointer;
 	}


macros/core/PGbasicmacros.pl
```diff
@@ -2889,7 +2889,7 @@ sub image {
 			my $altattrib = '';
 			if (defined $alt_list[0]) { $altattrib = 'alt="' . encode_pg_and_html(shift @alt_list) . '"' }
 			$out =
-				qq!<IMG SRC="$imageURL" class="image-view-elt $valign" tabindex="0" role="button"$width_attrib$height_attrib $out_options{extra_html_tags} $altattrib>!;
+				qq!<IMG SRC="$imageURL" class="image-view-elt $valign" style="max-width:100%;" tabindex="0" role="button"$width_attrib$height_attrib $out_options{extra_html_tags} $altattrib>!;
```
drgrice1, line 2892: Why use inline style for this?  There is a class `image-view-elt` that this could be applied to in the `imageview.scss` file.

Alex-Jordan: I commented above why I leaned inline. The motivating example has an inline width specification already, in the \ HTML for the table. And the purpose of this change is to play well with that inline specification.

If this styling were maintained in `imageview.scss`, what would happen if the HTML for the motivating problem ended up somewhere without `imageview.css`? I'm thinking of a third party application like PreTeXt, if it did not want to carry along the baggage of WeBWorK/PG styling files.

It doesn't matter for PTX for now. So if you don't think that reasoning outweighs avoiding inline styling, I will change it.

drgrice1: I don't see why an inline style is needed because the other style is inline.  In general, inline styles should be avoided if at all possible.

If a system like PreTeXt does not want to use the intended rendering pipeline, it is that systems task to deal with handling style issues that arise as a result of doing so.  Ideally, there would only be one rendering system, and all that use PG problems would use that.  The point is to get what we render to work for everyone (with some options if needed).

drgrice1 (approved):
Looks good now.  I will approve this pending the unit test fix.

drdrew42 (approved):

pstaabp (approved):
👍

PR #999, opened 2024-02-01
Add single answer message to MultiAnswer problems using singleResult.

somiaj:
This allows MultiAnswer checkers to set a single answer message for all answer rules tied together using singleResult instead of having per answer part messages. This is done via the setMessage method: `$multianswer->setMessage(0, $message);`

This messages can be combined with per answer rule messages.

This came about from an instructor wanting to give feedback and partial credit for a MultiAnswer. This can be done via post filters, but is a little clunky with how MultiAnswers work. I thought others might also want to be able to set a message for the full singleResult vs per rule messages.

### DIFF
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -201,6 +201,14 @@ sub single_check {
 			. join('<TR><TD HEIGHT="4"></TD></TR>', @errors)
 			. '</TABLE>';
 	}
+	if ($self->{single_ans_message}) {
+		if ($ans->{ans_message}) {
+			$ans->{ans_message} = $ans->{error_message} =
+				'<DIV>' . $self->{single_ans_message} . '</DIV>' . $ans->{ans_message};
+		} else {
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
I think that this is okay.

However, it is not as useful as I would like.  Often in the process of grading the parts of a multi-answer problem there are numerous messages that are generated.  So I usually need to store the messages in an array, and then join them later.  I usually use a custom key in the answer hash for this, and then join the messages in a post filter.  The advantage of that approach is that you can `return` from the checker at any time, and still have access to the messages and add them to the answer hash `ans_message`.  If this method could be called with a new message, and it would just append the message to the existing messages it would be much more useful.

It is perhaps interesting to note that without this pull request, calling `$ma->setMessage(0, 'hello')` works, and does not die and say "Answer 0 is not defined" as you might think when you look at the code.  The reason it works is because `$self->{ans}[-1]` is what it checks is defined, and `-1` for a Perl array index means the last element of the array which is defined.

somiaj:
@drgrice1 would the approach in #1001 allow for your use case and be more versatile than this approach?

somiaj:
Closing in favor of #1001

PR #1000, opened 2024-02-02
repeated macro

Alex-Jordan:
Sometimes I want to check if there are repeats in some list. This macro will make that easy to code in a PG problem.

Use `!repeated(@array)` to test (boolean true or false) if `@array` has repeats. If it does and you actually want to see them, `repeated(@array)` will be the list of repeats. Subject to the special consideration noted in the POD.

### DIFF
--- a/macros/core/PGauxiliaryFunctions.pl
+++ b/macros/core/PGauxiliaryFunctions.pl
@@ -37,6 +37,7 @@ This macro creates the following functions that are available for PG:
     random_pairwise_coprime($ar1, $ar2, ... )
     random_coprime($ar1, $ar2, ... )
     random_subset($n, @set)
+    repeated(@list)
 =cut
 
 # ^uses loadMacros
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
Looks good.

drdrew42 (approved):
👍

macros/core/PGauxiliaryFunctions.pl
```diff
@@ -555,5 +556,32 @@ sub random_subset {
...
+However occasionally with MathObjects, x == y is true while y == x is false, and then
+only x (the one that makes the relation true when it is on the left) will be included
+in the returned list.
+
+=cut
+
+sub repeated {
+	my @return;
+	for my $x (@_) {
+		push(@return,$x)  if (grep { $_ == $x } (@_)) > 1;
```
somiaj, line 581: This is now failing perltidy tests here, there is a misplaced space: `push(@return, $x) if (grep { $_ == $x } (@_)) > 1;`.

PR #1001, opened 2024-02-03
Create addMessage method for parserMultiAnswer.pl (alternative to #999)

somiaj:
This is an alternative to #999 addressing the comments by @drgrice1. This adds a new method, `addMessage`, that can add multiple messages which will all be joined together in the end. This also uses a different method, so calling `setMessage(0, $message)` will still function like it use to and add a message to the last message rule.

### DIFF
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -55,20 +55,21 @@ sub new {
 		}
 	}
 	bless {
-		data              => [@data],
-		cmp               => [@cmp],
-		ans               => [],
-		isValue           => 1,
-		part              => 0,
-		singleResult      => 0,
-		namedRules        => 0,
-		checkTypes        => 1,
-		allowBlankAnswers => 0,
-		tex_separator     => $separator . '\,',
-		separator         => $separator . ' ',
-		tex_format        => undef,
-		format            => undef,
-		context           => $context,
+		data                => [@data],
+		cmp                 => [@cmp],
+		ans                 => [],
+		isValue             => 1,
+		part                => 0,
+		singleResult        => 0,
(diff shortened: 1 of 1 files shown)


somiaj:
Here is the test problem I was using for this.

```
DOCUMENT();
loadMacros('PGstandard.pl', 'PGML.pl', 'parserMultiAnswer.pl', 'PGcourse.pl');

$ma = MultiAnswer(1, 2, 3)->with(
    singleResult => 1,
    checker => sub {

        $self->setMessage('2', 'Message2');
        $self->addMessage('First single message');
        $self->addMessage('Second single message');
        return 0.5;
    }
);

BEGIN_PGML
a) Enter '1': [_]{$ma}

b) Enter '2': [_]{$ma}

c) Enter '3': [_]{$ma}
END_PGML
ENDDOCUMENT();
```

drgrice1:
I think that you misinterpreted my second comment in the other pull request.  That was not meant to point out behavior that I believe is correct or good, but rather to point out behavior that I think is contrary to the intention of the code (as I interpret it, although @drdrew42 may have meant otherwise).  I don't think that having the zero case set the message for the last answer is a good thing, and I think it is meant for the `die` statement to occur if an index does not match the existing answer indices counting from one.  It is just what happens with the code prior to these pull request (and with the other pull request if single result is not set).

drgrice1 (approved):
Although, this does address my first comment that I think to me is more important.  This is more how I envision doing this also.  I have been meaning to basically implement what you have in this pull request.

somiaj:
I wasn't thinking of the behavior of `0` in `setMessage` good or bad, but addressed that comment to keep current behavior.

pstaabp (approved):
Looks good.

PR #1003, opened 2024-02-04
num and den methods for Fraction objects

Alex-Jordan:
This adds `num` and `den` methods for a Fraction object as a slightly more convenient and much more readable way to access a fraction's parts compared with how I usually do it.

Ex:
```
Context("Fraction");
$f = Fraction(15,20);

TEXT("The denominator is ", $f->den, " and the numerator is ", "$f->num);
```


dpvc:
I don't have any problem with this.

pstaabp (approved):

Issue #1007, opened 2024-02-08
checkboxlist within a RadioMultiAnswer


Alex-Jordan:
I'm pushing the limits of `parserRadioMultiAnswer.pl` with the problem below. It renders fine, but there are two issues:

- When you have not selected the second radio button (which contains checkboxes), only the first checkbox is grayed out. I think the intent is that all would be grayed out until you select that radio button.
- When I submit the correct answer (which is indeed the correct answer as displayed by show correct answers) it is not accepted as correct.

Does it go beyond the scope of `parserRadioMultiAnswer.pl` to use checkboxes within an option?

```
# WeBWorK problem written by Alex Jordan 
DOCUMENT();

loadMacros(qw(
    PGstandard.pl PGML.pl PGtikz.pl parserCheckboxList.pl parserRadioMultiAnswer.pl
));

$x = list_random(-2,2);
$y = list_random(-2,2);

$image = createTikZImage();
$image->BEGIN_TIKZ
\fill[color=blue!20] (0,0) -| ($x,$y) -| cycle;
\draw[{stealth}-{stealth}] (-2,0) -- (2,0);
\draw[{stealth}-{stealth}] (0,-2) -- (0,2);
\draw[-{stealth}, ultra thick, color=blue] (0,0) -- ($x,0);
\draw[-{stealth}, ultra thick, color=blue] (0,0) -- (0,$y);
\node at ($x/2,$y/2) {\(H\)};
END_TIKZ

$checks = CheckboxList(
    [
        '\(\vec{0}\) is not in \(H\)',
        'There are vectors \(\vec{u}\) and \(\vec{v}\) in \(H\) with \(\vec{u}+\vec{v}\) not in \(H\).',
        'There is a vector \(\vec{u}\) in \(H\) and a scalar \(c\) in \(\mathbb{R}\) with \(c\vec{u}\) not in \(H\).',
    ],
    [2]
);

$rma = RadioMultiAnswer(
    [
        ['\(H\) is a subspace.'],
        ['\(H\) is not a subspace because (check all that apply): %s', $checks]
    ],
    1
)->with(labels => ['Yes','No']);

BEGIN_PGML
In the image below, a subset [`H`] of [`\mathbb{R}^2`] is shaded. Is [`H`] a subspace of [`\mathbb{R}^2`]?

>> [! R^2 with one quadrant and its boundary shaded !]{$image} <<

[_]{$rma}
END_PGML

ENDDOCUMENT();
```

[...]

PR #1011, opened 2024-02-10
Fix the feedback class elements for checkbox or radio part answers of a RadioMultiAnswer.

drgrice1:
This was observed while looking at issue #1007 (but is not that issue).

When a `RadioMultiAnswer` has checkbox or radio button parts, then the feedback class is only being added to the first checkbox or radio button.  This updates the `parserRadioMultiAnswer.pl` `feedback_options` method to look for these, and correctly mark those that are checked (or all of them if none are checked).

### DIFF
--- a/macros/parsers/parserRadioMultiAnswer.pl
+++ b/macros/parsers/parserRadioMultiAnswer.pl
@@ -423,7 +423,22 @@ sub cmp {
 				my $contents = $selected->first->parent->at('div.radio-content[data-radio]');
 				if ($contents) {
 					my $partNames = JSON->new->decode($contents->attr('data-part-names'));
-					push(@$selected, $contents->at(qq{[name="$_"]})) for (@$partNames);
+					for (@$partNames) {
+						my $ansRules = $contents->find(qq{[name="$_"]});
+						if (@$ansRules > 1) {
+							# IF there is more than one input with this name, then this is a radio or checkbox answer
+							# group.  So only add the feedback class to the checked inputs if any are checked.
+							# Otherwise add the feedback class to all of them.
+							my $selectedParts = $ansRules->grep(sub { exists $_->attr->{checked} });
+							if (@$selectedParts > 0) {
+								push(@$selected, @$selectedParts);
+							} else {
+								push(@$selected, @$ansRules);
+							}
+						} else {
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):
It's a bit strange to put classes on the individual checkboxes that report "correct" or "incorrect", when "correct" or "incorrect" actually refers to the collection of submitted checks. Like if the checkboxes are A,B,C,D, with A,B the correct choices, and a student submits A,C, is it right to tell them that their selection of A was "incorrect"?

I realize this observation applies to plain old CheckboxList answers too, not just ones inside a RadioMultiAnswer option. I am unsure how to communicate to the user "as a package deal, your set of selections is not correct, and this is not saying one thing or another about each individual selection you made".

I am noticing something else that is odd. In my example above, the student submitted A,C, and now A,C are still checked and have a red glow. The student unchecks C, and it's red glow goes away. Then they check D, and C's red glow comes back. That's odd. Then check C again, and the red glow goes away. More checking and unchecking of boxes sometimes toggles the red glow on C and sometimes does not. I haven't determined the pattern.

I'm approving this because I'm sure my observations are independent of the improvement this is making.

drgrice1:
The current behavior is to add the correct/incorrect classes to only the checkboxes that are checked if there are any checked, and to add the classes to all of them if none are checked.  The idea being that the styles say what you have checked as a group is correct or incorrect.  The only other option is to just add the classes to all checkboxes in any case.  In your example with answers A, B, C, and D with A and B correct and the student checks A and C, you certainly wouldn't want to mark C as incorrect and leave out A, or even worse, mark A as correct.  That gives away to much information.

Your last observation is just the focus behavior.  It is not that the red glow goes away, but rather that it is focused.  If you move focus elsewhere in the page while C is still unchecked, you will see that the glow is still there.  The correct/incorrect styles only apply while the inputs are not focused.  The focus styles override those.

pstaabp:
I was baffled by this, but chalk it up to another Safari weirdism/bug.  It doesn't seem like Safari is paying attention to these classes.  Here's an example (I'm using this PR) of Safari:
image

And the same in Chrome:

image

The browser inspect has detected all the correct classes, but is not displaying them.  👎

PR #1012, opened 2024-02-12
add elementary, permutation, and zero matrix constructors

Alex-Jordan:
The Matrix class already has an identity matrix constructor. For example: `Value:Matrix->I(4)`. Or if `$A` is already a 4x4 matrix, then `$A->I` will do the same.

This PR adds a few more constructors for special matrices, so you can just ask for them rather than have to build them element by element in a problem. I would like feedback on whether or not it is OK to put these directly into the Matrix class, instead of into some linear algebra macro file.

* Elementary scalar matrix:  `Value:Matrix->E(4, [2], 3)` makes a 4x4 matrix such that left multiplication with a 4xn matrix scales the 2nd row by 3.
* Elementary row swap matrix:  `Value:Matrix->E(4, [2,3])` makes a 4x4 matrix such that left multiplication with a 4xn matrix swaps the 2nd and third rows.
* Elementary row replacement matrix:  `Value:Matrix->E(4, [2,3], 5)` makes a 4x4 matrix such that left multiplication with a 4xn matrix replaces the 2nd row with the 2nd row added to 5 times the 3rd row.
* Permutation matrix: `Value:Matrix->P(4, [1,2,3])` makes a 4x4 matrix corresponding to the cycle (123).
* Permutation matrix: `Value:Matrix->P(4, [1,2,3], [1,4], ...)` makes a 4x4 matrix corresponding to the product of the given cycles.
* Zero matrix: `Value:Matrix->Zero(4, 5)` makes a 4x5 matrix of all zeros.
* Zero matrix: `Value:Matrix->Zero(4)` makes a 4x4 matrix of all zeros.

All of the constructors can be called as a method from an existing matrix like `$A`. However in most cases you still must supply all the arguments. The exception is with a zero matrix, where `$A->Zero` will give you a zero matrix with the same dimensions as `$A`.

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -494,6 +494,110 @@ sub I {
 }
 
 #
+#  Get an elementary matrix of the requested size and type
+#  E(n,[i,j])   nxn, swap rows i and j
+#  E(n,[i,j],k) nxn, replace row i with row i added to k times row j
+#  E(n,[i],k)   nxn, scale row i by k
+#
+sub E {
+	my ($self, $d, $rows, $k, $context) = @_;
+	$context = $self->context unless $context;
+	Value::Error("You must provide a dimension for an Elementary matrix") unless defined $d;
+	Value::Error("Dimension must be a positive integer")                  unless $d =~ m/^[1-9]\d*$/;
+	my @ij = @{$rows};
+	Value::Error("Either one or two rows must be specified for an Elementary matrix") unless (@ij == 1 || @ij == 2);
+	Value::Error(
+		"If only one row is specified for an Elementary matrix, then a number to scale by must also be specified")
+		if (@ij == 1 && !defined $k);
+	for (@ij) {
+		Value::Error("Row numbers must be integers between 1 and $d")
+			unless ($_ =~ m/^[1-9]\d*$/ && $_ >= 1 && $_ <= $d);
+	}
+	@ij = map { $_ - 1 } (@ij);
+
+	my @M    = ();
+	my @Z    = split('', 0 x $d);
+	my $REAL = $context->Package('Real');
+
+	foreach my $i (0 .. $d - 1) {
+		my @row = @Z;
+		$row[$i] = 1;
+		if (@ij == 1) {
+			$row[$i] = $k if ($i == $ij[0]);
+		} elsif (defined $k) {
+			$row[ $ij[1] ] = $k if ($i == $ij[0]);
+		} else {
+			($row[ $ij[0] ], $row[ $ij[1] ]) = ($row[ $ij[1] ], $row[ $ij[0] ]) if ($i == $ij[0] || $i == $ij[1]);
+		}
+		push(@M, $self->make($context, map { $REAL->new($_) } @row));
+	}
+	return $self->make($context, @M);
+}
+
+#
+#  Get a permutation matrix of the requested size
+#  E.g. P(3,[1,2,3])  corresponds to cycle (123) applied to rows of I_3
+#  And  P(6,[1,4],[2,4,6]) corresponds to cycle product (14)(246) applied to rows if I_6
+#
+sub P {
+	my ($self, $d, @cycles) = @_;
+	my $context = $self->context;
+	$d = ($self->dimensions)[0] if !defined $d && ref($self) && $self->isSquare;
+	Value::Error("You must provide a dimension for a Permutation matrix") unless defined $d;
+	Value::Error("Dimension must be a positive integer")                  unless $d =~ m/^[1-9]\d*$/;
+	for my $c (@cycles) {
+		Value::Error("Permutation cycles should be array references") unless (ref($c) eq 'ARRAY');
+		for (@$c) {
+			Value::Error("Permutation cycle indices must be integers between 1 and $d")
+				unless ($_ =~ m/^[1-9]\d*$/ && $_ >= 1 && $_ <= $d);
+		}
+		my %cycle_hash = map { $_ => '' } (@$c);
+		Value::Error("A permutation cycle should not repeat an index") unless (@$c == keys %cycle_hash);
+	}
+	my @M    = ();
+	my @Z    = split('', 0 x $d);
+	my $REAL = $context->Package('Real');
+
+	foreach my $i (0 .. $d - 1) {
+		my @row = @Z;
+		$row[$i] = 1;
+		push(@M, $self->make($context, map { $REAL->new($_) } @row));
+	}
+
+	for my $c (@cycles) {
+		my $n = @$c;
+		my $swap;
+		for my $i (0 .. $n - 1, 0) {
+			($swap, $M[ $c->[$i] - 1 ]) = ($M[ $c->[$i] - 1 ], $swap);
+		}
+	}
+
+	return $self->make($context, @M);
+}
+
+#
+#  Get an all zero matrix of the requested size
+#
+sub Zero {
+	my ($self, $m, $n, $context) = @_;
+	$context = $self->context unless $context;
+	$n       = $m                     if !defined $n && defined $m;
+	$m       = ($self->dimensions)[0] if !defined $m && ref($self);
+	$n       = ($self->dimensions)[1] if !defined $n && ref($self);
+	Value::Error("You must provide dimensions for the Zero matrix") unless defined $m          && defined $n;
+	Value::Error("Dimension must be a positive integer")            unless $m =~ m/^[1-9]\d*$/ && $n =~ m/^[1-9]\d*$/;
+	my @M    = ();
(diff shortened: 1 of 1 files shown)


pstaabp:
I have some updates to Matrix objects as well.  I started a discussion at #991 but no one weighed in.  I think this is a good idea--there was some discussion about moving more into the MathObjects directly instead of using some of the macros.

Alex-Jordan:
I did not know about this macro: https://webwork.pcc.edu/webwork2/pod/macros/math/MatrixReduce.pl

It has some of the same functionality here that I'm adding to lib.

pstaabp:
> I did not know about this macro: https://webwork.pcc.edu/webwork2/pod/macros/math/MatrixReduce.pl
> 
> It has some of the same functionality here that I'm adding to lib.

I've got a few things started on this as well.  Let's talk so we're not duplicating things.

lib/Value/Matrix.pm
```diff
@@ -493,6 +493,110 @@ sub I {
...
+		Value::Error("Row numbers must be integers between 1 and $d")
+			unless ($_ =~ m/^[1-9]\d*$/ && $_ >= 1 && $_ <= $d);
+	}
+	@ij = map { $_ - 1 } (@ij);
+
+	my @M    = ();
+	my @Z    = split('', 0 x $d);
+	my $REAL = $context->Package('Real');
+
+	foreach my $i (0 .. $d - 1) {
```
pstaabp, line 522: I think for consistency, we're trying to move foreach over to for. There are a few places here to do that.

Alex-Jordan:
I opted for consistency with the macro that was already here for making an identity matrix. Should I change that one too?

lib/Value/Matrix.pm
```diff
@@ -493,6 +493,110 @@ sub I {
...
+		}
+		push(@M, $self->make($context, map { $REAL->new($_) } @row));
+	}
+	return $self->make($context, @M);
+}
+
+#
+#  Get a permutation matrix of the requested size
+#  E.g. P(3,[1,2,3])  corresponds to cycle (123) applied to rows of I_3
+#  And  P(6,[1,4],[2,4,6]) corresponds to cycle product (14)(246) applied to rows if I_6
```
drgrice1, line 540:
```suggestion
#  And  P(6,[1,4],[2,4,6]) corresponds to cycle product (14)(246) applied to rows of I_6
```
I would also suggest ending the previous line with a comma, and start this line with "and" instead of "And".  And so you aren't starting a sentence with "and".

lib/Value/Matrix.pm
```diff
@@ -493,6 +493,110 @@ sub I {
...
+		push(@M, $self->make($context, map { $REAL->new($_) } @row));
+	}
+
+	for my $c (@cycles) {
+		my $n = @$c;
+		my $swap;
+		for my $i (0 .. $n - 1, 0) {
+			($swap, $M[ $c->[$i] - 1 ]) = ($M[ $c->[$i] - 1 ], $swap);
+		}
+	}
```
drgrice1, line 573: This should be
```suggestion
	for my $c (@cycles) {
		my $swap;
		for my $i (0 .. $#$c, 0) {
			($swap, $M[ $c->[$i] - 1 ]) = ($M[ $c->[$i] - 1 ], $swap);
		}
	}
```
Perl already saves the index of the last entry in an array.  So just use that instead of computing it with `@$c - 1` and saving it in `$n`.

lib/Value/Matrix.pm
```diff
@@ -493,6 +493,110 @@ sub I {
...
+	Value::Error("You must provide dimensions for the Zero matrix") unless defined $m          && defined $n;
+	Value::Error("Dimension must be a positive integer")            unless $m =~ m/^[1-9]\d*$/ && $n =~ m/^[1-9]\d*$/;
+	my @M    = ();
+	my @Z    = split('', 0 x $n);
+	my $REAL = $context->Package('Real');
+
+	foreach my $i (0 .. $m - 1) {
+		my @row = @Z;
+		push(@M, $self->make($context, map { $REAL->new($_) } @row));
+	}
```
drgrice1, line 596: This could be done quite concisely with
```suggestion
	my $REAL = $context->Package('Real');

	foreach my $i (0 .. $m - 1) {
		push(@M, $self->make($context, map { $REAL->new($_) } 0 .. $n - 1));
	}
```

drgrice1 (approved):
I think these are good additions.  I added a few code suggestions.

One problem with adding to this file (and a problem with MathObjects to begin with) is that there is no POD.  As such we have to rely on adding documentation to the wiki (or elsewhere).  A big project that is needed is to add (good) POD documentation to these files so that it is available in the POD that can be accessed from the PG problem editor.

pstaabp:
> I think these are good additions. I added a few code suggestions.
> 
> One problem with adding to this file (and a problem with MathObjects to begin with) is that there is no POD. As such we have to rely on adding documentation to the wiki (or elsewhere). A big project that is needed is to add (good) POD documentation to these files so that it is available in the POD that can be accessed from the PG problem editor.

I started doing this for this file when I added another method to this module.  After we get this in, I can add another PR to do this.  And clearly, we should do more with the other MathObjects.

Alex-Jordan:
Some of what you see might just be that I tried to keep the new methods consistent with the existing identity matrix method. But I'm going to go in now and make changes throughout.

drgrice1:
I see.  I didn't look around elsewhere in the file.

Alex-Jordan:
I just made updates as suggested by @pstaabp and @drgrice1. I made one additional change that mostly removes the need for `@row` altogether except for the elementary matrix method. @drgrice1 had suggested this for the zero method, and that made me realize it could be done for the identity and permutation methods too using a conditional operator.

One question I have (because now is the time) is if this format for the elementary matrix syntax can be improved upon. One of the existing macro libraries can make these matrices too, but it uses different methods to produce a row-scale matrix, a row-swap matrix, and the other kind. I wanted something where they all fall under the same method name (`E` is what I went with.) But this makes the syntax not very intuitive. Like `E(4, [2], 3)`, `E(4, [2,3])`, and `E(4, [2,3], 5)` in the PR opening post. If anyone can imagine having a more intuitive syntax for this, let's do that.

drgrice1:
Are you referring to the methods in the `MatrixReduce.pl` macro?  That has methods to create elementary matrices, and uses a flat argument list.  If you think that would be more intuitive, then you could switch to that with multiple methods.  I leave that to you.  I am not opposed to your current syntax though.

One thing I noticed is that some of these methods allow calling them with a `Value::Matrix` object.  For example, to obtain an identity matrix of the same size as one you are working with you would use
```perl
$matrix = Matrix([[1, 2, 3], [4, 5, 6], [7, 8, 9]]);
$I = $matrix->I;
```
However, the new `E` method does not.  Should that be adapted to do so?  Also, if the `P` method is used in this way, the arguments needed are rather counterintuitive to obtain a permutation matrix of the same size.  You need to call it as in the following example.
```perl
$matrix1 = Matrix([[1, 2, 3], [4, 5, 6], [7, 8, 9]]);
$matrix2 = $matrix1->P(undef, [2, 3]);
```

Alex-Jordan:
Now you can do all of these:

```
$A = Matrix([1,2,3],[4,5,6]);

$I = $A->I;
$I2 = Value::Matrix->I(3);

$E = $A->E([1],3);
$E2 = Value::Matrix->E(3,[1],3);

$P = $A->E([1,2]);
$P2 = Value::Matrix->P(3,[1,2,3]);

$Z = $A->Zero;
$Z2 = Value::Matrix->Zero(3);
$Z3 = Value::Matrix->Zero(3,4);
```

The zero matrix is the only one allowed to be nonsquare, so if generated from `$A` it will keep the dimensions of `$A`.

For the others, they must be square. Since these matrices (I, E, P) are most often used to multiply on the left, I made it so that if `$A` is mxn, these matrices (I, E, P) will be mxm. For example `$I` above is 2x2. This is a change in behavior for I, because previously it would not be allowed unless `$A` were square.

Alex-Jordan:
I have not touched POD, because I do not know what the greater POD scheme should be for files like this.

drgrice1:
Awesome.  You read my mind on allowing `E` and `P` to be non-square.  Changing that for `I` makes sense, and I don't see it being a problem that this is a change in behavior from before.

The POD can be left for later.

pstaabp:
Got around to testing again. I think the basic API for these functions is good.  It would be nice to not have to do the `Value::Matrix->E` for example.  Should we provide the shortcut `E` or `Elem` as well?

pstaabp (approved):

Alex-Jordan:
Note that you can do things like:

```
$I = Matrix(1)->I(2);
```
to avoid doing:
```
$I = Value::Matrix->I(2);
```

PR #1016, opened 2024-02-16
Fix uninitialized value + small fixes (hotfix)

taniwallach:
Was being triggered by a Hebrew problem.
Line 2096 is probably the most critical fix.
Four changes provide an initial/fallback value.
Along the way, I also made changes to have the empty string and single space strings be defined using single quotes.

I'm relatively certain that one of the problems triggering such errors is the one attached.
[GivenSmallTable.pg.txt](https://github.com/openwebwork/pg/files/14312457/GivenSmallTable.pg.txt)

```
Use of uninitialized value $in in substitution (s///) at line 2101 of [PG]/macros/core/PGbasicmacros.pl.
Use of uninitialized value $in in substitution (s///) at line 2102 of [PG]/macros/core/PGbasicmacros.pl.
Use of uninitialized value $in in substitution (s///) at line 2105 of [PG]/macros/core/PGbasicmacros.pl.
Use of uninitialized value $in in substitution (s///) at line 2106 of [PG]/macros/core/PGbasicmacros.pl.
Use of uninitialized value $in in substitution (s///) at line 2107 of [PG]/macros/core/PGbasicmacros.pl.
Use of uninitialized value $in in substitution (s///) at line 2108 of [PG]/macros/core/PGbasicmacros.pl.
Use of uninitialized value $in in substitution (s///) at line 2109 of [PG]/macros/core/PGbasicmacros.pl.
Use of uninitialized value $in in substitution (s///) at line 2128 of [PG]/macros/core/PGbasicmacros.pl.
Use of uninitialized value $in in substitution (s///) at line 2132 of [PG]/macros/core/PGbasicmacros.pl.
Use of uninitialized value $in in substitution (s///) at line 2133 of [PG]/macros/core/PGbasicmacros.pl.
Use of uninitialized value $in in pattern match (m//) at line 2136 of [PG]/macros/core/PGbasicmacros.pl.
Use of uninitialized value $in in concatenation (.) or string at line 2139 of [PG]/macros/core/PGbasicmacros.pl.
Use of uninitialized value $in in concatenation (.) or string at line 2143 of [PG]/macros/core/PGbasicmacros.pl.
```

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -33,7 +33,7 @@ BEGIN {
 	be_strict;
 }
 
-my $displayMode;
+my $displayMode = '';
 
 my (
 	$PAR,               $BR,       $BRBR,    $LQ,                  $RQ,
@@ -1658,7 +1658,7 @@ sub processDivSpanOptions {
 		%{$option_ref},
 	);
 
-	my $LangVal = "";
+	my $LangVal = '';
 	if ($options{allowLang} && defined($options{lang})) {
 		# The standard for how the lang tag should be set is explained in
 		# https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang
@@ -1677,7 +1677,7 @@ sub processDivSpanOptions {
 		}
 	}
 
-	my $DirVal = "";
+	my $DirVal = '';
 	if ($options{allowDir} && defined($options{dir})) {
 		# the ONLY allowed values are: ltr rtl auto
 		if (($options{dir} eq "ltr")
@@ -1692,7 +1692,7 @@ sub processDivSpanOptions {
 		}
 	}
 
-	my $ClassVal = "";
+	my $ClassVal = '';
 	if ($options{allowClass} && defined($options{class})) {
 		# Class names which are permitted here must start with a letter [A-Za-z]
 		# and the rest of the class name can be characters in [A-Za-z0-9\-\_].
@@ -1729,7 +1729,7 @@ sub processDivSpanOptions {
 		}
 	}
 
-	my $StyleVal = "";
+	my $StyleVal = '';
 	if ($options{allowStyle} && defined($options{style})) {
 		# The value is validated in a very minimal sense only - use with great care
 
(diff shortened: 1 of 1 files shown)


macros/core/PGbasicmacros.pl
```diff
@@ -1773,7 +1773,7 @@ sub openDivSpan {
 		return ();
 	}
 	my $option_ref = {};
```
drgrice1, line 1775: This needs to be
```suggestion
	my $option_ref   = {};
```
for perltidy.  This is needed because you added the `= ''` on the line below, and so now this needs to line up with that.

macros/core/PGbasicmacros.pl
```diff
@@ -2093,6 +2093,7 @@ sub old_safe_ev {
 
 sub FEQ {    # Format EQuations
 	my $in = shift;
+	return '' if ( ! defined($in) ); # No need to do all the string replacements in this case
```
drgrice1, line 2096: This should be
```suggestion
	return '' if !defined $in;    # No need to do all the string replacements in this case

```
The spacing is incorrect for perltidy, and the parentheses aren't needed.  The new empty line is needed to ensure that the comment on the next line is not indented to match the comment you added.  perltidy will do that.  Generally comments at the end of a line aren't the best practice and shouldn't be used except for very short identifier comments in lists.

macros/core/PGbasicmacros.pl
```diff
@@ -33,7 +33,7 @@ BEGIN
 	be_strict;
 }
 
-my $displayMode;
+my $displayMode = '';
```
drgrice1, line 36: This isn't really needed.  This variable is always initialized on line 61 in any case.  The important thing is that the `$displayMode` in `main` is defined (set on line 80 of PG.pl), or that will override this anyway and make this `$displayMode` variable be undefined.

drgrice1 (commented):
The changes look fine.  Just some perltidy issues to fix.

drgrice1:
I was looking at this again, and I think that you should drop the clean up changes in this (like changing double quotes to single quotes).  That is not appropriate for a hotfix pull request.  If you want to make those changes, make a pull request to develop with that.  I noticed that one of those changes conflicts with develop where that has already been done (with other changes).

taniwallach:
I'll clean this up soon.
Note: I do have some log entries with errors from `$displayMode` undefined. I'll need to check that more carefully.
```
Use of uninitialized value $displaystyle in concatenation (.) or string at line 69 of (eval 4469)
   from within main::EV3P called at line 77 of Technion/ode/DirectionsField1stOrder/q01_q02/he/q01_v04.pg
```

taniwallach:
I am closing this.

When I look at the problem I had noted as causing the warnings with WW 2.19 RC - I do not get the warnings. I'm not sure why not - but if the warnings are not being triggered, there seems no good reason to add the main change on line 2096 which would run the test on every call of `FEQ`. If the problem comes up when I move my production servers to WW 2.19 - I can revisit this matter.

PR #1020, opened 2024-02-20
Fix Value::Error calls in a checker for a RadioMultiAnswer object.

drgrice1:
This fixes the issue observed in https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8489.  See the problem posted there for a problem to test this with. There were two problems really that caused the issue.  First `$self->{ans}` is a doubly nested array.  So `$self->{ans}[0]` won't be an answer evaluator, but rather an array of answer evaluators.  That should have been `$self->{ans}[0][0]`.  However, that still won't work in this case, because for the problem in that forum post there are no answers in the first part.  Thus the `$self->{ans}[0]` is the emtpy array, and so `$self->{ans}[0][0]` is undefined, and still not an answer evaluator. So another approach was taken, and the context error is just added via the RadioMultiAnswer `appendMessage` instead.

This also adds the context swapping added for MultiAnswer objects in pull request #632.  The same issues that led to that pull request can occur for a RadioMultiAnswer.

Furthermore, this fixes an issue in the changes from #632.  That is that the `$ans` used on line 307 of parserMultiAnswer.pl is a variable that is not declared. There is no error for this because `strict` is not (yet) in effect for this macro.

### DIFF
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -303,12 +303,12 @@ sub perform_check {
 	$rh_ans->{isPreview} = $inputs->{previewAnswers}
 		|| ($inputs_{action} && $inputs->{action} =~ m/^Preview/);
 
-	Parser::Context->current(undef, $context);                                 # change to multi-answer's context
-	my $flags = Value::contextSet($context, $self->cmp_contextFlags($ans));    # save old context flags
-	$context->{answerHash} = $rh_ans;                                          # attach the answerHash
+	Parser::Context->current(undef, $context);                                    # change to multi-answer's context
+	my $flags = Value::contextSet($context, $self->cmp_contextFlags($rh_ans));    # save old context flags
+	$context->{answerHash} = $rh_ans;                                             # attach the answerHash
 	my @result = Value::cmp_compare([@correct], [@student], $self, $rh_ans);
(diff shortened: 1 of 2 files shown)


Alex-Jordan (approved):
Thanks for this, I tested it with a few exercises I wrote today that use RadioMultiAnswer and it fixed them without having to stop using `Value::Error`.

I didn't test anything with MultiAnswer.

Alex-Jordan:
Something came up for me while testing this that is unrelated to this issue and its fix, but I will post about here just in case it sparks an idea for an additional commit for this PR. It is an issue we have seen versions of before: when a math object getting reused as an answer. Imagine a question that goes something like the following. Forgive any issues with the actual pedagogy/mathematics.

```
$f = Formula("1/x");
$I = Interval('[-1, 1]`);
...
Why is it inappropriate to ask for the Riemann integral \int_{-1}^1 $f dx?

a. 'Actually, that's completely appropriate.'
b. 'Because %s is not continuous on %s.',   $f, $I
c. 'Because %s is not differentiable on %s.',   $f, $I

```

The point is there are reasonable reasons to reuse `$f` and `$I` among the answer blanks internal to button selections. I had something like this (but more complicated) where it was just not working correctly, and I found that if I replaced the latter answers with like `$f->copy, $I->copy` then it started working as I expected it to.

PR #922, opened 2023-08-31
Update numerical methods

pstaabp:
This adds some functionality to the macro `PGnumericalmacros.pl` as well as includes a test suite for the macro.  Also, cleaned up the code and improved the documentation. 

Can provide some problems if needed.

PR #1022, opened 2024-02-20
Clean up numerical macros code and improve documentation

pstaabp:
This does some basic code cleanup of the `PGnumericalmacros.pl` macro, improves documentation and adds a file of unit tests.  

This replaces #922.  There is new code in #922 that will be introduced in another PR.

### DIFF
--- a/macros/math/PGnumericalmacros.pl
+++ b/macros/math/PGnumericalmacros.pl
@@ -17,7 +17,13 @@
 
 Numerical methods for the PG language
 
-=head2 Interpolation  methods
+=cut
+
+BEGIN { strict->import; }
+
+sub _PGnumericalmacros_init { }
+
+=head2 Interpolation methods
 
 =head3 Plotting a list of points (piecewise linear interpolation)
 
@@ -30,13 +36,9 @@ Usage:
     plot_list([x0,x1,x2...], [y0,y1,y2,...]);
     plot_list(\@xarray,\@yarray);
 
-
+It is important that the x values in any form are unique or this method fails.  There is no
+check for this however.
 =cut
-
-BEGIN { strict->import; }
-
-sub _PGnumericalmacros_init {
-}
 
 sub plot_list {
 	my ($xref, $yref) = @_;
@@ -50,10 +52,7 @@ sub plot_list {
 	}
 	my (@x_vals, @y_vals);
 	unless (defined($yref)) {    #with only one entry we assume (x0,y0,x1,y1..);
-		if (@$xref % 2 == 1) {
-			die "ERROR in plot_list -- single array of input has odd number of
-        	elements";
-		}
+		die "ERROR in plot_list -- single array of input has odd number of elements" if (@$xref % 2 == 1);
 
 		my @in = @$xref;
 		while (@in) {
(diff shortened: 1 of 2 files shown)


Alex-Jordan (approved):
I read the changes to `macros/math/PGnumericalmacros.pl` and I do not see any errors introduced from the changes.

I did not look into the new testing file.

macros/math/PGnumericalmacros.pl
```diff
@@ -17,7 +17,13 @@ =head1 NAME
...
-=head2 Interpolation  methods
+=cut
+
+BEGIN { strict->import; }
+
+sub _PGnumericalmacros_init { }
+
+=head2 Interpolation methods
 
 =head3 Plotting a list of points (piecewise linear interpolation)
```
drgrice1, line 28: This header is to long, and it would be better if it were just the method name.  The current header could be a remark below.  So make this something like
```suggestion
=head3 plot_list

Plot a list of points (piecewise linear interpolation).
```
So throughout this file, all of the `=head3` lines should just be the method name.  Then when a problem author is trying to find a method and what it does, it will be listed in the sidebar in the generated POD.  In general this is the scheme that I would like to work toward in the POD for all of the macro files.

macros/math/PGnumericalmacros.pl
```diff
@@ -64,48 +63,53 @@ sub plot_list {
...
+    &$fn(x) = q0 + q1*(x-x0) +q2*(x-x1)*(x-x0) + ...;
+
+Generates a subroutine which evaluates a polynomial passing through the points
+C<(x0,q0), (x1,q1), (x2, q2)>, ... using Horner's method.
 
-Generates a subroutine which evaluates a polynomial passing through the points C<(x0,q0), (x1,q1),
-... > using Horner's method.
+The array refs for C<x> and C<q> can be any length but must be the same length.
+
+=head4 Example
```
drgrice1, line 100: This shouldn't be a header at all.  Change this to
```suggestion
For example,
```
Do the same with the other `=head4 Example` on line 570.  Those two headers are out of place in the sidebar.  Examples shouldn't be headers in any case.

drgrice1 (approved):
I think this kind of clean up is good.

Here are some suggestions that I think should be done if the point is to clean up this file.

You fixed some perlcritic related issues, but there are some others that would be good to fix.  Those are as follows.

Change the arguments of the `javaScript_cubic_spline` starting on line 336 to `my ($x_ref, $y_ref, %options) = @_;`, and change the arguments of the `rungeKutta4` method to `my ($rf_fun, %options) = @_;`.

Give the `javaScript_cubic_spline` an explicit return value by changing line 393 to `return $output_str;`.

Use a negative array index on line 345.  So change that line to `rlimit => $x_ref->[-1],`.

I also have given some suggestions for the cleaning up the POD.

By the way, here is a good perlcritic rc file to use for macros.  With my suggestions above and this rc file, perlcritic gives this file a clean bill of health.
```
severity = 4

# Can not "use safe" or "use warnings" in the safe compartment.
[-TestingAndDebugging::RequireUseStrict]
[-TestingAndDebugging::RequireUseWarnings]
[-Community::StrictWarnings]
[-Freenode::StrictWarnings]

# Macros commonly have packages that don't match the filename, have multiple packages, and contain code outside of
# package delclarations.
[-Modules::RequireExplicitPackage]
[-Modules::ProhibitMultiplePackages]
[-Modules::RequireFilenameMatchesPackage]

# Macros and problems use $a and $b all of the time.
[-Community::DollarAB]
[-Freenode::DollarAB]
[Variables::RequireLocalizedPunctuationVars]
allow = $a $b @a @b
```

drgrice1:
By the way, the test file looks good.

pstaabp:
Updated the code suggested by @drgrice1 with updates to all of the POD to include the method names. 

Should we include the `.perlcriticrc` file you have above in the repository?

macros/math/PGnumericalmacros.pl
```diff
@@ -384,12 +390,12 @@ sub javaScript_cubic_spline {
...
 END_OF_JAVA_TEXT
 
-	$output_str;
+	return $output_str;
 }
 
 =head2 Numerical Integration methods
 
-=head3 Integration by Left Hand Sum
+=head3 lefthandsum (Left Hand Riemann Sum)
```
drgrice1, line 398: Please move the parenthetical below.  So make this
```suggestion
=head3 lefthandsum

 Left Hand Riemann Sum
```
Either that or just drop it.  The method name is self explanatory, and the explanation below tells what it is anyway.

Do the same for `righthandsum` and `inv_romberg`.  The parentheticals don't look good in the sidebar.

drgrice1:
You have addressed the issues I asked to be dealt with, and this now has two approvals.  So I will merge it.

As to including the `.perlcriticrc` file, I think that would be good.  However, this is actually my `.pg-perlcriticrc` file, and I have another `.perlcriticrc` file.  The first is for pg problem files and macro files, and the second is for `lib` files.  Different standards are needed because pg problem files and macro files are evaluated in the safe compartment, and `lib` files are not.

PR #1028, opened 2024-02-29
force list context when random_subset used in dprecated macros

Alex-Jordan:
`random_subset` can return an array reference (instead of an array) if the context is scalar. The deprecated macros `NchooseK` and `shuffle` only ever returned lists. With 2.18, their code was changed to use `random_subset`. Now the problem is that with like:

`$a = $myarray[NchooseK(2,1)];`

the `NchooseK(2,1)` is returning an array ref instead of an array with one number in it. And `$a` comes out undefined.

This commit forces list context, so it restores the original behavior. Perhaps this should be considered for a hotfix to 2.18 because it broke at least one PG problem in a Runestone subscriber's course. It's not an OPL problem, and I am unsure how extensive the issue is. That is certainly not a recommended way to randomly choose an array index for a scalar assignment.


drgrice1 (approved):
Looks good.

somiaj (approved):
Looks good to me too, and I think a hot fix is fine to ensure these deprecated functions work as they use to.

pstaabp:
I would agree, let's do a hot fix.

drgrice1:
I agree that a hotfix would be good for this.  I can create a branch for that unless you want to do it.

PR #1029, opened 2024-03-06
Matrix boolean

Alex-Jordan:
This builds on #1012, adding some boolean methods: `isUpperTriangular`, `isLowerTriangular`, `isDiagonal`, `isSymmetric`, `isREF`, `isRREF`, and `isOrthogonal`. In some places, the code could have been shorter and more elegant, but I tried to lean toward efficiency. For example, `isSymmetric` could have simply asked if `$self == $self->transpose`, but I suspect what is here is a little faster.

These will help with custom answer checkers in a linear algebra course. For example asking a student to orthogonally diagonalize a matrix.

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -494,6 +494,110 @@ sub I {
 }
 
 #
+#  Get an elementary matrix of the requested size and type
+#  E(n,[i,j])   nxn, swap rows i and j
+#  E(n,[i,j],k) nxn, replace row i with row i added to k times row j
+#  E(n,[i],k)   nxn, scale row i by k
+#
+sub E {
+	my ($self, $d, $rows, $k, $context) = @_;
+	$context = $self->context unless $context;
+	Value::Error("You must provide a dimension for an Elementary matrix") unless defined $d;
+	Value::Error("Dimension must be a positive integer")                  unless $d =~ m/^[1-9]\d*$/;
+	my @ij = @{$rows};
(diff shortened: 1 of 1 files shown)


pstaabp:
Overall, this looks good.  I think we do need to add POD to this and I would recommend a test file, but I have started one and can add it later.

pstaabp (approved):

drgrice1 (commented):
This looks good other than the lack of the `min` method.

lib/Value/Matrix.pm
```diff
@@ -291,6 +291,119 @@ sub isZero {
...
+#  See if the matrix is triangular, diagonal, symmetric, orthogonal
+#
+
+sub isUpperTriangular {
+	my $self = shift;
+	my @d    = $self->dimensions;
+	return 1 if scalar(@d) == 1;
+	return 0 if scalar(@d) > 2;
+	for my $i (2 .. $d[0]) {
+		for my $j (1 .. min($i - 1, $d[1])) {
```
drgrice1, line 304: There is not `min` method in the `Value::Matrix` package.  So this causes an error.  I am not sure that there is a `min` method to pull in, so you might need to implement it directly.  You could change this to
```suggestion
		for my $j (1 .. ($i - 1 < $d[1] ? $i - 1 : $d[1])) {
```

Alex-Jordan:
> This looks good other than the lack of the min method.

I committed your suggestion for doing it directly.

drgrice1 (approved):
Looks good now!

drgrice1:
This can be merged I think.

PR #1036, opened 2024-03-19
Close a feedback popover when escape is pressed.

drgrice1:
This only works if either the button or the popover is focused, and only closes that popover, not all popovers that are open on the page.  If the button or popover is not focused, then escape will not close it.

This is to address https://github.com/openwebwork/webwork2/issues/2362 (which was closed). This is a variant of what I think was desired in that issue that I think I could tolerate.  I don't want a global key handler for the page that closes all popovers when escape is pressed. There are several issues with global key handlers like that.  They can interfere with other things on the page for one thing. They often end up resulting in unintended consequenses.

### DIFF
--- a/htdocs/js/Feedback/feedback.js
+++ b/htdocs/js/Feedback/feedback.js
@@ -18,6 +18,12 @@
 			});
 		}
 
+		const escapeClose = (e) => {
+			if (e.key === 'Escape') feedbackPopover.hide();
+		};
+
+		feedbackBtn.addEventListener('keydown', escapeClose);
+
 		feedbackBtn.addEventListener('shown.bs.popover', () => {
 			// Execute javascript in the answer preview.
 			feedbackPopover.tip?.querySelectorAll('script').forEach((origScript) => {
@@ -27,6 +33,8 @@
 				origScript.parentNode.replaceChild(newScript, origScript);
 				setTimeout(() => feedbackPopover.update());
 			});
+
+			feedbackPopover.tip?.addEventListener('keydown', escapeClose);
 
 			const moveToFront = () => {
 				if (feedbackPopover.tip) feedbackPopover.tip.style.zIndex = 18;


Alex-Jordan:
Thanks for this, and I'd like to take the discussion back to the staff here and make sure they understand some things that would not have been obvious to them. One thing they may have not considered is that it will be common for a user to have multiple popovers open at the same time. That might change things.

One time I witnessed a blind staff member do some WeBWorK testing. He had a masters degree in engineering or physics (I can't recall) which is to say that he was experienced with reading technical content. At the time, the WeBWorK feedback table had an extended table description (via a `summary` attribute on the table). I remember (a) how he was very impressed with the description he was getting about this table, (b) that he did not seem to know about the inner working of things like a "summary attribute on an HTML table", and (c) that I personally felt the "summary" attribute was being abused in a technical sense, giving a lot of instructions to the user that went beyond just summarizing the table.

My point is the AT users don't always think in terms of what they are technically dealing with. "Exercise feedback" might taste more like a "modal dialogue interaction pattern" than it tastes like a "popover interaction pattern",  in some hard-to-describe way that makes more sense to a regular AT user. Maybe if I circle back to the team here I can get some tips on making it "taste" more like a "popover interaction pattern" in the first place.

somiaj (approved):
I tested that the only the focused popup closed when escape is pressed, and if no popup has focus, nothing is closed with escaped. Works as described.

I have no comment on the accessibility side of things, only tested functionality.

pstaabp (approved):
Tested that this does close popovers with focus with esc, but only if it has focus.

Alex-Jordan:
Closing this now. @drgrice1 is right that we shouldn't do this, and our disability services staff suggest that we have something like an "accessibility tips" help page somewhere that could explain how to dismiss the popovers as part of a larger help page. I'll think about making such a help page separately.

PR #1037, opened 2024-03-19
Accessibility tweak for the MathQuill toolbar.

drgrice1:
This adds the `aria-hidden` attribute to the span containing the static MathQuill element in the toolbar buttons, and changes the `data-bs-title` attribute of the button itself to a `title`. I don't know if the latter change makes much difference since bootstrap will move the title from the button to the tooltip in any case.  But adding `aria-hidden` will prevent the contents of the span from being read by screen readers.  The title of the tooltip or button should be read instead.

Maybe this will work to address https://github.com/openwebwork/mathquill/issues/24?

### DIFF
--- a/htdocs/js/MathQuill/mqeditor.js
+++ b/htdocs/js/MathQuill/mqeditor.js
@@ -332,10 +332,11 @@
 				button.classList.add('symbol-button', 'btn', 'btn-dark');
 				button.dataset.latex = buttonData.latex;
 				button.dataset.bsToggle = 'tooltip';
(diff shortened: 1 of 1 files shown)


pstaabp:
This looks reasonable.  @Alex-Jordan can your accessibility team retest this PR?

Alex-Jordan:
I think more will be needed. I will be at the meeting a bit late and can demo what is missing.

Alex-Jordan (approved):

pstaabp (approved):

PR #1038, opened 2024-03-19
have aria-label on feedback button indicate when there is a message

Alex-Jordan:
This changes the aria label on a feedback button to be like "Incorrect with message" when there is a message. This is for people who can't see the little yellow pip in the upper corner.

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -1186,53 +1186,37 @@ sub ENDDOCUMENT {
 
 				# Create the screen reader only span holding the aria description, create the feedback button and
 				# popover, and insert the button at the requested location.
-				my $feedback = (
-					# Add a visually hidden span to provide feedback to screen reader users immediately.
-					$showResults || ($rh_envir->{showMessages} && $ansHash->{ans_message})
-					? Mojo::DOM->new_tag(
-						'span',
-						class => 'visually-hidden',
-						id    => "ww-feedback-$answerLabel",
-						sub {
-							($showResults ? Mojo::DOM->new_tag('span', $options{resultTitle}) : '')
-								. ($rh_envir->{showMessages} && $ansHash->{ans_message}
-									? ' ' . Mojo::DOM->new_tag('span', $ansHash->{ans_message})
-									: '');
-						}
-						)->to_string
-					: ''
-					)
-					. Mojo::DOM->new_tag(
-						'button',
-						type  => 'button',
-						class => "ww-feedback-btn btn btn-sm $options{btnClass} $options{btnAddClass}"
+				my $feedback = Mojo::DOM->new_tag(
+					'button',
+					type  => 'button',
+					class => "ww-feedback-btn btn btn-sm $options{btnClass} $options{btnAddClass}"
 						. ($rh_envir->{showMessages} && $ansHash->{ans_message} ? ' with-message' : ''),
-						'aria-label' => (
-							$rh_envir->{showMessages}
-							&& $ansHash->{ans_message} ? maketext('[_1] with message', $options{resultTitle})
-							: $options{resultTitle}
-						),
-						data => {
-							bs_title               => $options{resultTitle},
-							bs_toggle              => 'popover',
-							bs_trigger             => 'click',
-							bs_placement           => 'bottom',
-							bs_html                => 'true',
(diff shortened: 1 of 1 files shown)


macros/PG.pl
```diff
@@ -1207,8 +1207,9 @@ sub ENDDOCUMENT {
 						type  => 'button',
 						class => "ww-feedback-btn btn btn-sm $options{btnClass} $options{btnAddClass}"
 						. ($rh_envir->{showMessages} && $ansHash->{ans_message} ? ' with-message' : ''),
-						'aria-label' => $options{resultTitle},
-						data         => {
+						'aria-label' => $options{resultTitle}
+						. ($rh_envir->{showMessages} && $ansHash->{ans_message} ? ' with message' : ''),
```
drgrice1, line 1211: This should be translated.  If you need help with that, let me know.

Alex-Jordan: Is it just ` maketext(' with message')`? Or should it encompass the `$options{resultTitle}` too?

drgrice1: It should include the title.  So this should be `maketext('[_1] with message', $options{resultTitle})` in the case that there is a message.

drgrice1:
Although, there is something else that seems wrong with this change.  The message that this `aria-label` is being made to refer to is also added in a screen reader only span also with the result title (on lines 1192-1201).  So a screen reader user would already have heard the message that they are now being told is yet to come since that is before this in the DOM.  So should that screen reader only span be removed with this change?

Alex-Jordan:
This is a sidebar. In testing things for this, I discovered I needed to restart `webwork2` to see the edits to `macros/PG.pl` take effect. Is that how it should be? In my head, anything in `macros/` (as opposed to `lib/`) is not part of the startup process.

drgrice1:
Nothing in `macros/` requires a restart to `webwork2` with the exception of `macros/PG.pl`.  The safe compartment is initialized and cached at the time that the server starts, and that precompiles `macros/PG.pl`.  Both `webwork2` and the standalone renderer use this process now.  This is why when you are using `morbo` it hot reloads when there is a change to `macros/PG.pl` (but that is not done for any other macro).

drgrice1:
I highly recommend that you try to get a local setup of webwork2 using morbo for development.  It is highly more efficient than trying to develop in the production environment with hypnotoad.

Alex-Jordan:
The screen reader user moves past the answer blank and hears a feedback message like "Incorrect Your answer is not a reduced fraction" (or whatever). Then they move into the feedback button where they hear something like "Incorrect with message, button". Is that really a problem? They have already heard the message, and are also being told that if they activate the button they can hear it again.

Alex-Jordan:
BTW, "Incorrect Your answer is not a reduced fraction" is actually `<span>Incorrect</span><span>Your answer is not a reduced fraction</span>`. with no space between the first tow words. The span delimiters get the screen reader to read it correctly, but in a visualization (like from VoiceOver) you see "IncorrectYour answer is not a reduced fraction".

If there is no objection, I will add a space between the two spans as part of this PR.

drgrice1:
> The screen reader user moves past the answer blank and hears a feedback message like "Incorrect Your answer is not a reduced fraction" (or whatever). Then they move into the feedback button where they hear something like "Incorrect with message, button". Is that really a problem? They have already heard the message, and are also being told that if they activate the button they can hear it again.

That is fine I guess.  I just don't understand why they need both.

As to your question about the space ... yes, go ahead and add a space.

Alex-Jordan:
I think it will raise issues for an accessibility tester if there is a visual signal of something on the button (the pip), but there is no corresponding signal for the AT user.

drgrice1:
That is why I suggested removing the screen reader only span with the title and message.  It seems to me that doing so makes things more equal for those using a screen reader to the rest of us.  Why do they get the title and message before requesting it?

Alex-Jordan:
OK, this has that space and translation on the aria.

Alex-Jordan:
Oh I missed that suggestion. I think it would be good to remove it. If you are not looking for the feedback message, I can see it being unwanted and having it put in front of you as an intrusion. I'll remove it.

drgrice1 (approved):

pstaabp (approved):

PR #1039, opened 2024-03-19
make title of preview feedback button be 'Preview Information'

Alex-Jordan:
This is to avoid confusion for an AT user with the Preview button (the button alongside the Submit button) and the button used to reveal preview feedback. Maybe there is a better way to do this, or a better phrase than "Preview Information", but this is a start.

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -817,12 +817,12 @@ The options that can be modified are as follows.
 =item *
 
 C<resultTitle>: This is the title that is displayed in the feedback popover for
-the answers in the response group. By default this is "Preview", "Correct",
-"Incorrect", or "n% correct", depending on the status of the answer and the type
-of submission. Those strings are translated via C<maketext>. Usually this should
-not be changed, but in some cases the default status titles are not appropriate
-for certain types of answers. For example, the L<PGessaymacros.pl> macros
-changes this to "Ungraded" for essay answers.
+the answers in the response group. By default this is "Answer Preview",
+"Correct", "Incorrect", or "n% correct", depending on the status of the answer
+and the type of submission. Those strings are translated via C<maketext>.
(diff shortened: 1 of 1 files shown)


drgrice1:
I am not exactly sure what the confusion is.  The button says "Preview My Answers".  Then this says "Preview".  That seems like a natural progression to me.  First the student enters an answer and then comes to the "Preview My Answers", so it seems that then they would be searching for that "Preview" in the next page that loads.

drgrice1:
Or perhaps change feedback title to "Answer Preview".  So the student enters an answer and then comes to the "Preview My Answers", then in the next page that loads they would be looking for that "Answer Preview".

It just seems to me that the title of "Preview Information" is misleading when in most cases all that is in the popover is a preview of the entered answer.  There isn't really any information there.

drgrice1:
It also seems that in combination with #1038, this becomes even worse.  With that the `aria-label` on the feedback button will become "Preview Information with message" in the case that there is a feedback message.  It seems that the case that there is a feedback message is the only case there would actually be information in the popover, but that is already indicated by "with message" that is added.  So "Answer Preview" (or perhaps "Preview of Answer" or even "Preview of Entered Answer"?) would be better.  Then you have "Answer Preview with message" in the case there is a message.

Alex-Jordan:
I think it has to do with a user who surveys all the buttons on the page before they are actually navigating the page. And is possibly trying to jump to a particular button by its name. I do like `Answer Preview` better than `Preview Information`. That's just what I could think of at the time.

pstaabp:
@Alex-Jordan Should we switch the name to "Answer Preview"?  I think then, this looks good.

Alex-Jordan:
This now has "Answer Preview" for the button text.

drgrice1 (approved):

pstaabp (approved):

PR #1040, opened 2024-03-21
mechanism to declare many rows within one row in niceTables.pl

Alex-Jordan:
This lets you declare many rows of a niceTables table all at once within one row of the table, using a `rows` attribute. Consider this sample code, with both a PGML marked up table, as well as a `DataTable` within a PGML command structure:

```
$rows = [ map {[ $_, $_**2 ]} (1..4)];

BEGIN_PGML

[@ DataTable(
    [
        ['x', 'y'],
        [{rows => $rows}],
        ['x', 'y'],
        [{rows => [ map {[ $_, $_**3 ]} (1..4)]}],
        ['x', 'y'],
    ]
) @]*

[#
    [. [`x`] .] [. [`y`] .]*
    [. .]*{rows => $rows}
    [. [`x`] .] [. [`y`] .]
#]

END_PGML
```

The space character (or any nonzero amount of whitespace) in the `[. .]` is necessary because without that, PGML skips over the cell.

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -47,7 +47,7 @@ Command for a typical table:
     );
 
 The cell entries above like C<a> may be simple cell content,
-a hash reference with C<data =E<gt> cellContent> and options,
+a hash reference with C<< data => cellContent >> and options,
 or an array reference where the 0th entry is the the cell content
 and it is followed by option key-value pairs.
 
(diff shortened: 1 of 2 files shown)


macros/ui/niceTables.pl
```diff
@@ -369,6 +369,23 @@ =head2 Options for ROWS
 Override table's overall vertical alignment for this row.  Can be C<'top'>, C<'middle'>,
 or C<'bottom'>.
 
+=item C<rows =E<gt> 2D array reference>
```
pstaabp, line 372: FYI: If you don't want to use the E<gt> format, this can be done with `C<< rows => 2D array reference >>`.  Note there needs to be a space after the opening `<<` and before the closing `>>`.

pstaabp (approved):
Nice option.

macros/ui/niceTables.pl
drgrice1: I agree with @pstaabp.  I have preferred that way of doing this since it is more readable when you are looking at the code.  Of course in the generated POD it is equivalent.

Alex-Jordan:
OK, this now has a commit with all the `E<gt>` converted.

drgrice1 (approved):
This looks good.  This adds a versatile way to use niceTables in PGML.

PR #1044, opened 2024-03-25
for PTX output, skip help links

Alex-Jordan:
The help links (from AnswerFormatHelp.pl) should be omitted from PTX output. That output is meant to be used in static (non-interactive) contexts like PDF or a preview of the exercise before the reader reaches the interactive HTML version. Giving the reader help on input syntax in that context is unnecessary and distracting.

So this change skips over such links for PTX output.

Perhaps the same should be done for TeX? I am fine only doing this for PTX though.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -2429,7 +2429,9 @@ sub knowlLink {
 	MODES(
 		TeX  => "{\\bf \\underline{$display_text}}",
 		HTML => qq!<a href="#" class="knowl" $properties>$display_text</a>!,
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This looks fine.

I want to clarify though, that the `helpLink` method does not output the help from `AnswerFormatHelp.pl`.  It has its own html help files that it uses (those are the files in `pg/htdocs/helpFiles`).  The `AnswerFormatHelp.pl` used to have its own help embedded directly in it, but now it just calls this method.

pstaabp (approved):
Didn't test--don't have things set up for pretext, but looks good.

Alex-Jordan:
One thing you can still do is use the PG editor to make PTX output and just
examine the result by sight.

pstaabp:
Thanks for the reminder about this.  Looks good.  Will merge.

PR #1045, opened 2024-03-25
tweaks affecting problems rendered in an iframe with iFrameResizer

Alex-Jordan:
This pairs with a PR to `webwork2`:  https://github.com/openwebwork/webwork2/pull/2378. This affects how problems behave when they are in an iframe that uses iFrameResizer. With this PR and the one in `webwork2`, the feedback popover should trigger the iframe resizing, but the MathQuill editor should not. And the MQ editor should never have a horizontal scrollbar. And when there is a vertical scrollbar, it should not obscure the buttons too much.

### DIFF
--- a/htdocs/js/Feedback/feedback.js
+++ b/htdocs/js/Feedback/feedback.js
@@ -44,6 +44,8 @@
 				?.querySelector('.popover-header')
 				?.addEventListener('click', () => feedbackPopover.hide());
 
+			if (feedbackPopover.tip) feedbackPopover.tip.dataset.iframeHeight = '1';
+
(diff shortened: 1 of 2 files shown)


drgrice1:
Make sure to run `npm run prettier-format` in the `pg/htdocs` directory when modifying javsacript, css, and html files therein.

drgrice1:
By the way, the issue is the double quotes on `1` in `if (feedbackPopover.tip) feedbackPopover.tip.dataset.iframeHeight = "1";`.  Those should be single quotes.

Alex-Jordan:
Got it, and pushed anew.

drgrice1 (approved):
Looks good.

pstaabp (approved):

Issue #1048, opened 2024-04-05
PTX "image" tag getting changed to "img"


Alex-Jordan:
Heads up that the following may be a `webwork2` issue, but I can't diagnose it well enough.

I experience the following on `develop`, but not 2.18. Put the following problem in the PG editor:
```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "PGML.pl"
);

BEGIN_PGML

[@ image( "imagename.png" ) @]*

[!alt text!]{"imagename.png"}

END_PGML
ENDDOCUMENT();
```

and set "View/Reload" to use the "PTX" display mode. What I get is:
```
<statement>

<img source="" width="33%" />
<img source="" width="33%" /><description>alt text</description>
</statement>
```
(and expected errors because there is no such image file). But in 2.18, the `img` above are `image`, and they *should* be `image` for PTX. Furthermore, the second of the two tags closes after the `description` (as it should): `<image width="33%" source=""><description>alt text</description></image>`.

In `develop`, the PTX output for the images should be coming from here:
https://github.com/openwebwork/pg/blob/7a0350da85067d7a80a1b40a7fb6ecdaa8f91f00/macros/core/PGbasicmacros.pl#L2895
Indeed, if I alter that code to produce something like `<TACO` instead of `<image`, then it comes out faithfully as `<TACO`. Something, somewhere, is recognizing these `image` tags and converting them to `<img` tags. And we don't want that. PTX is expecting to find an `<image` element and then use its `source` attribute to download the image files.

I can't work out what is changing the `image` tags to `img` tags. I've grepped for `image.*img` and similar patterns in both repos, and I cannot locate what is making this change.

Alex-Jordan:
Not that it matters, but I had that `PGbasicmacros.pl` link pointing to the `main` branch. Now I updated it to point to the same place in `develop`.

drgrice1:
This is  PG issue.  I suspect this may be caused by the `Mojo::DOM` in post processing.  I will look into it.

drgrice1:
So the problem is that the `XML` semantics for the `Mojo::DOM` object is set too late.  It is set after the problem text is parsed.  It needs to be set before that.

drgrice1:
By the way, attached is a simple script that you can use to render a problem.  Note that it assumes that `PG_ROOT` environment variable is set with the PG root location (and does not die if it isn't -- it just will fail).  Of course remove the `.txt` extension.
[...]

PR #1049, opened 2024-04-06
Fix the translator content post processing for the PTX display mode.

drgrice1:
The `XML` semantics option needs to be activated before the problem text is parsed.

This fixes issue #1048.


Alex-Jordan:
Thanks for figuring that out. It's working for me.

PR #1051, opened 2024-04-07
don't remove excess whitespace from student answers

Alex-Jordan:
This is motivated by a problem I had for a linear algebra course, where students encrypt a short message. The alphabet in play included a space character. One student's code word (the answer they needed to enter) had space characters that were getting trimmed because of the line here. I can't recall if it was because their answer started with a space or had two consecutive spaces inside.

I can't think of a bad side effect from removing this, although it is possible somewhere something assumes that excess white space has already been removed.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -301,14 +301,13 @@ sub NAMED_ANS_RULE {
 		$rh_sticky_answers->{$name} = \@answers;                # Store the rest.
 	}
 
-	$answer_value =~ s/\s+/ /g;                                 # Remove excessive whitespace from student answer.
 	$name = RECORD_ANS_NAME($name, $answer_value);
 	my $previous_name = "previous_$name";
 	$name          = ($envir{use_opaque_prefix}) ? "%%IDPREFIX%%$name"          : $name;
 	$previous_name = ($envir{use_opaque_prefix}) ? "%%IDPREFIX%%$previous_name" : $previous_name;
 
-	my $tcol = $col / 2 > 3 ? $col / 2 : 3;                     # get max
-	$tcol = $tcol < 40 ? $tcol : 40;                            # get min
+	my $tcol = $col / 2 > 3 ? $col / 2 : 3;    # get max
+	$tcol = $tcol < 40 ? $tcol : 40;           # get min
 
 	return MODES(
 		TeX => "{\\answerRule[$name]{$tcol}}",


Alex-Jordan:
The other change here is just some change with perltidy on this file.

somiaj:
@Alex-Jordan I think this is needed, all of the MathObjects operators expect at most one space, so you have things like `* `, as various strings that are matching. Having multiple spaces will probably break normal MathObject parsing.

I think there should be a flag that stops this for when using strings only (or maybe other contexts multiple spaces could have meaning).

drgrice1:
@somiaj:  This will not break anything for MathObject parsing.  The MathObject gets the answer input before this is done.  This only affects the input value that is in the generated HTML returned by PG to display the problem again.

somiaj:
@drgrice1 Thanks for the clarification, I thought this was done before other parsing to make parsing easier by only assuming at most one white space. Since that is the case, I can't think of any other possible side effects.

Alex-Jordan:
> This only affects the input value that is in the generated HTML returned by PG to display the problem again.

If that's true then something else was going on. Because my student's answer was not accepted on the first submission. Maybe I lost track of the details, and it was MQ that was trimming spaces before submission? I will dig up the thread I had with that student and report back.

drgrice1:
Yeah, it does seem that when we had that discussion we tested and this fixed it.  I might be mistaken though.  Though this really does only affect the returned input value, and I can't see that it would have effect on the first submission for MathObject parsing.

Alex-Jordan:
I'll track down the original issue, but even so this would still be something to change. So that upon the second submission, an answer where the white space matters does not become incorrect.

Alex-Jordan:
It's worth noting that this change was on my production server since whenever that meeting happened, I think in February. So far no side effect have been noticed (or at least, not reported).

Alex-Jordan:
I'm attaching the problem file where this arose. If you use seed 1889, the third answer blank is expecting `UU  CTZEK,` with two spaces. In production, if I do not have the change from this PR, it is marked incorrect even on the first submission. After this change, it is marked correct.
[cipher.pg.txt](https://github.com/openwebwork/pg/files/14898327/cipher.pg.txt)


As long as anyone is looking at this file, there is a separate issue I will open a thread about.

drgrice1:
I missed that the `$answer_value` is passed to `RECORD_ANS_NAME` on the next line after the one removed in this pull request.  Thus this does get recorded in the PGresponsegroup for the answer.

drgrice1:
I don't think that removing the spaces will break anything though for MathObjects.  It certainly doesn't affect the operators involving spaces like `* ` or just ` ` (implied multiplication).  I have tested those.

somiaj:
I just tested those as well, and tested the older `fun_cmp` method with additional spaces, and they seem to all parse just fine. Also sounds like this has been used for a while by students, so nothing obvious seems to rely on this code. I too have wanted to use spaces in encryption like problems, just haven't gotten around it, and stuck to strings without spaces. This is a good change to me.

somiaj (approved):

drgrice1 (approved):

pstaabp (approved):

PR #1052, opened 2024-04-07
add an alternative 'such that' pipe

Alex-Jordan:
This is pretty niche, but a student managed to copy-paste a special unicode vertical "such that" character from somewhere when entering an answer like `{x | x < 1}`. So this adds that character to be recognized as an alternative to the keyboard pipe character. Only in the `InequalitySetBuilder` context when `UseVerticalSuchThat` (as opposed to a colon) is in use.

### DIFF
--- a/macros/contexts/contextInequalitySetBuilder.pl
+++ b/macros/contexts/contextInequalitySetBuilder.pl
@@ -155,7 +155,8 @@ sub UseVerticalSuchThat {
 				type          => "bin",
 				string        => " | ",
 				TeX           => ' \mid ',
-				class         => "InequalitySetBuilder::BOP::suchthat"
+				class         => "InequalitySetBuilder::BOP::suchthat",
+				alternatives  => ["\x{2223}"]
 			},
 			"_suchthat" => { alias => "|", hidden => 1 },
 		);


pstaabp:
This works well.  I did notice there are other unicode vertical line symbols.  Wondering if we should also include U+2758 (Light Vertical Bar) or U+2758 (medium vertical bar) as well?

dpvc:
@pstaabp, I think the purpose, here, was that MathJax output for the `\mid` used when `UseVerticalSuchThat` is in play is U+2223, and so copy and past would introduce that character.  I don't think it is necessary (or wise) to include all vertical lines from unicode (there are several others).

pstaabp:
@dpvc That sounds reasonable.  I think covering the standard MathJax output is important.

pstaabp:
Should this be documented anywhere?

Alex-Jordan:
I think this is more about protecting against the student who copy-pastes from MathJax, because they are not aware that `|` is on a keyboard. So I don't see a need to put this in release notes. Especially since it is not for general contexts, it is just for the InequalitySetBuilder context.

pstaabp (approved):

drgrice1 (approved):
I think this should be fine.

drgrice1:
This branch was also targeted to develop.  It has been added to PG-2.19.

PR #1054, opened 2024-04-07
Remove the feedback header click close behavior and instead add a close button.

drgrice1:
This is a possible fix to issue #2362 that might be better than the attempt in #1036.

This adds a close button to feedback popover headers.  The previous behavior of clicking on the header to close it has been removed in favor of this close button.


pstaabp (approved):

Alex-Jordan:
This is a good improvement, and I'll merge it now.

PR #1055, opened 2024-04-08
Improve input feedback styles and add focus style to checkboxes.

drgrice1:
The correct/incorrect/partial-correct and focus box-shadows are also made a bit more pronounced by removing the blur radius and making the spread radius a tad bigger.  Basically, the same box-shadow used by bootstrap for buttons and inputs.

To make this work well the margins between buttons and inputs needed to be increased a bit so that the box-shadows of the two don't overlap.

The focus style has also been added to checkboxes.  I think this style accidentally didn't get added for checkboxes when the correct/incorrect/partially-correct styles were added for checkboxes. This caused an undesirable lack of style when a checkbox was focused that could be confusing (@Alex-Jordan observed this a while back).

The focus colors have been synchronized with those of MathQuill. MathQuill will also need a change to its style so that MathQuill inputs have the same box-shadow when focused.  See https://github.com/openwebwork/mathquill/pull/26.

### DIFF
--- a/htdocs/js/MathQuill/mqeditor.js
+++ b/htdocs/js/MathQuill/mqeditor.js
@@ -137,7 +137,7 @@
 			// Find the preview button container, and add the equation editor button to that.
 			const buttonContainer = document.getElementById(`${answerLabel}-latexentry-button-container`);
 			if (buttonContainer) {
-				buttonContainer.classList.add('d-flex', 'gap-1');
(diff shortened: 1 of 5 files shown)


Alex-Jordan (approved):

pstaabp (approved):

drgrice1:
It would be good to consider the change to MathQuill in https://github.com/openwebwork/mathquill/pull/26 that is associated with this pull request.

PR #1056, opened 2024-04-11
Fix a possible undefined value used in string comparison.

drgrice1:
This was missed in #1044.  The knowLink type option may not be defined. In fact it is only defined for solutions, hints, and helpLinks. The knowlLink method can be called directly, and may not have the type defined.  This causes a warning (even if the display mode is not PTX).

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -2429,7 +2429,7 @@ sub knowlLink {
 	MODES(
 		TeX  => "{\\bf \\underline{$display_text}}",
 		HTML => qq!<a href="#" class="knowl" $properties>$display_text</a>!,
-		PTX  => ($options{type} eq 'help')
+		PTX  => ($options{type} && $options{type} eq 'help')
 		? ''
 		: '<url ' . ($options{url} ? 'href="' . $options{url} . '"' : '') . ' >' . $display_text . '</url>',
 	);


drgrice1:
@Alex-Jordan:  Please look at this and determine what should be done if the `$options{type}` is not set.  With this it still outputs the knowl link.  Should it not?

drgrice1:
By the way you can see this warning if you are an admin user and check the show "Auxiliary Resources" checkbox and click "Check Answers" in a problem that has auxiliary resources (for example a problem that uses one of the PGtikz.pl or PGlateximage.pl macros), since that uses a direct call to `knowlLink`.

Alex-Jordan:
I think I need to know more about how this works in general. If you pass a `value` to this, is that some chunk of content that could just otherwise be dropped into the output?

And when you click the knowl, does that content appear immediately after the clickable, or like after the end of whatever block level element the clickable was in?

The PTX mode is for static content. Think print, or a static preview of the problem in HTML before you get to an interactive version. It looks like with TeX output, we just print the display text and whatever would be revealed is just not there at all, even if it was `value` (as opposed to a `url`). I'd say for sure if there is a `url`, the PTX should make a PTX `url` out of this. But if there is a `value`, I don't know how it should work. With #1044, I was really only trying to "fix" `helpLink` and it took me into `knowlLink`. But how `knowlLink` is used aside from hints, solutions, and help links is not something I see often, if ever.

drgrice1:
There are example in `http://your.server.edu/webwork2/sampleproblems/problem-techniques/AnswerHints` and `http://your.server.edu/webwork2/sampleproblems/problem-techniques/Knowls`.  They are not good examples, and probably should be removed, but they are there.  They were converted from https://webwork.maa.org/wiki/AnswerHints and https://webwork.maa.org/wiki/Knowls.

In any case, if `value` is passed, then that is the content of the knowl.  Otherwise if `url` is passed, then the content of the knowl will be the page that url links to.

Where the knowl content is displayed, depends on where the knowl link appears in the DOM.  If the link is in a table cell, then a new row is injected into the table with a single cell that spans the width of the table, and the content is put into that cell.  If the link is in a list element, then the content is appended to the end of that list element.  Otherwise, the javascript searches after the element and up to an ancestor (that is a crude description of the loop) until it finds a display block element.  If it finds an element after the link that is display block, then the content is injected before that next element.  If it goes up to an ancestor that is display block, then it appends the content into the end of that ancestor.

drgrice1:
Really, it is all pretty much a massive kludge to deal with bad authoring practices.  We really should enforce that all `knowlLink`s (including those called from `helpLink`) never be placed in a table or list, and always should be on a line by themself.  Then we can do what we do with solutions and hints with them, and really eliminate the extremely bad idea of a knowl.

drgrice1:
Note that there are thousands of problems in the OPL that need to be modified to fix this.  Knowls may be the biggest mistake ever introduced into the WeBWorK/PG problem authoring scheme.

drgrice1:
You should note also that if the `value` option is not passed, and a `url` is passed instead (as is done for `helpLink`s), then the `url` is `fetch`ed, and the resulting data is injected into the DOM, and javascript in that content is executed.  It is assumed that the url contents are not an actual valid page on their own.  The `helpLink` files are not.  They are designed to be inserted into another page.  So I don't think that even in that case it is correct for the PTX output to consider them urls.

drgrice1:
One solution for this that can work with having knowl links inline would be to show the content in a modal dialog instead of trying to place it somewhere in the DOM of the parent page.  For PTX and TeX output, this would just be omitted.  That is probably the best solution that I can offer that does not involve rewriting thousands of OPL problems.

drgrice1:
There is another issue that I noticed with the `PTX` output of the `knowlLink` method.  It assumes that `$options{url}` is set.  However, only one of `$options{url}` or `$options{value}` should ever be set, and the `PTX` output does not deal with the case that `$options{url}` is not set and `$options{value}` is.

drgrice1:
Oh wait.  I guess it does deal with the case that `$options{url}` is not set, but then it would output `<url>$display_text</url>`.  Is it correct to have a `url` without an `href`?

drgrice1:
Closing in favor of #1060.

PR #1058, opened 2024-04-12
Fix a couple of minor issues with feedback.

drgrice1:
First, the `previewAnswer` and `feedbackLine` methods should not be nested subs.  They should be closures named by a variable instead.

Second, the `aria-describedby` attribute added to the inputs needs to be removed since its value is for an element that is not in the DOM until the feedback popover is opened.  So until the popover is opened, the `aria-describedby` attribute is invalid HTML.  Since a screen reader user would come to the input typically before the popover is opened that attribute will not work right.


drgrice1:
Actually, I realized that the id that was referred to by this `aria-describedby` attribute was for the visually hidden span that was previously removed, and not to the popover itself.  So it just was invalid.

Alex-Jordan (approved):

pstaabp (approved):

PR #1059, opened 2024-04-12
Remove some accidentally duplicated POD.

drgrice1:
I did this at some point according to git blame.  The removed POD was also not up to date.  The POD before it is.


pstaabp (approved):

pstaabp:
This is minor enough.  I'll just merge.

PR #1060, opened 2024-04-12
Change knowls to being buttons that open modal dialogs.

drgrice1:
This definitively solves the problem of placement of the knowl content. The old method of trying to find a place to put the content in an accordion is entirely unreliable, and depends on the problem author not doing bad things.

Technically a knowl link is not a link at all anymore.  In fact it is now a button.  Although, it really wasn't a link before either.  It was an anchor with a href to the root fragment, and the default action for that was disabled by JavaScript.

There is a new option that can be passed to the `knowlLink` method. That is a `title`.  This `title` will be used for the title of the modal dialog if provided.  Otherwise first required argument of the `knowlLink` (the `$display_text`) will be used for the modal dialog title.

Note that this includes the same change for PTX as #1056 (which will be closed in favor of this pull request).  The PTX output still needs to be evaluated by @Alex-Jordan.


drgrice1:
Note that this also definitively solves the header depth issue.

All of the help files start with header levels of `h2` (and some have `h3`s below that).  That typically was fine in webwork2 since there is an `h1` for the page title, and then this `h2` would fall under that.  However, that could be messed up if the author adds headers, or if in another page (even in an iframe where iframes are supposed to fit into the parent page).

Since a modal dialog is semantically a page on its own, and the title of the modal dialog is an `h1`, the `h2`s (and `h3`s) in the help files will always be correct.

pstaabp:
This looks good.  I tested a few of these and in the Library Browser, for short problems, I get the dialog to look like 
image

which isn't great, but I know it is this size due to the iframe.  Also, this is the instructor view, which probably isn't opening the help.

drgrice1:
It should be opening the help.  What problem is that?

pstaabp:
The dialog is very short to fit in the iframe.  Doesn't look great, but as I mentioned, instructors probably won't be opening these.

drgrice1:
Actually, it is probably opening the help, but since the problem is short it fits to the iframe window.  Most likely you can scroll the content of the dialog.

pstaabp:
> Actually, it is probably opening the help, but since the problem is short it fits to the iframe window. Most likely you can scroll the content of the dialog.

Yes, you can scroll the content, but pretty hard to read, because it is only a couple of lines of help at a time.  I don't think there is much we can do except increase the height of the iframe.   Again, I don't think it is too much of a problem.

drgrice1:
There are contexts where students see problems in an iframe (for instance PreTeXt).  So I will see what I can do about the size issue.

drgrice1:
I made an adjustment to the sizing.  So now a knowl dialog will be at least as tall as its content up to 600 pixels.  So if the content is more than 600 pixels (actually about 530 pixels with the header and possible footer height), then scrolling will still be needed.  But there should be plenty of height for reasonable viewing of the content (about 530 pixels).

Alex-Jordan (approved):
This looks good, and I'm marking for approval.

In defense of knowls (in general), they are supposed to let a reader see the original content at literally the same time as the revealed content. Like in a PTX textbooks setting, the ability to read and re-read a paragraph as well as some cross reference, without one thing obscuring the other, is good. And that's something that other methods don't do, when the revealed content obscures the original content.

But it relies on the author knowing what they are doing, and doing it well, and putting it into a PG macro is just asking for inexperienced authors to use it incorrectly. Even PTX doesn't allow authors to just "make a knowl". They make certain semantic things where it's appropriate for PTX to go aheaad and make a knowl.

drgrice1:
Yeah, and unfortunately authors have incorrectly utilized knowl placement.

I think for most of webwork's usage (like help files) it isn't a particularly big deal if the page content is obscured while the knowl content is read.

pstaabp (approved):
This looks good now.  Plenty of room for help documentation on short problems.

PR #1062, opened 2024-04-13
set iframeHeight on graphTool popover for iframeResizer

Alex-Jordan:
A graphTool help popover might extend low and iframeResizer does not adjust. This fixes that.

Before, except without the "Lorem ipsum".

Screenshot 2024-04-13 at 8 50 11 AM

After, except without the "Lorem ipsum".

Screenshot 2024-04-13 at 8 54 08 AM

The help popover texts are not all that long, and this is probably not an issue for regular WW use (in the limited places where an iframe is in play). But I noticed it in PTX output where the regular buttons below the problem are not within the iframe, so the bottom edge of the iframe can be much closer to the graph. Perhaps with the numberline help, since iirc that can be long. But I did not look at that.


drgrice1 (approved):
Looks good.

There are probably other places where the `data-iframe-height` attribute will be needed as well.

pstaabp (approved):

PR #1063, opened 2024-04-18
Update feedback popover position when typing in a MathQuill input.

drgrice1:
When a MathQuill input is typed in the width of the input grows.  So if a feedback popover is open, then the popover is no longer position correctly relative to the button.  So this updates the position of the popover so that it stays correctly positioned as the MathQuill input grows.


Alex-Jordan (approved):
Tested and works as described

drdrew42 (approved):

PR #1065, opened 2024-04-20
Add a special case to feedback to show correct answers only.

### LINKED ISSUE #1047: Viewing all correct answers at the same time
With the new feedback button tool, it is difficult/impossible to see multiple/all of the expected correct answers at the same time, which you could do with the old results table. There are times when it is reasonable to want to do that, like with `Library/Valdosta/APEX_Calculus/4.1/APEX_4.1_4.pg`.

I'm not sure what to do about this, but I can describe what PTX does. There is another button to reveal correct answers, and when you push it, all answer input fields get filled in with the expected correct answer. Of course it's not as simple as that, because there are regular inputs, MathQuill inputs,  and non-textual inputs like radio buttons, checkboxes, and select elements.

drgrice1:
If the translation options `showCorrectAnswers`,
`forceShowAttemptResults`, and `forceScaffoldsOpen` are all true, and the options `showAttemptAnswers`, `showAttemptPreviews`, and `showMessages` are all false, then correct answers will be shown with no other content in the feedback popover except a close button. Furthermore the popover will open automatically on page load. Obviously scaffold will all be open (since the `forceScaffoldsOpen` option is true).  Otherwise there would be a problem with popovers opening immediately from inside a closed scaffold.

A corresponding pull request to webwork2 will utilize this.

This is a potential approach to resolve #1047.

### DIFF
--- a/htdocs/js/Feedback/feedback.js
+++ b/htdocs/js/Feedback/feedback.js
@@ -40,9 +40,7 @@
 			moveToFront();
 
 			// Make a click on the popover header close the popover.
-			feedbackPopover.tip
-				?.querySelector('.popover-header .btn-close')
-				?.addEventListener('click', () => feedbackPopover.hide());
+			feedbackPopover.tip?.querySelector('.btn-close')?.addEventListener('click', () => feedbackPopover.hide());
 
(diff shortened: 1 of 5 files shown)


drdrew42 (approved):
👍

pstaabp (approved):
Checked a lot of different problems.  This looks good.

PR #1067, opened 2024-04-24
Update dependencies again.

drgrice1:
Most importantly a new tag has been created in the MathQuill repository for WeBWorK/PG 2.19, and the package.json file version of mathquill now points to that. Note that the tag will need to be updated if/when the pending pull requests to that repository are merged, but the change to the package.json file will still work.

Other updates are:

* plotly.js-dist-min: 2.27.0 to 2.32.0
* sortablejs: 1.15.0 to 1.15.2
* autoprefixer: 10.4.16 to 10.4.19
* chokidar: 3.5.3 to 3.6.0
* cssnano: 6.0.1 to 6.1.2
* postcss: 8.4.31 to 8.4.38
* prettier: 3.1.1 to 3.2.5
* sass: 1.69.5 to 1.75.0
* terser: 5.24.0 to 5.30.0

Note that jsxgraph is NOT updated to the latest version.  The latest version is 1.8.0, and we are using 1.7.0.  Although 1.7.0 has a console log issue that is annoying, that is better than the issue with the grid that I see in 1.8.0 that makes the grid thick and obtrusive in certain situations.  I reported the issue, and it has been fixed in their develop branch, but that is not in a release yet.

### DIFF
--- a/htdocs/package-lock.json
+++ b/htdocs/package-lock.json
@@ -1,2799 +1,2824 @@
 {
-    "name": "pg.javascript_package_manager",
-    "lockfileVersion": 2,
-    "requires": true,
-    "packages": {
-        "": {
-            "name": "pg.javascript_package_manager",
-            "license": "GPL-2.0+",
-            "dependencies": {
-                "jsxgraph": "^1.7.0",
-                "jszip": "^3.10.1",
-                "jszip-utils": "^0.1.0",
-                "mathquill": "github:openwebwork/mathquill#WeBWorK-2.18",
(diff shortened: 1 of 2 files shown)


drgrice1:
Something is wrong with the Ubuntu repositories that are being downloaded in the workflow.  It is finding an invalid "InRelease" file from one of the Ubuntu repositories.  So we will have to wait on them to fix it for the unit test workflows to start working again.

pstaabp (approved):

somiaj (approved):

PR #1070, opened 2024-04-24
adjust by larger amount in 'digits' comparisons

Alex-Jordan:
In `develop` (or `2.18`) try the following:

```
DOCUMENT();

Context("Numeric")->flags->set(tolType => 'digits');
$a = Real("500*(1+13/100)^2");

BEGIN_PGML
[`[$a]={}`] [_]{$a}
END_PGML

ENDDOCUMENT();
```

It will show you `638.45`, but will not accept `638.45` as correct.

The actual value of `$a` is a bit smaller due to machine rounding error. I believe what is happening is that `13/100` introduces rounding error. If that were all, then I think there would not be a problem. The code in `lib/Value/Real.pm` starting at line 150 gets around small rounding errors. But I believe that the squaring in this expression compounds the error enough so that those lines of code are no longer enough.

This change ties the approach to machine rounding issues to the value of the `zeroLevel` flag. With the default for that, this changes the `15` to `14`, making a larger bump at the new lines 152/153.

@dpvc, if you have time to think about this, does this seem like an OK solution?

Another approach I was thinking about was that for `digits`, that instead of comparing using a MathObject's value, use its string. Apply the same basic logic, but use whatever the context wants the string to look like as the basis for comparison. Then at least in the situation like this example, you would not have PG telling you the answer is `638.45`, but refusing to accept `638.45` as correct.

### DIFF
--- a/lib/Value/Real.pm
+++ b/lib/Value/Real.pm
@@ -147,22 +147,23 @@ sub compare {
 			my $digits      = (1 > int($tolerance) ? 1 : int($tolerance + 0.5)) - 1;
 			my $extraDigits = int($self->getFlag('tolExtraDigits'));
 			my $tdigits     = (0 < $extraDigits ? $extraDigits + $digits : $digits);
-			my $exp         = substr(sprintf("%E", $b), -3) - 15;    # Adjust $a by an amount in the round-off error
-			$a += ($a <=> 0) * "1E$exp";                             #   range so that it rounds better in sprintf
-			$b += ($b <=> 0) * "1E$exp";                             # Same for $b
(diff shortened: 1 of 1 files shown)


dpvc:
I have always been leery of trying to fiddle with the round off as a means of "correcting" the answer.  I just don't think it is reliable enough, and it makes me very uncomfortable.  As I was reading your message, I was thinking about using the string versions and truncating at the requested number of digits, so I think that is a good direction to pursue.

pstaabp (approved):

drgrice1 (approved):
I haven't tested as I don't use this tolType, but it should be fine.

drgrice1:
This branch was targeted to develop.  I added it to the PG-2.19 branch now also.

PR #1071, opened 2024-04-27
Add the `a` trig functions to the mqeditor autoCommands.

drgrice1:
This depends on https://github.com/openwebwork/mathquill/pull/29.  This temporarily has switched the version of MathQuill in the package.json file to the branch for that pull request.  That is why this is being created as a draft pull request and should not be merged until that is removed.  This is for ease of testing.  When you run `npm ci` it will install the MathQuill from that branch.


drgrice1:
Note that you can also test the other things that were recently merged into MathQuill with this branch.  Of course you can also test those with #1067.

pstaabp:
Seems to work and I think for the inverse trig functions, this seems like a good way to handle asin, acos, ...

Alex-Jordan (approved):
This together with https://github.com/openwebwork/mathquill/pull/29 works for me.

pstaabp (approved):

PR #1072, opened 2024-04-30
Add previous answers to MathQuill inputs on the next tick.

drgrice1:
This fixes https://github.com/openwebwork/mathquill/issues/30. Although, I am not sure what is going on here.    For some reason delaying filling in previous answers until the "next tick" (a timeout with 0 delay) fixes the problem.  Probably the MathQuill math field is not fully initialized until after the mqeditor setup method completes.

I am unable to reproduce the issue using the development build of MathQuill directly via `npm run serve`.

### DIFF
--- a/htdocs/js/MathQuill/mqeditor.js
+++ b/htdocs/js/MathQuill/mqeditor.js
@@ -533,7 +533,7 @@
 			answerQuill.mathField.latex(answerQuill.latexInput.value);
 			answerQuill.mathField.moveToLeftEnd();
 			answerQuill.mathField.blur();
-		}, 0);
+		}, 100);
 	};
 
 	// Set up MathQuill inputs that are already in the page.
--- a/htdocs/js/MathQuill/mqeditor.js
+++ b/htdocs/js/MathQuill/mqeditor.js
@@ -529,9 +529,11 @@
 		};
 		answerQuill.addEventListener('keydown', answerQuill.keydownHandler);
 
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
Fixes the strange behavior.

Alex-Jordan:
I checked out a copy of PG-2.19 branch and merged this branch into it. Then I ran `npm ci` and restarted `webwork2`. But I still experience the issue. I looked at the minimized mqeditor file that my browser console says is in play, and I do see that it has the setTimeout that was added here. So I think that I've set up testing correctly, but the bug described at https://github.com/openwebwork/mathquill/issues/30 still happens.

drgrice1:
Experimenting with this more I see that sometimes it happens, and sometimes it doesn't.  I increased the delay before the timeout executes to a tenth of a second.  I haven't been able to reproduce it with that yet.  However, it might still happen anyway.

I don't really think this pull request should be merged.  I think more investigation is needed to determine what is really going on.  I also don't think this is particularly high priority though.  I think typing `\\end` in text mode is an edge case that won't happen much.

Alex-Jordan:
OK, FWIW, I pulled with the 100ms delay, and now I am not seeing the issue.

And no, this is not high priority. IIRC I was trying things that might demonstrate an insertion vulnerability and that exposed this. But no reasonable exercise would lead to a user inputting this kind of answer.

Alex-Jordan:
Merging this for now. Maybe there's a better solution to find later.

PR #1076, opened 2024-05-17
More matrix features to MathObject matrices

pstaabp:
This builds on #1012 and includes
* new methods for setting elements, removing columns and rows and a submatrix. 
* A test suite for nearly all of the methods in the module. 
* updated POD.

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -6,128 +6,137 @@
 
 =head1 Value::Matrix class
 
-
 References:
 
-MathObject Matrix methods: L<http://webwork.maa.org/wiki/Matrix_(MathObject_Class)>
-MathObject Contexts: L<http://webwork.maa.org/wiki/Common_Contexts>
-CPAN RealMatrix docs: L<http://search.cpan.org/~leto/Math-MatrixReal-2.09/lib/Math/MatrixReal.pm>
+=over
 
-Allowing Matrices in Fractions:
-L<http://webwork.maa.org/moodle/mod/forum/discuss.php?d=2978>
+=item MathObject Matrix methods: L<http://webwork.maa.org/wiki/Matrix_(MathObject_Class)>
 
-     Context()->parens->set("[" => {formMatrix => 1});
+=item MathObject Contexts: L<http://webwork.maa.org/wiki/Common_Contexts>
 
-Files interacting with Matrices:
+=item CPAN RealMatrix docs: L<http://search.cpan.org/~leto/Math-MatrixReal-2.09/lib/Math/MatrixReal.pm>
 
-L<MatrixReal1>
+=back
 
-L<MatrixReduce.pl>
+For allowing Matrices in Fractions, see L<http://webwork.maa.org/moodle/mod/forum/discuss.php?d=2978>
 
-L<Matrix>
+    Context()->parens->set("[" => {formMatrix => 1});
 
-L<MatrixCheckers.pl> -- checking whether vectors form a basis
+=head2 Files interacting with Matrices:
 
-L<MatrixReduce.pl>  -- tools for  row reduction via elementary matrices
+=over
 
-L<MatrixUnits.pl>   -- Generates unimodular matrices with real entries
+=item L<MatrixReal1>
 
-L<PGmatrixmacros.pl>
+=item L<MatrixReduce.pl>
 
-L<PGmorematrixmacros.pl>
+=item L<Matrix>
 
-L<PGnumericalmacros.pl>
+=item L<MatrixCheckers.pl> -- checking whether vectors form a basis
 
-L<tableau.pl>
+=item L<MatrixReduce.pl>  -- tools for  row reduction via elementary matrices
 
-quickMatrixEntry.pl
+=item L<MatrixUnits.pl>   -- Generates unimodular matrices with real entries
 
-L<LinearProgramming.pl>
+=item L<PGmatrixmacros.pl>
 
-Contexts
+=item L<PGmorematrixmacros.pl>
 
-	Matrix -- allows students to enter [[3,4],[3,6]]
-	       -- formMatrix =>1 also allows this?
-	Complex-Matrix -- allows complex entries
+=item L<PGnumericalmacros.pl>
 
-Creation methods
+=item L<tableau.pl>
 
-		 $M1 = Matrix([1,2],[3,4]);
-   		 $M2 = Matrix([5,6],[7,8]);
-    	  $v = Vector(9,10);
-    	  $w = ColumnVector(9,10); # differs in how it is printed
+=item quickMatrixEntry.pl
+
+=item L<LinearProgramming.pl>
+
+=back
+
+=head2 Contexts
+
+=over
+=item C<Matrix> -- allows students to enter [[3,4],[3,6]]
+				 -- formMatrix =>1 also allows this?
+=item C<Complex-Matrix> -- allows complex entries
+
+=back
+
+=head2 Creation of Matrices
+
+Using the C<Matrix>, C<Vector> or C<ColumnVector> methods
+
+Examples:
+
+    $M1 = Matrix([1,2],[3,4]);
+    $M2 = Matrix([5,6],[7,8]);
+    $v = Vector(9,10);
+    $w = ColumnVector(9,10); # differs in how it is printed
 
 Commands added in Value::matrix
 
-	Conversion
-		$matrix->values produces [[3,4,5],[1,3,4]] recursive array references of numbers (not MathObjects)
-		$matrix->wwMatrix   produces CPAN MatrixReal1 matrix, used for computation subroutines
+Conversion:
+    $matrix->value produces [[3,4,5],[1,3,4]] recursive array references of numbers (not MathObjects)
+    $matrix->wwMatrix   produces CPAN MatrixReal1 matrix, used for computation subroutines
 
-	Information
-		$matrix->dimension:  ARRAY
+Information
+    $matrix->dimension:  ARRAY
 
-	Access values
+Access values
 
-		row : MathObjectMatrix
-		column : MathObjectMatrix
-		element : Real or Complex value
+    row : MathObjectMatrix
+    column : MathObjectMatrix
+    element : Real or Complex value
 
-	Assign values
+Update values
+    setElement.
 
-		these need to be added:
-
-see C<change_matrix_entry()> in MatrixReduce and L<http://webwork.maa.org/moodle/mod/forum/discuss.php?d=2970>
-
-	Advanced
-		$matrix->data:  ARRAY reference (internal data) of MathObjects (Real,Complex, Fractions)
-		                stored at each location.
-
+See C<change_matrix_entry()> in MatrixReduce and L<http://webwork.maa.org/moodle/mod/forum/discuss.php?d=2970>
 
 Passthrough methods covering subroutines in Matrix.pm which overrides or
 augment CPAN's MatrixReal1.pm.  Matrix is a specialized subclass of MatrixReal1.pm
 
 The actual calculations for these methods are done in C<pg/lib/Matrix.pm>
 
-	trace
-	proj
-	proj_coeff
-	L
-	R
-	PL
-	PR
+    trace
+    proj
+    proj_coeff
+    L
+    R
+    PL
+    PR
 
-Passthrough methods covering subroutines in C<pg/lib/MatrixReal1.pm>
-(this has been modified to handle complex numbers)
-The actual calculations are done in C<MatrixReal1.pm> subroutines
+Passthrough methods covering subroutines in C<pg/lib/MatrixReal1.pm> (this has been modified to handle complex numbers)
+The actual calculations are done in C<MatrixReal1.pm> subroutines.
+
 The commands below are Value::Matrix B<methods> unless otherwise noted.
 
+    condition
+    det
+    inverse
+    is_symmetric
+    decompose_LR
+    dim
+    norm_one
+    norm_max
+    kleene
+    normalize
+    solve_LR($v)    - LR decomposition
+    solve($M,$v)    - function version of solve_LR
+    order_LR        - order of LR decomposition matrix (number of non-zero equations)(also order() )
+    order($M)       - function version of order_LR
+    solve_GSM
+    solve_SSM
+    solve_RM
 
-
-	condition
-	det
-	inverse
-	is_symmetric
-	decompose_LR
-	dim
-	norm_one
-	norm_max
-	kleene
-	normalize
-	solve_LR($v)    - LR decomposition
-	solve($M,$v)    - function version of solve_LR
-	order_LR        - order of LR decomposition matrix (number of non-zero equations)(also order() )
-	order($M)       - function version of order_LR
-	solve_GSM
-	solve_SSM
-	solve_RM
+=head2 methods
 
 =cut
 
-#
 package Value::Matrix;
 my $pkg = 'Value::Matrix';
 
 use strict;
+use warnings;
 no strict "refs";
 use Matrix;
 use Complex1;
@@ -171,7 +180,7 @@ sub matrixMatrix {    #internal
 	my ($x, $m);
 	my @M         = ();
 	my $isFormula = 0;
-	foreach $x (@_) {
+	for $x (@_) {
 		if (Value::isFormula($x)) { push(@M, $x); $isFormula = 1 }
 		else {
 			$m = $self->new($context, $x);
@@ -180,7 +189,7 @@ sub matrixMatrix {    #internal
 		}
 	}
 	my ($type, $len) = ($M[0]->entryType->{name}, $M[0]->length);
-	foreach $x (@M) {
+	for $x (@M) {
 		Value::Error("Matrix rows must all be the same type")
 			unless (defined($x->entryType) && $type eq $x->entryType->{name});
 		Value::Error("Matrix rows must all be the same length") unless ($len eq $x->length);
@@ -199,7 +208,7 @@ sub numberMatrix {    #internal
 	my $context   = shift;
 	my @M         = ();
 	my $isFormula = 0;
-	foreach my $x (@_) {
+	for my $x (@_) {
 		$x = Value::makeValue($x, context => $context);
 		Value::Error("Matrix row entries must be numbers: $x ") unless _isNumber($x);
 		push(@M, $x);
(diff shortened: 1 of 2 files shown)


lib/Value/Matrix.pm
```diff
@@ -6,128 +6,137 @@
 
 =head1 Value::Matrix class
 
-
 References:
```
Alex-Jordan, line 9:
```suggestion
=head2 References:
```

lib/Value/Matrix.pm
```diff
@@ -6,128 +6,137 @@
...
+=item MathObject Contexts: L<http://webwork.maa.org/wiki/Common_Contexts>
 
-Files interacting with Matrices:
+=item CPAN RealMatrix docs: L<http://search.cpan.org/~leto/Math-MatrixReal-2.09/lib/Math/MatrixReal.pm>
 
-L<MatrixReal1>
+=back
 
-L<MatrixReduce.pl>
+For allowing Matrices in Fractions, see L<http://webwork.maa.org/moodle/mod/forum/discuss.php?d=2978>
```
Alex-Jordan, line 21: Is this about "Matrices in Fractions", or is it actually about "Fractions in Matrices"?

lib/Value/Matrix.pm
```diff
@@ -6,128 +6,137 @@
...
 
-L<Matrix>
+    Context()->parens->set("[" => {formMatrix => 1});
 
-L<MatrixCheckers.pl> -- checking whether vectors form a basis
+=head2 Files interacting with Matrices:
```
Alex-Jordan, line 25: It took me a sec to understand this phrase (which I know was already there before this PR). Not sure if the suggestion below is any better but maybe. 

```suggestion
=head2 Matrix-related Files:
```

lib/Value/Matrix.pm
```diff
@@ -6,128 +6,137 @@
...
+
+=back
+
+=head2 Contexts
 
-	Conversion
-		$matrix->values produces [[3,4,5],[1,3,4]] recursive array references of numbers (not MathObjects)
-		$matrix->wwMatrix   produces CPAN MatrixReal1 matrix, used for computation subroutines
+=over
+=item C<Matrix> -- allows students to enter [[3,4],[3,6]]
```
Alex-Jordan, line 58: This is not showing up for me in the POD. I see the "Contexts" h2, but then there is nothing there. It goes straight to "Creation of Matrices".

Screenshot 2024-07-16 at 10 48 54 AM

lib/Value/Matrix.pm
```diff
@@ -6,128 +6,137 @@
...
-		row : MathObjectMatrix
-		column : MathObjectMatrix
-		element : Real or Complex value
+Using the C<Matrix>, C<Vector> or C<ColumnVector> methods
+
+Examples:
+
+    $M1 = Matrix([1,2],[3,4]);
+    $M2 = Matrix([5,6],[7,8]);
+    $v = Vector(9,10);
```
Alex-Jordan, line 72: Do the Vector and ColumnVector examples really belong here? It seems misleading, since those are distinct MathObjects from a Matrix.

lib/Value/Matrix.pm
```diff
@@ -6,128 +6,137 @@
...
+    $w = ColumnVector(9,10); # differs in how it is printed
+
+Commands added in Value::matrix
 
-	Assign values
+Conversion:
+    $matrix->value produces [[3,4,5],[1,3,4]] recursive array references of numbers (not MathObjects)
```
Alex-Jordan, line 78: Some of the content here should be wrapped in code markup. Also the line break between the two items is not being respected in output.

Alex-Jordan:
I was looking at the POD since a lot of that is in the diff here, and I left some comments. But I'm going to pause because I think if you proofread the actual POD output you will see a lot of the same things that I'm seeing. I'll move on to testing the new tools now.

lib/Value/Matrix.pm
```diff
@@ -244,36 +278,68 @@ sub typeRef {
...
+
+    $A->isSquare; # is '' (false)
+    $B->isSquare; # is 1 (true);
+
+=cut
+
 sub isSquare {
 	my $self = shift;
 	my @d    = $self->dimensions;
-	return 0 if scalar(@d) > 2;
```
Alex-Jordan, line 253: Nice catch here on some incomplete logic.

lib/Value/Matrix.pm
```diff
@@ -485,17 +560,17 @@ sub mult {
 	Value::Error("Can only multiply 2-dimensional matrices") if scalar(@dl) > 2 || scalar(@dr) > 2;
 	Value::Error("Matrices of dimensions %dx%d and %dx%d can't be multiplied", @dl, @dr)
 		unless ($dl[1] == $dr[0]);
-	#
-	#  Do matrix multiplication
-	#
+
+	#  Perform atrix multiplication.
```
Alex-Jordan, line 564: Typo here.

lib/Value/Matrix.pm
```diff
@@ -720,26 +906,43 @@ sub Zero {
...
+
+=cut
+
 sub row {
 	my $self = (ref($_[0]) ? $_[0] : shift);
 	my $M    = $self->promote(shift);
 	my $i    = shift;
-	return if $i == 0;
-	$i--   if $i > 0;
+	Value::Error("Row must be a positive integer") unless $i =~ m/^[1-9]\d*$/;
```
Alex-Jordan, line 924: I can't recall where, but there is somewhere else like this where we support having an explicit plus sign in front of the integer.

Alex-Jordan: But maybe this would be better if if actually accessed the dimensions of the matrix, and required an integer that is a valid row index.

lib/Value/Matrix.pm
```diff
@@ -749,27 +952,171 @@ sub column {
...
+
+Assign an element in the matrix to a value.
+
+Inputs: indices as an arrayref and the value in a form that can be parsed.
+
+Note: this mutates the matrix itself.
+
+Usage:
+
+    my $A = Matrix([ [ 1, 2, 3, 4 ], [ 5, 6, 7, 8 ], [ 9, 10, 11, 12 ] ]);
```
Alex-Jordan, line 991: I think all these usage examples should not use `my`.

lib/Value/Matrix.pm
```diff
@@ -749,27 +952,171 @@ sub column {
...
+    $A->setElement([2,3],-5);
+
+
+=cut
+
+sub setElement {
+	my ($self, $ind, $value) = @_;
+
```
Alex-Jordan, line 999: Maybe there should be a check that the length of the `$ind` that is passed matches the dimension of the matrix.

lib/Value/Matrix.pm
dpvc: Since `$i` is usually a Perl real (not a string, for example), an initial plus sign will not be in its stringified version, so won't be part of the comparison in `$i =~ m/.../` even if you used `$M->row(+3)`.  It would only be there if you passed the row number as a string, as in `$M->row("+3")`, which I might consider to be an improper value to pass (it is *not* a positive integer, it is a string).

lib/Value/Matrix.pm
```diff
@@ -749,27 +952,171 @@ sub column {
...
+Usage:
+
+    my $A = Matrix([ [ 1, 2, 3, 4 ], [ 5, 6, 7, 8 ], [ 9, 10, 11, 12 ] ]);
+    $A->setElement([2,3],-5);
+
+
+=cut
+
+sub setElement {
```
Alex-Jordan, line 997: I tried this using the example above, and it is not working. The matrix `$A` still displays with `7` in the (2,3) entry instead of with `-5`. Looking below, it seems that the last line that sets `$el` is just changing the value of a local variable. It is not changing the value at some reference like you'd want it to.

lib/Value/Matrix.pm
```diff
@@ -749,27 +952,171 @@ sub column {
...
+    $A = Matrix([[1,2,3,4],[5,6,7,8],[9,10,11,12]]);
+
+    $A->subMatrix([2..3],[2..4]);  # returns a Matrix([[6,7,8],[10,11,12]])
+
+    $A->subMatrix(2,3);  # returns Matrix([ [ 1, 2, 4 ], [ 9, 10, 12 ] ]);
+
+    $A->subMatrix([3,1,2],[1,4,2]);  # returns Matrix([9,12,10],[1,4,2],[5,8,6]);
+=cut
+
+sub subMatrix {
```
Alex-Jordan, line 1027: This has some issues as well. If I try it on a 1D matrix, I get an error:
```
$A = Matrix([ 1, 2, 3, 4 ]);
$B = $A->subMatrix([3]);
```

And if I try it on a 3D matrix, it is not returning what I would expect it to:
```
$A = Matrix([ [[1,2],[3,4]], [[5,6], [7,8]]]);
$B = $A->subMatrix([1], [1,2], [2]);
# $B seems to be what I would expect from subMatrix([1], [1,2], [1,2]);
```

lib/Value/Matrix.pm
```diff
@@ -749,27 +952,171 @@ sub column {
...
+
+sub setElement {
+	my ($self, $ind, $value) = @_;
+
+	# Drill down into the matrix
+	my $el = $self->{data}[ $ind->[0] - 1 ];
+	for my $i (1 .. scalar(@$ind) - 1) { $el = $el->{data}[ $ind->[$i] - 1 ]; }
+
+	# update the value of $el
+	$el = Value::makeValue($value);
```
drgrice1, line 1005: This needs to be
```suggestion
	my $el = \($self->{data}[ $ind->[0] - 1 ]);
	for my $i (1 .. scalar(@$ind) - 1) { $el = \($$el->{data}[ $ind->[$i] - 1 ]); }

	# update the value of $el
	$$el = Value::makeValue($value);
```
```perl
	my $el = \($self->{data}[ $ind->[0] - 1 ]);
	for my $i (1 .. scalar(@$ind) - 1) { $el = \($$el->{data}[ $ind->[$i] - 1 ]); }

	# update the value of $el
	$$el = Value::makeValue($value);
```
in order to actually mutate the matrix.

lib/Value/Matrix.pm
```diff
@@ -749,27 +952,171 @@ sub column {
...
+}
+
+=head3 C<removeRow>
+
+Remove a row from a matrix.
+
+Usage:
+
+    $A = Matrix([ [ 1, 2, 3, 4 ], [ 5, 6, 7, 8 ], [ 9, 10, 11, 12 ], [13, 14, 15, 16] ]);
+    $A->removeRow(3);
```
drgrice1, line 1065: This method isn't working.  I used your example above, and it returned the matrix `[[1],[2],[3]]`.

drgrice1: Also, since `$A` is not mutated, your POD example should assign the return result to something.

lib/Value/Matrix.pm
pstaabp: Changed this, but odd that the test works for the previous version of the code as well as this one, however, I think the test needs some help.

lib/Value/Matrix.pm
pstaabp: I thought of subMatrix only on 2D matrices, so will rework for other dimensions.

lib/Value/Matrix.pm
pstaabp: I'll add this.

lib/Value/Matrix.pm
pstaabp: Rewrote this now.  Should be working on non-2D matrices.

lib/Value/Matrix.pm
```diff
@@ -6,128 +6,156 @@
...
 Commands added in Value::matrix
 
-	Conversion
-		$matrix->values produces [[3,4,5],[1,3,4]] recursive array references of numbers (not MathObjects)
-		$matrix->wwMatrix   produces CPAN MatrixReal1 matrix, used for computation subroutines
+Conversion:
 
-	Information
-		$matrix->dimension:  ARRAY
+    $matrix->value produces [[3,4,5],[1,3,4]] recursive array references of numbers (not MathObjects)
```
Alex-Jordan, line 81: This seems to be incorrect. (It was incorrect before this PR.) `$matrix->value` produces an actual array, not an array reference. Its entries are either numbers (if `$matrix` was 1D) or are array references. I don't have a suggestion for how to succinctly write that in the POD.

lib/Value/Matrix.pm
```diff
@@ -6,128 +6,156 @@
...
+    $matrix->wwMatrix   produces CPAN MatrixReal1 matrix, used for computation subroutines
 
-	Access values
+Information
 
-		row : MathObjectMatrix
-		column : MathObjectMatrix
-		element : Real or Complex value
+    $matrix->dimension:  ARRAY
```
Alex-Jordan, line 86: This is also incorrect (and was before). There is no `dimension` method. However there is `dimensions`, which returns an array of integers as you would expect. Usually two for what we think of normally as a matrix.

I think it would be goo if there were a `dimension` method though, that tells you if it is a 1D, 2D, 3D, etc matrix. It would just be evaluating `$matrix->dimensions` in a scalar context.

lib/Value/Matrix.pm
```diff
@@ -720,26 +929,43 @@ sub Zero {
...
+Extract a given column from the matrix.
+
+Usage:
+
+    $A1 = Matrix([ [ 1, 2, 3, 4 ], [ 5, 6, 7, 8 ], [ 9, 10, 11, 12 ] ]);
+    $A1->column(2);  # returns the column Matrix [[2],[6],[10]]
+
+=cut
+
 sub column {
```
Alex-Jordan, line 964: Similar to my comment about `row`, the `column` method could take multiple whole numbers as arguments.

Also I am seeing what I think is undesirable behavior here with 1D matrices. For an nD matrix with n > 1, `$matrix->column(1)` is still an nD matrix. Good. But with a 1D matrix, `$matrix->column(1)` is a `Real` rather than a 1D matrix. I think for consistency it should still be a 1D matrix.

lib/Value/Matrix.pm
```diff
@@ -6,128 +6,156 @@
...
-		element : Real or Complex value
+    $matrix->dimension:  ARRAY
 
-	Assign values
+Access values
 
-		these need to be added:
+    row : MathObjectMatrix
+    column : MathObjectMatrix
+    element : Real or Complex value
```
Alex-Jordan, line 92: See my other comment about `column`. Currently for a 1D matrix, `$matrix->column(1)` is a `Real`. I do think it should be changed to be a MathObject Matrix though.

With the `element` method, it is more complicated than this. If you have an nD matrix and you pass n whole numbers to `element`, then yes you get a `Real` (or `Complex`, I assume). But if you pass fewer than n whole numbers, you get a `Matrix`. (And if you pass more than n, you get null.)

Alex-Jordan:
See my recent comments just now. I feel like we need to work on this file in a more methodical manner, especially since there are things to fix that precede the new things here. I propose:

1. Go through the current methods and (a) make them better if we see opportunity for that, and (b) correct their POD without moving POD around.
2. Then reorganize the POD and provide/update a test suite for the existing methods. The test suite should be thorough with at least 1D, 2D, and 3D matrices.
3. Then add new methods, with new POD, and new tests.

When I look at the diff for this PR, it's just too overwhelming to sort out what is just POD moving around and what is new content.

pstaabp:
Would it make sense if we split this in two pieces?
1. update the POD for existing methods and functionality
2.  Update the methods as you mentioned and with them any needed POD.

Alex-Jordan:
We should catalog what is still here that needs to be added.

- `setElement` method
- `setExtractElements` method
- `subMatrix` method
- `removeRow` method
- `removeColumn` method

Did I miss anything? We should be careful to add them so that they work for a Matrix object of any degree. We should agree on what things like "removeRow" and "removeColumn" mean for degrees other than 2.

somiaj:
I think the methods in `ReduceMatrix.pl` would be nice to be part of matrices by default.

Alex-Jordan:
I've opened PRs that recover `setElement` (renamed as `set` and reworked) and `subMatrix` (nearly identical to as here). That leaves:

* `removeRow`
* `removeColumn`

and @somiaj also mentioned some methods from `ReduceMatrix.pl`. But regarding `removeRow` and `removeColumn`, I propose some conventions.

1. No matter what degree a matrix has, its "rows" are specified by the first index.
2. No matter what degree a matrix has, its "columns" are specified by the second index.

Good thing about this convention:

1. `removeRow` and `removeColumn` both have meaning for all degree matrices, except `removeColumn` doesn't make sense for degree 1 matrices.
2. These methods could be implemented without new code. Just call `subMatrix` in the appropriate way.

"Bad" thing about this convention:

1. For a degree 1 Matrix, it is presented as a row. In fact, the `isRow` method returns 1 if and only if the Matrix is degree 1. But I'm suggesting if `$M` is a degree 1 Matrix, then `$M->removeRow(2)` would remove its second entry. It would look like the second column is removed.

So an alternative would be to still make these methods illegal on degree 1 Matrices, yet still apply to degrees above 2.

Alex-Jordan:
> I think the methods in `ReduceMatrix.pl` would be nice to be part of matrices by default.

@somiaj Where is `ReduceMatrix.pl`? I don't find that in `pg` or the OPL.

somiaj:
It was my dyslexia, sorry about that, it is `macros/math/MatrixReduce.pl`. One I use is `rref` to do things like grade if a matrix is in Echelon form.

Alex-Jordan:
OK, I found `MatrixReduce.pl`. 

Most things in there are now implemented in `Matrix.pm`, maybe with different syntax. One thing that is not is row reduction. But I'm not sure that should be added to `Matrix.pm`. If all entries of a matrix are integers, then you could program a row reduction algorithm in a way that avoided machine rounding issues only by (a) using the Fraction context or (b) saving all division until the last step in the algorithm.

But even then, what about when your matrix has floating point values in the first place? Well, if you think they all come from fractions, you could apply Fraction() to them first, again using the Fraction context. 

But even then, what if floating point entries came from irrational numbers?

So I see too many floating point issues with doing a row reduction algorithm in `Matrix.pm`. 

We could put something like `is_rref` though into `Matrix.pm`, just to check if a matrix is reduced or not.

Alex-Jordan:
With #1421 and #1424 open now, I'm going to close this older PR. The conversation about rref can continue though.

somiaj:
I haven't ran into enough floating point issues on smaller 3x3 or 4x4 matrices for the method in `MatrixReduce.pl` to be a problem. Basically I check for echelon form, by checking the two matrices rref matches and if the zeros are in the right spot of the matrix.

It would be nice to not have to use that older macro for this, and I think since these are math objects, equality isn't exact so floating point issues have never gotten bad enough to not mark answers correct.

Alex-Jordan:
When you have used this, have you been also using Fraction context?

Based on your second paragraph, I'm not sure the issue is clear. It's not an issue about one value being slightly different from another value. In a row reduction algorithm, scaling and subtraction happens in a way that is engineered to make one element of a row become zero, while other elements in the row get the same treatment and may or may not come out as zero.  When one of those other elements in that row should become zero, but actually comes out nonzero because of machine rounding, it throws a big wrench in the algorithm. The algorithm reaches a cell that mathematically should have come to zero, and the algorithm should have moved to look at a cell to the right. But since it reaches a nonzero cell (owing to the rounding error), it normalizes the entire row so that cell has value 1. So what should not have been a pivot position comes out as a pivot position.

Try this MWE:

```
DOCUMENT();

loadMacros(qw(PGstandard.pl PGML.pl MatrixReduce.pl));

Context("Matrix");
$A = Matrix([ 0.2, 0.6], [ 10000, 30000 ]);
$R = rref($A);

BEGIN_PGML
[`[$A]`] reduces to [`[$R]`]
END_PGML

ENDDOCUMENT();
```

That matrix should row reduce to have only one pivot position, but I'm getting it to come out as the identity matrix. And it's just the classic machine rounding issue with an rref algorithm.

somiaj:
I haven't tested the fraction context, and yea my matrices are nice in that they won't create something that the issues can cause problems. I can keep loading this older macro, but it also means it is harder to deprecate.

To me I think it would be better to have the rref algorithm available with maybe a warning, and let authors decide if it is useful or not. If there are ways to improve it, that is good to. I also have algorithms that I use to put matrices in echelon form with no division to avoid such issues, but these are for writing solutions not graders.

PR #1078, opened 2024-06-12
Fix a security vulnerability that has appeared when running `npm ci`.

drgrice1:
Not much to say.  Just a minor maintenance issue.


Alex-Jordan:
I'm going to merge this and the webwork2 one like it.

PR #1083, opened 2024-07-04
Update the package-lock.json to point to the latest MathQuill commit.

drgrice1:
Pull request https://github.com/openwebwork/mathquill/pull/28 was merged into MathQuill.  That commit is now tagged for WeBWorK-2.19.  So if you run `npm install` it will install that commit.  However, if you run `npm ci` it will get the commit before it because that is what is in the package-lock.json file.

### DIFF
--- a/htdocs/package-lock.json
+++ b/htdocs/package-lock.json
@@ -804,8 +804,7 @@
 		},
 		"node_modules/mathquill": {
 			"version": "0.10.1",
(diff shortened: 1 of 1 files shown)


drgrice1:
Wait to merge this until all of the MathQuill changes have been merged.  Then I will update the commit hash for the one that will actually be the release.  Either that or I will close this pull request and instead do this in #1071.

pstaabp (approved):

PR #1088, opened 2024-07-25
let width specifications in niceTables be numbers from (0,1]

Alex-Jordan:
niceTables allows width specification for columns or cells using the LaTeX column type `p{width}`. With this PR, `width` can now be a decimal from 0 (exclusive) to 1 (inclusive). This is understood to be a portion of the available width.
* In HTML, something like `40%` should appear in CSS.
* In TeX, something like `0.4\linewidth` should appear in appropriate places.
* In PTX, something like `40%` should appear in either
    * a `col`'s `width` attribute (for a DataTable)
    * or an `sbsgroup`'s `widths` attribute (for a LayoutTable).

Note that PTX has no way to specify a width for an individual cell.

You can try with this test file:

```
DOCUMENT();

loadMacros(qw(PGstandard.pl PGML.pl PGcourse.pl));

BEGIN_PGML

[#
	[. This table should have a 40% width column, an unspecified column, and a 1-inch column. .]
	[. This table should have a 40% width column, an unspecified column, and a 1-inch column. .]
	[. This table should have a 40% width column, an unspecified column, and a 1-inch column. .]
#]{padding => [0,1], align => 'p{0.4}lp{1in}'}

----

[#
	[. This table should have a 40% width cell, an unspecified cell, and a 1-inch column. .]{halign => 'p{0.4}'}
	[. This table should have a 40% width cell, an unspecified cell, and a 1-inch column. .]
	[. This table should have a 40% width cell, an unspecified cell, and a 1-inch column. .]{halign => 'p{1in}'}
#]{padding => [0,1]}

----

[#
	[. This table should have a 40% width column, an unspecified column, and a 1-inch column. And then overrided by a 50% cell, a 1in cell, and a 20% cell .]{halign => 'p{0.5}'}
	[. This table should have a 40% width column, an unspecified column, and a 1-inch column. And then overrided by a 50% cell, a 1in cell, and a 20% cell .]{halign => 'p{1in}'}
	[. This table should have a 40% width column, an unspecified column, and a 1-inch column. And then overrided by a 50% cell, a 1in cell, and a 20% cell .]{halign => 'p{0.2}'}
#]{padding => [0,1], align => 'p{0.4}lp{1in}', center => 0}

----
# Layout Table

[#
	[. This table should have a 40% width column, an unspecified column, and a 30% column. .]
	[. This table should have a 40% width column, an unspecified column, and a 30% column. .]
	[. This table should have a 40% width column, an unspecified column, and a 30% column. .]
#]*{padding => [0,1], align => 'p{0.4}lp{0.3}', center => 0}

END_PGML

ENDDOCUMENT();
```

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -1586,6 +1586,7 @@ sub getPTXthickness {
 
 sub getWidthPercent {
 	my $width = shift;
+	return $width             if (substr($width, -1) eq '%');
 	return $width * 100 . '%' if ($width =~ /^0*(0(\.\d*)?|1(\.0*)?)$/);
 	my $x    = 0;
 	my $unit = 'cm';
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -88,8 +88,9 @@ C<c> for center-aligned column
 
 C<r> for right-aligned column
 
-C<p{width}> for a column with left-aligned paragraphs of fixed width.
-The width needs to be absolute to work in all output formats.
+C<p{width}> for a column with left-aligned paragraphs of the given width.
+The width can be an absolute width or (unlike in LaTeX) a positive decimal number at most 1.
+If it is a decimal, it will be interpreted as a portion of the available width.
 
 C<X> for a column that expands to fill (see C<Xratio> below),
 and will have left-aligned paragraphs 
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
To test this, you should load a problem (perhaps the one posted here) in the editor. And then examine the output for MathJax, TeX, and PTX. With the latter two, you can see directly where the width specifications are. With MathJax, you can use the inspector to make sure things are what they are supposed to be.

drgrice1:
The last case in your example (the layout table) does not seem to be working.  I see that the style for the first cell has "width: 0.4".  `0.4` is an invalid value for `width`.

Alex-Jordan:
Hmm, I see it as `40%`. 

Screenshot 2024-07-25 at 8 21 37 AM

This was the last things I fixed. So I thought for a second maybe I hadn't pushed this, but the commit hash here is b23bc052e29291cb471317e7c238c9fab5c311df and that matches what I have. Since this was the last thing, is there any chance you have the previous version before my last force push late last night?

drgrice1:
Heh!  It seems I just tested your sample problem with the PG-2.19 branch, and forgot to even check out your branch.  That was dumb.

drgrice1 (approved):
Looks good.  Everything seems to work as claimed.

Alex-Jordan:
I changed one more thing here that will only affect PTX output. First, I reviewed the actual definitions of lengths in TeX, and redefined the conversion from one unit to another based on those actual TeX definitions. And the conversion is all to `pt` instead of `cm`, since that is the more fundamental unit in TeX.

And that is also good because the default line with for a PTX document is 340pt. So it helps clarify how to translate an absolute width (which PTX disallows) to a percent width.

pstaabp (approved):

PR #1089, opened 2024-07-30
attempt to parse display math for proper PTX md/mrow structure when a…

Alex-Jordan:
…pplicable

In PTX, multi-lined math expressions are structured with `md` and `mrow` tags. Or you can just have the usual LaTeX multi-line math constructors inside an `me` tag. Up until this PR, we have just done the latter for PG display math being converted to PTX. This PR changes things so that if the display math is an `align`, `alignat`, or `gather` environment, it all gets parsed into an `md` with `mrow`s.

The following code demonstrates various math structures. Only the last three structures are meant to be affected by this PR, because those are what PTX converts `md/mrow` to. You can try this before the PR using the PG editor with PTX output, and you will see the display math is always wrapped in `me`. Then with these changes, the last three have `md/mrow` structure.

````
BEGIN_PGML
[`\pi`]

[``\pi``]

[```\pi```]

[```
\begin{aligned}
\pi&=3.14\ldots\\
&<\frac{22}{7}
\end{aligned}
```]

[```
\begin{align}
\pi&=3.14\ldots\\
&<\frac{22}{7}
\end{align}
```]

[```
\begin{alignat}{1}\pi&=3.14\ldots\\
&\approx\frac{22}{7}
\end{alignat}
```]

[```
\begin{gather}
\pi=3.14\ldots\\
\approx\frac{22}{7}
\end{gather}
```]

END_PGML
````

There is an observable (and desirable) difference in PDF output when the PTX source is structured this way. I can elaborate if anyone cares.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -1968,11 +1968,26 @@ sub general_math_ev3 {
 		$out = "`$in`"                                   if $mode eq "inline";
 		$out = '<DIV ALIGN="CENTER">`' . $in . '`</DIV>' if $mode eq "display";
 	} elsif ($displayMode eq "PTX") {
-		#protect XML control characters
+		# protect XML control characters
 		$in =~ s/\&(?!([\w#]+;))/\\amp /g;
 		$in =~ s/</\\lt /g;
-		$out = '<m>' . "$in" . '</m>'   if $mode eq "inline";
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Looks fine to me.  As usual, you are the only one that really knows what is needed here.

pstaabp (approved):

PR #1094, opened 2024-08-13
replace scaffold yellow with lightblue

Alex-Jordan:
This replaces the yellow background in scaffold problem headers with lightblue. This is the color that is used on a section's header/button when you are able to open/close that section and do not yet have that section correctly answered. Note that gray is used when you do not yet have access to a section, and lightgreen is used when you have answered that section correctly.

To me the yellow is just too...yellow.

This looks fine (to me) in all four themes, but perhaps the yellow theme needs special attention. Although, to me the current scaffold yellow is so bright that even in the yellow theme, I don't think it looks good.

If people like this and we're still doing non-critical hotfixes, I can open another PR. I'm assuming that this is noncritical and we have to stop somewhere.

### DIFF
--- a/htdocs/js/Scaffold/scaffold.scss
+++ b/htdocs/js/Scaffold/scaffold.scss
@@ -22,7 +22,7 @@
 		}
 
 		&.canopen > button.accordion-button {
-			background: yellow;
+			background: lightblue;
 		}
 
 		&.iscorrect > button.accordion-button {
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
This is fine with me.  I don't think that any special attention is needed for the yellow theme.  If anything the previous yellow was an issue with the yellow theme that this remedies.

pstaabp (approved):

PR #1097, opened 2024-08-18
Better method for updating the feedback popover position.

drgrice1:
This is an alternate and better approach to repositioning the feedback popover after the "Show Correct Answers" button is used.

This is to address issue https://github.com/openwebwork/webwork2/issues/2524.


somiaj (approved):
Fixes the issue the timeout was to quick on my setup.

PR #1099, opened 2024-08-20
allow characters for lte and gte in a few contexts

Alex-Jordan:
This is so a student can directly enter ≤ and ≥ in contexts where `<=` and `>=` are valid operators.

It affects a bit more than just the two files here, because at least one context (Inequality-SetBuilder) is built on top of `contextInequalities.pl`.

### DIFF
--- a/macros/parsers/parserLinearInequality.pl
+++ b/macros/parsers/parserLinearInequality.pl
@@ -104,6 +104,7 @@ sub Init {
 			string        => ' <= ',
 			TeX           => '\le ',
 			kind          => "le",
+			alternatives  => ["\x{2264}"],
 			class         => 'LinearInequality::inequality',
 			formulaClass  => "LinearInequality"
 		},
@@ -126,6 +127,7 @@ sub Init {
 			TeX           => '\ge ',
 			kind          => 'ge',
 			reverse       => 'le',
+			alternatives  => ["\x{2265}"],
 			class         => 'LinearInequality::inequality',
 			formulaClass  => "LinearInequality"
 		},
--- a/macros/contexts/contextInequalities.pl
+++ b/macros/contexts/contextInequalities.pl
@@ -148,6 +148,7 @@ sub Init {
 			TeX           => '\le ',
 			class         => 'Inequalities::BOP::inequality',
 			eval          => 'evalLessThanOrEqualTo',
+			alternatives  => ["\x{2264}"],
 			combine       => 1
 		},
 		'=<' => {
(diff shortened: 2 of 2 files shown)


macros/parsers/parserLinearInequality.pl
```diff
@@ -104,16 +104,7 @@ sub Init {
...
-			formulaClass  => "LinearInequality"
-		},
-		'=<' => {
-			precedence    => .5,
-			associativity => 'left',
-			type          => 'bin',
-			string        => ' <= ',
-			TeX           => '\le ',
-			kind          => "le",
+			alternatives  => [ '=<', "\x{2264}" ],
```
dpvc, line 107: I'm confused by the alternatives, as the first one is the same as the actual operator at line 10.  Was this meant to go in the previous operator `<=`?

One could make `=<` an alias for `=>` rather than an alternative.  I think that might not have been available for operators when this was written.

macros/parsers/parserLinearInequality.pl
```diff
@@ -126,17 +117,7 @@ sub Init {
...
-		},
-		'=>' => {
-			precedence    => .5,
-			associativity => 'left',
-			type          => 'bin',
-			string        => ' >= ',
-			TeX           => '\ge ',
-			kind          => 'ge',
-			reverse       => 'le',
+			alternatives  => [ '=>', "\x{2265}" ],
```
dpvc, line 120: Similarly here.

Alex-Jordan:
Yeah, sorry. I was manually copy-pasting from where I have this in my production server and messed that up. I'll fix it soon.

macros/parsers/parserLinearInequality.pl
Alex-Jordan: Wait, I'm looking at this again and I think it is right. Maybe the diff is misleading.

This removes `=<` as an operator altogether. And then it makes `=<` and `"\x{2264}"` be alternatives for `<=`. Because the original block for `<=` and `=<` are right next to each other and nearly identical, the diff that is displayed might be misleading.

Alex-Jordan: OK, I didn't think about how alternatives are not available by default. I was thinking I was just simplifying the code, but of course you are right that it would change behavior. I updated the commit to only add the ≤ and ≥ alternatives. (And ≠ in the other file where `!=` has meaning.)

drgrice1 (approved):
This looks fine.  There are no big changes, and the alternatives should be good.

somiaj (approved):

drgrice1:
I will merge this with two approvals.

PR #1102, opened 2024-08-22
preserve the order that is used for items in NAMED_POP_UP_LIST

Alex-Jordan:
This is about the forum issue that was revealed at:
https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8535

I think that once upon a time, perl did not randomize the order of a hash, and this macro presented the items in a dropdown list in the order the author gave them. But when perl moved to randomize hash order, then this macro started presenting options in random order on every page load. So the change here captures the original order of the hash keys and uses that.

I have not surveyed how `NAMED_POP_UP_LIST` is used, and whether the random ordering might be a good thing for some uses. Although, that seems unlikely.

I have also not surveyed surrounding macros to see if other things would benefit from a similar change.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -751,7 +751,8 @@ sub checkbox {
 sub NAMED_POP_UP_LIST {
 	my ($name, @list) = @_;
 
-	my %options = ref($list[0]) eq 'ARRAY' ? (map { $_ => $_ } @{ $list[0] }) : @list;
+	my %options      = ref($list[0]) eq 'ARRAY' ? (map { $_ => $_ } @{ $list[0] }) : @list;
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
I don't think your assessment is quite correct on the cause of the issue.  Comparing to the 2.17 code, I see that it did not use a hash.  So order would have been preserved even with newer versions of Perl.  So this is a regression.

This should be made a hotfix as well.

Alex-Jordan:
Ah, OK. I only looked closely at file changes to PGchoicemacros.pl before I realized the hash order issue was part of this and prematurely concluded it was about the perl version. Also I experienced this in the spring, so also convinced myself it was happening with 2.18. But now I recall that I had the PCC server running an early version of 2.19.

somiaj (approved):
Didn't test (unsure what problem to test this with), but the change is clear and needed to persevere order.

PR #1103, opened 2024-08-22
preserve the order that is used for items in NAMED_POP_UP_LIST (hotfix for #1102)

Alex-Jordan:
This is a hotfix for #1102. I'm short on time this morning, and I still haven't patrolled neighboring code to see if a similar change from array to hash structure was made that might want a similar update.


drgrice1:
I am certain this is the only place where this was done.

somiaj (approved):

drgrice1 (approved):

PR #1104, opened 2024-08-28
Update date and version in pg.sty

dlglin:
Make it match the version of pg.


somiaj (dismissed):
These need to be added to develop too. Do we make the string 2.20-develop there, then when we forget to update it, it will be more confusing.

drgrice1 (dismissed):

drgrice1 (approved):

pstaabp (approved):

Issue #1092, opened 2024-08-07
Number with units in MathObject List doesn't work.


drgrice1:
Here is a MWE:

```perl
DOCUMENT();

loadMacros('PGstandard.pl', 'PGML.pl', 'parserNumberWithUnits.pl');

$ans = List(NumberWithUnits(60, 'deg'), NumberWithUnits(240, 'deg'));

BEGIN_PGML
Find all exact solutions of the following equation in the interval
[` 0^{\circ} \leq \theta < 360^\circ `].  If there is more than one answer,
enter them as a comma separated list. 

	[`  \tan(\theta) = \sqrt{3}`]

[` \theta = `] [_]{$ans}{15}
END_PGML

ENDDOCUMENT();
```

If MathQuill is disabled and you enter `60 deg, 240 deg`, the message "'deg' is not defined in this context" is given.  If Mathquill is enabled and you enter the same thing (of course the deg turns into a degree symbol), then the message "Unexpected character '°'" is given. 

I found an old forum post at https://webwork.maa.org/moodle/mod/forum/discuss.php?d=3314 (that apparently was never answered) that is related.

It would be nice to be able to do this, but the MathObject parser fails when the unit is encountered.

dpvc:
Yes, the NumberWithUnits has never been a first-class citizen.  The handling of the units is a real kludge, as the units are stripped off before any other parsing is done, and the units are never actually parsed.  That means a NumberWithUnits can't be used in a compound setting, like in a list, as the parser will choke on the units, as you have found out.  

I had always hoped to get back to them and make an honest MathObject out of them, but never did.  One of the difficulties at the time was that the list of units were not available within the problem (they were in one of the modules in pg/lib, if I recall correctly), so the unit names would need to be duplicated in a MathObject, and that is a maintenance issue.  But I think that changes to the unit library since then have made the units available, so it would be possible to use them to define the needed units within the NumberWithUnits class.

It would be possible, then, to do things like `1 m + 3 cm`, and so on.  There would probably want to be contexts that allow that and other LimitedNumberWithUnits contexts that require only a number with a unit (like the other Limited* contexts).

Of course, some units have single letter names, like `m`, and you probably would want to avoid defining that as a unit in a situation where it wasn't needed.  So one might need to specify the category of units (length, volume, force, etc.) and only have the context define the units for that category (or several categories, if needed).  So it would probably take some thought.

But for now, you can't use NumberWithUnits in a list.  Sorry about that!

[...]

PR #1107, opened 2024-08-29
Add new contextUnits.pl implementing units as true MathObject classes

dpvc:
This PR adds the file `contextUnits.pl` to the `macros/contexts` directory.  It implements units as true MathObject classes, so that you can use units and numbers with units as you would any other MathObject.  With this context, you can do

``` perl
$n = Compute("3 m/sec");
```

rather than having to use `NumberWithUnit()` to create them.

It also means you can do things like

``` perl
$n = Compute("3m + 2cm");             # equivalent to Compute("3.02m")
$v = Compute("50 miles / (2 hours)"); # equivalent to Compute("25 mi/h")
$l = Compute("3m, 4cm, 2sec");        # a list of numbers with units
```

or even

``` perl
$v = Compute("50 miles per hour");
$a = Compute("9.8 meters per second squared");
```

Since units are now MathObjects, even without a number, you can use

``` perl
$u = Compute("m/s^2");
```

and ask a student what are the units for accelleration, without the need for a number in front of the units.

The `contextUnits.pl` file is based on the `contextExtensions.pl` framework for extending a context to include new features.  That means you can add units to existing contexts.  For example, you could add units to the `Matrix` context to allow the entries of a matrix to have units.

See the POD documentation at the top of the file for more details about how to use the Units contexts.

Note that the `contextExtension.pl` is included in this PR, even though I made a separate PR for that file.  Because the branch for that PR isn't in the PG repository, I can't target this PR to that one.  If that PR is merged, the `contextExtensions.pl` file should disappear from this PR.


macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2094 @@
...
+
+    Context("Units")->addUnits(
+      apples => { fruit => 1, aliases => ["apple"] },
+      oranges => { fruit => 1, aliases => ["orange"], factor => 2 }
+    );
+
+    Compute("1 apple") == Compute("2 oranges"); # returns 1
+
+will make an apple equivalent to two oranges by making both C<apples>
+and C<oranges> be examples of the fundamentaul unit C<fruit>.
```
drgrice1, line 126: Typo: fundametnaul -> fundamental

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2094 @@
...
+C<removeUnits()> method of the context.  For example
+
+    Context("Units")->withUnitsFor("length")->removeUnits("ft", "inch", "mile", "furlong");
+
+removes the English units and their aliases, leaving only the metric
+units.  To remove a unit without removing its aliases, use C<removeUnitsNotAliases()>
+instead.
+
+Note that the units are stored in the context as constants, so to list
+all the units, together with other contants, use
```
drgrice1, line 138: Typo: contants -> constants

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2094 @@
...
+    $n = Compute("5 m");
+    $r = Real($m);           # returns 5 as a Real MathObject
+    $u = Unit($m);           # returns "m" as a Unit MathObject
+
+You can get the numeric portion of the number-with-units
+object relative to the base units using the C<quantity> method:
+
+    $q = Compute("3 ft")->quantity;    # returns .9144
+
+Using C<$m->quantity> is equivalent to calling C<$m->toBaseUnits->number>.
```
drgrice1, line 345: When a `<` or `>` is inside a `C<...>` the `C<< ... >>` needs to be used instead.  So this will need to be
```suggestion
Using C<< $m->quantity >> is equivalent to calling C<< $m->toBaseUnits->number >>.
```
Note that the spaces are needed.

This is also needed on lines 412 and 426

drgrice1 (commented):
There are a couple of typos and POD issues noted.

This looks good, and will be a nice MathObject addition.

I did note one issue in testing that is related to issue #1092.  If you use the `Units` context in a `List` it works nicely.  However, if you use the `LimitedUnits` context with a `List` it does not.  If `Compute("3 deg, 4 deg")` is called, the error "Can't use ',' in this context" is thrown.  You can force it to work for creating the answer by calling `List("3 deg", "4 deg")`, but then if a student enters the answer "3 deg, 4 deg" the student gets the above error message.  Perhaps there is some way to accomplish this using the `contextExtension.pl` macro?  Although, it would be nice to not need additional steps for this.

dpvc:
> if you use the `LimitedUnits` context with a `List` it does not. If `Compute("3 deg, 4 deg")` is called, the error `"Can't use ',' in this context"` is thrown.

Yes, that is because the `LimitedUnits` context is based on the `LimitedNumeric` context, in which the comma has been removed (i.e., made undefined).  There are several possible approaches to this.

One would be to leave it as is and document the fact that

``` perl
$context->operators->redefine(',');
```

will make the comma available again.  Alternatively, this could be added to the `LimitedUnits` context initially, though that does mean that it differs from how `LimitedNumeric` works.  Finally, one could define a third context, `LimitedUnits-List`, corresponding to the `LimitedNumeric-List` that is in `contextLimitedNumeric.pl', that includes the comma operator.

Which route do you all think should be taken?

macros/contexts/contextUnits.pl
dpvc: Right!  Thanks!

drgrice1:
I should have thought of that.  I think you mentioned that before somewhere in a related setting.  It works for me to add `$context->operators->redefine(',');`.  Perhaps just make a comment on that in the documentation.   Maybe others will like the other approaches better though?

drgrice1 (dismissed):
Thanks for fixing the typos and pod issues, and thanks for this pull request to begin with!

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2056 @@
...
+    mass             (fundamental units "kg", except for those in "astronomy" below)
+    temperature      (fundamental units "defC", "defF", "K")
+    frequency        (fundamental units "rad/s")
+    force            (fundamental units "(kg m)/(s^2)")
+    energy           (fundamental units "(kg m^2)/(s^2)")
+    power            (fundamental units "(kg m^2)/(s^3)" except for those in "astronomy" below)
+    pressure         (fundamental units "kg/(m s^2)")
+    electricity      (fundamental units "amp", "amp/s", "(kg m)/(amp s^-3)", "(amp s^-3)/(kg m)",
+                                        "(amp^2 s^4)/(kg m^2)", "(kg m^2)/(amp^2 s^3)", and "(amp^2 s^3)/(kg m^2)")
+    magnatism        (fundamental units "kg/(amp s^2)" and "(kg m)/(amp s^2)")
```
pstaabp, line 52: Shouldn't this be `magnetism`?

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2056 @@
...
+	electricity       => [
+		{ amp => 1 },
+		{ amp => 1,  s => 1 },
+		{ kg  => 1,  m => 2,  amp => -1, s => -3 },
+		{ kg  => -1, m => -2, amp => 1,  s => 3 },
+		{ amp => 2,  s => 4,  kg  => -1, m => -2 },
+		{ kg  => 1,  m => 2,  amp => -2, s => -3 },
+		{ kg  => -1, m => -2, amp => 2,  s => 3 },
+	],
+	magnatism   => [ { kg => 1, amp => -1, s => -2 }, { kg => 1, m => 2, amp => -1, s => -2 }, ],
```
pstaabp, line 671: Shouldn't this be `magnetism`?

dpvc: OOPS!  Thanks for catching that!  I will fix.

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2056 @@
...
+will get the list of units.
+
+
+=head2 Adding units to other contexts
+
+The C<Units> and C<LimitedUnits> contexts are based on the C<Numeric>
+and C<LimitedNumeric> contexts.  You can add units to other contexts
+using the C<context::Units::extends()> function.  For example,
+
+    Context(context::Units::extending("Fraction")->withUnitsFor("length"));
```
pstaabp, line 155: To use this example, one also must include the 'contextFraction.pl' macro.  I think we may need to make a comment about this.

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2056 @@
...
+
+    keepNegativePowers => 1,     Preserve use of negative powers so C<m s^-1> will
+                                 not be shown as C<m/s> (but will still match it).
+    useNegativePowers => 0       Always use negative powers instead of fractions?
+    limited => 0                 Don't allow operations on numbers with units.
+    exactUnits => 0              Require student units to exactly match correct ones
+                                   in both order and use of negative powers
+    sameUnits => 0               Require student units to match correct ones
+                                   not scaled versions
+    partialCredit => .5          Partial credit if answer is right but units
```
pstaabp, line 181: I think you imply that 'units are incorrect'.

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2056 @@
...
+
+=head2 Creating unit and number-with-unit objects
+
+In the units contexts, units are first-class citizens, and unit and
+number-with-unit objects can be created just like any other
+MathObject.  So you can use
+
+    $n = Compute("3 m/s");
+
+to get a numer-with-units object for 3 meters per second.  You can also use
```
pstaabp, line 207: typo 'number-with-units'

pstaabp:
@dpvc Just tried one of your examples from the POD:
```
$u = Compute("meters per second per second");
```

I got the error 
```
Division is allowed only for Numbers or a Vector, Point, or Matrix and a Number; see position 8 of formula at line 61 of [PG]/macros/core/Parser.pl
```

I think this is because of the second squared in the bottom.  I had set the context to:
```
Context('Units')->withUnitsFor('length','velocity','time');
```

BTW, the shorthand `Unit('m/s^2')` worked fine.

Alex-Jordan:
Sometimes there is a question where the answer is like `5 %/yr` or `0.05/yr`.
* Is the second format already possible with this?
* Would it be possible for `%` to be a unit? Perhaps in its own category, or a "dimensionless" category? (I know there is a context for percent, but it doesn't help with units like `%/yr`.)
* And then if `%` were a unit, is there any hope that `5%` and `0.05` could be equivalent? Or`5 %/yr` and `0.05/yr`?

I'm just thinking openly and not requesting these things. I'm interested in your assessment of how the structure you have here already could support these things, or if it would involve more.

macros/contexts/contextUnits.pl
dpvc: Fixed.

macros/contexts/contextUnits.pl
dpvc: Fixed.

macros/contexts/contextUnits.pl
dpvc: Fixed.  Thanks for the close reading of all of this.

dpvc:
> Just tried one of your examples from the POD:
> 
>      $u = Compute("meters per second per second");

Argh.  This was due to a last-minute reorganization of the initialization of the context, and it caused the `per` operator to get the wrong class definition.  I've fixed it, so the example now works, as does `Compute("m/s/s")` and similar constructs.

dpvc:
@Alex-Jordan:

> Is the second format already possible with this?

It wasn't but is now with the latest commit.

> Would it be possible for `%` to be a unit?

You could add it using

``` perl
Context("Units")->withUnitsFor("time")->addUnits("%" => {factor => .01, "%" => 1});
```

for example.  Then you could do `5%/yr`, but the formatting will be strange (`5 %/yr` as a string, and `5\,\frac{%}{\text{yr}}` in TeX).  Also, you could do things like `5/%` and `5 % yr` and `5 yr/%` and `5 %^2` and other such nonsense.  And `5 %` and `.05` will not be considered the same.

I don't think using `%` as a unit is the right approach.  It's not a unit, it is a postfix unary operator (essentially `/100`).  Adding units to the Percent context is the way to go, I think.

``` perl
Context(context::Units::extending("Percent")->withUnitsFor("time"));
```

would allow `5%/yr` and it will format properly.  Further `5%/yr` and `.05/yr` will be equivalent.

You are right that you can't do just `%/yr` this way, but if you need that as an answer, then you can use the technique of adding `%` to the Units context as in the first example above and make just that answer be in that specialized context.

macros/contexts/contextUnits.pl
dpvc: fixed

pstaabp:
@dpvc The latest commit fixed my other errors.  Also testing your other example in the POD

```
   Context("Units")->addUnits(
      apples => { fruit => 1, aliases => ["apple"] },
      oranges => { fruit => 1, aliases => ["orange"] }
    );
    Compute("3 apples") == Compute("3 oranges"); # returns 1
```

isn't working for me.  It all makes sense to me, but not getting the expected results.

dpvc:
> It all makes sense to me, but not getting the expected results.

There was a typo in a variable name (when removing the `aliases` from the unit definition), and that meant the two definitions weren't the same.  I had done my testing without the aliases, so didn't catch it.  I pushed a commit to fix it.

somiaj:
Also regarding TeX output, further testing seems to suggest that some times units are wrapped in `\text{}` and some are not. I see that `miles per hour` becomes `\frac{\text{miles}}{\text{hour}}`,  but `miles/hour` is formatted just as written.

dpvc:
> n the TeX output, should units be wrapped in a `\text{}`

All the units already should be in `\text{}` in TeX output, unless the unit definition provides its own TeX formatting.

> but `miles/hour` is formatted just as written

I can't reproduce that; the TeX output is `\frac{\text{miles}}{\text{hour}}` for me, and I don't see how you would be able to not get that, as `per` and `/` both use the same class internally.  Are you sure you were asking for TeX output and not string output?

-----

> seems the category velocity isn't working directly

It is actually doing what it is supposed to, as the various categories add the *named* units for that category, not all possible ways of obtaining compound units for that category.  You are right that `velocity` only adds `c`, `knots` and `mph`, as those are the only named units in the `Units.pm` file that have m/s fundamental units.  If you want allow `m/s` as an answer, you need to include the length and time units as well, as in

``` perl
Context("Units")->withUnitsFor("velocity", "length", "time");
$v = Compute("5 m/s");
```

It would be possible to define the `velocity` category using

``` perl
velocity => [{m => 1, s => -1}, {m => 1}, {s => 1}]
```

to get length and time as well, or to have `withUnitsFor()` include the individual categories for any fundamental units involved in a category, if people felt that it is too hard to specify the collections you want.  I tried to hint at this with the example

``` perl
Context("Units")->withUnitsFor("length", "volume");
```

so that you could do `5 m^3`, but perhaps should have been more explicit about that.

it is possible that the category idea needs better handling.  If there are other ideas about how to specify the units to be included, that would be worth a discussion.

-----

> Is there a way to add all units to a problem

You could do

``` perl
Context("Units")->addUnitsNotAliases(keys %context::Units::Context::UNITS);
```

but I would not recommend it.  One reason is that there are a number of single-letter units (`A`, `c`, `F, `G`, `g`, `H`, `h`, `J`, `K`, `L`, `m`, `N, `S, `s`, `T`, `v`, `W`), and these can easily conflict with variables or other values within a problem.  That is why I limited the units to the specific category that is needed rather than have them all defined all the time.  Remember that these can be used internally in the problem (e.g., `3m + 5cm`), not just at the end, like in the traditional `NumberWithUnits()` where conflicts with other variables would be minimized.

-----

> students who enter in units I didn't add to the context get the message, "[Variable] foo is not defined in this context",

Yes, that is correct, just as they would if they typed in some other letter that is not defined in the context.  There is no way for the MathObject parser to know that the student might have been thinking about a unit rather than a function or variable or some other thing when it sees a letter that it doesn't have a definition for.  I'm not sure how this could be handled better.

> having an message like "Answer is dimensionally incorrect" that could be toggled via some `showDimensionWarnings` flag I would find more useful.

Perhaps in the `LimitedUnits` context, it would make sense to change the message about undefined variables to one about units, but remember that this will be given got *any* unknown letter (or string of letters), not just ones that are units that just aren't in this context.

Note that you *do* get a message about units being wrong if you enter units (that are defined) but are not correct, in case that is a help.

dpvc:
I did find a typo in the definition for the temperature category, and pushed a commit to fix it.

somiaj:
Here is a screenshot if it helps, as you can see the TeX inside the problem output is just fine, it is just the processing of the student answers that didn't get wrapped in `\text{}`.

image

somiaj:
> > Is there a way to add all units to a problem
> 
> You could do
> 
> ```perl
> Context("Units")->addUnitsNotAliases(keys %context::Units::Context::UNITS);
> ```
> but I would not recommend it. One reason is that there are a number of single-letter units (A, c, F, G, g, H, h, J, K, L, m, N, S, s, T, v, W), and these can easily conflict with variables or other values within a problem. That is why I limited the units to the specific category that is needed rather than have them all defined all the time. Remember that these can be used internally in the problem (e.g., 3m + 5cm), not just at the end, like in the traditional NumberWithUnits() where conflicts with other variables would be minimized.

This makes sense, as I was considering a case where I just wanted a numerical answer with correct units, and didn't want to give students hints as to what the units might be, so I thought having a quick way to add all units would be nice.

---

> > having an message like "Answer is dimensionally incorrect" that could be toggled via some showDimensionWarnings flag I would find more useful.
>
> Perhaps in the LimitedUnits context, it would make sense to change the message about undefined variables to one about units, but remember that this will be given got any unknown letter (or string of letters), not just ones that are units that just aren't in this context.

Yes, the case I was thinking of would be best done with some LimitedUnits context where I'm not expecting variables but want to be able to control if or if not hints are given to students who enter in dimensionally incorrect units.

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+
+Since units are themselves MathObjects, you can work with units
+without a preceding number.  These can be created through C<Compute()>
+just as with other MathObject, or you can use the C<Unit()> constructor.
+
+    $u = Compute("meters per second per second");
+    $u = Unit("m/s^2");
+
+This allows you to ask a student to say what units should be used for
+a particular setting, without the need for a quntity.
```
drgrice1, line 285:
```suggestion
a particular setting, without the need for a quantity.
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+
+    $m = Compute("5 m");
+    $ft = $m->toUnits("ft");                 # returns "16.4042 ft"
+
+    $cm = Compute("5.21 m")->toUnits("cm");  # returns "521 cm"
+
+    $a = Compute("32 ft/s^2")->toBaseUnits;  # returns "9.7536 m/s^2"
+
+For a given number with units, you may wish to obtain the numeric
+portion or the units portion separatly.  This can be done using the
```
drgrice1, line 329:
```suggestion
portion or the units portion separately.  This can be done using the
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+C<number> and C<unit> methods:
+
+    $n = Compute("5 m");
```
drgrice1, line 332: Should this be `$m = Compute("5 m"):`?  I think this is the variable referenced below.

The same thing for the `$n` on line 339.

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+
+    $v = Compute("sin(30 deg)");
+
+will return 0.5, and so will
+
+    $a = Compute("60 deg");
+    $sin_a = sin($a);
+
+as the perl functions have been overloaded to handle numbers with
+units when the units are anglular units.
```
drgrice1, line 373:
```suggestion
units when the units are angular units.
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+
+The other exception is C<abs()>, which can be applied to numbers with
+units, and returns a number with units hacing the same units, but the
```
drgrice1, line 376:
```suggestion
units, and returns a number with units having the same units, but the
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+in the same way that you use any other MathObject.  For example
+
+    BEGIN_PGML
+    What are the units for acceleration? [_______]{"m/sec^2"}
+    END_PGML
+
+Here, the student can answer any equivalent units, such as C<ft/s^2>
+or even C<mi/h^2>, and get full credit.  If you wish to require the
+units to being the same as the correct answer, you can use the
+C<sameUnits> option on the answer checker (ot set the C<sameUnits>
```
drgrice1, line 392:
```suggestion
C<sameUnits> option on the answer checker (to set the C<sameUnits>
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+
+    $u = Compute("m/s^2");
+    BEGIN_PGML
+    What are the metric units for acceleration? [_______]{$u->cmp(sameUnits => 1)}
+    END_PGML
+
+If the student entered C<ft/sec^2>, they would get partial credit, and
+a message indicating that their units are correct but are not the same
+as the expected units.  The amount of partial credit is determined by
+the C<partialCredit> answer-checkeroption (or context flag), whose
```
drgrice1, line 403:
```suggestion
the C<partialCredit> answer-checker option (or context flag), whose
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+
+to increase the credit to 75%, or
+
+    $u->cmp(sameUnits => 1, partialCredit => 0)
+
+to give no partial credit.
+
+Similarly, if the correct answer is given with units of C<m>, then
+when C<< sameUnits => 1 >> is set, an answer using C<cm> instead will be
+given only partical credit.
```
drgrice1, line 416:
```suggestion
given only partial credit.
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+
+In the case where the units include products of units, like C<m s>,
+the C<sameUnits> option requires both be present, but they can be in
+either order.  So a student can enter C<s m> and still get full
+credit.  If you want to require the order to be the same as in the
+correct anser, then use the C<exactUnits> option.  Again, partial
```
drgrice1, line 422:
```suggestion
correct answer, then use the C<exactUnits> option.  Again, partial
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+	return $self;
+}
+
+#
+#  Alias for addUnitsFor
+#
+sub withUnitsFor { (shift)->addUnitsFor(@_) }
+
+#
+#  Remove the named units and thier aliases
```
drgrice1, line 839:
```suggestion
#  Remove the named units and their aliases
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+#  The MathObject class for units (single or compound)
+#
+package context::Units::Unit;
+our @ISA = ('Value');
+
+#
+#  Create a new Unit object, either by parsing a string version of
+#  the units, or by giving the name of a known unit, or as name => unit_def,
+#  where unit_def is an object like the known units.  You can also use this
+#  to objectin the Unit from a Number-with-Unit, or to make a copy of an
```
drgrice1, line 890:
```suggestion
#  object in the Unit from a Number-with-Unit, or to make a copy of an
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+			nfunds         => $nfunds,
+			dfunds         => $dfunds,
+			isConsant      => 1,
+		},
+		$class
+	);
+}
+
+#
+#  Copy a Unit by duplicating the internal hashs and arrays.
```
drgrice1, line 963:
```suggestion
#  Copy a Unit by duplicating the internal hashes and arrays.
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+#  the $other hash if the power ends up being negative.
+#
+sub addUnitPower {
+	my ($self, $units, $other, $u, $n) = @_;
+	$units->{$u} = ($units->{$u} // 0) + $n;
+	$self->checkUnitPower($units, $other, $u);
+}
+
+#
+#  Check if there is cancelation between the $key1 and $key2 lists,
```
drgrice1, line 1071:
```suggestion
#  Check if there is cancellation between the $key1 and $key2 lists,
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+	my ($self, $x) = @_;
+	return CORE::sin($self->quantity) if $self->fString eq 'rad';
+	context::Units::fnError('sin');
+}
+
+#############################################################
+
+#
+#  Convert a Number with Units to one using the base units (in
+#  alphabetial order)
```
drgrice1, line 1598:
```suggestion
#  alphabetical order)
```

macros/contexts/contextUnits.pl
```diff
@@ -0,0 +1,2076 @@
...
+	return unless $lunit;
+	my ($rnum, $runit) = $self->splitNumberUnit($self->{rop});
+	return unless $runit && $lunit->string eq $runit->string;
+	$self->{lop} = $self->Item("BOP")->new($self->{equation}, $self->{bop}, $lnum, $rnum);
+	$self->{rop} = $lunit;
+	$self->makeMult;
+}
+
+#
+#  For string output, add parenthese if the precedence is the same
```
drgrice1, line 1812:
```suggestion
#  For string output, add parentheses if the precedence is the same
```

drgrice1 (commented):
Spell checking...

macros/contexts/contextUnits.pl
dpvc: I'll fix in a separate commit.

dpvc:
> unsure if mathquill is the issue here

@somiaj, I'm pretty sure it is, but I don't have an installation running MathQuill to try it out.  I'm not able to reproduce the problem without it, but don't have any idea why it would be producing this result.  Does the MathQuill code set the `preview_latex_string` in its own, by any chance?

drgrice1:
> > unsure if mathquill is the issue here
> 
> @somiaj, I'm pretty sure it is, but I don't have an installation running MathQuill to try it out. I'm not able to reproduce the problem without it, but don't have any idea why it would be producing this result. Does the MathQuill code set the `preview_latex_string` in its own, by any chance?

This seems to be caused by the different precedence that is used when MathQuill is enabled.  When MathQuill is enabled, then the precedence of the ' *' operator is 2.9 instead of 3.

dpvc:
> This seems to be caused by the different precedence that is used when MathQuill is enabled.

I had seen that special `mq_precedence` added in the Parser code, but never knew what the issue was that it was supposed to resolve.  Can anyone give me the backstory on that?

It turns out that the different precedence did lead to an issue with the TeX output for the unit objects as well as the division operator within formulas.  It turns out the `preview_latex_string` is created from the answer hash's `student_formula` rather than the `student_value`, and so that is why the value wasn't being set properly.

I've added a commit to resolve the issues.  I think that should take care of the problem with MathQuill.

drgrice1:
> I had seen that special `mq_precedence` added in the Parser code, but never knew what the issue was that it was supposed to resolve. Can anyone give me the backstory on that?

The precedence for `' *'` is 2.9 (lower than the 3 for division) so that a vertical fraction from a MathQuill input has the division performed before it is multiplied with something typed in front of it.  For example, if $5\frac{x}{3}$ is entered in a MathQuill input, then `5 x/3` is submitted for the answer, and this is interpreted with the `mq_precedence` as what the student entered, i.e., `5(x/3)`.  Without the `mq_precedence` it becomes `(5x)/3`.  So the student sees in the preview what the student sees in the MathQuill input.  This is even worse with something like $\frac{x-3}{3}\frac{x+5}{7}$.  In that case ` (x-3)/3  (x+5)/7 ` is submitted for the answer, and without the `mq_precedence` that becomes $\frac{\frac{x-3}{3}(x+5)}{7}$ in the preview. Yes, they are equal mathematically speaking, but it can be confusing to students to enter one thing and be showed a preview that looks quite different.

Before the `mq_precedence` was implemented this was handled by adding extra parentheses around fractions in the text that was submitted for the answer.  However, that caused issues with numbers with units (if the unit is a fraction) and mixed numbers (in the `Fraction` context).

drgrice1:
The issue with MathQuill does seem to be resolved with the last commit by the way.  The preview shows the units properly formatted when MathQuill is in use.

dpvc:
Thanks for the explanation for the 2.9 precedence.  That is what I figured it was.

I'm wondering whether the precedence for ` *` should be set to 2.9 regardless of whether MathQuill is used, as I would guess students would want `(x-3)/3 (x+5)/7` to be rendered $\frac{x-3}{7}\,\frac{x+5}{7}$ regardless.  That would also make `2 x/3` be $2\frac{x}{3}$, but still leave `2x/3` as $\frac{2x}{3}$.

Having  the precedence change based on the presence/absence of MathQuill feels wrong to me, and has the potential to break things of someone is adjusting the precedences and doesn't know to adjust the `mq_precedence` as well.

drgrice1:
I think that should be fine.  The intent with the `mq_precedence` was to avoid changing existing behavior.  But if you think that changing the wouldn't cause issues for existing behavior then that would be fine.

Note that there is one other place that the `mq_precedence` is used.  That is for the `' '` implied multiplication operator with `contextFraction.pl` that is needed to make mixed numbers work right in combination with the general `' *'` precedence.

drdrew42 (dismissed):

somiaj (approved):

dpvc:
> Note that there is one other place that the `mq_precedence` is used. That is for the `' '` implied multiplication operator with `contextFraction.pl` that is needed to make mixed numbers work right in combination with the general `' *'` precedence.

It turns out that neither of the precedences for `' '` have an effect (neither `precedence` nor `mq_precedence') since the changes that made it possible to have unicode alternatives for the operators.  The line

https://github.com/openwebwork/pg/blob/10746b18719cfea63f335dcd4ed4240613d285ed/lib/Parser.pm#L300

replaces the definition of an operator that has a `space` property with the definition of its spaced version.  Since `' '` has `space` set to `' *'`, when `' '` is found, it is replaced by `' *'` before any of the precedences are used in the code below that.  That means the precedences in the line

https://github.com/openwebwork/pg/blob/10746b18719cfea63f335dcd4ed4240613d285ed/macros/contexts/contextFraction.pl#L245

can be removed (that is, if we keep a legacy copy of `contextFraction.pl`).

It was actually the line

https://github.com/openwebwork/pg/blob/10746b18719cfea63f335dcd4ed4240613d285ed/macros/contexts/contextFraction.pl#L246

that sets the `mq_precedence` for `' *'` to 2.8 that was providing the precedence for the implicit multiplication for a space.  But since

https://github.com/openwebwork/pg/blob/10746b18719cfea63f335dcd4ed4240613d285ed/lib/Parser/Context.pm#L218

sets the `mq_precedence` to 2.9 for **all** contexts (that are descendants of the `Numeric` context), and since there are no other 2.9 precedences, that makes 2.8 and 2.9 effectively the same, so you don't need the `mq_precedence` in the definition for the Fraction context.

Since 2.9 has been used for all contexts by many people who are using MathQuill, I suspect that setting the main `precedence` to 2.9 rather than 3 (and removing all the `mq_precedence` references) should be OK.

Since this has been merged, I guess that should be a separate PR.

[I had written this up earlier, but apparently neglected to press the "Comment" button, as it doesn't seem to have made it into the list above, so I've typed it up again.  Apologies if it came through to anyone a second time.]

PR #1109, opened 2024-09-01
Add an en-GB localization file.

drgrice1:
This isn't exactly needed, but is added for consistency.  As for webwork2, translation isn't really needed in the `.po` file.  However, if someone wants to do so it could be done.

Note that the `en.po` file was updated as well.  This file hasn't been updated for a while since it really doesn't need to be.


somiaj (approved):
I don't think there is anything to actually test here, since there are no dates on the pg side of things, and no transitions are being used yet.

drgrice1:
Yeah, this is just for completeness. Also, I think it would be an error if the file did not exist and the language was selected when a problem renders.

somiaj:
I did test a few problems and got no errors, so that at least is good.

somiaj:
FWIW, it seems using `en-GB` in webwork without this applied still works without any error when rendering problems, so looks like `maketext` just returns the inputted string without any error if given localization which doesn't have a `.po` file.

pstaabp:
If anyone else is testing this, you also need to checkout openwebwork/webwork2#2553

pstaabp (approved):
Not much to test here without translations.

PR #1111, opened 2024-09-02
Typos

Alex-Jordan:
This is similar to https://github.com/openwebwork/webwork2/pull/2554, but for `pg`.

### DIFF
--- a/lib/Match.pm
+++ b/lib/Match.pm
@@ -11,7 +11,7 @@ All items accessed by $out = $ml -> item( $in );
 
 =head1 SYNOPSIS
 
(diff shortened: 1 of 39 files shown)


lib/Match.pm
```diff
@@ -11,7 +11,7 @@ All items accessed by $out = $ml -> item( $in );
...
+Match.pm is intended to be used to create standard matching questions in which the student is given a list of questions and a list of answers and is asked to match the correct answers to those questions. Some answers may be used more than once while others are not used at all. The order of answers is usually random but some answers can be appended to the end of the list in a set order (i.e. 'None of the above', 'All of the above'). Answers are not directly typed in but are given a corresponding letter that is the answer that the system expects. (i.e. 'the answer to #1 is A' not 'the answer to #1 is the square root of 2').  Also, students can be given different sets of questions (to avoid students sharing answers) by entering many questions and then using choose with a number less than the total so that each student only receive a sub-set of those questions.
```
somiaj, line 14: While you are fixing things, want to make this multiple lines less than 120 characters as well (since perltidy leaves POD alone)?

somiaj (approved):
Looks good, though there is one very long line that could used split up.

lib/Match.pm
drgrice1: Yeah, I think there are still quite a few of these long lines in the code that need to be fixed.  If they occur in POD or just a long string, then perltidy won't fix them.  So those need to be manually fixed.  Line 113 is another such long line in this file.

drgrice1 (approved):

PR #1112, opened 2024-09-03
Add option to set size for which parserPopUp prints an inline string.

somiaj:
In hardcopy output, allow the size (number of characters) that an inline string [A/B/C] vs a bullet list is printed to be configurable via a new option `inlineSize`. Note, this slightly changes the default behavior, if the total characters is `inlineSize = 25`, then the list is now inline instead of a bullet list.

### DIFF
--- a/macros/parsers/parserPopUp.pl
+++ b/macros/parsers/parserPopUp.pl
@@ -137,6 +137,15 @@ of answer options might make printing the answer option list
 unnecessary in a static output format.)  Default: 1, except 0 for
 DropDownTF.
 
+=item C<S<< inlineSize => number >>>
+
+In PDF output, this controls break point between printing the list of
+options as an inline list [A/B/C] or multiline bullet list. If the total
+number of characters (of all items combined) is greater than or equal to
+this setting, or an item contains a '[', '/', or ']' character, then the
+list of options is printed as a bullet list, otherwise an inline list is
+used. Default: 25
+
 =back
 
 To insert the drop-down into the problem text when using PGML:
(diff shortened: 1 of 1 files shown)


somiaj:
I have various problems that use drop down menus to fill in the blank in sentences. I do like these being displayed inline in hardcopy rather than a bullet list, hence adding this feature to configure the size this happens, as a few of my cases where just barely over the break point.

I thought about maybe using this for a longer list of choices to keep them in line, but ran into another issue where a single line `[A/B/C]` can get a little long, and was thinking of allowing configuring the separator, `/`, so I can put a space on each end, and the list `[A / B / C]` can wrap at more natural breaks. I do know I could use `showStatic => 0` in those cases, but I was just seeing what others would think about allowing configuration of the inline separator as well.

Alex-Jordan:
This seems fine, although I think that before you have to be strictly less than 25 characters long to get it printed inline, but  here its inline when 25 or less. I didn't test, just reading the code. I may be missing a detail.

drgrice1:
@Alex-Jordan is correct.  The previous condition was `length(join('', @list)) < 25 && !grep(/(\/|\[|\])/, @list)`.  So when that was negated to work with `unless` and the new `inlineSize` option added it should have become `length(join('', @list)) >= $self->{inlineSize} || grep(/(\/|\[|\])/, @list)`.

somiaj:
I was aware that I changed the behavior, and even mentioned it in the PR description (as I thought is simplified the logic and the description a bit). I have changed it so now the behavior is the same.

drgrice1 (approved):
I have no problem with this change, and don't see any reason not to allow it.

pstaabp (approved):
Looks good.

PR #1114, opened 2024-09-03
Update contextReaction.pl with several new features

dpvc:
This PR updates the Reaction context to include several new features.  These are documented in the changes to the POD section, but the main additions are:

* The ability to give ions with postfix notation, as in `Cl^{2-}` or `Na^+`.
* Ability to enter complexes using brackets, as in `[CoCl_4(NH_3)_2]^−`
* The ability to combine copies of the same element, molecule, etc., so that `O + O` and `2O` are equivalent.
* The ability to specify whether students must include states when the correct answer does, or whether a stateless answer is marked correct.  This is controlled by a new flag: `studentsMustUseStates` whose default is true.

This is work I did a while ago, but never made a PR for it until now.

### DIFF
--- a/macros/contexts/contextReaction.pl
+++ b/macros/contexts/contextReaction.pl
@@ -7,38 +7,45 @@ contextReaction.pl - Implements a MathObject class for checmical reactions.
 
 This file implements a Context in which checmical reactions can be
 specified and compared.  Reactions can be composed of sums of integer
-multiples of elements (possibly with subscripts), separated by a right
-arrow (indicated by "-->").  Helpful error messages are given when a
-reaction is not of the correct form.  Sums of compounds can be given
-in any order, but the elements within a compound must be in the order
-given by the correct answer; e.g., if the correct answer specifies
-CO_2, then O_2C would be marked incorrect.
+multiples of elements, molecules, ions, compounds, and complexes
+separated by a right arrow (indicated by C<< --> >>).  Helpful error
+messages are given when a reaction is not of the correct form.  Sums
+of compounds can be given in any order, but the elements within a
+compound must be in the order given by the correct answer; e.g., if
+the correct answer specifies C<CO_2>, then C<O_2C> would be marked
+incorrect.
 
 To use the context include
 
     loadMacros("contextReaction.pl");
     Context("Reaction");
 
-at the top of your PG file, then create Formula() objects for your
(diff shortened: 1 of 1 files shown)


pstaabp:
Testing some of the examples in the POD, the complexes that you mentioned, I tried to create:

```
Formula("[CoCl_4(NH_3)_2]^−");
```
and got the error:
```
Unexpected character '−'; see position 18 of formula at line 15 of setunits/reaction.pg
   from within main::Formula called at line 15 of setunits/reaction.pg
```
and that line cited above is line 15. 

I tested other  cases using `^-` without problems.

pstaabp:
@drgrice1 when testing this with MQ, for the --> in the reactions, the `-->` show up in MQ as 3 characters.  Could we add the --> to create a \longrightrarrow ?  Not a huge thing, but would look nice.

drgrice1:
Perhaps.  Add a feature request issue to https://github.com/openwebwork/mathquill, and I will look into it when I have time.

dpvc:
> I tried to create:  `Formula("[CoCl_4(NH_3)_2]^−");` and got the error: ...

It turns out that the minus is the Unicode minus at U+2212 rather than the hyphen-minus at U+002D.  Since U+2212 is not defined as an operator, you get that error.  

I must have copied the TeX output or something.  Sorry about that.  I suppose we could add the alternatives like in other contexts.  I'll make a PR to fix the comment.

drgrice1 (approved):
I think that this pull request looks good.

I see a few typos in the POD and comments (most of them not new).  Rather than commenting in the code, I will just list them by line number to prevent the lengthy comments in the GitHub conversation.

* lines 4 and 8: `checmical` -> `chemical`
* line 41: `brakets` -> `brackets`
* line 75: `corect` -> `correct`
* line 82 `Context()->flags-set` -> `Context()->flags->set`
* line 377: `somone` -> `someone`
* line 458: `chemcicals` -> `chemicals`
* line 810: `othewise` -> `otherwise`

The updated POD looks nice.

I did some basic testing with the examples in the POD, and at least those seem to be working.  My memory of chemistry from classes way back when I was an undergrad is rather faint, so I don't think I can really test this particularly thoroughly.

dpvc:
@drgrice1, I have fixed the spelling errors.  Thanks for the close reading!

drdrew42 (approved):

PR #1117, opened 2024-09-06
Use `@openwebwork/mathquill` from npm instead of directly using github.

drgrice1:
A beta version of an `@openwebwork/mathquill` package has been published on npm.  This switches to using that instead of from github.  Before the next release, a non-beta version will be published and this updated to use that version.

The github repository will also be updated to not include the distribution build anymore.  This fixes the mess of always needing to reubild the `dist` files for every change to the mathquill code.  That is a real pain when different pull requests are made, particularly unrelated pull requests, as they have to be put in order with sequential builds.

### DIFF
--- a/htdocs/package-lock.json
+++ b/htdocs/package-lock.json
@@ -1,2823 +1,2863 @@
 {
-	"name": "pg.javascript_package_manager",
-	"lockfileVersion": 2,
-	"requires": true,
-	"packages": {
-		"": {
-			"name": "pg.javascript_package_manager",
-			"license": "GPL-2.0+",
-			"dependencies": {
-				"jsxgraph": "^1.9.2",
-				"jszip": "^3.10.1",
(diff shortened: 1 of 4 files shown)


pstaabp:
I had to do a `nom ci` to get mathquill pulled down from npm.

drgrice1:
Generally anytime anything in the `htdocs` directory is changed (other than the `helpFiles`), you need to run `npm ci` to make changes take effect.  I didn't mention it, because I though by now that would be obvious.

pstaabp (approved):

somiaj (approved):

somiaj:
Two approvals so I'll merge. Works in my testing in both using MathQuill and past answers.

drgrice1:
Since this is merged, https://github.com/openwebwork/mathquill/pull/32 should also be reviewed and merged since this technically depends on that.

PR #1118, opened 2024-09-12
prevent extra p surrounding multiple choice ul in PTX

Alex-Jordan:
This should have been included in #1090.

Without getting into the weeds too much, consider something like:
```
$rational = DropDown(["is","is not",], 1);

BEGIN_PGML
The number [`\sqrt{2}`] [_]{$rational}{5} rational.
```

Without this commit, the PTX output is like:
```
<p>The number <m>\sqrt{2}</m> <p><ul form="popup" ...> ...
```
Note the `p` within a `p`.

With this commit, the PTX output is:
```
<p>The number <m>\sqrt{2}</m> <ul form="popup" ...> ...
```

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -2193,7 +2193,10 @@ sub PTX_cleanup {
 	if ($displayMode eq 'PTX') {
 		#encase entire string in <p>
 		#except not for certain "sub" structures that are also passed through EV3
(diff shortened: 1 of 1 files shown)


somiaj (approved):
I can see the additional regex matches your description, but again don't have a way to test.

drgrice1 (approved):
Looks good.

PR #1123, opened 2024-09-17
move the essay help text to a general helpLink

Alex-Jordan:
In `PGessaymacros.pl` there is a subroutine to print help text. This moves that to be like other help links, and at the same time changes the text content to mention the equation editor, among other things.

### DIFF
--- a/htdocs/helpFiles/Entering-Essays.html
+++ b/htdocs/helpFiles/Entering-Essays.html
@@ -1,0 +1,12 @@
+<h2 style="text-align: center">Entering Essays</h2>
+
+<p>
+	In an essay answer text box, you can type a long answer. After after you hit submit, it will be saved so that your
+	instructor can read it and assign a score at a later time.
+</p>
+<p>If your instructor makes any comments on your answer, those comments will appear on the exercise page.</p>
+<p>
+	You can use the MathQuill equation editor to make your math content look pretty. Once you have constructed a math
+	expression, click the Insert button. In the answer text box, your math content will look like code, but the Preview
+	button will show you what your isntructor will see.
+</p>
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -2280,13 +2280,15 @@ sub PTX_cleanup {
                     # form(from the webpage) inside another (the defining form for the problem
 A wide variety of google widgets, youtube videos, and other online resources can be imbedded using this macro. In HTML mode it creates an iframe, in TeX mode it prints the url.
 
-    helpLink($type)     allows site specific help. specified in global.conf or course.conf
+    helpLink($type)
+                   Allows site specific help.
                    The parameter localHelpURL  must be defined in the environment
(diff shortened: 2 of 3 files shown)


somiaj:
Overall looks good, I can't fully test (not an issue of this PR) because on my development server `$pg_root_url` used in `conf/pg_config.dist.yml` isn't defined and I just get a 404 when viewing any help file since the link is just `/helpFiles/Entering-Essays.html` instead of `/pg_files/helpFiles/Entering-Essays.html`. I tried grepping for where `$pg_root_url` is set set and cannot find it. So unsure why this isn't set on my development server but works just fine on my live server (for other help links).

somiaj:
Hmm, even manually set `localHelpURL: /pg_files/helpFiles` in `conf/pg_config.yml` had no effect on the url that is being used for the help files. Could this have something to do with `/webwork2-morbo`? (I even restarted the binary with no change).

htdocs/helpFiles/Entering-Essays.html
```diff
@@ -0,0 +1,12 @@
+<h2 style="text-align: center">Entering Essays</h2>
+
+<p>
+	In an essay answer text box, you can type a long answer. After after you hit submit, it will be saved so that your
+	instructor can read it and assign a score at a later time.
+</p>
+<p>If your instructor makes any comments on your answer, those comments will appear on the exercise page.</p>
```
somiaj, line 7: I find "on the exercise page" a bit strange. I would call it the "problem page", or maybe just as the help before this does, "on this page". Maybe specify the comment will be at the top of the page.

htdocs/helpFiles/Entering-Essays.html
```diff
@@ -0,0 +1,12 @@
...
+<p>
+	You can use the MathQuill equation editor to make your math content look pretty. Once you have constructed a math
+	expression, click the Insert button. In the answer text box, your math content will look like code, but the Preview
+	button will show you what your isntructor will see.
+</p>
```
somiaj, line 12: I would also mention that latex can be directly entered using `\( ... \)`, or use the equation editor to insert formulas. Could be useful for users who know a little latex but unsure about how to insert it.

somiaj (approved):
Overall looks good. The issues I had were related to my development setup and not this PR at all. I left some comments about the actual help file, but am approving this as is.  The comments are just things that came to mind when reading it over, and I'm okay with leaving it as is or addressing them.

Alex-Jordan:
Thanks @somiaj for looking at this. I will think about rewording that help page. I try to use "exercise" rather than "problem", but I think something good can be found. I would prefer to not mention LaTeX though. I think it would make a worse problem, confusing/intimidating the students who don't know what that means. Also it would get pretty verbose, because even students who know LaTeX may not know about using `\(` instead of `$`. Once they use the editor and see actual LaTeX show up, I feel that studnets who know LaTeX will know what to do.

pstaabp (approved):

drgrice1 (approved):
Looks good.

PR #1127, opened 2024-09-24
Fix an uninitialized value in the openDivSpan method of PGbasicmacros.pl

drgrice1:
This was reported in https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4767#p21510. See some of the problems posed in that thread for testing.


somiaj (approved):
Tested and this fixes the warnings.

pstaabp (approved):
This fixes the issue. 

Should this also be a hot fix?

PR #1131, opened 2024-10-01
When typing in a MathQuill input, update positions of all feedback popovers.

drgrice1:
This fixes the issue reported by @somiaj in
https://github.com/openwebwork/pg/pull/1110#issuecomment-2380357222.


somiaj (approved):
Confirmed it fixed the issue by moving all popups when mathquill inputs resize.

Issue #1129, opened 2024-09-30
Feedback button for 1x1 matricies appear inside the matrix brackets.


somiaj:
Adding this here just so it is known. Here is a test problem of different matrix sizes.

```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl));

$M1 = Matrix([1]);
$M2 = Matrix([ [1], [2], [3] ]);
$M3 = Matrix([ [1, 2, 3] ]);
$M4 = Matrix([ [1, 2], [3, 4] ]);

BEGIN_PGML
[``[$M1]=``] [_]*{$M1}

[``[$M2]=``] [_]*{$M2}

[``[$M3]=``] [_]*{$M3}

[``[$M4]=``] [_]*{$M4}
END_PGML
ENDDOCUMENT();
```

Here is the feedback button for a 1x1 matrix. All the rest appear outside the last bracket.

![image](https://github.com/user-attachments/assets/c56de072-ba8c-4d1a-a07f-e30bb5d3f446)

drgrice1:
This was fixed with #1132.

PR #1132, opened 2024-10-01
Move feedback button for one by one matrix out of matrix.

drgrice1:
A one by one matrix does not have a last answer rule with the MaTrIx prefix or the row and column indices for a suffix.  In this case, the last answer name is of course also the first answer name.  As such the data attribute added was not naming the correct answer rule.

This fixes issue #1129.


somiaj (approved):
Confirmed this puts the feedback button outside the matrix brackets.

PR #1133, opened 2024-10-09
Update WeBWorK::PG::Tidy to handle content after ENDDOCUMENT.

drgrice1:
The content is removed but saved before perltidy is called.  Then put back after perltidy is done.  So its contents will not change.  That is with the exception of possibly adding `();` if that is not there.  Note that if you have something like `ENDDOCUMENT('invalid arg');` the `'invalid arg'` will be removed.

### DIFF
--- a/lib/WeBWorK/PG/Tidy.pm
+++ b/lib/WeBWorK/PG/Tidy.pm
@@ -42,13 +42,20 @@ use warnings;
 use Perl::Tidy;
 use Mojo::File qw(curfile);
 
-our @EXPORT = qw(pgtidy);
+our @EXPORT_OK = qw(pgtidy);
 
 my $perltidy_pg_rc = curfile->dirname->dirname->dirname->dirname . '/bin/perltidy-pg.rc';
+
+my $contentAfterEnd;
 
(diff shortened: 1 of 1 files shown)


drgrice1:
I also add a minor correctness tweak.  Instead of placing the `~~` first, place it last.  So if the original problem has `\\~~`, that becomes `\\\\\` in pre processing.  Then it becomes `\\~~` again in post processing.  However, if the original problem has `~~\\`, then that also becomes `\\\\\` in pre processing, but becomes `\\~~` in post processing.

This is for correctness of escaping when interpreting `~~` as the escape sequence for PG.  After translation both `\~~` and `~~\` becomes `\\\`, so the effect is identical.  However, what that means after translation is that the first backslash escapes the second, and the third backslash escapce whatever is after it.  So `\~~` is more accurate in meaning when thinking of `~~` as the escape sequence because the first `\` is now `\\`, and the `~~` escapes what is next the same as the last `\` does
after translation.

somiaj (approved):
Works as described.

pstaabp (approved):
This works well.

PR #1135, opened 2024-10-16
Center the RadioMultiAnswer error messages.

drgrice1:
The left aligned style worked well with the attempts table, but does not look so good with the new feedback buttons.  Having the messages centered looks better now since everything else is centered.

A problem to test this with is attached.  The problem is set up to give a message for any answer.
[appendMessageMinimal.zip](https://github.com/user-attachments/files/17388760/appendMessageMinimal.zip)

### DIFF
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -178,7 +178,7 @@ sub single_check {
 		push(
 			@errors,
 			main::tag(
-				'tr', main::tag('td', style => 'text-align:left', colspan => '2', $self->{ans}[0]{ans_message})
+				'tr', main::tag('td', style => 'text-align:center', colspan => '2', $self->{ans}[0]{ans_message})
 			)
 		);
 		$self->{ans}[0]{ans_message} = "";
@@ -211,7 +211,7 @@ sub single_check {
 		$ans->{ans_message} = $ans->{error_message} = main::tag(
 			'table',
 			class => 'ArrayLayout',
-			style => 'width:100%',
+			style => 'margin-left:auto;margin-right:auto;',
 			join(main::tag('tr', style => 'height: 4px', main::tag('td')), @errors)
 		);
 	}
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -13,7 +13,7 @@
 # Artistic License for more details.
 ################################################################################
 
-loadMacros("MathObjects.pl");
+loadMacros('MathObjects.pl', 'PGbasicmacros.pl');
 
 sub _parserMultiAnswer_init {
 	main::PG_restricted_eval('sub MultiAnswer {parser::MultiAnswer->new(@_)}');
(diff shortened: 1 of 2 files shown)


somiaj:
Think MulitAnswer should be updated as well, their tags still left align error messages.

somiaj:
Though I guess MultiAnswer is wrapped in a table, so it isn't quite the same situation.

drgrice1:
The RadioMultiAnswer is also wrapped in a table.  I also had thought that MultiAnswer had this issue also, but in a quick test I didn't see the issue.  I will look at it again.

drgrice1:
I see.  So I rarely (if ever) use the MultiAnswer `setMessage` method.  Instead I use the `addMessage` method (with `singleResult` set).  The messages added by `addMessage` (which go into the `single_ans_messages` array) are placed in consecutive `div`s instead of the table.  Those messages are centered.  So that is why my brief test didn't show an issue.

I tested using the `setMessage` method instead, and see that those messages aren't centered.  I think the only thing that can be done here is to set `width:100%` on the table.

drgrice1:
I added another commit to this pull request that sets `width:100%` on the MultiAnswer error message table.  It also switches to using the `PGbasicmacros.pl` `tag` method for generation of the HTML, and removes the deprecated attributes that were used.  Note that the `ArrayLayout` css class already does what the deprecated attributes were doing.

somiaj:
I notice a few things testing this with parserMultiAnswer. First minor thing, `setMessage(1, 0);`, doesn't actually show a message and I was confused for a while until I caught that (was setting messages in a loop). Probably not a big deal, I doubt anyone would set the message to zero outside of testing.

Second, I don't think the table width being 100% looks correct. Here is what I'm seeing.

scrot1

If I use the inspector to remove the width from the table, but set the `mx-auto` class instead it looks better to me. Though I assume we need to not use the bootstrap class but its equivalent in pg.

scrot2

Last, I use `Value->Error('message')` out of habit (need to switch to `addMessage`), and those are still left aligned which also looks off to me.

scrot3

somiaj:
Here is the test case I was using.

```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl parserMultiAnswer.pl));

$ma = MultiAnswer(0 .. 3)->with(
    singleResult => 1,
    checker      => sub {
        my ($c, $s, $self, $ans) = @_;
        Value->Error('Test');
        for my $i (0 .. 3) {
            $self->setMessage($i + 1, (A .. D)[$i]);
        }
        return $score;
    }
);

BEGIN_PGML
0 = [_]{$ma}, 1 = [_]{$ma}, 2 = [_]{$ma}, 3 = [_]{$ma}
END_PGML
ENDDOCUMENT();
```

drgrice1:
I switched to centering the tables for the messages, instead of making them width 100%.

As to the issue with `setMessage(1, 0)`, that has nothing to do with this pull request.  Also, I don't really see that as being something that we need to support even for test usage.

drgrice1:
I also centered the error messages that occur for parserMultiAnswer.pl.

drgrice1:
The error messages for parserRadioMultiAnswer.pl are also centered.

somiaj:
I agree, no need to deal with a 0 value here, just something that initially confused me so I pointed it out.

somiaj (approved):

PR #1140, opened 2024-11-06
Add MultiAnswer convenience methods (default checker and passing cmp options).

somiaj:
This adds two conveniences to `parserMultiAnswer.pl`.

* A `cmpOpts` option (taken from `parserRadioMultiAnswer.pl`) to pass a hash of options to all cmp methods.
* A default checker that is used if no checker is defined. This checker just checks which answer parts are equal, and can either return an array of 0s and 1s for each answer part, or a single score of 0 or 1, depending on if the new option `partialCredit` is set or not. `partialCredit` defaults to the value of `$showPartialCorrectAnswers`.

### DIFF
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -103,6 +103,7 @@ sub cmp {
 	}
 
 	unless (ref($self->{checker}) eq 'CODE') {
+		die "Your checker must be a subroutine." if defined($self->{checker});
 		$self->{checker} = sub {
 			my ($correct, $student, $self, $ans) = @_;
 			my @scores;
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -62,6 +62,7 @@ sub new {
 		part                => 0,
 		singleResult        => 0,
 		namedRules          => 0,
+		cmpOpts             => undef,
 		checkTypes          => 1,
 		allowBlankAnswers   => 0,
 		tex_separator       => $separator . '\,',
@@ -89,8 +90,10 @@ sub setCmpFlags {
 #  the individual answer checkers.
 #
 sub cmp {
-	my $self    = shift;
-	my %options = @_;
+	my ($self, %options) = @_;
+
+	%options = (%options, %{ $self->{cmpOpts} }) if ref($self->{cmpOpts}) eq 'HASH';
+
 	foreach my $id ('checker', 'separator') {
 		if (defined($options{$id})) {
 			$self->{$id} = $options{$id};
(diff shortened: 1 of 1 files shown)


macros/parsers/parserMultiAnswer.pl
```diff
@@ -488,6 +518,13 @@ =head2 namedRules
 if you need to intersperse other rules with the ones for the C<MultiAnswer>. In this case, you must
 use C<NAMED_ANS> instead of C<ANS>. Default: 0.
 
+=head2 cmpOpts
+
+This is a hash of options that will be passed to the cmp method of all answers. For example,
+C<< cmpOpts => { weight => 0.5 } >>. This option is provided to make it more convenient to pass
+options to all answers when utilizing PGML. To pass options to a single answer, use the
+C<< setCmpFlags($which_rule, %flags) >> method instead. Default: undef (no options are sent).
```
drgrice1, line 526: This documentation is not correct.  Regardless of if `singleResult` is set, this is not a substitute for the `setCmpFlags` method, and in no case are these flags actually passed to the `cmp` method for all answers.  There are two sets of answer evaluators that a MultiAnswer object uses.  There is a set of answer evaluators that are associated with the answer rules (there is only one of these if `singleResult` is set, and one for each answer otherwise), and there are answer evaluators for each answer that are called internally by the `entry_check` or `single_check` methods (there are always as many of these as there are answers regardless of the `singleResult` setting).  The only way to set flags for the internally called answer evaluators you must use the `setCmpFlags` method.  The things set in `cmpOpts` won't be passed to those evaluators.

somiaj: Thanks for the explanation. I got confused when I thought it was implied at the meeting I could have used that method to add cmp options. I have updated the docs to mirror what you have in parserRadioMultiAnswer.

macros/parsers/parserMultiAnswer.pl
```diff
@@ -89,15 +91,33 @@ sub setCmpFlags {
...
+
 	foreach my $id ('checker', 'separator') {
 		if (defined($options{$id})) {
 			$self->{$id} = $options{$id};
 			delete $options{$id};
 		}
 	}
-	die "You must supply a checker subroutine" unless ref($self->{checker}) eq 'CODE';
+
+	unless (ref($self->{checker}) eq 'CODE') {
```
dpvc, line 105: You may wish to produce an error message if `checker` is defined but not a code reference, rather than just replacing it silently.

somiaj: Thanks for the suggestion.

dpvc: I think you could put the check inside the `unless` clause, as 

``` perl
unless (ref($self->{checker} eq 'CODE') {
  die "Your checker must be a subroutine." if defined($self->{checker});
```

instead of having to check for `CODE` two times.

drgrice1 (approved):
This looks good.

Alex-Jordan:
Is it right that the default checker does not take into consideration whether `singleResult` is true or false?

drgrice1:
The checker should work for both values of `singleResult`.  In either case an array of numbers from 0 to 1 or a single number from 0 to 1 is accepted.  The initial handling of scores is done by the `perform_check` method in either case, and it assigns the scores to all answers.  If an array of numbers is returned then the array must have the same size as the number of answers, and those scores are assigned to the respective answers.  If a single number is returned, then that score is assigned to all answers.  That is true regardless of if `singleResult` is true or false.  In the case that `singleResult` is true, the `single_check` method then puts those together into a single score and sets the score of the primary evaluator.

Alex-Jordan:
OK, thanks for that explanation. I have always assumed that if `singleResult` is 1, then I must write a checker that returns a number. And if `singleResult` is 0, that I must write a checker that returns an array reference. But I understand now that is not true.

Alex-Jordan (approved):

PR #1141, opened 2024-11-07
Fix issue with variable TeX output with var = value inequality.

somiaj:
In contextInequalities, the $context variable could be undefined, making it so the TeX formatting of a variable in inequalities of the form Inequality("var = value") was not used. This defines the variable $context = $self->context in the string and TeX methods of Inequalities::Set to ensure any variable definitions are used in the string and TeX output.

Here is a test problem to use, without this the variable's TeX definition is not used.

```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl contextInequalities.pl));
Context('Inequalities')->variables->are(gamma => [ 'Real', TeX => '\gamma ' ]);
$test = Inequality("gamma = 5");

BEGIN_PGML
[`[$test]`]
END_PGML
ENDDOCUMENT();
```

### DIFF
--- a/macros/contexts/contextInequalities.pl
+++ b/macros/contexts/contextInequalities.pl
@@ -890,7 +890,8 @@ sub type {"Set"}
 sub string {
 	my $self     = shift;
 	my $equation = shift;
-	my $x        = $self->{varName} || ($self->context->variables->names)[0];
+	my $context  = $self->context;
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Looks good.  Clearly, the `$context` would never be defined with the current code which is an obvious issue.  I tested with the provided example and it checks out.

PR #1143, opened 2024-11-17
Add usesAllOf method to Value::Formula.

somiaj:
This method is similar to usesOneOf, but only returns true if all of the variables in the list are used. This is a convenience method to not have to call usesOneOf multiple times to check if multiple variables are being used.


drgrice1 (approved):
This looks good.

PR #1146, opened 2024-11-22
Switch to using the latest Perl::Tidy (version 20240903) alternate approach.

drgrice1:
A new `--valign-signed-numbers` option has been added in the version of Perl::Tidy and is the default.  That option has been disabled to reduce the number of code changes.
    
All of the other changes to the code made by the updated version are definite improvements.
    
Note that the `bin/dev_scripts/run-perltidy.pl` script now insists on this specific version of perltidy instead of this version or newer.  This is because every time a new version of perltidy comes out, it changes things.  So developers really need to be using the same version or it will not be consistent with the github workflow result.

Finally, the `-vxl='q'` option to disable alignment of qw quotes has been removed from the `bin/perltidy-pg.rc` file.  The upside of this is that those using the `WeBWorK::PG::Tidy` module (either via the `bin/perldity-pg.pl` script or via webwork2) can use any version of Perl::Tidy, even those from before the 20220613 version.

The first commit makes the changes for the switch to the new version, and the second commit is merely the result of running `perltidy` with the new version and rules.

### DIFF
--- a/.github/workflows/check-formats.yml
+++ b/.github/workflows/check-formats.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
       - name: Install dependencies
-        run: cpanm -n Perl::Tidy@20220613
+        run: cpanm -n Perl::Tidy@20240903
       - name: Run perltidy
         shell: bash
         run: |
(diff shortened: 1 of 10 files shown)


somiaj:
Going with #1145.

PR #1147, opened 2024-11-24
Add a `parserLogb.pl` macro for logarithms with base b.

drgrice1:
This is similar to the `parserRoot.pl` macro.  To enable the usage of the `logb` function in the current context call `Parser::Logb->Enable`.

Then you can call `$n = logb(3, 5)` to get the value of the logarithm with base 3 evaluated at 5, or `$ans = Compute("logb(3, 5)")` to  use as an answer.

You can also call `Parser::Logb->EnableComplex` to allow negative bases and evaluation at negative numbers. Note that attempting to use a negative base or evaluate at a negative number would otherwise produce an error message.

An answer in this context will automatically have the MathQuill option `logsChangeBase => 0` set.  To facilitate this without the need to have numerous special cases for contexts in the `ENDDOCUMENT` method of `PG.pl`, there is a new `mathQuillOpts` context flag.  The default value for this flag is a reference to a hash with no keys.  Any context can set keys in this hash, and those will be tranferred to the MathQuill options for any answer rule in the context.  This is now used by the the `parserRoot.pl` macro to set `rootsAreExponents => 0`, and by the new `parserLogb.pl` macro to set `logsChangeBase => 0`.

To facilitate students entering a logarithm with a base, if the `logsChangeBase` option is 0 (or not set), then a subscript button is added to the MathQuill toolbar.

Note that MathQuill is already set up to output the text `logb(b, x)` if `log_b(x)` is entered.

### DIFF
--- a/htdocs/js/MathQuill/mqeditor.js
+++ b/htdocs/js/MathQuill/mqeditor.js
@@ -309,6 +309,9 @@
 			{ id: 'sqrt', latex: '\\sqrt', tooltip: 'square root (sqrt)', icon: '\\sqrt{\\text{ }}' },
 			{ id: 'nthroot', latex: '\\root', tooltip: 'nth root (root)', icon: '\\sqrt[\\text{ }]{\\text{ }}' },
 			{ id: 'exponent', latex: '^', tooltip: 'exponent (^)', icon: '\\text{ }^\\text{ }' },
+			...(cfgOptions.logsChangeBase
+				? []
+				: [{ id: 'subscript', latex: '_', tooltip: 'subscript (_)', icon: '\\text{  }_\\text{  }' }]),
 			{ id: 'infty', latex: '\\infty', tooltip: 'infinity (inf)', icon: '\\infty' },
 			{ id: 'pi', latex: '\\pi', tooltip: 'pi (pi)', icon: '\\pi' },
 			{ id: 'vert', latex: '\\vert', tooltip: 'such that (vert)', icon: '|' },
--- a/lib/Parser/Context/Default.pm
+++ b/lib/Parser/Context/Default.pm
@@ -443,6 +443,7 @@ $flags = {
 	parseAlternatives          => 0,              # 1 = allow parsing of alternative tokens in the context
 	convertFullWidthCharacters => 0,              # 1 = convert Unicode full width characters to ASCII positions
 	useMathQuill               => 0,
+	mathQuillOpts              => {},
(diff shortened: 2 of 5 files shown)


Alex-Jordan:
I find that I have to set
```
Context()->flags->set(mathQuillOpts => {});
```
before I run `Parser::Logb->Enable;`, or else I get an error like:
```
Can't use an undefined value as a HASH reference at line 92 of [PG]/macros/parsers/parserLogb.pl
   from within Parser::Logb::Enable called at line 6 of undefined
```

Here is the file I am testing with:
```
DOCUMENT();

loadMacros(qw(PGstandard.pl PGML.pl parserLogb.pl));

#Context()->flags->set(mathQuillOpts => {});
Parser::Logb->Enable;
$n = logb(3, 5);

BEGIN_PGML
Enter a value for [`[$n]`].

[_]{$n}{5}
END_PGML

ENDDOCUMENT();
```

Alex-Jordan:
Something came up today where a person used `lg()`, meaning the logarithm base 2. So it comes to mind when looking atht his. We have `log()`, `log10()`, and `ln()`. Is it appropriate to add `lg()`? And if so, is it appropriate to include that as part of this macro?

drgrice1:
I tested with your sample code, and I do not see any errors or warnings about undefined HASH references.  The `mathQuillOpts` context flag should always be defined to be a reference to an empty hash.  It is set in `lib/Parser/Context/Default.pm`.

I am not sure that I agree to adding `lg` for a logarithm with base 2.  That notation seems to be primarily a computer science notation, and it seems to be rather heavily disputed as to whether it is the logarithm with base 2 or base 10 (see for instance https://en.wikipedia.org/wiki/Binary_logarithm#Notation).  This macro is designed to handle logarithms with explicit bases given.  Special notations like `log`, `ln`, `log10` and something like `lg` are not the intent for this macro.

drgrice1:
Ahh, you may need to restart the webwork2 app to get the changes in this pull request to take effect (if you are not using `morbo`).  The change in `Default.pm` occurs when the safe compartment is compiled, and that happens when the webwork2 app is first started.  That may be why you got the error about the undefined value used as a HASH reference.

Alex-Jordan:
Yeah, oops. I had not restarted `webwork2`. This looks good and works in all the ways.

Alex-Jordan (approved):

somiaj (approved):

PR #1148, opened 2024-11-25
Remove a local link in the `package-lock.json` file.

drgrice1:
This is a remnant from testing the recent MathQuill changes locally, that seems to have stayed there even after uninstalling the local link. This was never supposed to have been committed.

### DIFF
--- a/htdocs/package-lock.json
+++ b/htdocs/package-lock.json
@@ -26,43 +26,6 @@
                 "yargs": "^17.7.2"
             }
         },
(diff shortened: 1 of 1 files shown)


somiaj (approved):
I'll just merge this, seems obvious we shouldn't be linking to stuff on your computer.

drgrice1:
Yeah, that was a stupid mistake.  I thought that when the link was uninstalled it would remove that, but it didn't.  In fact, I found the only way to remove it was to manually delete it from the file.  So I will need to remember to check that if I use a local link again.

PR #1149, opened 2024-11-26
Update the sample problems to use the pg-codemirror-editor with CodeMirror 6.

drgrice1:
This updates the script and generated tutorial code to use the pg-codemirror-editor package which uses CodeMirror 6, rather than using CodeMirror 5.  This script is used to generate the html sample problem pages linked to on the webwork wiki.


tutorial/templates/problem-template.mt
```diff
@@ -7,10 +7,11 @@
 	<title><%= $filename %></title>
 
 	<link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha3/dist/css/bootstrap.min.css" rel="stylesheet">
-	<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/codemirror@5.65.11/lib/codemirror.min.css">
-	<script src="https://cdn.jsdelivr.net/npm/codemirror@5.65.11/addon/runmode/runmode-standalone.min.js" defer>
+	<script
+		src="https://cdn.jsdelivr.net/npm/@openwebwork/pg-codemirror-editor@0.0.1-beta.26/dist/pg-codemirror-editor.js"
+		defer
+	>
```
somiaj, line 13: I assume there is no better way than to hard code the version here, just feel this could get overlooked during a release.

drgrice1: Unfortunately, there isn't.  This isn't like the packages installed by npm where the version is automatically handled by the package system.  I don't really like it either, although, it was previously a hard coded version of the CodeMirror 5 package that was there.  So it doesn't really change anything with that.

I plan to remove the beta version before the next release, and once there is a stable version here it should be fine.

Alex-Jordan (approved):

drgrice1:
Note that the version of the `pg-codemirror-editor` that this updates to also includes some fixes for syntax highlighting in interpolation (the fixes are really in the `codemirror-lang-pg`, `codemirror-lang-perl`, and `codemirror-lang-mt` packages).

somiaj (approved):

PR #1150, opened 2024-11-26
Only show non empty messages in feedback.

drgrice1:
Some problems prevent messages by overriding the default message with the string consisting of a single space.  See the sample problem `tutorials/sample-problems/Algebra/EquationImplicitFunction.pg` for example.  If those messages are used, then the feedback button shows the dot indicating the presence of a message even if there really isn't a message.  To fix this, the answer messages are only shown if they do not consist entirely of whitespace.

Note that I do not consider this a good practice for problem authors to employ.  Even the previous attempts table in WeBWorK 2.18 or before would have shown the message column in this case even though there are no actual messages.  Furthermore, in most cases there are better ways to prevent these messages from occuring.  In the case of the sample problem mentioned above, I believe that there is something wrong with `paserImplicitEquation.pl` macro's `createPoints` method that causes the message to be displayed for some answers where it should not be, and that is a bug that should be looked into. So I would be completely fine with not accepting this pull request.

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -1198,9 +1198,9 @@ sub ENDDOCUMENT {
 					'button',
 					type  => 'button',
 					class => "ww-feedback-btn btn btn-sm $options{btnClass} $options{btnAddClass}"
-						. ($rh_envir->{showMessages} && $ansHash->{ans_message} ? ' with-message' : ''),
+						. ($rh_envir->{showMessages} && $ansHash->{ans_message} =~ /\S/ ? ' with-message' : ''),
 					'aria-label' => (
-						$rh_envir->{showMessages} && $ansHash->{ans_message}
+						$rh_envir->{showMessages} && $ansHash->{ans_message} =~ /\S/
 						? maketext('[_1] with message', $options{resultTitle})
 						: $options{resultTitle}
 					),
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):
I think regardless of what motivated you doing this, it is good to do this. If the message is just white space, there really is no message.

drgrice1:
Yeah, that is true.  Generally, there really shouldn't be such messages though.

somiaj (approved):

somiaj:
I agree that this is a hack to work around incorrectly setting error messages, but this is seems easy enough to just suppress messages with white space to allow hacks like the example problem to work. Many problem authors won't always use best practices, but just want to get things to work, so I'm okay with allowing such hacks.

drgrice1:
Yeah, that is for sure.

somiaj:
(note we should probably see if we can fix the example problem mention to not need this, and hopefully this won't hide issues and is merely just a convenience to work around things for problem authors)

drgrice1:
I am actually working on the sample problems now.  That is what led to this.  Unfortunately, the particular issue with that problem is not so easy.

drgrice1:
I think it would be nice to be able to disable certain messages without having to use a hack like overriding the message to be a string consisting only of white space.

somiaj:
I have been wanting a better error system. A more standardized way to associated errors with a key, so they can be translated, have ways to be overridden or suppressed. It would be nice to have ways users can customize the error messages easier than some post filter modifying them.

somiaj:
Though I guess this is the first time I've seen this approach, do only some macros set error messages like this? Though be nice if the key wasn't just the error message string.

drgrice1:
MathObject's in general do this.  All of the messages can be overridden in this way.  This is defined in the `Error` method in Value.pm.

drgrice1:
Sorry, it is in the `setError` method of `Value/Context.pm`.

PR #1151, opened 2024-11-27
make specialAngle return 0 when it would otherwise return 0pi

Alex-Jordan:
Prior to this commit, `specialAngle` can return `Formula("0pi")` and that is not simplified automatically. This makes it return `Formula("0")` instead.

### DIFF
--- a/macros/math/specialTrigValues.pl
+++ b/macros/math/specialTrigValues.pl
@@ -17,7 +17,8 @@ Imaginary parts separately.
 
 C<specialAngle(x)> returns a MathObject Formula in Numeric context of the form
 "a pi/c" that is the closest possible to x, where a is an integer, and c is from
-a specified set of positive integers. By default, c comes from [1,2,3,4,6].
+a specified set of positive integers. By default, c comes from [1,2,3,4,6]. If
+a equals 0, then this returns Formula('0').
 
 =head2 Options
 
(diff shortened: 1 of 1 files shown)


macros/math/specialTrigValues.pl
```diff
@@ -106,7 +107,7 @@ sub specialAngle {
 	} else {
 		$a = $closest->[0];
 	}
-	my $return = Formula("$a pi $divc");
+	my $return = ($a ne '0') ? Formula("$a pi $divc") : Formula("0");
```
somiaj, line 110: You can just use single quotes here, `Formula('0')`.

Alex-Jordan: OK, I changed that.

somiaj (approved):

macros/math/specialTrigValues.pl
Alex-Jordan: OK, single quotes in the POD now too.

somiaj: Oh, I always overlook this too, but `Formula` already treats the input as a string, so `Formula(0)` would have been sufficient. I'm unsure out of what it looks like in the code if there is any advantage here.

pstaabp (approved):

drgrice1 (approved):

drgrice1:
Three approvals. Merging.

Issue #1155, opened 2024-12-03
Currency with multiple Context()->currency->set()


Alex-Jordan:
The problem below works as it is. It comes from trying to use Kenyan currency notation, with "sh" in place of the dollar sign. However if you uncomment the commented lines to replace the preceding line, there are errors and the problem does not render. The error message indicates that something expects a dollar sign to still be the currency symbol.

```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl contextCurrency.pl));
Context("Currency");

Context()->currency->set(symbol=>'sh', comma => ':');
#Context()->currency->set(symbol=>'sh');
#Context()->currency->set(comma => ':'); 

$a = Compute("sh 1000");

ENDDOCUMENT();
```

drgrice1:
Interestingly, changing the order of the `set` calls as in the following works.

```perl
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl contextCurrency.pl));

Context("Currency");
Context()->currency->set(comma  => ':');
Context()->currency->set(symbol => 'sh');

$a = Compute('sh 1000');

ENDDOCUMENT();
```

Issue #1154, opened 2024-12-03
Currency with comma=>' '


Alex-Jordan:
If I try this problem below, the space that is declared to be the separator does not appear in the math mode output. It is there if you inspect the math, but it is `{ }` instead of `{\ }`. And you can't just use `comma => '\ '` because then in text mode you see the backslash. You cannot use `comma => $SPACE` either because apparently that comes out as `&nbsp;` and causes other problems.

```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl contextCurrency.pl));
Context("Currency")->currency->set(comma=>' ');
$m = Currency(1000.99);

BEGIN_PGML
[`[$m]`]

[$m]
END_PGML

ENDDOCUMENT();
```

Issue #1153, opened 2024-12-03
Currency with comma=>''


Alex-Jordan:
If I try this problem below, processing times out. It does not like setting the "comma" to the empty string. I think that would be a reasonable thing to do if someone were trying to just not have digit block separators. Note also that the stall only happens if something tries to make the string for `$m`.

```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl contextCurrency.pl));
Context("Currency")->currency->set(comma=>'');
$m = Currency(1000.99);

BEGIN_PGML
[$m]
END_PGML

ENDDOCUMENT();
```

PR #1156, opened 2024-12-03
Fix problems with currency->set() calls (#1153, #1154, #1155)

dpvc:
This PR fixes some problems with setting the currency `symbol` and `comma` values reported in #1153, #1154, and #1155, and also cleans up some spacing issues with TeX output.

Issue #1153 is resolved by adding an if-then around old line 535, which assumed the string was being changed by inserting the commas, which isn't the case when the comma is empty.

Issue #1154 is resolved by calling `Context::quoteTeX()` on the comma value in the `format()` method at (new) line 538.  That should have been done, but wasn't.

Issue #1155 is resolved by the addition of the line 374, which tracks what the previous symbol was, but wasn't being changed when the new symbol is created.

The remaining changes are for spacing improvements.  The `quoteTeX()` method now converts a space to `\,` rather than `\ `, which produces a more appropriate spacing.  The symbol string now only adds a space on one side or the other when it is a text string, rather than on both sides; this improves the TeX output (the spaces were trimmed from string output, but not TeX output).  A new `symbolString()` method is used for this to reduce redundant code.

### DIFF
--- a/macros/contexts/contextCurrency.pl
+++ b/macros/contexts/contextCurrency.pl
@@ -202,9 +202,10 @@ sub quoteTeX {
 	my $s      = shift;
 	my $isText = ($s =~ m/[a-z]/i);
 	$s =~ s/\\/\\backslash /g;
-	$s =~ s/([\#\$%^_&{} ])/\\$1/g;
+	$s =~ s/([\#\$%^_&{}])/\\$1/g;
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):
Thanks, these all do as described!

PR #1158, opened 2024-12-07
Fix the values option of parserPopUp.pl.

drgrice1:
The current code sets the correct choice to be the label, and it should be the value.  As such if the values are specified, none of the answers are accepted as correct because the correct label is not equal to the correct value.

Here is a minimal test case:

```perl
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl parserPopUp.pl PGcourse.pl));

$popup = DropDown([ 'first', 'second', 'third' ], 1, values => [ 'a', 'b', 'c' ]);

BEGIN_PGML
Select "second": [_]{$popup}
END_PGML

ENDDOCUMENT();
```

On the develop branch (or main) no answers will be accepted with this problem.  With this pull request the second answer will be accepted as correct.  If you delete the `values` option from the `DropDown` call, then the second answer will be accepted as correct with the develop branch (and main), and with this pull request.

### DIFF
--- a/macros/parsers/parserPopUp.pl
+++ b/macros/parsers/parserPopUp.pl
@@ -269,7 +269,7 @@ sub addLabelsValues {
 	my $self    = shift;
 	my $choices = $self->{orderedChoices};
 	my $labels  = [];
-	my $values  = $self->{values};
+	my $values  = [];
 	my $n       = $self->{n};
 
 	foreach my $i (0 .. $n - 1) {
@@ -278,7 +278,7 @@ sub addLabelsValues {
 			$values->[$i] = $choices->[$i]{ $labels->[$i] };
 		} else {
 			$labels->[$i] = $choices->[$i];
-			$values->[$i] = $choices->[$i] unless (defined($values->[$i]) && $values->[$i] ne '');
+			$values->[$i] = $self->{values}[ $self->{order}[$i] ] // $choices->[$i];
 		}
 
 	}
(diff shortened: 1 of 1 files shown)


drgrice1:
I also noticed that the values are not randomized with the labels if given as I did in the comment above.   For example, with the following code the values are assigned in order to the labels after they are randomized.

```perl
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl parserPopUp.pl PGcourse.pl));

$popup = DropDown([[ 'first', 'second', 'third' ]], 1, values => [ 'a', 'b', 'c' ]);

BEGIN_PGML
Select "second": [_]{$popup}
END_PGML

ENDDOCUMENT();
```

If you instead use
```perl
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl parserPopUp.pl PGcourse.pl));

$popup = DropDown(
    [ [ { 'first' => 'a' }, { 'second' => 'b' }, { 'third' => 'c' } ] ], 1);

BEGIN_PGML
Select "second": [_]{$popup}
END_PGML

ENDDOCUMENT();
```
then the labels and values stay together.

With `parserRadioButtons.pl` and `parserCheckBox.pl` the values are randomized with the labels using either format.

That is not fixed with this pull request obviously, but is something to look into.

drgrice1:
Okay, so now this also fixes the randomization of values with labels.

drgrice1:
One other tweak was added.  There was a typo in the `data_feedback_insert_element` (it was `data_feedback_insert_elt`), and the `span`  around the select was switch to a display inline `div`.  The feedback popover is injected into that span which result in invalid html.

somiaj (approved):
Double checked this fixes both the randomization error and selection error when values are set.

Issue #440, opened 2019-10-30
vec_dot not working


jirilebl:
In PGmatrixmacros.pl it seems to me like =cut is missing and so vec_dot doesn't work (or well, it's not working for some reason anyway).  Yeah I know it's deprecated, so probably not a huge deal, but existing problems do use it.

Alex-Jordan:
It looks like this was fixed with #1162.

PR #1162, opened 2024-12-14
Fix `vec_dot` in the PGmatrixmacros.pl so issue #440 can be closed.

drgrice1:
Not much here.  Just add the `=cut` suggested in the issue and one other issue that the Perl language server I use revealed.


somiaj (approved):
Didn't test, but the fix is obvious.

Issue #1105, opened 2024-08-29
Scaffold problems and preview answer will inform students they are correct or not.


somiaj:
When previewing answers (or as an instructor showing correct answers), the scaffold part headings are still colored and opened based on the correctness of the inputted answer, so students can get info about if an answer is correct or not without using any submits by previewing their answers.

To see this, test a scaffold problem as a student, enter in the correct answer to the first part, and click preview answer. The scaffold will turn green to state they got it correct, and the next part will open up. In essence a student could work though the whole scaffold problem only using the preview answer button, then use a single submit at the end when they get it correct.

This might not be fully a pg bug, could be with its interaction with webwork2.

drgrice1:
This is how scaffolds have always been.  This is also what #506 is all about.

somiaj:
Thanks, that had slipped my mind.

pstaabp:
#1165 fixes this.

PR #1163, opened 2024-12-14
Prevent scaffolds from revealing correct answers when feedback is not permitted (really)

drgrice1:
In #986 I believed that this was set.  But either something has changed since then, or I didn't have it right at that point.

In that pull request I claimed that when an answer preview occurs, scaffolds will open but not receive the green color if all answers within are correct.  In particular, if a preview occurs it should not reveal that answers are correct.  Unfortunately, it still is.

This fixes that by only giving the correct colors when attempt results are allowed.

### DIFF
--- a/macros/core/scaffold.pl
+++ b/macros/core/scaffold.pl
@@ -593,9 +593,8 @@ sub add_container {
 					'div',
 					class => 'accordion-header'
 						. (
-							$iscorrect && ($main::envir{showFeedback} || $main::envir{forceShowAttemptResults})
-							? ' iscorrect'
-							: ' iswrong'
+							$iscorrect && (($main::envir{showFeedback} && $main::envir{showAttemptResults})
+								|| $main::envir{forceShowAttemptResults}) ? ' iscorrect' : ' iswrong'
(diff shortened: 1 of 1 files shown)


drgrice1:
Note that as stated in #986, the main point of this is for tests.  In homework, although the correct colors will not be shown on preview, there is still an indication of correctness that the savvy student can figure out (depending on the scaffold settings of the problem).  That is that later scaffolds will open when previous parts are correct.  For tests even that will not reveal that parts are correct since all scaffolds are forced open in tests to begin with.

somiaj (approved):
Works as described.

I'll close #1105 when this is merged, as I didn't realize that the correct color wasn't suppose to be shown which is part of what that issue was about. Sure a clever student who realizes the state of the scaffolds in homework can still use preview to figure out correct answers, but this is no longer possible in tests with my testing.

I'd suggest a hotfix for this as it is small and probably worthwhile to have in 2.19 for the time.

drgrice1:
I would agree to a hot fix since this was supposed to be something new for 2.19.

Issue #1166, opened 2024-12-19
possible bug with recent contextFraction.pl changes


Alex-Jordan:
The problem below will render on 2.19, but not on develop. On develop there is an error about not finding enough test points.

I suspect it has to do with the changes to `contextFraction.pl` but I haven't been able to track it down.

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "MathObjects.pl",
  "PGML.pl",
  "contextFraction.pl",
  "parserAssignment.pl",
  "bizarroArithmetic.pl",
);

Context("Fraction")->variables->add(y => 'Real');
parser::Assignment->Allow;
Context()->operators->set(
  '*' => {class => 'bizarro::BOP::multiply', isCommand => 1},
  '/' => {class => 'bizarro::BOP::divide', isCommand => 1},
);
$f = Formula("y =  2x + 1")->reduce;

ENDDOCUMENT();
```

Alex-Jordan:
Note, this is an MWE. There is an exercise that this came from with much more going on.

dpvc:
This issue is due to mishandling the implied multiplication in the new fraction context.  I've made a PR (#1167) to resolve the issue.

pstaabp:
Fixed in #1167

PR #1167, opened 2024-12-21
Fix problem with implied multiplication in fraction context, and some typos. (#1166)

dpvc:
This PR fixes the problem raised in #1166.  This was due to mishandling of the "space" operator used for implied multiplication.  The issue resulted from setting the cached operator definition to the `string` operator, but mutating the operator to the wrong class in some instances (it should be to the class for the new operator, not the original space operator's class).

This PR checks whether to mutate to the original space class or to bless the operator to the `string` operator's class.

It also fixes a couple of typos (one in an unrelated file -- hope that is OK).

Resolves issue #1166.

### DIFF
--- a/lib/Parser/Function.pm
+++ b/lib/Parser/Function.pm
@@ -214,7 +214,7 @@ sub checkComplexOrMatrix {
 	my $self = shift;
 	my $op   = $self->{params}[0];
 	return if ($self->checkArgCount(1));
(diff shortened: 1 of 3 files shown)


Alex-Jordan (approved):
I tested and the exercises that were not working from #1166 are working now. Thanks for this!

drgrice1 (approved):

Issue #1168, opened 2025-01-07
bug with derivative operator in Fraction context


Alex-Jordan:
On `develop`, here is a MWE of something that causes an error, but it does not cause an error in 2.19:

```
DOCUMENT();

loadMacros(qw(PGstandard.pl MathObjects.pl contextFraction.pl));
Context("Fraction")->flags->set(reduceConstants=>0);
$f = Formula('1/2x')->D('x');

ENDDOCUMENT();
```

I boiled this down from a complicated exercise that was reported broken since I moved our server to develop. I can avoid the error by not setting `reduceConstants=>0`. However doing so causes some other issue (cosmetic issues, not PG errors).

Alex-Jordan:
I added `sub DESTROY {};` immediately under your patch and that at least clears up the warnings. Unsure if this was wise. At least one other `sub DESTROY` in the repo is empty like this.

dpvc:
Try changing the `AUTOLOAD` function to

``` perl
#
#  Trap calls to methods that aren't overridden in the subclass
#  and perform the super-class method with the given arguments.
#
sub AUTOLOAD {
	my $self = @_[0];
	my $name = $AUTOLOAD;
	$name =~ s/.*:://;
	my $meth = $self->super($name);
	goto &$meth if ref($meth) eq 'CODE';
}
```

and see if that is better.

somiaj:
I too can confirm that adding `sub DESTROY {}` removed the warnings.

somiaj:
@dpvc Your recent suggestion didn't stop the warnings, I still see them. In addition I get the following `Scalar value @_[0] better written as $_[0]`. Would the first line as `my ($self) = @_;` be better there?

dpvc:
OOPS, yes, `$_[0]` is correct, but your version is also OK.

I don't see how you can still be getting the message because the line that produced is has been removed.  Are you sure the correct macro file is loading?  Or are they just similar messages?  The message 

```
Undefined method DESTROY for context::Fraction::BOP::divide at line 462 of [PG]/macros/contexts/contextExtensions.pl
    from within context::Extensions::Super::AUTOLOAD called at line 86 of [PG]/lib/Parser/BOP.pm
```

was generated by the `die` command that should have been removed in the new version of `AUTOLOAD`.

dpvc:
OK, here is another try:

``` perl
sub AUTOLOAD {
	my ($self) = @_;
	my $name = $AUTOLOAD;
[...]

PR #1172, opened 2025-01-08
Add D() methods to fractions, and allow parentheses in limited context

dpvc:
This PR addresses the differentiation issue raised in #1168.  The solution presented here is to add `D()` methods to the division and space classes that return 0 (since a fraction or a mixed number is constant).  My [comment](https://github.com/openwebwork/pg/issues/1168#issuecomment-2578680976) in that discussion explains why my earlier idea of using `AUTOLOAD` is not the right one, and why this is the correct one.

This also redefines the parenthesis in the `LimitedFraction` context so that `-(1/2)` will parse from MathQuill.

I may need to make a similar PR for `contextUnits.pl` that adds differentiation there (I haven't looked into that yet).

### DIFF
--- a/macros/contexts/contextFraction.pl
+++ b/macros/contexts/contextFraction.pl
@@ -354,6 +354,7 @@ sub Init {
 		showMixedNumbers  => 1,
 	);
 	$context->{cmpDefaults}{Fraction} = { studentsMustReduceFractions => 1 };
+	$context->parens->redefine('(');
 
 	$context = $main::context{LimitedProperFraction} = $context->copy;
 	$context->flags->set(requireProperFractions => 1);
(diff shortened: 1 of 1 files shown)


dpvc:
PS, I suppose tests should be added for `-(1/2)` in `LimitedFraction` context, and for the `D()` methods.  I'm afraid I haven't looked into the test files to see how that is done.

somiaj (approved):
I checked that this fixes both issues raised with contextFraction.

pstaabp (approved):
Issue fixed.

drgrice1 (approved):
This looks good to me.  I agree that unit tests would be nice.  I have a branch with unit tests for differentiation that I have had sitting locally for over a year now.  I should put that in and add tests for this as well.

PR #1177, opened 2025-01-11
Update units context to include differentiation

dpvc:
The main purpose of this PR is to add the ability to differentiate formulas with units.  

I also made a pass over the documentation, and corrected some errors, moved the section on removing units from a context to earlier (right after adding them), and added a section on the new differentiation features.

In the past, the `D()` method provided for multiple derivatives like `D(2)` or `D(2, "x")`, but only allowed one to be specified.  I have extended the `D()` method to allow more than one numeric repeat operation, so `$f->D(2, "x", 3, "y")` would be $d\over dx^2 dy^3$, and is equivalent to `$f->D("x", "x", "y", "y", "y")` or `$f->D(2, "x")->D(3, "y")`.  This should not affect existing problems, as it is an extension of the syntax, not a change.

While working on this, I noticed that I had handled `abs()` for arguments that are numbers with units, but didn't handle the `|...|` syntax for that.  So I've added that here.  That also required adding another option to the `contextExtensions.pl` extension mechanism to override the list classes (`|...|` are handled using the list delimiters infrastructure).

Finally, I fixed an error with `abs()` with number-with-units arguments where formula comparisons would fail (the type was not being set properly), simplified the evaluation of trig functions with angle units by removing unneeded tests that are already made in the `_check()` function, and added a `quantity` method to the Formula-with-Units class to match the one in the Number-with-Units class.

### DIFF
--- a/lib/Parser/Differentiation.pm
+++ b/lib/Parser/Differentiation.pm
@@ -12,28 +12,26 @@
 #
 sub Parser::D {
 	my $self = shift;
-	my $d;
-	my @x = @_;
-	my $x;
-	if (defined($x[0]) && $x[0] =~ m/^\d+$/) {
-		$d = shift(@x);
-		$self->Error("You can only specify one variable when you give a derivative count")
-			unless scalar(@x) <= 1;
-		return ($self) if $d == 0;
-	}
-	if (scalar(@x) == 0) {
-		my @vars = keys(%{ $self->{variables} });
-		my $n    = scalar(@vars);
-		if ($n == 0) {
-			return $self->new('0') if $self->{isNumber};
-			$x = 'x';
-		} else {
-			$self->Error("You must specify a variable to differentiate by") unless $n == 1;
-			$x = $vars[0];
+	my @vars = keys %{ $self->{variables} };
+	my $X    = @vars == 1 ? $vars[0] : undef;
+	push(@_, $X) if @_ == 0 && defined $X;
+	$self->Error("You must specify a variable to differentiate by") unless @_;
+
+	my @x;
+	while (@_) {
+		my $d = 1;
+		my $x = shift;
+		($d, $x) = ($x, shift) if $x =~ m/^\d+$/;
+		if (!defined $x) {
+			$self->Error([ "You must specify a variable to differentiate by following %s", $d ])
+				unless defined $X;
+			$x = $X;
 		}
-		CORE::push(@x, $x);
+		my $def = $self->context->variables->get($x);
+		$self->Error([ "Variable of differentiation not defined: %s", $x ]) unless $def;
+		push(@x, ($x) x $d) if $d;
 	}
-	@x = ($x[0]) x $d if $d;
+
 	my $f = $self->{tree};
 	foreach $x (@x) {
 		return (0 * $self)->reduce('0*x' => 1) unless defined $self->{variables}{$x};
--- a/lib/Parser/Differentiation.pm
+++ b/lib/Parser/Differentiation.pm
@@ -14,7 +14,7 @@ sub Parser::D {
 	my $self = shift;
 	my @vars = keys %{ $self->{variables} };
 	my $X    = @vars == 1 ? $vars[0] : undef;
-	push(@_, $X) if @_ == 0 && defined $X;
+	CORE::push(@_, $X) if @_ == 0 && defined $X;
 	$self->Error("You must specify a variable to differentiate by") unless @_;
 
 	my @x;
@@ -29,7 +29,7 @@ sub Parser::D {
 		}
 		my $def = $self->context->variables->get($x);
 		$self->Error([ "Variable of differentiation not defined: %s", $x ]) unless $def;
-		push(@x, ($x) x $d) if $d;
+		CORE::push(@x, ($x) x $d) if $d;
 	}
 
 	my $f = $self->{tree};
--- a/macros/contexts/contextUnits.pl
+++ b/macros/contexts/contextUnits.pl
@@ -388,7 +388,7 @@ In order to be able to differentiate a formula that returns a number
 with units, the MathObjects library needs to know the units of the
 variable you are differentiating by.  For example, if you have
 
-    $s = Compute("(3t^2 - 2x) m");
+    $s = Compute("(3t^2 - 2t) m");
 
 as a function of time, C<t>, then you would like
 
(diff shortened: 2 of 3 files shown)


macros/contexts/contextUnits.pl
```diff
@@ -377,6 +382,43 @@ =head2 Working with numbers with units
 quantity is the absolute value of the original quantity.
 
 
+=head2 Differentiation of numbers with units
+
+In order to be able to differentiate a formula that returns a number
+with units, the MathObjects library needs to know the units of the
+variable you are differentiating by.  For example, if you have
+
+    $s = Compute("(3t^2 - 2x) m");
```
pstaabp, line 391: I think you want `$s = Compute("(3t^2 - 2t) m");`

macros/contexts/contextUnits.pl
```diff
@@ -377,6 +382,43 @@ =head2 Working with numbers with units
...
+
+You can pass as many unit assignments to a single C<assignUnits()>
+call as you like.  E.g.
+
+    Context("Units")
+      ->withUnitsFor("length", "time")
+      ->assignUnits(
+        t => "s",
+        s => "m"
+      );
```
pstaabp, line 415: Also, with this example, I think adding `->variables->are(t => Real, s=> Real);` will help.

pstaabp: Never mind. I guess the `assignUnits` will defined the variables.

pstaabp:
Overall, this looks great.  I was testing the higher-order derivatives with the non-realistic function 

```
Context('Units')->withUnitsFor('length','mass')->assignUnits(
        x => 'm',
        y => "ft",
    z => 'kg'
    );

$z = Compute("(3x^2 * y^4) kg");

$dz = $z->D('x',3,'y');
```

and the result was 6x*24y kg/(m ft^3) as expected.  I was thinking that using `->reduce` would change this to 144 x*y kg/(m ft^3), but left it unchanged.  Is this because the reduction rules don't automatically apply to these objects?   Not a big deal, but just curious.

Alex-Jordan:
@pstaabp Assuming implicit multiplication has higher precedence than explicit, I think that `6x * 24y` would be parsed as `(6x)*(24y)`. 

There is a relatively new reduction rule:  `m(nx) -> (mn)x` for when `m` and `n` are numbers. There is also a reduction rule for `x*n -> n*x`. But neither of those apply here so I think there just is no reduction rule for this.

Note that if implicit and explicit multiplication had the same precedence, then maybe you have:
`((6x)24)y`
and then by `x*n -> n*x`:
`(24(6x))y`
and then by `m(nx) -> (mn)x`:
`(144x)y`

dpvc:
> Assuming implicit multiplication has higher precedence than explicit

It doesn't.  In fact, the MathQuill settings have implicit multiplication as **lower** precedence than explicit multiplication, if I recall correctly.

The actual result from the differentiation is `[(6x)*(24y)] kg`, and there is no reduction rule for `(nx)*(mx)` so it is not further reduced.

Also note that differentiation does a `reduce` automatically, so adding it afterword will never do anything.

> Is this because the reduction rules don't automatically apply to these objects? 

You are right, the Number-with-Units object does not have a `reduce()` method, and it probably should (one that reduces its formula).  I will add one and make a commit for that (plus the documentation fixes that you spotted -- thanks).

drgrice1:
Note that MathQuill precedence is only used when parsing student answers.

lib/Parser/Differentiation.pm
```diff
@@ -12,28 +12,26 @@
...
-		my $n    = scalar(@vars);
-		if ($n == 0) {
-			return $self->new('0') if $self->{isNumber};
-			$x = 'x';
-		} else {
-			$self->Error("You must specify a variable to differentiate by") unless $n == 1;
-			$x = $vars[0];
+	my @vars = keys %{ $self->{variables} };
+	my $X    = @vars == 1 ? $vars[0] : undef;
+	push(@_, $X) if @_ == 0 && defined $X;
```
drgrice1, line 17: I think that this should be `CORE::push(@_, $X) if @_ == 0 && defined $X;`, and line 32 should use `CORE::push` as well.

I am seeing the warnings `Ambiguous call resolved as CORE::push(), qualify as such or use &` from these lines  when running the unit tests.  Perl wants it clarified that this is not the `Parser::push` method.

dpvc:
> Note that MathQuill precedence is only used when parsing student answers.

I seem to recall a discussion where we considered making this the precedence all the time.  An one could certainly take derivatives of student answers, although I know that isn't what Peter was doing.

In any case, I was only trying to say that Alex's assumption about the precedences wasn't true, and that it is more likely to be the other way around, if anything.

lib/Parser/Differentiation.pm
dpvc: I will fix it.

dpvc:
> You are right, the Number-with-Units object does not have a `reduce()` method, and it probably should

It turns out that reducing formulas that return number-with-unit objects already works fine, and that the actual NumberWithUnit object doesn't need a `reduce()` function after all, since it is an actual number (not a formula) with units.  The formula returning a number-with-units is implemented as a formula joined to a unit via the unit "multiplication" operation, and that multiplication inherits the usual reduction method, so there is no need to do anything special for that case.

drgrice1:
I know, I was just pointing out that the MathQuill precedence wouldn't have had effect in @pstaabp's calls.

macros/contexts/contextUnits.pl
dpvc: Change made.  Thanks!

macros/contexts/contextUnits.pl
dpvc: I've added a sentence to the documentation indicating that that's what happens.

dpvc:
> The new units context and the context extensions stuff in general is really awesome. So much new potential.

Thanks!  I had wanted to do a real units object since the very beginning (the old `NumberWIthUnits()` was a stop-gap measure that was never intended to be permanent), and it seemed like units should be able to be added to any (numeric) context, so I had to come up with a way to do that.  Thus the extension framework seemed a natural thing to do.

I agree, that this opens up a lot of new possibilities.

I've made the POD change you requested.  Thanks!

pstaabp (approved):
Looks good.  Wondering if we'd want to implement any of the other reduction rules that @Alex-Jordan mentioned above.  Not to this PR, but in the future.

dpvc:
> Wondering if we'd want to implement any of the other reduction rules that @Alex-Jordan mentioned above

They are already implemented as part of the standard Numeric context, on which the standard Units context is based, so you get them for free.  But it might be good to add a `(n*x)(m*y) => (nm)*(x*y)` rule to the Numeric rules.  There are lots of rules that could be added.

Issue #1178, opened 2025-01-15
links not working within hint or solution


Alex-Jordan:
If a link is inside a hint or solution, when I click on it nothing happens. This is perhaps some bug with the summary/details HTML elements being used now, but I have not investigated. Here is a sample problem.

```
DOCUMENT();

loadMacros(qw(PGstandard.pl PGML.pl));

BEGIN_PGML
[@ htmlLink('https://openwebwork.org/', "This link works.") @]*
END_PGML

BEGIN_PGML_HINT
[@ htmlLink('https://openwebwork.org/', "This link does not work.") @]*
END_PGML_HINT

BEGIN_PGML_SOLUTION
[@ htmlLink('https://openwebwork.org/', "This link does not work.") @]*
END_PGML_SOLUTION

ENDDOCUMENT();

```

somiaj:
I can confirm this with 2.19 as well. Testing it a bit, it appears the click event on the accordion covers the body as well, so the event is not passing through to the link.

drgrice1:
Odd.  I am not sure what is blocking the link.  Note that middle clicking on the link does work, and the link opens in a new tab or window.  Of course that is a linux thing, and may not work elsewhere.

somiaj:
I was able to fix the issue by limiting when the default action of an event was prevented. Probably a better way to go about this, but this fixes the issue in my testing. The default action needs to be prevented to cause the container to toggle, but that should only happen when clicking on the accordion button or header. At least that is what my limited understanding was able to figure out.

```diff
diff --git a/htdocs/js/Problem/details-accordion.js b/htdocs/js/Problem/details-accordion.js
index 621de1e95..7a1b8653a 100644
--- a/htdocs/js/Problem/details-accordion.js
+++ b/htdocs/js/Problem/details-accordion.js
@@ -8,7 +8,10 @@
                const collapse = new bootstrap.Collapse(collapseEl, { toggle: false });
                button.addEventListener('click', () => collapse.toggle());
 
-               details.addEventListener('click', (e) => e.preventDefault());
+               details.addEventListener('click', (e) => {
+                       if (e.target.matches('[class*="accordion"]'))
+                               e.preventDefault();
+               });
                collapseEl.addEventListener('show.bs.collapse', () => {
                        details.open = true;
                        button.classList.remove('collapsed');
```

PR #1179, opened 2025-01-15
Prevent the `click` event on a solution or hint accordion button instead of the details.

drgrice1:
<s>Only prevent the default click behavior when the user clicks on the summary or something within the summar. Allow the click event default to occur otherwise.</s>

Instead add `e.preventDefault()` to the `summary.accordion-button` `click` event listener instead of preventing all clicks on the `details` tag.

This fixes issue #1178.

This could be made into a hotfix if desired.


somiaj (approved):
Confirmed this fixes the issue. A hot fix sounds reasonable.

Alex-Jordan (approved):

PR #1183, opened 2025-01-22
Add the listAnswer.pl macro.

drgrice1:
This macro was originally omitted because it is not directly used by any problems in the OPL or Contrib.  However, the CofIdaho_macros.pl macro loads it if the `Picky_equation_cmp` method is used.  So any problem that used that macro and method don't work right.  Apparently several of the problems in the `Library/CollegeOfIdaho/setAlgebra_01_05_ProblemSolvingAndFormulas` use this.

See https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8632#p21807.

### DIFF
--- a/macros/deprecated/listAnswer.pl
+++ b/macros/deprecated/listAnswer.pl
@@ -1,0 +1,613 @@
+sub _listAnswer_init { };    # don't reload this file
+
+######################################################################
+#
+#  This file implements a general list answer checker, which allows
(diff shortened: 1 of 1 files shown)


drgrice1:
I am going to close this issue.  I just pulled the latest changes to the OPL, and I see that all of those problems have been rewritten to not use the `CofIdaho_macros.pl` macro.

PR #1185, opened 2025-01-29
Fix coordinate hints in the GraphTool for restricted sine wave points.

drgrice1:
After construction of a sine wave is completed, if the point that determines amplitude is selected and moved with the mouse, then the x-coordinate of the point can't change.  However, the x-coordinate of the coordinate hints in the lower right corner of the board does change if the mouse is moved to the left and right while dragging that point, and should not be.

The same thing happens with the point that determines period on relative to the vertical axis.  The y-coordinate of that point can't change, but the y-coordinate of the coordinate hints do change if the mouse is moved up or down.

This pull request fixes the issue by overriding the default graph object `updateTextCoords` method for the sine wave graph object.  A flag set on the point when it is dragged, and the `updateTextCoords` override checks for the flag and sets the coordinate hints with the coordinates of the point instead of using the mouse cursor coordinates which might be incorrect.

This is similar to the issue that @Alex-Jordan noticed in #1157, although that was during the construction phase of the sine wave.

To test you can use the attached problem:  [GraphSineWavePi.pg.txt](https://github.com/user-attachments/files/18582191/GraphSineWavePi.pg.txt)

In that problem, graph a sine wave by plotting all three points.  Then click on either the amplitude or period point (the second or third point graphed during the construction phase), and try to drag the point in the direction it can't move.  With the develop branch you will see the coordinate change in the direction the point can't go.  With this pull request that coordinate stays fixed.

### DIFF
--- a/htdocs/js/GraphTool/cubictool.js
+++ b/htdocs/js/GraphTool/cubictool.js
@@ -8,8 +8,16 @@
 			preInit(gt, point1, point2, point3, point4, solid) {
 				[point1, point2, point3, point4].forEach((point) => {
 					point.setAttribute(gt.definingPointAttributes);
-					point.on('down', () => (gt.board.containerObj.style.cursor = 'none'));
-					point.on('up', () => (gt.board.containerObj.style.cursor = 'auto'));
+					if (!gt.isStatic) {
+						point.on('down', () => {
+							point.dragging = true;
+							gt.board.containerObj.style.cursor = 'none';
+						});
+						point.on('up', () => {
+							delete point.dragging;
+							gt.board.containerObj.style.cursor = 'auto';
+						});
+					}
 				});
 				return gt.graphObjectTypes.cubic.createCubic(point1, point2, point3, point4, solid, gt.color.curve);
 			},
@@ -163,6 +171,7 @@
 
 				groupedPointDrag(gt, e) {
 					gt.graphObjectTypes.cubic.adjustDragPosition(e, this, this.grouped_points);
+					gt.setTextCoords(this.X(), this.Y());
 					gt.updateObjects();
 					gt.updateText();
 				},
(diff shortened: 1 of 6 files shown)


drgrice1:
I have converted this to a draft pull request because I found that this issue is not new with the sine wave tool.  This affects all graph objects that have restricted point drags.  I will fix that issue more generally in this pull request.

drgrice1:
This is ready now.

Basically the full issue is that any time that a point's drag is restricted the coordinate hints can be wrong. This happens if the drag is in a range that the point can't be moved to and can happen for both mouse or keyboard drags.
    
For an extreme example where this is easy to see, with the 4-point cubic tool graph the points (-1, 0), (0, 0), (1, 0), and (2, 0).  Then click on the point at (2, 0) and slowly drag it to the left. Without this pull request you will see the coordinate hints show (1, 0), then (0, 0), etc., even though that is not where the point being dragged is. 

Or graph the same points for a cubic as above, then click on the fourth point graphed (or use `Tab` to keyboard navigate to that point) after completion of graphing the cubic.  Then use the left arrow on the keyboard to move the point left.  It will move to the left of the other three points, but the coordinates shown will be incorrect without this pull request.  If you then use the right arrow to move it back to the other side of the other three points the coordinates will change, but will still be incorrect without this pull request.

drgrice1:
Also, these incorrect coordinates can happen during the construction phase as well (again without this pull request).

drgrice1:
By the way, here is a general problem with all of the tools for testing: 
[Demo.pg.txt](https://github.com/user-attachments/files/18592005/Demo.pg.txt)

Alex-Jordan (approved):
Tested and works as described.

pstaabp (approved):
Looks like it fixes the issue.

Issue #1096, opened 2024-08-16
Feature request, show green checkmarks in feedback when partial feedback is hidden but answers are all correct.


somiaj:
I have a problem with `$showPartialCorrectAnswers=0` which contains four drop down menus. I expected to see green checkmarks if everything was correct, but instead these were also hidden. Here is the result, with the submit earning 100%.

![image](https://github.com/user-attachments/assets/e8ee4b72-271a-48d8-bb7f-b7d90be1b884)

somiaj:
Here is what I came up with to show all green checkmarks when all parts are correct, but to keep the feedback hidden when `$showPartialCorrectAnswers` is false.

```diff
diff --git a/macros/PG.pl b/macros/PG.pl
index e3f4a802..dfef13e4 100644
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -1035,6 +1035,13 @@ sub ENDDOCUMENT {
                                && !$rh_envir->{showAttemptPreviews}
                                && !$rh_envir->{showMessages};
 
+                       # Compute $numCorrect here, since we need to know if everything is correct when deciding
+                       # to display correct feedback when $showPartialCorrectAnswers is false.
+                       for my $answerLabel (@answerNames) {
+                               my $answerScore = $PG->{PG_ANSWERS_HASH}{$answerLabel}{ans_eval}{rh_ans}{score} // 0;
+                               ++$numCorrect if $answerScore >= 1;
+                       }
+
                        for my $answerLabel (@answerNames) {
                                my $response_obj = $PG->{PG_ANSWERS_HASH}{$answerLabel}->response_obj;
                                my $ansHash      = $PG->{PG_ANSWERS_HASH}{$answerLabel}{ans_eval}{rh_ans};
@@ -1080,8 +1087,13 @@ sub ENDDOCUMENT {
                                        push(@{ $options{feedbackElements} }, @$elements);
                                }
 
-                               if (($rh_envir->{showAttemptResults} && $PG->{flags}{showPartialCorrectAnswers})
-                                       || $rh_envir->{forceShowAttemptResults})
+                               if (
+                                       (
+                                               $rh_envir->{showAttemptResults}
+                                               && ($PG->{flags}{showPartialCorrectAnswers} || $numCorrect == @answerNames)
+                                       )
+                                       || $rh_envir->{forceShowAttemptResults}
+                                       )
                                {
                                        if ($showCorrectOnly) {
                                                $options{resultClass} = 'correct-only';
[...]

PR #1187, opened 2025-01-31
If `$showPartialCorrectAnswers = 0` and all answers are correct, then show all correct.

drgrice1:
This makes it so that if `$showPartialCorrectAnswers = 0` and all answers are correct, then the feedback buttons are all green revealing that all answers are correct.

This works by passing the problem state found when evaluating answers into the post content processor methods (it is a new third parameter passed to the callback). The ENDDOCUMENT post processor then uses this to quickly determine that the score for the problem is 1 (or more), and in that case shows the usual feedback showing that all answers are correct.

This avoids looping multiple times to determine correctness of the problem.

This is as discussed in https://github.com/openwebwork/pg/issues/1096.  This is a change in behavior, so may need further discussion and a consensus.

### DIFF
--- a/lib/WeBWorK/PG/Translator.pm
+++ b/lib/WeBWorK/PG/Translator.pm
@@ -1174,9 +1174,12 @@ sub post_process_content {
 		$self->{safe}->share_from('main', [qw(%Mojo::Base:: %Mojo::Collection:: %Mojo::DOM::)]);
 		our $problemDOM =
 			Mojo::DOM->new->xml($self->{rh_pgcore}{displayMode} eq 'PTX')->parse(${ $self->{PG_PROBLEM_TEXT_REF} });
-		our $pageHeader = Mojo::DOM->new(${ $self->{PG_HEADER_TEXT_REF} });
-		$self->{safe}->share('$problemDOM', '$pageHeader');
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
At the same time you were doing this, I made PTX js handle it too. (That won't be a problem, as what I did is for use with 2.19 and this here will be for 2.20).

But in addition to making green buttons when all is correct, I also made things happen when `$showPartialCorrectAnswers = 0` and the answers are not all correct. I changed the card title from 'Answer Preview' to 'One or more answers incorrect'. And set the aria-label to 'One or more answers incorrect' as well. Is there something like that that can be done here? Of course different wording may be better.

Alex-Jordan (approved):

drgrice1:
> But in addition to making green buttons when all is correct, I also made things happen when `$showPartialCorrectAnswers = 0` and the answers are not all correct. I changed the card title from 'Answer Preview' to 'One or more answers incorrect'. And set the aria-label to 'One or more answers incorrect' as well. Is there something like that that can be done here? Of course different wording may be better.

There are two reasons not to do this.

First, webwork doesn't need this because webwork shows the summary that has this information and is better.

Second, in my opinion that is confusing for the user, and not enlightening at all.  That means that the feedback that is meant for a single answer (or answer group) is now feedback for the problem as a whole.

PR #1188, opened 2025-02-04
correct some POD

Alex-Jordan:
This is phase 1 of my attempt to replace #1076. It seemed to me that some of the POD for `lib/Value/Matrix.pm` is just incorrect or unclear. So this tries to fix that.

I started thinking of MathObject Matrices as n-dimensional tensors, to help with some confusion when is not a "normal" 2D matrix. I'm using "tensor" some in the POD. There are inconsistencies with how things work when you have a 1D tensor versus nD with n > 1. The inconsistencies would be resolved if 1D tensors were presented as column vectors instead of row vectors. But I suppose we can't make a change like that for backward compatibility reasons, for problems out there that use 1D matrices. With the inconsistencies kept, a lot of the POD is going to need to be like "it works this way for a 1D tensor/matrix but it works this other way for higher D".

Phase 2 will be to rearrange/clean up the POD in the manner of #1076. But that should be easier to review as it will just be moving things around with perhaps some minor rewording.

Phase 3 will be to enhance some things in the existing methods. Specifically I think that `row()` and `column()` should be allowed to take multiple index arguments to produce submatrices (e.g. the submatrix of the 1st and 3rd rows).

Phase 4 will be to add `@pstaabp's` new methods.

Somewhere in there, also add the tests. Maybe some with Phase 3 and some with Phase 4.

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -52,25 +52,43 @@ Contexts
 
 Creation methods
 
-		 $M1 = Matrix([1,2],[3,4]);
-   		 $M2 = Matrix([5,6],[7,8]);
-    	  $v = Vector(9,10);
-    	  $w = ColumnVector(9,10); # differs in how it is printed
+	$M1 = Matrix(1, 2, 3);    # 1D tensor (row vector)
+	$M1 = Matrix([1, 2, 3]);
+
+	$M2 = Matrix([1, 2], [3, 4]);    # 2D tensor (matrix)
+	$M2 = Matrix([[1, 2], [3, 4]]);
+
+	$M3 = Matrix([[1, 2], [3, 4]], [[5, 6], [7, 8]]);    # 3D tensor
+	$M3 = Matrix([[[1, 2], [3, 4]], [[5, 6], [7, 8]]]);
+
+	$M4 = ...
 
 Commands added in Value::matrix
 
 	Conversion
-		$matrix->values produces [[3,4,5],[1,3,4]] recursive array references of numbers (not MathObjects)
-		$matrix->wwMatrix   produces CPAN MatrixReal1 matrix, used for computation subroutines
+		$matrix->value produces an array of numbers (for a 1D tensor) or array refs representing the rows.
+		These are perl numbers and array refs, not MathObjects.
+		$M1->value is (1, 2, 3)
+		$M2->value is ([1, 2], [3, 4])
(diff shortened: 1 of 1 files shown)


pstaabp (approved):

drgrice1 (approved):
This looks fine to me.

dpvc:
> The inconsistencies would be resolved if 1D tensors were presented as column vectors instead of row vectors.

Do you mean that you would like `Matrix(1, 2, 3)` to produce what `Matrix([1], [2], [3])` currently does as opposed to what `Matrix([1, 2, 3])` does?  This would actually be a 2D matrix (3 x 1) rather than a what is essentially a row vector.

Or do you mean that 1D matrices should be treated more like `ColumnVector(1, 2, 3)` (in terms of multiplication, string and TeX output, etc.)?

Would `Matrix(1, 2, 3)` be different from `Matrix([1, 2, 3])`?  

Would `Matrix(1, 2, 3)->row(1)` return just `1` and `Matrix(1, 2, 3)->column(1)` return `[1, 2, 3]`?

The original idea was that you passed `Matrix()` the rows, so `Matrix([1, 2, 3])` should be what amounts to a 1 x 3 matrix, but I could see how that would suggest that `Matrix(1, 2, 3)` should be a 3 x 1 matrix.

Alex-Jordan:
Sorry, I messed up at least one thing typing that and not being careful.

> Of course with that change, to make something that looks like a row vector would need Matrix([1], [2], [3]). But that is the complication you currently need for something that looks like a column vector.

Should have saud:

Of course with that change, to make something that looks like a row vector would need `Matrix([1, 2, 3])` instead of the simple `Matrix(1, 2, 3)`. But right now to get something that looks like a column vector you need the complicated `Matrix([1], [2], [3])` and while that would still work, you could instead just do `Matrix(1, 2, 3)`.

Alex-Jordan:
This has two approvals. If it can be merged, I will proceed with phase 2. That's just rearranging the POD to be like @pstaabp's in #1076.

I'll pause after that for discussion. But I would like to review the existing methods and possibly enhance them. For example `$M->row([2,3])` or `$M->row(2,3)` could return the submatrix of `$M` built from the 2nd and 3rd rows.

Also if @dpvc doesn't think my idea about 1D matrices would be a disaster for existing exercises, maybe I would do that as well.

And only after existing methods have all been reviewed/revised, I'd return to the new methods that @pstaabp is adding in #1076.

drgrice1:
I will go ahead and merge this, so you can proceed with the next phase.

PR #1191, opened 2025-02-08
Add triangle and quadrilateral tools to the graph tool.

drgrice1:
These were requested by Robin Cruz in https://webwork.maa.org/moodle/mod/forum/discuss.php?d=8646.

The usage of the new tools is documented in the parserGraphTool.pl macro.

Example problems for testing this are attached.

[Quadrilateral.pg.txt](https://github.com/user-attachments/files/18722536/Quadrilateral.pg.txt)
[Triangle.pg.txt](https://github.com/user-attachments/files/18722537/Triangle.pg.txt)

### DIFF
--- a/htdocs/js/GraphTool/graphtool.scss
+++ b/htdocs/js/GraphTool/graphtool.scss
@@ -235,6 +235,14 @@
 			&.gt-sine-wave-tool {
 				background-image: url('images/SineWaveTool.svg');
 			}
+
+			&.gt-triangle-tool {
+				background-image: url('images/Triangle.svg');
+			}
+
+			&.gt-quadrilateral-tool {
+				background-image: url('images/Quadrilateral.svg');
+			}
 		}
 	}
 
--- a/htdocs/js/GraphTool/images/Quadrilateral.svg
+++ b/htdocs/js/GraphTool/images/Quadrilateral.svg
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="32px" height="32px" viewBox="0 0 32 32" version="1.1" id="SVGRoot" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
+    <g id="layer1">
+        <path style="fill:none;stroke:black;stroke-width:1.38113px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1" d="M 5.6,11 26,6.1 24,21 12,25 Z" id="path1" />
+    </g>
+</svg>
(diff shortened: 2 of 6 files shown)


pstaabp:
The triangle tool worked well.

When I used your sample problem with many different seeds (248, 6775), the correct answer wasn't accepted.  Those both had internal crossings, so I thought that might be an issue, but I even hardcoded a simple square and didn't work.

drgrice1:
There was a mistake that I fixed recently that made it so the quadrilateral `cmp` method wasn't even being called.  It should work now.

pstaabp (approved):
This works now.  Looks good.

Alex-Jordan:
What should happen if you use the quadrilateral tool to make a "bad" quadrilateral like in the picture, and then fill? The current behavior is to not respect where the edges cross. Just checking that is what we'd want.

Screenshot 2025-02-25 at 1 52 25 PM

Options:

1. The way it works now is what we want.
2. We'd want it to only fill one of the triangles.
3. It won't let you make a quadrilateral like this because it always plots the convex quadrilateral, no matter the point sequence.

I am not able to think through which option it should be...

Alex-Jordan:
I notice the tool won't let me make a degenerate quadrilateral (like with three collinear points). This is also something I am wondering about. Could there be an exercise where in the abstract you are asking for a quadrilateral (so you provide this tool) but the answer could be a degenerate one?

Is the question of allowing degenerate quadrilaterals a separate question from allowing "hourglass" quadrilaterals like in my previous post?

Alex-Jordan:
> It won't let you make a quadrilateral like this because it always plots the convex quadrilateral, no matter the point sequence.

I realized that is not an option. I had to remember about non-convex quadrilaterals where specifying four points does not uniquely define the quadrilateral. Like three points in an equilateral triangle with a fourth point at their center.

drgrice1:
That is how it pretty much how the fill has to work for now.  In order to do better it requires rewriting the general graphtool fill code.

Currently there is an assumption that objects are two sided.  Basically there are three states a point can satisfy.  Either the point is on one side of the object, the other side of the object, or on the object.  Basically, the `fillCmp` method is designed to be like Perl's `cmp` or `<=>` operators.  It returns -1 for all points on one side of an object, 0 for points on the object, and 1 for all points on the other side of the object.  The original design of the fill tool was that it is an inequality fill.  It fills the region that satisfies a given inequality (`> 0` or `< 0`) for the equation defining the object, and there was an assumption that the things graphed were curves.

Of course quadrilaterals are not two sided.  They can have intersection points which result in three regions as in your example.  Furthermore, the inequality fill concept doesn't work with this implementation.  In order for the inequality approach to work, something more is needed than the current `cmp` type approach.  See the comment on line 68 of `htdocs/js/GraphTool/quadrilateral.js`.

As to degeneracy, I went with the design of the existing tools to not allow degeneracy.  For example. the line tool does not allow graphing the same point twice.  The triangle tool added here also does not allow three collinear points.  Note that the "hourglass" quadrilateral is a valid quadrilateral.  It is not degenerate, and so it is allowed.  Three consecutive collinear vertices are not allowed in a polygon.  In general, a vertex is not allowed to be on the line segment between any two other vertices in a polygon.  However, two edges of a polygon are allowed to intersect (just not at a vertex).  Note that JSXGraph does allow polygons to be degenerate, and this could be changed to allow degeneracy.  The code that prevents degeneracy is in the `adjustDragPosition` method in `htdocs/js/GraphTool/quadrilateral.js`.

Issue #1193, opened 2025-02-12
Alias units in contextUnits.pl


pstaabp:
Writing some new problems using the new context Units and have a possible bug with plurals (aliases).  Here's an MWE:

```
DOCUMENT();
loadMacros('PGstandard.pl', 'PGML.pl', 'contextUnits.pl');

Context('Units')->withUnitsFor('volume');

$vol = Compute("100 ml");

BEGIN_PGML
Convert [$vol] to cups. [_]{$vol->toUnits('cups')->with(sameUnits =>1)}
END_PGML

ENDDOCUMENT();
```
Entering the correct answer "0.42267 cups" gives the right answer, but "0.42267 cup" does not give credit for the units.  Is this intended?

dpvc:
I've made a PR that should address the issue for you.  Aliased units were not considered the same for `sameUnits => 1`, as you have pointed out, but they should be, and the PR should make that happen.

Thanks for the careful testing!

pstaabp:
fixed in #1207

PR #1194, opened 2025-02-12
Mark units that are aliases, and use alias string for uString comparisons so sameUnits will allow plurals, etc. (#1193)

dpvc:
This PR addresses the issue raised in #1193, where aliases for units are not accepted when `sameUnits => 1` is specified.

Because the `lib/Units.pm` file removes the `aliases` array from the `%Units::known_units` hash, we can't currently determine what is an alias and what isn't, so this PR adds new hashes to `Units.pm` that the units context can use to identify aliases.  One maps the original to the array of aliases for it, and one maps from aliases to the original that they are aliases for.  The former isn't used in the units context, so can be removed if you want.

The changes to the units context itself are to include a new `ustring` property for units that are aliases of another unit, and changes to the `uString()` method to cause it to use the `ustring` value, if present when constructing the unit string.  This is only used internally to compare the units for a correct and student answer, so this doesn't affect the string or TeX value that is produced, but does normalize the unit names for aliases so that any alias will match.

### DIFF
--- a/macros/contexts/contextUnits.pl
+++ b/macros/contexts/contextUnits.pl
@@ -685,7 +685,11 @@ our @ISA = ('Parser::Context');
 #  The units from the original Units package
 #
 our %UNITS = (%Units::known_units);
-$UNITS{$_} = $UNITS{L} for ('litre', 'litres', 'litre', 'litres');    # add these extras
+our %ALIAS = (%Units::unit_aliased_to);
+for ('litre', 'litres', 'liter', 'liters') {
+	$UNITS{$_} = $UNITS{L};
+	$ALIAS{$_} = 'L';
+}
 
 #
 #  The categories of units that can be selected.
@@ -782,6 +786,7 @@ sub addUnit {
 				isConstant => 1
 			}
 		);
+		$constants->set($name => { ustring => $ALIAS{$name} })      if $ALIAS{$name};
 		$constants->add(map { $_ => { alias => $name } } @$aliases) if $aliases;
 		$self->addUnitAliases($name) unless $options{noaliases};
 	} else {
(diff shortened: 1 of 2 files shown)


pstaabp (approved):
Looks good and fixes the error.

macros/contexts/contextUnits.pl
```diff
@@ -685,7 +685,11 @@ package context::Units::Context;
 #  The units from the original Units package
 #
 our %UNITS = (%Units::known_units);
-$UNITS{$_} = $UNITS{L} for ('litre', 'litres', 'litre', 'litres');    # add these extras
+our %ALIAS = (%Units::unit_aliased_to);
+for ('litre', 'litres', 'liter', 'liters') {
+	$UNITS{$_} = $UNITS{L};
+	$ALIAS{$_} = 'L';
+}
```
drgrice1, line 692: Why are these extras added here instead of to the `Units.pm` file?  I am curious as to why this is done differently for liters than for all other units that use the `aliases` key in `Units.pm`.

drgrice1 (commented):
This looks fine.

It seems that there is more work to do on adding aliases though, particularly for plural and full word forms.  I noticed that things like "gram" and "grams" are not accepted, and there are lots of other units like that.  Other forms have been added for some units, but not for many others.

drgrice1 (approved):

macros/contexts/contextUnits.pl
drgrice1: I see.

I didn't mean to say that you should handle the work on aliases in this pull request.  I just wanted to point that out, and that we need to work on it.

Interestingly, I noticed that I get somewhat different behavior from trying to use `grams` with the `parserNumberWithUnits.pl` versus with the new `contextUnits.pl` macro.  With the `parserNumberWithUnits.pl` macro the message is `'gra' is not defined in this context`, while with the `contextUnits.pl` macro the message is `'rams' is not defined in this context`.  The legacy macro sees `ms` as a unit and has a problem with `gra`, but the new macro sees `g` as a unit and has a problem with `rams`.

dpvc: Yes, the two worked very differently about how they determined the units.  The old units context tried to remove units from the end of the student answer and parse the rest as a formula, so it was essentially right-to-left handling of the units.  But the new contest integrates units into the parsing as actual MathObjects, so it is done left-to-right, as with all the parsing.  When there is a run of characters like "grams", MathObjects will take the maximal substring that matches one of its defined values (constants, functions, etc.), so in this case, since "gr", "gra", "gram", and "grams" are not defined in `Units.pm`, the `g` is the longest match.

The units context currently defines `squared` and `square` so that you can do `square meters` or `seconds squared`.  I suppose one could add `milli`, `centi`, `deci`, etc., so that you could use `millimeters`, `decilitres` and so on without having to add all the separate aliases.

PR #1196, opened 2025-02-17
Remove obsolete display modes.

drgrice1:
The remaining display modes are HTML_MathJax, HTML_dpng, HTML, TeX, PTX, and HTML_tth (see below).  These correspond to MathJax, images, plainText, tex, PTX, and nothing (HTML_tth is only a fallback mode). This mapping is in lib/WeBWorK/PG.pm.

HTML_tth is left for now due to the issue with problems incorrectly using some PGbasicmacros variables outside of math mode as document at https://webwork.maa.org/wiki/Installation_Manual_for_2.19_on_Ubuntu#PGbasicmacros.pl.

I have had this sitting around long enough.  I am putting it in.


drgrice1:
I found them.  They had incorrect case which is why I missed them before.  Note there is one case where they have to stay in the comments for the obsolete `M3` method in `PGbasicmacros.pl` and backwards compatibility reasons.

pstaabp (approved):

Alex-Jordan (approved):

PR #1197, opened 2025-02-17
more POD cleanup for Value/Matrices.pm

Alex-Jordan:
This continues #1188, with POD changes (and minor code changes like `foreach` to `for`) that are largely the same as in #1076. For that reason I credited the commit to @pstaabp (hope that's OK). The diff between this branch and the #1076 branch should be a bit smaller than between #1076 and develop, only showing the new additions in #1076 plus some consequences of small things cleaned up here or in #1188.

To review this, you should just confirm that it is moving or lightly editing POD. And that any code changes are trivial/minor.

The next phase for this would be enhancing or changing existing methods, if we want to do any of that. For example enhancements like letting the `row` method take multiple arguments. And changes like treating 1D Matrix objects as column vectors instead of row vectors (regarding their presentation and interactions with 2D Matrix objects).

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -6,146 +6,182 @@
 
 =head1 Value::Matrix class
 
+This is the Math Object code for a Matrix.
 
-References:
+=head2 References:
 
-MathObject Matrix methods: L<http://webwork.maa.org/wiki/Matrix_(MathObject_Class)>
-MathObject Contexts: L<http://webwork.maa.org/wiki/Common_Contexts>
-CPAN RealMatrix docs: L<http://search.cpan.org/~leto/Math-MatrixReal-2.09/lib/Math/MatrixReal.pm>
+=over
 
-Allowing Matrices in Fractions:
-L<http://webwork.maa.org/moodle/mod/forum/discuss.php?d=2978>
+=item L<MathObject Matrix methods|http://webwork.maa.org/wiki/Matrix_(MathObject_Class)>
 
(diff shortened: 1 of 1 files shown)


pstaabp (approved):

somiaj (approved):
Skimmed over the changes, and overall looks good.

Should the links be updated to point to the new location for the wiki?

Alex-Jordan:
> Should the links be updated to point to the new location for the wiki?

I'm not caught up on this. Where is the new location? There are local relative links here that just will go to a different POD page. And there are links here to the forums at webwork.maa.org/moodle/mod/forum. Which kind of thing do you mean?

somiaj:
It is the `webwork.maa.org` being moved to `openwebwork.org`. @dlglin knows the details on when this move will be completed, but the location of the wiki I think is now here <https://wiki.openwebwork.org/wiki/WeBWorK_Main_Page>

drgrice1:
Don't change the links at this point.  The actual details of the move are not certain yet.

dlglin:
The new wiki location is not in production yet.  It will live at the address you quoted, but right now that has an old copy of the wiki.  Until we officially migrate please continue to make any wiki updates on the webwork.maa.org site.

The intention is to have webwork.maa.org point to the new server, so for the time being existing links shouldn't break.  We'll set up redirects to try to get people to use the new URLs once they're ready to go.

Alex-Jordan:
Two approvals now, merge this? Phase 3 will be to review the existing methods and look for ways to improve them (either more efficient or more flexible).

PR #1199, opened 2025-02-20
macros to show a text file and a problem file alongside its code

Alex-Jordan:
With this, have a problem file like

```
DOCUMENT();
loadMacros("PGbasicmacros.pl");
problem_with_code('newProblem.pg');
ENDDOCUMENT();
```

That is, assuming `newProblem.pg` is a problem file in your templates folder. You will see the problem file's text, then a horizontal rule, then the problem rendered.

I'm unsure if this would be useful, but perhaps in some place where we'd want to show the sample problem code alongside rendered output.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -3220,6 +3220,24 @@ sub tag {
 	return "<$tag" . ($attributes_str ? " $attributes_str" : '') . '>' . ($SELF_CLOSING{$tag} ? '' : "$content</$tag>");
 }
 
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I guess this is not very useful since the problem editor does a fine job of showing problem code alongside a rendered problem.

pstaabp (approved):
works fine.  I'm not sure how I would use it though.

Alex-Jordan:
I'm going to close this. The `text_file` macro is perhaps useful but the record of how to do this will always be here.

PR #1200, opened 2025-02-24
Fix a typo that causes reductions of constants to persist where they shouldn't.

drgrice1:
This typo dates back to 2005, but is an obvious typo.


somiaj (approved):

pstaabp (approved):
Oh my.  Bad typo.

PR #1201, opened 2025-02-27
New sample problems and updates

pstaabp:
Added new Sample problems and updated existing problems.  

Updates includes
* removing the subjects of 'answer' and 'graphs' and reclassifying appropriate places for these problems. These two should never have been subjects. 
* running perltidy again on the sample problems.
* adding some additional links (categories, see_also, ..)

Note: for the new VectorGraphTool problem, #1195 needs to be merged. 

I've also updated the list of sample problems that @drgrice1 posted earlier to 
[setSample_Problems.def.txt](https://github.com/user-attachments/files/19011067/setSample_Problems.def.txt)


Alex-Jordan:
I have merged #1195, and it changed since you opened this PR. I haven't looked but I suspect you may want to revise the sample problem for that tool.

pstaabp:
Just updated the sample problem for the vector graph tool and added a line segment graph tool problem.

tutorial/sample-problems/Algebra/GraphToolCustomChecker.pg
```diff
@@ -159,10 +159,10 @@ The equation of the circle given by the equation
...
     [`[$circle_eq_lhs] = [$r ** 2]`]
 
-has center at [`([$h],[$k])`] and radius [$r].  To enter the graph, click the
-circle tool, then click on the center at [`([$h],[$k])`] and then click on a
-second point that is [$r] units from the center.  This is easiest going left,
-right, up or down from the center.
+has center at [`([$h],[$k])`] and radius [$r]. To enter the graph,
+click the circle tool, then click on the center at [`([$h],[$k])`]
+and then click on a second point that is [$r] units from the center.
+This is easiest going left, right, up or down from the center.
```
drgrice1, line 165: Why did you reformat this?  This already fit nicely into an 80 character limit (which is consistent with the default perltidy character width).  So there is no reason for this reformatting to be done.

tutorial/sample-problems/Arithmetic/UnitConversion.pg
```diff
@@ -0,0 +1,68 @@
...
+#: Don't forget to use `helpLink('units')` so your students will have access to the
+#: complete list of units that WeBWorK understands.
+BEGIN_PGML
+Perform the following unit conversions:
+
+a) Convert [$vol] to cups.  You may use [`1\;\text{L} = 1000\;\text{ml}`], [`1\;\text{qt} =  0.9464\;\text{L}`] and [`1\;\text{qt}=4\;\text{cups}`].
+
+    Answer: [_]{$vol_cups->with(sameUnits => 1)}
+
+b) Convert [$s] to miles per hour.  Recall that [`1\;\text{mile}=5280\;\text{feet}`].
```
drgrice1, line 56: This should be
```suggestion
b) Convert [$speed] to miles per hour.  Recall that [`1\;\text{mile}=5280\;\text{feet}`].
```
Otherwise this reads something like `Convert 110 to miles per hour.` which is not valid without a unit on `110` to begin with.

tutorial/sample-problems/Arithmetic/UnitConversion.pg
```diff
@@ -0,0 +1,68 @@
...
+
+    Answer: [_]{$speed->toUnits('mi/h')->with(sameUnits => 1)}
+
+Note: use units in all answers. [@ helpLink('units') @]*
+END_PGML
+
+#:% section = solution
+BEGIN_PGML_SOLUTION
+[$vol->toUnits('cups')]
```
drgrice1, line 65: This doesn't work and gives the error `Error evaluating variable $vol->toUnits('cups'): 'cups' is not defined in this context`.  You have changed to another context on line 39 above that does not have the volume units in it.  So you would need to switch back to that context for this to work.  Although it would be better to just use `$vol_cups` here since it has already been computed.

tutorial/sample-problems/DiffCalc/AnswerWithUnits.pg
```diff
@@ -22,26 +24,25 @@
...
 
 #:% section = setup
+#: First, we will set the variable for the functions in the problem to be t.
+#: Since there is a variable of t, we will assign it to be "seconds".
+#:
 #: Use the differentiation operator `->D('t')` to compute the derivated, and the
-#: evaluation method `->eval()` to evaluate it as a function.
+#: evaluation method `->eval()` to evaluate it as a function. All of these will
+#: now have the expect units.
 Context()->variables->are(t => 'Real');
```
drgrice1, line 36: In order for this problem to work this needs to be
```suggestion
Context('Units')->withUnitsFor('length', 'time')->variables->are(t => 'Real');
```

tutorial/sample-problems/Geometry/LineSegmentGraphTool.pg
```diff
@@ -0,0 +1,65 @@
...
+DOCUMENT();
+
+loadMacros('PGstandard.pl', 'PGML.pl', 'parserGraphTool.pl', 'PGcourse.pl');
+
+#:% section = setup
+#: This problem first finds two points with unequal x-values and unequal y-values.
+#: Below we will ask to drap the line segment between these two and add the correct answer
+#: `{ segment, solid, ($sx1, $sy1), ($sx2, $sy2) }` creates this line segment as a solid line.
+#:
+#:
```
drgrice1, line 31: Please delete this unnecessary empty line.

tutorial/sample-problems/Geometry/LineSegmentGraphTool.pg
```diff
@@ -0,0 +1,65 @@
...
+#:% see_also = [VectorGraphTool.pg, TriangleGraphTool.pg, QuadrilateralGraphTool.pg]
+
+#:% section = preamble
+#: Load the PODLINK('parserGraphTool.pl') macro to be able to use the GraphTool.
+DOCUMENT();
+
+loadMacros('PGstandard.pl', 'PGML.pl', 'parserGraphTool.pl', 'PGcourse.pl');
+
+#:% section = setup
+#: This problem first finds two points with unequal x-values and unequal y-values.
```
drgrice1, line 27: This should say "This problem first finds two distinct points."  The problem in fact does not find two points with unequal x and unequal y values.  The x-values may be equal, and in that case the y-values must be different, and vice versa.

tutorial/sample-problems/Geometry/LineSegmentGraphTool.pg
```diff
@@ -0,0 +1,65 @@
...
+#: Below we will ask to drap the line segment between these two and add the correct answer
```
drgrice1, line 28: How do you "drap" a line segment?

tutorial/sample-problems/Geometry/LineSegmentGraphTool.pg
```diff
@@ -0,0 +1,65 @@
...
+END_PGML
+
+#:% section = solution
+#: The important thing to note is that it is simple to add the correct graph using the
+#: method `generateAnswerGraph` as shown in the solution here.  The PGML directive [ ]*
+#: needs a * for the graph tool to be rendered correctly.
+BEGIN_PGML_SOLUTION
+The correct answer is
+
+[$gt->generateAnswerGraph]*
```
drgrice1, line 62: This should not use the `generateAnswerGraph` method.  Instead this should be
```suggestion
[!graph of correct answer!]{$gt}
```
Of course "graph of correct answer" is not good alternate text.  So either change that to better alternate text or make a comment in the documentation that  more descriptive wording should be used for the alternate text.

Make this change in all of the GraphTool problems that you added that have this code.

tutorial/sample-problems/IntegralCalc/GraphShading.pg
```diff
@@ -32,11 +32,13 @@ loadMacros('PGstandard.pl', 'PGML.pl', 'PGtikz.pl', 'PGcourse.pl');
 #: The shading in done with the `\filldraw` command with the
 #: `fill opacity = 0.5`. Also note that one of the sides of the filled area is
 #: the square root function.
-$a    = random(-4, 2);
-$b    = random(1,  4);
-$g    = Compute("$b + sqrt(x - $a)");
-$gtex = "$b + \sqrt{" . nicestring([ 1, -$a ]) . '}';
-$ans  = Compute("3*$b + 14/3");
+
+Context('Numeric');
```
drgrice1, line 36: Delete this line.  I just finished removing all of these unnecessary calls from the sample problems.  The `Numeric` context is the default context.  There is not need to activate it if that is the context to be used.

tutorial/sample-problems/Parametric/PolarGraph.pg
```diff
@@ -55,7 +55,7 @@ BEGIN_PGML
 Find the area enclosed by one petal of the rose curve
 [`r = f(\theta) = \cos(5\theta)`].
 
->>[!a rose curve!]{$graph}{300}<<  
+>>[!a rose curve!]{$graph}{300}<<
```
drgrice1, line 58: Put this line back to the way that it was.  It was that way for a reason.  The two spaces at the end of the line give a PGML break.  That makes `Graph of ...` centered on the next line instead of on the same line to the right of the image which looks quite bad.

tutorial/sample-problems/Parametric/SpaceCurveGraph.pg
```diff
@@ -13,7 +13,7 @@
 
 #:% name = Space Curve Graph
 #:% type = Sample
-#:% subject = [parametric, graph]
+#:% subject = [parametric, vector calculus]
```
drgrice1, line 16: Instead of just removing the `graph` subject, make `graph` a category.  This is still a problem with a graph.

tutorial/sample-problems/ProblemTechniques/RestrictAnswerToFraction.pg
```diff
@@ -13,7 +13,7 @@
 
 #:% name = Restrict Answers to a Fraction
 #:% type = [technique, sample]
-#:% subject = [answer]
+#:% subject = [algebra, precalculus]
```
drgrice1, line 16: Add the `answers` category to this problem instead of just removing the `answer` subject.

tutorial/sample-problems/ProblemTechniques/RestrictingFunctions.pg
```diff
@@ -13,7 +13,7 @@
 
 #:% name = Restrict Answers to a Constant
 #:% type = [technique, sample]
-#:% subject = [answer]
+#:% subject = [trigonometry]
```
drgrice1, line 16: Same here.  Add the `answers` category.

tutorial/sample-problems/Statistics/BarGraph.pg
```diff
@@ -0,0 +1,97 @@
...
+#:% section = setup
+#: First, we're going to use the 'Percent' context to allow students to enter percents.
+
+#: The graph is created dynamically. The axes are created with the given labels.
+#: Extending the number of bars or the height of the bars should be similar to this code.
+#: The heights of the bars are the grades which are generated randomly with the perl
+#: `map` function. If you want a non uniform distribution, you can create a different
+#: range of random numbers for each grade level.
+#:
+#: Make sure that alt text is added to all graph for accessibility. Be detailed.
```
drgrice1, line 36: This should be
```suggestion
#: Make sure that an alternate text is added to all graphs for accessibility. Be detailed.
```

tutorial/sample-problems/Statistics/BarGraph.pg
```diff
@@ -0,0 +1,97 @@
...
+    . "bar labelled D is $grades[3].";
+
+# This code adds up all the numbers in the @grades array.
+$num_students = 0;
+$num_students += $_ for @grades;
+
+$perA = Real($grades[0] / $num_students);
+
+#:% section = statement
+#: Int this case, we plot the bar graph and ask a question.
```
drgrice1, line 81:
```suggestion
#: In this case, we plot the bar graph and ask a question.
```

tutorial/sample-problems/VectorCalc/GraphToolVectors.pg
```diff
@@ -0,0 +1,57 @@
...
+DOCUMENT();
+
+loadMacros('PGstandard.pl', 'PGML.pl', 'parserGraphTool.pl', 'PGcourse.pl');
+
+#:% section = setup
+#: This problem asks for a vector with random endpoint $vx1, $vy1.  Create this with
+#: `{ vector, solid (0,0), ($vx1, $vx2) }`.
+#:
+#: The answer check will give credit for a vector with same length and point in the
+#: same direction.  The starting point does not need to be the origin.
```
drgrice1, line 30: This should be
```suggestion
#: This problem asks the student to graph the random vector <$vx1, $vy1>.  Set
#: this as the correct answer by passing `{ vector, solid (0,0), ($vx1, $vx2) }`
#: to the `GraphTool` method.
#:
#: The checker will give credit for any vector that is graphed with the same
#: length and direction.  The graphed vector's initial point does not need to be
#: the origin.
```

tutorial/sample-problems/VectorCalc/GraphToolVectors.pg
```diff
@@ -0,0 +1,57 @@
...
+$vx1 = random(-8, 8);
+$vy1 = random(-8, 8);
+
+$gt = GraphTool("{vector, solid, (0, 0), ($vx1, $vy1)}")
+    ->with(availableTools => [ 'VectorTool', 'FillTool', 'SolidDashTool' ]);
+
+#:% section = statement
+#: First, ask the question with the given points.  Then add the `[_]{$gt}` to produce
+#: the graph tool within the problem.
```
drgrice1, line 39: Reword this.  What are the "points" this is talking about?

tutorial/sample-problems/VectorCalc/VectorLineSegment2.pg
```diff
@@ -41,14 +42,14 @@ $ans = Compute("$P + $speed * t * $disp / norm($disp)");
...
 #:% section = statement
 BEGIN_PGML
-A particle starts at the point [`P = [$P]`] when [`t = 0`] and moves along a
-straight line toward [`Q = [@ Point($P + $disp) @]`] at a speed of [`[$speed]`]
-centimeters per second.  Assume that [`x`], [`y`], and [`z`] are measured in
-centimeters.
+A particle starts at the point [`P = [$P]`] when [`t = 0`]
+and moves along a straight line toward [`Q = [@ Point($P + $disp) @]`]
+at a speed of [`[$speed]`] centimeters per second . Assume that [`x`],
+[`y`], and [`z`] are measured in centimeters.
```
drgrice1, line 48: Put this back to the way it was.  This is unnecessary reformatting, and isn't even formatted as well as it was.

tutorial/sample-problems/VectorCalc/VectorLineSegment2.pg
```diff
@@ -41,14 +42,14 @@ $ans = Compute("$P + $speed * t * $disp / norm($disp)");
...
 
 Find a vector equation for the position of the object.
 
-[` \vec{r}(t) = `] [____]{$ans}
+[` \vec{r}(t) = `] [__]{$ans}
```
drgrice1, line 52: Why did you change this?  You made the answer rule shorter.  If anything the answer rule should be made longer for this problem.  Also it would be better to use the `[_]{$ans}{10}` syntax for the answer rule length.  It is clearer what the length setting is than having to count underscores.  Note that this only affects the answer rule length when MathQuill is not in use.

drgrice1 (commented):
There are several changes needed here.

First , please reformat the problems that you added.  Always try to fit into an 80 character text width whenever possible.  That is what we have the pg perltidy script set to use.  One reason for that is that even with a wide browser window anything more will wrap and look bad in the problem editor.

Note that the `sample-problems/Geometry/LineSegmentGraphTool.pg` problem is not in the set definition file that you posted.  I am just noting this so that others testing this pull request will know to add it (or look at it in a different way).

tutorial/sample-problems/Arithmetic/UnitConversion.pg
drgrice1: Actually, in thinking about this more.  It would be better if you deleted line 65.  This is a bad practice that should not be encouraged in the sample problems.  What you are showing here is NOT a solution.  It is a correct answer.  Correct answers are already shown when solutions are available, and the solution should not be just the correct answer again.  A solution should show how to work the problem, and not just be an answer.

tutorial/sample-problems/Algebra/GraphToolCustomChecker.pg
pstaabp: Not sure?  reverted it.

pstaabp:
I think the couple that I reformatted to be more narrow was just due to a narrow screen I was using in the ProblemEditor.   If I use vscode, I get the 80 limit as a vertical line.  Do you know if CodeMirror6 has the ability to include a vertical line at a given width.  That might be helpful. 

I have made updates on these and just pushed.

tutorial/sample-problems/DiffCalc/AnswerWithUnits.pg
```diff
@@ -22,26 +24,25 @@
...
-loadMacros(
-    'PGstandard.pl',            'PGML.pl',
-    'parserNumberWithUnits.pl', 'parserFormulaWithUnits.pl',
-    'PGcourse.pl'
-);
+loadMacros('PGstandard.pl', 'PGML.pl', 'contextUnits.pl', 'PGcourse.pl');
 
 #:% section = setup
+#: First, we will set the variable for the functions in the problem to be t.
+#: Since there is a variable of t, we will assign it to be "seconds".
```
drgrice1, line 31: Change this to something like

```suggestion
#: First, set the variable for the functions in the problem to be t,
#: and assign the unit "seconds" to it.
```

Generally, try to avoid using the "royal we" in the documentation.  That is bad practice.

drgrice1 (commented):
When I asked that the problems be formatted to fit into the 80 character limit, I mean the entire problem.  That means documentation and PGML sections as well.  That is what I did with the existing sample problems in the last related pull request.  I see this still has not been done.

pstaabp:
I double checked all of the problems. 

The only cases that I think remain is when the categories or sample problems list goes beyond, but I think we'd have to rewrite the script to handle that.

drgrice1:
That is better.  Yeah, the categories have to be on the same line at this point.

The next thing to look at is the wording in the documentation of the new problems.  For example, in the `LineSegmentGraphTool.pg` it says

```
Below we will ask to graph the line segment between these two and add the
correct answer `{ segment, solid, ($sx1, $sy1), ($sx2, $sy2) }` creates this
line segment as a solid line.
```

First it has the "royal we" I said to try and get rid of.  Second, it is a run on sentence.

Then a few lines later in the same documentation it states:

```
Mulitple line segments can be added to the graph tool, but note that there is
a Triangle Graph Tool, PROBLINK('TriangleGraphTool.pg') and a Quadrilateral
Graph too, PROBLINK('QuadrilateralGraphTool.pg') as well.
```

Technically that is not a run on sentence, but it is a conjunction between two sentences that are really unrelated, and reads quite poorly.

pstaabp:
I do use the royal "we" and will look overthing again.

pstaabp:
Made some more changes.  Got rid of the royal "we" and tried to be more consistent with the style of the other problems.

somiaj (approved):
Example updates are good, no real comment on the actual problems though. Just one minor question below.

tutorial/sample-problems/DiffCalc/AnswerWithUnits.pg
```diff
@@ -12,6 +12,8 @@
 ## MO(1)
 ## KEYWORDS('differential calculus', 'units')
 
+# Updated 2025.02 to use the contextUnits.pl macro.
+
```
somiaj, line 16: Do we want to get into the habit of adding changelog comments to some of these examples? I don't initially see the point to this.

drgrice1: I also did not like that this comment was added.  It should be removed.

Alex-Jordan:
```suggestion
```

PR #1202, opened 2025-02-27
Make crossed quadrilaterals fill only the region containing the fill point.

drgrice1:
This is to fix a comment I placed in the code about filling quadrilaterals, and the comment made by @Alex-Jordan in https://github.com/openwebwork/pg/pull/1191#issuecomment-2683381959.

The code implemented in #1191 fills the entire interior of a quadrilateral even if it is a crossed quadrilateral (that is a quadrilateral with two edges that intersect). That is inconsistent with all of the other graph tool objects in some sense.  To be more consistent only the region inside the quadrilateral that satisfies the same inequalities relative to the borders should be filled.  So that is what this does.

I also was doing some benchmarking of the TikZ fill code used for generating hardcopy and noticed that some things were rather slow. The slow down it turns out was using MathObject Reals in computations. There is no need for the MathObject structure in those computations, so all of the fill code now uses the perl values instead. It isn't even funny how much faster computations are with pure perl numbers versus MathObject Reals. The difference is rather extreme. Performing something like 10000 basic arithmetic operations with MathObject Reals takes more than five seconds, but with pure perl numbers it takes less than 20 milliseconds.

### DIFF
--- a/htdocs/js/GraphTool/quadrilateral.js
+++ b/htdocs/js/GraphTool/quadrilateral.js
@@ -65,11 +65,9 @@
 				].join(',');
 			},
 
-			// Note that this is an interior fill which is a bit inconsistent with the other fill cmp methods for other
-			// graph objects.  Other grap objects use an inequality fill.
 			fillCmp(gt, point) {
 				// Check to see if the point is on the border.
-				for (i = 0, j = this.definingPts.length - 1; i < this.definingPts.length; j = i++) {
+				for (let i = 0, j = this.definingPts.length - 1; i < this.definingPts.length; j = i++) {
 					if (
 						point[1] <= Math.max(this.definingPts[i].X(), this.definingPts[j].X()) &&
 						point[1] >= Math.min(this.definingPts[i].X(), this.definingPts[j].X()) &&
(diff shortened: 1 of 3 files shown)


drgrice1:
That is exactly the expected result with the current fill algorithm.  All of the points that are shaded are in the bottom region of the larger quadrilateral, and outside the other quadrilateral.

Remember the current fill algorithm was designed to be an inequality fill.  So here are a couple of examples to demonstrate this.

The following screenshot is the graph of the parabolas $y=x^2-4$ and $x=y^2-4$, and the region filled is the solution to the system of inequalities $y < x^2 -4$ and $x < y^2 - 4$.

parabola-fill

The following is the graph of the lines $y=4$, $y=-4$, $x=4$, and $x=-4$, and the circle $x^2+y^2 = 16$.  The region filled is the solution to $y<4$, $y>-4$, $x<4$, $x>-4$, and $x^2+y^2 <16$.

circle-square-fill

The way that quadrilaterals are designed with respect to filling for the develop branch is an exterior versus interior fill.  So all points inside the quadrilateral are filled together assuming they all satisfy the same inequalities relative to the other objects graphed.

In a sense, the fill implemented in this pull request is similar to the second example above.

It seems that you are expecting a completely different kind of fill than what is implemented.  I am guessing that you are thinking of a flood fill.  I have a partial flood fill implementation that at some point I want to add as an alternate optional algorithm.  Unfortunately, it doesn't entirely work, and there are some serious obstacles to really getting it to work correctly and efficiently.

Alex-Jordan:
I'm not trying to express opinions, and I am happy to just have these tools. I'm guessing that general uses will be expecting flood fills, so there may just be some confusion and/or need for education. Although, these scenarios with multiple non-line objects that need regions filled will be rare.

drgrice1:
At this point a flood fill is not an option.  I have been working for about 10 years to actually get that to work.

Alex-Jordan (approved):
This all makes sense, and works.

pstaabp (approved):
I checked some regions on develop and didn't notice it being slow, but the PR looks good.  Makes sense to use pure perl reals.

PR #1205, opened 2025-03-07
Bump the version of MathQuill to pull in the removal of default `autoOperatorNames`.

drgrice1:
See https://github.com/openwebwork/mathquill/pull/37.

All default `autoOperatorNames` are removed to address https://github.com/openwebwork/webwork2/issues/2203 and https://github.com/openwebwork/webwork2/issues/2681.

Any of those operators can be added back if desired, but having them by default continues to cause issues with typing other things.

### DIFF
--- a/htdocs/package-lock.json
+++ b/htdocs/package-lock.json
@@ -7,7 +7,7 @@
             "name": "pg.javascript_package_manager",
             "license": "GPL-2.0+",
             "dependencies": {
-                "@openwebwork/mathquill": "^0.11.0-beta.3",
+                "@openwebwork/mathquill": "^0.11.0-beta.5",
                 "jsxgraph": "^1.10.1",
                 "jszip": "^3.10.1",
                 "jszip-utils": "^0.1.0",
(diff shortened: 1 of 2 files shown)


pstaabp:
Here's a problem that I tested this with: 
```
DOCUMENT();

loadMacros('PGstandard.pl', 'PGML.pl', 'contextUnits.pl');

Context('Units')->withUnitsFor('length', 'time');

$v = Compute("5 min/mi");

BEGIN_PGML
Enter [$v]

[_]{$v}
END_PGML

ENDDOCUMENT();
```

pstaabp (approved):
This looks good.

pstaabp:
@drgrice1: you can add back operators to math quill on a problem-level (or course-level) as needed, correct?  Where is this documented?

drgrice1:
Yes, you can do that.  However, it isn't really documented anywhere at this point.

The method used in https://webwork.maa.org/wiki/NamedAnswerRules gives the basic technique to modify the properties of the mqeditor for a specific answer rule.  That will work with `window.answerQuills.$ansName.mathField.addAutoOperatorNames` to add operators back.  Code can also be added in a problem to add operators for all answers, or in `PGcourse.pl` to do that for all problems in a course.

I thought I added a version the problem linked above to the sample problems, but now I am not seeing it. Hmm...

Alex-Jordan (approved):

PR #1206, opened 2025-03-07
Add a flood fill algorithm to the GraphTool.

drgrice1:
That is a fill algorithm that fills to borders.  The current algorithm is an "inequality" fill.  That means that it shades all points that satisfy the same inequalities relative to the graphed objects as the fill point.  The previous inequality fill algorithm is still available, but the new flood fill algorithm is the default.  To use the inequality algorithm set `useFloodFill => 0`.

The inequality fill does not entirely make sense with triangles, quadrilaterals, vectors, or segments.  The flood fill is more intuitive in general for all available objects.  However, the inequality fill will work better for some scenarios.  For example, if the problem is to solve a system of inequalities.  That is of course what the inequality fill was designed for. However, with the addition of these new graph objects for which there isn't a single inequality that describes the "sides" of the object (like triangles and quadrilaterals) or that don't divide the plane into regions (like vectors and segments) the flood fill is needed.

There is one caveat with the flood fill to watch out for.  The flood fill considers the edges of the board to be borders.  That means that if an object goes off the edges of the board that fills will stop there. For instance, if the board's ranges are -10 to 10 in both directions, and the circle centered at the origin with radius 10 is graphed, then a fill that is placed at the point (8, 8) will only shade the upper right corner of the board outside the circle, and not the upper left, lower left, and lower right corners.  This is a case that the inequality fill is better.  It will shade all four corners since they all satisfy the inequality x^2 + y^2 > 100 that the point (8, 8) satisifies.

So authors will need to consider the options carefully and decide which is best for their problem.  If the flood fill is used, in many cases care will need to be taken to ensure that the expected region is filled and that random parameters won't generate something that could cause unexpected results.

In addition, if the inequality fill will work, then that should be chosen. It is much more efficient and reliable both for javascript client side, and for hardcopy generation server side. Although, I believe the flood fill should work quite well in most cases now. The TikZ code performs the same flood fill algorithm that the javascript does, and then uses an edge detection algorithm to convert the filled region into an outer curve that is filled and inner curves (if there are holes) that are clipped out.  The edge detection seems to be working rather well in the scenarios that I have come up with, but there may be issues that may require that the algorithm be tweaked further.

The default grader for fills if the flood fill is used also performs the flood fill. Basically, it starts at the correct fill point and performs the flood fill algorithm until it reaches the student's fill point.  As such it can't use MathObject Reals because they are way to slow. MathObject Reals really aren't needed for this anyway though. These are answers provided from the javascript, not things entered by students into a text or MathQuill input.

I also noticed that concave quadrilaterals were being filled in the interior an odd way with #1202 and the inequality fill.  Only the portion of the interior that satsifies the same inequalities relative to the borders was being filled, and not the entire interior.  So this also fixes that issue. Now the entire interior of a quadrilateral is filled unless it is a crossed quadrilateral. Although, as noted above, using the inequality fill with quadrilaterals doesn't make much sense.

### DIFF
--- a/htdocs/js/GraphTool/cubictool.js
+++ b/htdocs/js/GraphTool/cubictool.js
@@ -19,7 +19,6 @@
 			postInit(_gt, point1, point2, point3, point4) {
 				this.definingPts.push(point1, point2, point3, point4);
 				this.focusPoint = point1;
-				this.floodFillCheckSides = true;
 			},
 
 			stringify(gt) {
@@ -33,10 +32,6 @@
 
 			fillCmp(gt, point) {
 				return gt.sign(point[2] - this.baseObj.Y(point[1]));
-			},
-
-			hasPoint(gt, point) {
-				return Math.abs(point[2] - this.baseObj.Y(point[1])) < 0.5 / gt.board.unitY;
 			},
 
 			restore(gt, string) {
--- a/htdocs/js/GraphTool/cubictool.js
+++ b/htdocs/js/GraphTool/cubictool.js
@@ -19,6 +19,7 @@
 			postInit(_gt, point1, point2, point3, point4) {
 				this.definingPts.push(point1, point2, point3, point4);
 				this.focusPoint = point1;
+				this.floodFillCheckSides = true;
 			},
 
 			stringify(gt) {
@@ -32,6 +33,10 @@
 
 			fillCmp(gt, point) {
 				return gt.sign(point[2] - this.baseObj.Y(point[1]));
+			},
+
+			hasPoint(gt, point) {
+				return Math.abs(point[2] - this.baseObj.Y(point[1])) < 0.5 / gt.board.unitY;
 			},
 
 			restore(gt, string) {
--- a/htdocs/js/GraphTool/graphtool.js
+++ b/htdocs/js/GraphTool/graphtool.js
@@ -837,6 +837,15 @@ window.graphTool = (containerId, options) => {
 	class GraphObject {
 		supportsSolidDash = true;
 
+		// If this is true for an object, then the flood fill algorithm will check for and not allow side changes.  This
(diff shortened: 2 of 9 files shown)


drgrice1:
Here is a problem that I use to test things: 
[FloodFill.pg.txt](https://github.com/user-attachments/files/19134687/FloodFill.pg.txt)

There are groups of commented out answers (and corresponding PGML text) you can uncomment to test the grader and there are commented out `staticObjects` you can uncomment in various combinations to test hardcopy generation.

drgrice1:
If you want to see a real mess, uncomment all of the `staticObjects` in that problem.  That doesn't give a very good test for hardcopy generation though (but does work).  Once you have that much junk the algorithm doesn't have a problem because not much is actually filled.  There are several groups separated by empty lines that give some good tests, and then the big group mix and match.

drgrice1:
This is now a draft pull request.  There are tolerance issues that can cause bleeding of the fill into other regions.  I will work on a fix, but for now this isn't acceptable.

Alex-Jordan (approved):
I played around testing some things from that testing file and it all worked well. Hardcopy too.

Predictably, when using the flood fill and producing hardcopy, I see a spike in CPU usage from `xelatex`. What would you think about keeping the inequality fill as the default?

I notice that if the only tools and static objects are a subset of {Point, Line, Fill} then it's impossible for flood fill to create a different looking result than inequality fill. Is it worth checking for this and forcing inequality fill?

Kudos on completing a challenging project.

pstaabp (approved):
I think this looks good as well. 

I tested also with a few problems that I have that would fit in the inequality fill that seems to work fine with the default `floodFill => 1`.  There are so few graph tool problems in the OPL (if any) that I don't think switching the default is a problem, but in general it's always good to keep the previous defaults the same.

Issue #1204, opened 2025-03-07
Fraction problem broken on develop


Alex-Jordan:
Try the MWE problem below. I don't think seed is relevant because it's the same with several seeds I tried. But in case it matters, for me this all started with seed 3444.

- It works on a 2.19 server.
- On develop, it does not work. You enter the correct answer and it is declared incorrect. If I change the limits like with `Context()->flags->set(limits => [0,1]);
`, then it works.

It seems likely that this has to do with some side effect from the changes that were made to Fractions recently.

```
DOCUMENT();

loadMacros(
  "PGstandard.pl",
  "PGML.pl",
  "parserAssignment.pl",
  "contextFraction.pl",
);

Context("Fraction")->variables->add(y => 'Real');
parser::Assignment->Allow;
#Context()->flags->set(limits => [0,1]);

$equation = Formula("y = 8/3 x - 3");

BEGIN_PGML
Enter [`[$equation]`].

[_]{$equation}{16}
END_PGML

ENDDOCUMENT();
```

dpvc:
I know what's happening and can fix it, but have to run right now so can't write it up just yet.  But I didn't want anyone else to waste time on it.  I'll make a PR soon.

dpvc:
I've made PR #1208 to resolve the problem.

This actually *is* related to the problem seed, but in a subtle way.  In the correct answer, because `reduceConstants` is true, the division in `8/3`is performed during the parsing of the answer, producing a fraction object.  It turns out that one of the tests points turned out to have `x = -2`, which means the `8/3 x` will be handled as a fraction times an integer producing another fraction, and then that minus 3 is another fraction, so the result is -25/3 as a Fraction object.

On the other hand, when the student answer is parsed, `reduceConstants ` is false (so that the student answer preview will show what the student actually entered), and that means the `8/3` is left as a division in the parsed answer, not a Fraction object.  So when that is evaluated, the `8/3` is computed first, producing a decimal number, not a fraction, and that is multiplied by -2 then 3 is subtracted, as decimal numbers.  It turns out that these have slightly different round-off errors (the truncation error from 8/3 is moved up one bit by when multiplied by -2 compared to the truncation error for -25/3).

[...]

PR #1208, opened 2025-03-08
Fraction eval method should return Real MathObject not perl real

dpvc:
This PR corrects an error where the `eval()` method of a Fraction MathObject would return a perl real rather than a MathObject Real.  That lead to comparisons that didn't use the fuzzy tolerances, which in turn lead to values that differ by small round-off errors being considered not equal when they should be equal.  E.g., Fraction(-25,3) not equaling (8/3)*(-2)-3.

This also changes the `compare()` method to use a better comparison for fractions, and a more precise tolerance value for comparing a fraction and a real, given by the new `fractionTolerance` flag.  Comparisons between a fraction and anything other than a fraction, real, or infinity will produce error messages.

This resolves issue #1204 without making it too easy to use a decimal value to match a fraction.

### DIFF
--- a/macros/contexts/contextFraction.pl
+++ b/macros/contexts/contextFraction.pl
@@ -47,7 +47,7 @@ must enter a whole number or a fraction in this context.  It is also
 permissible to enter a whole number WITH a fraction, as in C<2 1/2> for
 "two and one half", or C<5/2>.
 
-The fourth is the same as LimiteFraction, but students must enter proper
+The fourth is the same as LimitedFraction, but students must enter proper
 fractions, and results are shown as proper fractions.
 
 It is also possible to add fractions to an existing context using
--- a/macros/contexts/contextFraction.pl
+++ b/macros/contexts/contextFraction.pl
@@ -205,6 +205,20 @@ indicating that mixed numbers are not allowed.  This flag is off by
 default in all four contexts.  You should not set both
 C<requirePureFractions> and C<requireProperFractions> to 1.
 
+=item S<C<< fractionTolerance >>>
+
+Thie determines the tolerance to use when comparing a fraction to a
+real number.  The fraction will be converted to a real, and then this
+is used as the tolerance in a relative-tolerance comparison of the two
+reals.  The default is 10E-10, meaning the decimal must match to
+roughly 10 digits.
+
+=item S<C<< contFracMaxDen >>>
+
+This is the largest denominator to allow for continued-fraction
+computations when trying to produce a fraction from a real number.
+The default is 10**8.
+
 =back
 
 Fraction objects have two methods that can be useful when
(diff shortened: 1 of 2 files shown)


macros/contexts/contextFraction.pl
```diff
@@ -205,6 +205,20 @@ =head1 DESCRIPTION
 default in all four contexts.  You should not set both
 C<requirePureFractions> and C<requireProperFractions> to 1.
 
+=item S<C<< fractionTolerance >>>
+
+Thie determines the tolerance to use when comparing a fraction to a
+real number.  The fraction will be converted to a real, and then this
+is used as the tolerance in a relative-tolerance comparison of the two
+reals.  The default is 10E-10, meaning the decimal must match to
+roughly 10 digits.
```
drgrice1, line 214: I think you mean `1E-10` here.  In the code you use `10**-10` which is equal to `1E-10`.  Furthermore, that would actually mean roughly 10 digits of accuracy, while `10E-10` only means roughly 9 digits of accuracy.

Not that you don't know that.  Probably just a typo, or mental conversion error from the `**` notation to the `E` notation.

dpvc: I made a lot minute change from `10**-10`, which just didn't look good with the two 10s and negative sign, and didn't change it correctly.  I will fix.

pstaabp: Another minor typo, not from this PR is on line 50 of contextFraction.pl. `LimiteFraction`.  Since you have this open, instead of making another PR.

Alex-Jordan:
I went to try this out and something strange is happening (even before I check this out).

- On my production server (which is basically develop from January with a few patches) the MWE still shows the bug.
- On my develop server (a clean up-to-date) I can't actually reproduce the bug.

More out of curiosity that anything else, could this be explained by my perl version? On that develop server the perl version is 5.26, but on my production it's 5.36.

Alex-Jordan:
At last week's developer meeting, it seemed like two or three (@drgrice1, @pstaabp, @dlglin) could confirm this worked. I had trouble since I couldn't reproduce the actual bug (that I reported!). But do we have enough to merge this?

drgrice1 (approved):
I think this is good.

dpvc:
I think there is a POD typo that I still need to fix.  Sorry.  Will get to that soon.

dpvc:
OOPS, looks like I already did that.  Sorry, I thought I hadn't.  I think you are good to go.

drgrice1:
I noticed your edit to `PGalias.pm` that you just removed.  You can leave that in if you want.  There is no problem with that being there.  It does make the seed consistent with the other variables.  I assume you have a local script that you test things with that doesn't set the seed, and that is why you added that.

dpvc:
Indeed, you are correct about the reason.  Ok, I can undo the last commit.  It is a bit of a pain to remember not to add that file in when making changes, and you can see that I forgot this time.  I only noticed it because I was changing branches and didn't see that file in my modified list, and wondered where it went.  Then I found it in that previous commit.

I'm happy to have it merged into develop, so I'll put it back.

Alex-Jordan (approved):

Issue #1207, opened 2025-03-08
Issue with `parserAssignment.pl`.


drgrice1:
This issue occurs with WeBWork/PG 2.18, 2.19, and develop, but not with 2.17 (or before).  The following MWE demonstrates the issue:

```perl
DOCUMENT();

loadMacros('PGstandard.pl', 'PGML.pl', 'parserAssignment.pl', 'PGcourse.pl');

parser::Assignment->Allow;

$answer = Compute('x = -4');

BEGIN_PGML
Enter [`-4`] (but the correct answer is the equation [`[$answer]`]).

[_]{$answer}{10}
END_PGML

ENDDOCUMENT();
```

If you enter `-4` for the answer, then a score of 50% is given.  Even though the correct answer is the equation `x = -4`.  Furthermore, the message "There should be more words in your assignment" is given which doesn't make much sense.

I don't think this is related to #1193 since this occurs for WeBWorK/PG 2.18 and 2.19, and not just for develop.

drgrice1:
I forgot to mention.  I also discovered that if you use `[_]{$answer->cmp(partialCredit => 0)}`, then this does not happen.  The `Value::List::cmp_equal` method is handling this, but I don't know what has changed since 2.17 that causes this.

drgrice1:
Actually, you also need to add `showLengthHints => 0` to get rid of the message.  My actual problem doesn't need that because it uses a post filter to grade, and it replaces the message.

somiaj:
Seems `$showPartialCorrectAnswers = 0;` also makes it not give partial credit, so maybe a change in partial credit with macro?

drgrice1:
`$showPartialCorrectanswers` is the default for those options.  So that is the same thing.

dpvc:
I've made a PR that should address the problem (it adds `partialCredit => 0` and `shoeLengthHints => 0` to the default options for assignment answers.

drgrice1:
This was fixed by #1209.

PR #1209, opened 2025-03-08
Set default options to prevent partial credit or length hints for assignments

dpvc:
This PR sets the defaults for `partialCredit` and `showLenghtHints` for assignments so that unwanted messags and credit aren't given.

This resolves issue #1207.

### DIFF
--- a/macros/parsers/parserAssignment.pl
+++ b/macros/parsers/parserAssignment.pl
@@ -325,7 +325,7 @@ sub TeX {
 #
 sub cmp_defaults {
 	my $self = shift;
-	{$self->SUPER::cmp_defaults(@_), showLengthHints => 0, partialCredit => 0};
+	{ $self->SUPER::cmp_defaults(@_), showLengthHints => 0, partialCredit => 0 };
 }
 
 #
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
Fixes the issue.

drgrice1 (approved):
I think this is fine and fixes the problem.  However, I would like to know what changed after PG 2.17 that caused the grading to change.  Do you have any ideas on that?

dpvc:
I don't know for sure, but I think it may have been when you corrected a misspelling in `PGcore.pm` in commit f1e654f that changed `showpartialCorrectAnswers` to `showPartialCorrectAnswers`, as it is the value of `showPartialCorrectAnswers` that seeds the values for the `partialCredit` and `showLengthHints` values.  There were no important changes to either of those options between PG-2.17 and PG-2.18, and this is the only important change involving `showPartialCorrectAnswers`.  I haven't checked that that is the actual culprit, but that would be my best bet.

drgrice1:
Ahh.  That makes sense.  I am not entirely sure that is the actual culprit either, but I am betting it is.  I suspected it was something like that that was causing this.

I will assume that is the cause, and in that case, I can put full support behind this pull request.

PR #1210, opened 2025-03-10
Fix filling of concave quadrilaterals.

drgrice1:
Currently only the portion of a concave quadrilateral that the fill point is in is filled, and not the entire interior because the current check separates portions that satisfy different inequalities with respect to the borders.  So this makes it so the interior parts are only distinquished for crossed quadrilaterals.

Here is a screen shot demonstrating the issue:
bad-fill

Also fix the documentation about how to use the `%graphToolObjectCmps` hash entry for vectors.  The GraphTool object must be passed so that it can get the `vectorsArePositional` option. This is only for those using this for a custom checker (if anyone does).

This was included in #1206.  However, I am (at least temporarily) changing that pull request to a draft until (and if) some issues are resolved.  These things are unrelated to that pull request really.


pstaabp (approved):
This fixes the issue.

drgrice1:
I forgot to add the corresponding TikZ code for this.  I added that now.

Alex-Jordan:
Working for me.

PR #1211, opened 2025-03-13
Fix bracket ends for graphtool intervals with the `useBracketEnds` setting.

drgrice1:
In #1157 I made it so that the `up`, `down`, and `drag` handlers for the defining points of graph objects are not added to static graph objects.  However, in doing so for intervals I also made it so that the paired point association is not made.  For most graph objects those associations are only needed when the handlers are added since they are used to prevent points from being moved in such a way to prevent degeneracy for the object.  But for intervals that association is also used to determine which direction the brackets or parentheses should face when the `useBracketEnds` option is used.  As such, now correct answer graphs always show those facing to the right even if they are on the right end.


Alex-Jordan (approved):

pstaabp (approved):
Fixes the issue.

PR #1213, opened 2025-03-21
Update the Rserve interface.

drgrice1:
The primary change in this pull request is how the `rserve_get_file` method in the `RserveClient.pl` macro works.  Previously it accepted a second optional local filename parameter.  That parameter is no longer honored.  If it is given it is ignored.  I don't think that there are any problems in the OPL or Contrib that pass that argument, but even if they do it will still work.  The filename just won't be the chosen one. The filename should not be a problem author choice.  Furthermore, previously the method fell back to using the Rserve remote filename. That is also no longer used.  Instead, the `getUniqueName` method is used that creates a filename that is dependent on the problem seed, psvn, problem UUID, etc. That means that the filename will not change each time that the problem is loaded. Unfortunately, the file is still downloaded every time the problem is loaded and overwrites the existing file (with the same contents), and that can't really be fixed without a major revision of the Rserve setup.

Note that all changes in this pull request are backwards compatible, and all existing problems will continue to function correctly. However, the courses html temporary directory will no longer be filled with files that are only used once before another copy of the file is created by another name the next time the problem is rendered (due to a preview, answer submission, etc.).

In addition there are two new methods added that are more convenient than the previous methods for generating R images or csv files. Those are the `rserve_data_url` and `rserve_plot` methods.  These handle the details of creating these things in a much nicer way.  See the added POD and examples in the POD for the usage and comparison to the old approaches.

The macro code and POD are also rather heavily clean up.  The `Rserve.pm` module is also cleaned up. The `_inherits` method is removed from the `Rserve.pm` module, because contrary to the comments the `inherits` method of the `$rserve` object does work in the safe compartment, and so the `_inherits` method is not needed.

There is one other related change. The `htmlLink` method now has a new calling convention for adding additional attributes to the generated `a` tag. Previously these were added by passing a single string argument. For example, `htmlLink($url, 'dataset', 'download="dataset.csv"')`.  That will still work, but passing the attributes in that way should be considered deprecated.  Instead the attributes should be passed as additional attribute/value pairs after the url and link text. For example, `htmlLink($url, 'dataset', download => 'dataset.csv')`.  Of course, this is useful for Rserve csv files to specify the filename that will be offered when the user clicks to download a csv file.  Particularly since the default will now be something like `ed721060-00b2-37fb-87c4-e3eb36c7ced2___5643740d-c34f-37c8-afc5-c2369fec26e1.csv`. Previously it would have been something like `file1231asdfasdf.csv`. The new alias is worse, but that wasn't good either.

### DIFF
--- a/conf/pg_config.dist.yml
+++ b/conf/pg_config.dist.yml
@@ -139,6 +139,12 @@ specialPGEnvironmentVars:
   # Size in pixels of dynamically-generated images, i.e. graphs.
   onTheFlyImageSize: 400
 
+  # To enable Rserve (the R statistical server), uncomment the following two
+  # lines. The R server needs to be installed and running in order for this to
+  # work. See http://webwork.maa.org/wiki/R_in_WeBWorK for more info.
+  #Rserve:
+  #  host: localhost
+
   # Locations of CAPA resources. (Only necessary if you need to use converted CAPA problems.)
   CAPA_Tools: $Contrib_dir/CAPA/macros/CAPA_Tools/
   CAPA_MCTools: $Contrib_dir/Contrib/CAPA/macros/CAPA_MCTools/
--- a/lib/Rserve.pm
+++ b/lib/Rserve.pm
@@ -11,53 +11,34 @@ my $rserve_loaded = eval {
 sub access {
 	die 'Statistics::R::IO::Rserve could not be loaded. Have you installed the module?'
 		unless $rserve_loaded;
-
-	Statistics::R::IO::Rserve->new(@_);
+	return Statistics::R::IO::Rserve->new(@_);
 }
 
-## Evaluates an R expression guarding it inside an R `try` function
-##
-## Returns the result as a REXP if no exceptions were raised, or
-## `die`s with the text of the exception message.
+# Evaluates an R expression guarding it inside an R `try` function
+#
+# Returns the result as a REXP if no exceptions were raised, or
+# `die`s with the text of the exception message.
 sub try_eval {
 	my ($rserve, $query) = @_;
 
-	my $result = $rserve->eval("try({ $query }, silent=TRUE)");
-	die $result->to_pl->[0] if _inherits($result, 'try-error');
-	# die $result->to_pl->[0] if $result->inherits('try-error');
+	my $result = $rserve->eval("try({ $query }, silent = TRUE)");
+	die $result->to_pl->[0] if $result->inherits('try-error');
 
-	$result;
+	return $result;
 }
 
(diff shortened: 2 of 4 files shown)


dlglin:
The instructions on the wiki (https://webwork.maa.org/wiki/R_in_WeBWorK#Transferring_files_from_the_R_server) will need to be updated to match.
Do we have sample problems for this?

drgrice1:
Note that the instructions on the wiki will still work as is.  Although it would be good to add examples of using the new method.  There are examples in the POD that are quite complete.

drgrice1:
I don't think there are any Rserve sample problems at this point.

drgrice1:
I removed the `chmod` changes to `LaTeXImage.pm` and `PGcore.pm` from this pull request.  I will add that in another pull request.

Alex-Jordan (approved):
Tested and it works as described.

drgrice1:
In case anyone is up for extensive testing of this, here are two set definition files.   The first contains all OPL problems that use the `RserveClient.pl` macro, and the second contains all Contrib problems that use the macro.  Note that there are other problems that load the macro, but don't use it that were not included in those set definitions.
[RserveProblems.zip](https://github.com/user-attachments/files/19457466/RserveProblems.zip)

Alex-Jordan:
When I tested, I just tested the two sample problems in the wiki. I looked closely at the file names (for the image and csv).

drgrice1:
Yeah, that is really enough.  The problems in the OPL and Contrib don't do anything different really.  But if you want to test extensively, then you can try the set definitions I just posted.

conf/pg_config.dist.yml
```diff
@@ -139,6 +139,12 @@ specialPGEnvironmentVars:
   # Size in pixels of dynamically-generated images, i.e. graphs.
   onTheFlyImageSize: 400
 
+  # To enable Rserve (the R statistical server), uncomment the following two
+  # lines. The R server needs to be installed and running in order for this to
+  # work. See http://webwork.maa.org/wiki/R_in_WeBWorK for more info.
+  #Rserve:
+  #  host: localhost
+
```
pstaabp, line 147: It looks like this is needed for the standalone server only?   Either way, we should update the wiki for this.

pstaabp (approved):
Tested.

conf/pg_config.dist.yml
drgrice1: This is needed for anyone not using webwork2 to render PG problems.

The standalone renderer has its own `pg_config.yml` file.  This will need to be added to that file.

I don't think there is a wiki page where this is discussed.  Where in particular are you thinking this should be commented on in the wiki?

PR #1215, opened 2025-03-23
tensor multiplication, powers, idenity, inverses

Alex-Jordan:
This continues enhancements that I am making to `Value/Matrix.pm`. I have more than this, but what I gathered here for this PR seems more manageable for a PR. I could add the rest now or keep things simple with smaller PRs and wait for this one to be reviewed.

* clean up the language for "Matrix" objects that aren't regular 2D matrices by referring to them as "degree n". There is a new `degree` method to get this number from a Matrix.
* Most vocabulary we apply to degree 2 Matrix objects (square, orthogonal, upper triangular, etc) makes sense for degree n Matrix objects as well, if you apply it to each degree 2 matrix you get by only looking at the "frontal slice". Here the `isSquare` method now applies to degree n Matrix objects in this way. I also have this for `isOrthogonal`, `isUpperTriangular` etc, but `isSquare` is the only one that is used by the other things I'm including here.
* Well, not quite. I also included changes to `isOne` here since the next item is about identity matrices.
* The identity Matrix constructor will now produce a degree n identity Matrix of the appropriate dimensions. What does it mean to be an identity Matrix of, say, degree 4? It means your dimensions are i by j by k by k and the matrices you get by looking at the frontal slice are kxk identity matrices.
* Multiplication between two degree n Matrix objects makes sense if one is (whatever by i by j) and the other is (sameWhatever by j by k). The result is (sameWhatever by i by k). As should happen, for any Matrix A, if you generate `I = A->I`, then `I*A == A`.
* Powers of square Matrix objects now can happen, regardless of degree. I also replaced the power algorithm with one based on repeated squaring. It might make an efficiency improvement for a matrix raised to a high power.
* Inverses of square Matrix objects now can happen, regardless of degree. Provided all the matrices in the frontal slice are invertible. As should happen, `A*A->inverse == A->I`.

I'm attaching a file ([matrixTesting.txt](https://github.com/user-attachments/files/19411785/matrixTesting.txt)) you can use to begin testing.


Alex-Jordan:
One more perhaps controversial change is here. Before this commit, a `Matrix*Vector` or `Matrix*Point` would produce a `Matrix` object. Now the result will be a `Vector` or `Point`. Similarly, a degree2 `Matrix` times a degree-1 `Matrix` would produce a degree-2 `Matrix`. In the spirit of viewing a degree-1 `Matrix` as a row vector, this commit changes things so the result will be a row vector two (that is, a degree-1 `Matrix`).

dpvc:
Alex, I'm glad to see you taking on this project.  The Matrix MathObject was always a bit of a mess, so it is good to see you straightening out some of the kinks.  It sounds like your approach is a good one.  I'm not going to go through it all myself, but trust you and the others to handle this.

A couple of things to think about if you haven't already.  There may be some implications for the Parser's Matrix object (pg/lib/Parser/List/Matrix.pm) or the main List object of which it is a subclass (pg/lib/Parser/List.pm), though I don't think so.  The latter is also something that is messier than I would like, due to the fact that the original library wasn't nearly as extensible as it is now, and so it really was made only to handle the original Point, Vector, Matrix, and List objects (something to straighten out in the future).  The Matrix answer checker (pg/lib/Value/AnswerChecker.pm) should also be checked to make sure nothing needs modification for these changes, though again, probably nothing is needed.

One thing that has been requested in the past is to allow a Matrix of Formulas.  The current implementation turns that into a Formula returning a Matrix, and so it doesn't have the methods of a Matrix.  One of the reasons for this is that some of those methods rely on the values being constants (like the check for identity matrix or zero matrix), which can't be done in general for matrixes of formulas.  But some other ones would be useful in that setting.

I haven't worked with the Matrix code in quite a while, but since you have been getting your hands dirty with it, do you have a sense of how much of the code depends on the values being constants?  Would it make sense to handle a Matrix of Formulas and convert to a Formula of Matrices only when necessary?  Are there significant enough advantages to that to make it worth the effort?

Another alternative that might help avoid the necessity would be to add some reduction rules to the Parser's BOP objects that could combine non-constant matrices.  Rules for matrix addition, subtraction, multiplication and scalar multiplication might be enough to make working with matrix formulas more useful than they are now.

Alex-Jordan:
> There may be some implications for the Parser's Matrix object (pg/lib/Parser/List/Matrix.pm) or the main List object of which it is a subclass (pg/lib/Parser/List.pm)

I looked and I don't see anything obvious to me, but it is easy to imagine myself overlooking something. Where I've made changes here, there is a lot where the real action is still just handled the way it was always handled, and I added code in the surrounding layers to make it work for degree > 2. I've tested a lot, though never exhaustively. I reached the point where I think if there's something broken, it's only going to be some gotcha from some obscure situation. And I've tested with Real, Complex, and Fraction numbers in the entries.

> The Matrix answer checker

All seems to work here as it has in the past according to testing. The one thing that still doesn't exist is an `ans_array` for a Matrix with degree > 2. I didn't add that for now. I could work on it though. It just looks like it will be a bit of a slog to get that sorted out with names for the answer blanks. Doable, but tedious to work through correctly.

In "Matrix" context, without an answer array, these things can still be entered like `[[1,2],[3,4]],[[5,6],[7,8]]`.

> do you have a sense of how much of the code depends on the values being constants?

For example, with `isOne`, the crucial comparison (is a certain element 1 or 0) is made using `eq`. If the element entry could be a formula, then I suppose `eq` needs to become `==`?

With for example `isZero`, that appears to just reduce repeatedly until `isZero` is called from the individual elements. So if `isZero` is working on Formula objects, that would be enough. I note however that `Formula("x-x")->isZero` is returning false, so I'm unsure if `isZero` is working there as intended.

> Would it make sense to handle a Matrix of Formulas and convert to a Formula of Matrices only when necessary?  Are there significant enough advantages to that to make it worth the effort?

I think this would make sense and be useful in some linear algebra settings. To the degree that it's been requested repeatedly. I think it is a surprise when you set up something like `[[1, 'x'], [0, 1]]^2` and expect to see `[[1, '2x'], [0, 1]]` but you will just see `[[1, 'x'], [0, 1]]^2`. I should not be trusted to estimate how much work this would take though. 

Do you know if there is some switch I can make so that such things are left as Matrices of Formulas, but of course with all the methods broken on them? If I could toggle such a switch, then I could explore upgrading the methods one by one.

> Rules for matrix addition, subtraction, multiplication and scalar multiplication might be enough to make working with matrix formulas more useful than they are now.

And powers too? But one thing I could easily see wanting is a method to extract an element. Like that `2x` in the example above.

dpvc:
> Do you know if there is some switch I can make so that such things are left as Matrices of Formulas

These changes should allow you make a matrix of formulas.  

``` diff
diff --git a/lib/Value.pm b/lib/Value.pm
index 64b3d955..5e3545f2 100644
--- a/lib/Value.pm
+++ b/lib/Value.pm
@@ -1023,7 +1023,7 @@ sub string {
        foreach my $x (@{ $self->data }) {
                if (Value::isValue($x)) {
                        $x->{format} = $self->{format} if defined $self->{format};
-                       push(@coords, $x->string($equation));
+                       push(@coords, $x->string($x->isFormula ? () : $equation));
                } else {
                        push(@coords, $x);
                }
@@ -1068,7 +1068,7 @@ sub TeX {
        foreach my $x (@{ $self->data }) {
                if (Value::isValue($x)) {
                        $x->{format} = $self->{format} if defined $self->{format};
-                       push(@coords, $x->TeX($equation));
+                       push(@coords, $x->TeX($x->isFormula ? () : $equation));
                } elsif (defined($str->{$x}) && $str->{$x}{TeX}) {
                        push(@coords, $str->{$x}{TeX});
                } else {
@@ -1092,7 +1092,7 @@ sub perl {
        my $perl;
        my @p = ();
        foreach my $x (@{ $self->data }) {
-               if (Value::isValue($x)) { push(@p, $x->perl(0, $mtype)) }
+               if (Value::isValue($x)) { push(@p, $x->perl($x->isFormula ? () : (0, $mtype))) }
                else                    { push(@p, $x) }
        }
        @p = ("'" . $self->{open} . "'", @p, "'" . $self->{close} . "'") if $self->classMatch('Interval');
diff --git a/lib/Value/Matrix.pm b/lib/Value/Matrix.pm
index 74e1b9f5..842fc3e1 100644
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -239,7 +239,7 @@ sub matrixMatrix {    #internal
                        unless (defined($x->entryType) && $type eq $x->entryType->{name});
                Value::Error("Matrix rows must all be the same length") unless ($len eq $x->length);
        }
-       return $self->formula([@M]) if $isFormula;
+#      return $self->formula([@M]) if $isFormula;
        bless { data => [@M], context => $context }, $class;
 }
 
@@ -259,7 +259,7 @@ sub numberMatrix {    #internal
                push(@M, $x);
                $isFormula = 1 if Value::isFormula($x);
        }
-       return $self->formula([@M]) if $isFormula;
+#      return $self->formula([@M]) if $isFormula;
        bless { data => [@M], context => $context }, $class;
 }
 
@@ -1273,7 +1273,7 @@ sub TeX {
                for my $x (@{ $self->data }) {
                        if (Value::isValue($x)) {
                                $x->{format} = $self->{format} if defined $self->{format};
-                               push(@entries, $x->TeX($equation));
+                               push(@entries, $x->TeX($x->isFormula ? () : $equation));
                        } else {
                                push(@entries, $x);
                        }
@@ -1285,7 +1285,7 @@ sub TeX {
                        for my $x (@{ $row->data }) {
                                if (Value::isValue($x)) {
                                        $x->{format} = $self->{format} if defined $self->{format};
-                                       push(@entries, $x->TeX($equation));
+                                       push(@entries, $x->TeX($x->isFormula ? () :$equation));
                                } else {
                                        push(@entries, $x);
                                }
```

There may be more changes that are needed.  With this, you can do operations on the results, but I haven't tested most of the methods, and one probably needs to use `reduce()` on the entries to make meaningful results (one might want to add a `reduce()` method to the Matrix object for that).

I'll reply to your other comments later -- out of time at the moment.

pstaabp:
I commented on PR #1217, but should have gone here.  

The testing should be formal unit testing instead of a problem.   If you like me to set this up, I can do that.

drgrice1 (approved):
This looks good.

I tested that things generally work and don't break anything that I do with Matrices.  That is still not extensive testing though.

I comment on a few minor details.

lib/Value/Matrix.pm
```diff
@@ -327,24 +336,31 @@ sub typeRef {
...
 =cut
 
 sub isSquare {
 	my $self = shift;
 	my @d    = $self->dimensions;
-	return 1 if scalar(@d) == 1 && $d[0] == 1;
-	return 0 if scalar(@d) != 2;
-	return $d[0] == $d[1];
+	if (scalar(@d) == 1) { return $d[0] == 1 }
+	return ($d[-1] == $d[-2]);
```
drgrice1, line 363: I just noticed this (while looking at #1216).  I would change this to
```suggestion
	return $d[0]  == 1 if scalar(@d) == 1;
	return $d[-1] == $d[-2];
```
The first line is consistent with how it was before and how conditionals like this are generally done in this file.
The parentheses are just unnecessary on the second line.

Alex-Jordan:
This now has unit tests that @pstaabp added. And I addressed the items that @drgrice1 noticed.

There are at least two things in this thread that I plan to work on, but not as part of this PR: making an answer array for a degree n Matrix with n > 2, and letting Matrices of Formula objects be possible. I just want to clear up not to wait on this for those things.

drgrice1:
Looks good.

I know this is Peter's work, but I would suggest consistency in the test names.  Most of the test names are full sentences with capitalized first word and a period at the end.  However, a few are not.  The worst is `extract an element from a 2D matrix.`  There is a period at the end, but the first word is not capitalized.

drgrice1:
That would be fine.

Alex-Jordan:
@dpvc :  Peter noticed something. The documentation here in `Value/Matrix.pm` says that for a matrix `$M`, then `$M->value` should be either an array of perl reals or an array of (possibly nested) array refs containing perl reals. And that would be consistent with what I think `value` does with other MathObjects.

However testing back as far (at least) as 2.17, that's not what you get. `$M->value` comes back as the same sort of thing, but with MathObject Reals, not perl reals.

Should this be changed to do as the documentation says?

drgrice1:
I think that the documentation should be changed to say what it really does.  I think that it would be inconsistent, as you say, to change the behavior to do what the documentation says.  For a point `$point`, if you call `$point->value` you get an array of MathObject Reals.

@dpvc can weigh in if there is a reason to change this behavior though.

Alex-Jordan:
Oh, I was under the impression that you get perl reals in other cases. Is
that what happens with a Fraction? (I'm not in a position to check.) I most
often use value on a Fraction.

drgrice1:
That is not what happens for a `Fraction`.  If `$fraction` is a `Fraction`, then the elements in the array returned by the `value` method are Perl numbers.

dpvc:
> $M->value comes back as the same sort of thing, but with MathObject Reals, not perl reals.
> 
> Should this be changed to do as the documentation says?

No, according to the blame listing, the documentation was added 8 years ago by Mike Gage, and was incorrect at the time, and has been maintained that way ever since.  The `value` method has always returned MathObject Reals.

The main idea of the `value` method is to obtain the data needed to recreate the given object.  So, for example, an `Interval` returns not just the left and right endpoints, but also the parentheses or brackets that determine whether those ends are open or closed.

Usually, this is just the array of the data stored in the `data` property of the MathObject.  For a `Point` or `Vector`, that is an array of MathObject Reals (not perl reals).  For a `Real`, that data is the corresponding perl real, and for a `Fraction`, it is a pair of perl reals (not MathObject Reals), since that is what is stored in its `data` array.

For a `Matrix`, things are a bit more complicated, rather than returning an array of Matrices of one dimension lower (which is what the `data` array holds), those rows are turned into perl arrays.  But the lowest level entries are MathObject Reals.

One should be able to recreate the object (minus any changed properties) from the what the `value` returns.  For example,

``` perl
Interval(Compute("(1,2]")->value);
```

should produce another interval `(1,2]`.  Similarly,

``` perl
Matrix(Matrix([1,2], [3,4])->value)
```

would produce a the `[[1,2],[3,4]]` matrix again.  That *could* have been done with Perl reals as well, but that is never what has happened.  It probably wouldn't hurt if you changed it, but I am not certain of that (this is all over 20 years old now, and I don't remember all the details any more).

> I suppose a `Complex` is not a basic numerical type then.

Well, there isn't really a sense of "basic" types in MathObjects.  The real and imaginary parts are MathObject Reals, so that's what `value` returns.  There are a number of places where that is used, so it would not be a good idea to change that.

The upshot is, the documentation should be changed, not the `value` method.

Alex-Jordan:
Thanks @dpvc , I made a light edit to the POD so that it says what actually happens now.

@drgrice1 I added some tests here and went ahead and tried my hand at re-titling tests.

For reviewers, I think this tier is back to being ready for review.

drgrice1:
This still looks good to me.  I think you made the test titles more accurate, but didn't really fix the mix of casing and punctuation.  That is fine though, at least for this pull request.

pstaabp (approved):
I think this looks good.  

Before all of these go in, we should remove the Data::Dumper code.  I should have done that before sending to you @Alex-Jordan.

Alex-Jordan:
> One more perhaps controversial change is here. Before this commit, a Matrix*Vector or Matrix*Point would produce a Matrix object. Now the result will be a Vector or Point. Similarly, a degree2 Matrix times a degree-1 Matrix would produce a degree-2 Matrix. In the spirit of viewing a degree-1 Matrix as a row vector, this commit changes things so the result will be a row vector two (that is, a degree-1 Matrix).

This has led to an issue. Before, if `$M` and `$b` are degree-2 Matrices and `$v` is a Vector, then `$M*$v == $b` would be an acceptable comparison. Because both `$b$ and `$M*$v` would be degree-2 Matrices. But now `$M*$v` is a Vector, and the comparison `$M*$v == $b` promotes `$M*$v` to a degree-1 Matrix. And then there is an error that Matrices of different degrees cannot be compared.

What I propose is to adjust the `compare` subroutine for matrices so that if either the left or right operand is a degree 1 Matrix, it is promoted to degree 2 via a `->transpose` before the comparison happens.

With an explicit example, before this made sense: `[[1, 2], [3,4]] * <1,2> == [[5],[11]]` because it was the same as `[[5],[11]] == [[5],[11]]`. But now that reduces to `<5,11> == [[5],[11]]` which is effectively altered to `[5,11] == [[5],[11]]`. And I am proposing that any comparison like `[5,11] == [[5],[11]]` be effectively altered to `[[5],[11]] == [[5],[11]]`, to restore the behavior of some comparisons that are coded in some problems.

PR #1216, opened 2025-03-24
update several matrix methods to apply to degree n matrices

Alex-Jordan:
This builds on #1215. If #1215 is updated following feedback, I will update here too. Once #1215 is merged, this diff should be more manageable.

In addition to changes from #1215, this has:

- some more POD and code comments updated to reflect how methods can be applied to degree n matrices, as well as specify in some places where the method only applies for degree 2 matrices
- make the following methods apply to degree n matrices: `isUpperTriangular`, `isLowerTriangular`, `isDiagonal`, `isSymmetric`, `isOrthogonal`, `isREF`, `isRREF`. For a degree 1 Matrix, it is interpreted as a 1xk degree 2 Matrix. For degree > 2, these features can apply to all of the degree 2 matrices in the frontal slice.

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -103,17 +103,17 @@ Examples:
 
 =head3 Access values
 
-    row(i) : MathObjectMatrix
-        For a 1D tensor, produces a 1D tensor
-        For nD tensor with n > 1, produces a (n-1)D tensor
+    row(i) : MathObject Matrix
+        For a degree 1 Matrix, produces a degree 1 Matrix
+        For a degree n Matrix with n > 1, produces a degree (n-1) Matrix
 
-    column(j) : MathObjectMatrix or Real or Complex
-        For a 1D tensor, produces a Real or Complex
-        For nD tensor with n > 1, produces an nD tensor where 2nd dimension is 1
+    column(j) : MathObject Matrix or Real or Complex
+        For a degree 1 Matrix, produces a Real or Complex
+        For a degree n Matrix with n > 1, produces a degree n Matrix where the 2nd dimesion is length 1
 
-    element : Real or Complex value when passed the same number of arguments as the dimension of the tensor.
-        If passed more than n arguments, null. If the dimension of the tensor is n and element is passed
-	k arguments with k < n, then this produces the corresponding (n-k)-dimensional tensor.
+    element : Real/Complex/Fraction value when passed the same number of arguments as the degree of the Matrix.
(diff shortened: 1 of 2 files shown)


lib/Value/Matrix.pm
```diff
@@ -416,58 +451,105 @@ sub isZero {
...
+		for my $i (2 .. $d[0]) {
+			for my $j (1 .. ($i - 1 < $d[1] ? $i - 1 : $d[1])) {
+				return 0 unless $self->element($i, $j) == 0;
+			}
+		}
+	} else {
+		for my $row (@{ $self->{data} }) {
+			if (!$row->isUpperTriangular) {
+				return 0;
+			}
```
drgrice1, line 474: I would make this
```suggestion
			return 0 if !$row->isUpperTriangular;
```
Just a suggestion.  I like the more condensed form, and this is consistent with how these sort of checks are done elsewhere in the file.

lib/Value/Matrix.pm
```diff
@@ -416,58 +451,105 @@ sub isZero {
...
+		for my $i (1 .. $d[0]) {
+			for my $j ($i + 1 .. $d[1]) {
+				return 0 unless $self->element($i, $j) == 0;
+			}
+		}
+	} else {
+		for my $row (@{ $self->{data} }) {
+			if (!$row->isLowerTriangular) {
+				return 0;
+			}
```
drgrice1, line 503: Same comment here as above.  I see this also applies to several places below as well.

lib/Value/Matrix.pm
```diff
@@ -479,50 +561,77 @@ sub isOrthogonal {
...
+		for my $row (@{ $self->{data} }) {
+			if (!$row->isREF) {
+				return 0;
 			}
 		}
 	}
 	return 1;
 }
 
+=head3 C<isREF>
```
drgrice1, line 598: This should be
```suggestion
=head3 C<isRREF>
```

drgrice1 (approved):
I think this generally looks good.  Again some minor code comments.

I tested this by adding some columns to the test problem you gave in #1215, and things are working as expected.

lib/Value/Matrix.pm
drgrice1: `unless` is certainly preferred.  Several of us have expressed preference for `statement unless condition` to `statement if !condition`.  If the condition is more complicated though, I prefer using `statement if complicated condition`.  `unless` can be confusing when you have a lot of conditions some of which are negated and others not with `&&`s and `||`s mixed in.

Alex-Jordan:
If we merge this one I'll rebase #1217.

PR #1217, opened 2025-03-24
transpose degree n matrices

Alex-Jordan:
This builds on #1216 (which builds on #1215). If #1216 is updated following feedback, I will update here too. Once #1216 is merged, this diff should be more manageable.

In addition to changes from #1216, this has:

* a `slice` method for a Matrix object. This produces a degree (n-1) Matrix from a degree n Matrix. You specify an index and a value for that index, and it produces the degree (n-1) Matrix that is thereby defined. This is currently only used as a helper for the update to `transpose`. It is similar to `row` and `column` methods, but more general. And always produces something of the next lower degree.
* to transpose a degree n Matrix, the word "transpose" is perhaps a misnomer. Instead you have some permutation of the index dimensions in mind. For the usual transpose of a degree 2 matrix, this is just the permutation (12). For a degree n matrix, you might want to permute the indices in a different way. Doing this once can make it more efficient to repeatedly access data in the Matrix, depending on the scenario. So now you can transpose a degree n Matrix. If you don't specify a permutation, the default is to just swap the last two indices. This will look like a regular transposition on each of the degree 2 matrices in the frontal slice. Otherwise, for the permutation you may specify a cycle in the form of an array reference, or a product of cycles in the form of an array reference of array references.

This is the last of the updates I wanted to do before adding new methods for setting elements, removing columns and rows and a submatrix that @pstaabp has in #1076. We should make sure new tools work with degree n Matrix objects as we add them. Unless it is for sure only appropriate for degree 2 matrices.

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -837,29 +837,155 @@ sub twiddle {
 	return $self->make(@coords);
 }
 
-=head3 C<transpose>
+=head3 C<slice>
 
-Take the transpose of a matrix.
+Apply this to a degree n Matrix, passing (m, k), and produce the degree (n-1) Matrix defined by
+taking all entries whose position has mth index with value k. For example if C<$M> is a 4x5x6
+Matrix, then m can be 1, 2, or 3. If m is 2, then k can be 1, 2, 3, 4, or  5. C<$M-<gt>slice(2,3)>
+is the 4x6 Matrix such that the entry at position (i,j) is the entry of C<$M> at position (i,3,j).
 
 Usage:
 
     $A = Matrix([ [ 1, 2, 3, 4 ], [ 5, 6, 7, 8 ], [ 9, 10, 11, 12 ] ]);
-    $A->transpose;
+    $A->slice(1, 2)
+    # Index 1 identifies the 1st, 2nd, or 3rd row above, and with value 2 we get the second one:
+    # Matrix([5, 6, 7, 8])
+
+    $B = Matrix([ [ [ 1, 2 ], [ 3, 4 ] ], [ [ 5, 6 ], [ 7, 8 ] ] ]);
+    $B->slice(1, 2)
+    # Index 1 identifies the two arrays at depth 1, and with value 2 we get the second one:
+    # Matrix([ [ 5, 6 ], [ 7, 8 ] ])
+    $B->slice(2, 1)
+    # Here we take all entries from $B where the 2nd index is 1: the 1 at position (1,1,1),
+    # the 2 at position (1,1,2), the 5 at position (2,1,1), and the 6 at position (2,1,2):
+    # Matrix([ [ 1, 2 ], [ 5, 6 ] ])
+    $B->slice(3, 1)
+    # Here we take all entries from $B where the 3rd index is 1: the 1 at position (1,1,1),
+    # the 3 at position (1,2,1), the 5 at position (2,1,1), and the 7 at position (2,2,1):
+    # Matrix([ [ 1, 3 ], [ 5, 7 ] ])
 
 =cut
 
+sub slice {
+	my ($self, $index, $value) = @_;
+	my @d = $self->dimensions;
+	my $d = scalar(@d);
+	my $w = $d[0];
+	Value::Error("index must be an integer from 1 to $d") unless ($index == int($index) && $index >= 1 && $index <= $d);
+	my $M = $self->data;
+	if ($index == 1) {
+		Value::Error("value must be an integer from 1 to $w")
+			unless ($value == int($value) && $value >= 1 && $value <= $w);
+		return $M->[ $value - 1 ];
+	} else {
+		my @rows;
+		for (1 .. $w) {
+			push @rows, $M->[ $_ - 1 ]->slice($index - 1, $value);
+		}
+		return $self->make(@rows);
+	}
+}
+
+=head3 C<transpose>
+
+Take the transpose of a matrix. For a degree 1 Matrix, first promote to a degree 2 Matrix.
+For a degree n Matrix, apply a permutation of the indices. The default permutation transposes the
+last two indices. To specify a permutation, provide an array reference representing a cycle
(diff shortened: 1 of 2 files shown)


pstaabp:
Looks good, but we should bring the test suite in soon.  There was one in my branch that probably works for a lot of this, but should be update. 

Although I'm guessing you have a number of problems that you've been testing with, it's more efficient to have the test suite.

drgrice1 (commented):
I haven't tested this at all yet, but I see some code changes that are needed.  Some things that perlcritic pointed out.

lib/Value/Matrix.pm
```diff
@@ -689,67 +848,177 @@ sub twiddle {
...
+	my @d = $self->dimensions;
+	my $d = scalar(@d);
+	my $w = $d[0];
+	Value::Error("index must be an integer from 1 to $d") unless ($index == int($index) && $index >= 1 && $index <= $d);
+	my $M = $self->data;
+	if ($index == 1) {
+		Value::Error("value must be an integer from 1 to $w")
+			unless ($value == int($value) && $value >= 1 && $value <= $w);
+		return $M->[ $value - 1 ];
+		return $self->make($M->[ $value - 1 ]);
```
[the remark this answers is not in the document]
Alex-Jordan: I'm not sure what happened here during my editing. But the first line should be good. Certainly while I was testing, that's the line that was used!

lib/Value/Matrix.pm
```diff
@@ -837,29 +837,117 @@ sub twiddle {
...
 
     $A = Matrix([ [ 1, 2, 3, 4 ], [ 5, 6, 7, 8 ], [ 9, 10, 11, 12 ] ]);
-    $A->transpose;
+    $A->transpose              # will be [ [ 1, 5, 9 ], [ 2, 6, 10 ], [ 3, 7, 11 ], [ 4, 8, 12 ] ]
 
+    $B = Matrix([ [ [ 1, 2 ], [ 3, 4 ] ], [ [ 5, 6 ], [ 7, 8 ] ] ]);
+    $B->transpose([1, 2, 3])   # will be [ [ [ 1, 3 ], [ 5, 7 ] ], [ [2 , 4 ], [ 6, 8 ] ] ]
+
+    $C = Matrix([ [ [ [ 1, 2 ], [ 3, 4 ] ], [ [ 5, 6 ], [ 7, 8 ] ] ], [ [ [ 9, A ], [ B, C ] ], [ [ D, E ], [ F, 0 ] ] ] ]);
+    $C->transpose([ [ 1, 2], [3, 4] ])   # will be [ [ [ [ 1, 3 ], [ 2, 4 ] ], [ [ 9, B ], [ A, C ] ] ], [ [ [ 5, 7 ], [ 6, 8 ] ], [ [ D, F ], [ E, 0 ] ] ] ]
```
drgrice1, line 895: Please change this to:
```suggestion
    $C = Matrix([
        [ [ [ 1, 2 ],  [ 3,  4 ] ],  [ [ 5,  6 ],  [ 7,  8 ] ] ],
        [ [ [ 9, 10 ], [ 11, 12 ] ], [ [ 13, 14 ], [ 15, 16 ] ] ]
    ]);
    $C->transpose([ [ 1, 2], [3, 4] ])
    # will be
    # [
    #     [ [ [ 1, 3 ], [ 2, 4 ] ], [ [ 9,  11 ], [ 10, 12 ] ] ],
    #     [ [ [ 5, 7 ], [ 6, 8 ] ], [ [ 13, 15 ], [ 14, 16 ] ] ]
    # ];
```

That is much more readable and the long lines will easily go out of the window in many cases which requires scrolling to see.  That makes it even harder to read in that case.

Also `A, B, C, ...` shouldn't be used.  Someone might cut and paste this into a problem thinking it should work, and it won't.

I think the examples above this could also be reformatted better, but this one was extremely bad. as it even exceeds what I consider a hard limit of a maximum 120 characters.

lib/Value/Matrix.pm
```diff
@@ -837,29 +837,117 @@ sub twiddle {
 	return $self->make(@coords);
 }
 
+=head3 C<slice>
+
+Produce the degree (n-1) Matrix defined by a given index and value for that index. If n is 1,
+this produces a Real/Complex/Fraction.
```
drgrice1, line 843: I think this could use some clarification as to what this really means.  What do `index` and `value` mean in this context?

Alex-Jordan: I tried to do better here. I will push and you can see. Imagine a 5x7x9 matrix. Entry positions are defined by three indices i,j,k. The first index, second index, and third index. So "index" here would be 1, 2, or 3. Say it is 3. Then a value for that index can be 1, 2, ..., or 9. Take a look and see if what I changed to is OK or can be improved upon.

drgrice1: I think what you changed it to is better.  What you said here really is what really worked for me though.  Perhaps adding an example like that to the POD would better.  Terminology with these "tensor" things can get confusing!

lib/Value/Matrix.pm
```diff
@@ -837,29 +837,155 @@ sub twiddle {
 	return $self->make(@coords);
 }
 
-=head3 C<transpose>
+=head3 C<slice>
 
-Take the transpose of a matrix.
+Apply this to a degree n Matrix, passing (m, k), and produce the degree (n-1) Matrix defined by
+taking all entries whose position has mth index with value k. For example if C<$M> is a 4x5x6
+Matrix, then m can be 1, 2, or 3. If m is 2, then k can be 1, 2, 3, 4, or  5. C<$M-<gt>slice(2,3)>
```
pstaabp, line 844: The code at the end of this can be made with C<<$M->slice(2,3)>>, a bit easier than the <gt> notation.

drgrice1: That actually needs to be `C<< $M->slice(2, 3) >>`.  There must be spaces after `<<` and before `>>`.

pstaabp: Good point.

pstaabp (approved):

PR #1218, opened 2025-03-25
Graphtool graph math object

drgrice1:
Switch to a better extension system for the GraphTool.

For JavaScript instead of piecing together the parts of the class definition, use a function that returns the class definition directly.  This has the advantage of being able to use `super` directly to access methods of the parent class, and is just much easier to work with.

All of the graph objects and graph tools are now in separate files (except the SelectTool and SolidDashTool which are core functionality).

For Perl, all graph objects are packages that derive from the `GraphTool::GraphObject` package instead of the previous method of defining cmp and tikz subroutines in a hash. A mapping from the string identifier to the package is maintained and used to construct the correct perl object for a graph tool object. There is a compatibility layer for objects added the old way.

Furthermore this take the Perl package approach further and the `GraphTool::GraphObject` package derives from the `Value::List` package. This means they are MathObjects.  This makes things much simpler when writing a custom checker.  The elements of the correct and student answer lists in the list_checker are all objects that derive from a `GraphTool::GraphObject` (which derives from a Value::List). See the new POD on CUSTOM CHECKERS for details.

### DIFF
--- a/htdocs/js/GraphTool/circletool.js
+++ b/htdocs/js/GraphTool/circletool.js
@@ -1,0 +1,205 @@
+/* global graphTool, JXG */
+
+'use strict';
+
+(() => {
+	if (graphTool && graphTool.circleTool) return;
+
+	graphTool.circleTool = {
+		Circle(gt) {
+			return class Circle extends gt.GraphObject {
+				static strId = 'circle';
+
+				constructor(center, point, solid) {
+					super(
+						gt.board.create('circle', [center, point], {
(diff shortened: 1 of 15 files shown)


pstaabp (approved):
Tested all of the graph tool problems in the Sample problem set. 

Also, we should update the custom checker graph tool sample problem, since this simplifies the checking.

drgrice1:
Yeah, I will add that to this pull request.

PR #1219, opened 2025-03-25
Fix a vulnerabilities with the `PGcore::insertGraph` and `surePathToTmpFile` methods.

drgrice1:
This vulnerability makes it possible for a problem to create a file anywhere that the server has write permission, overwriting a file if it already exists. I have confirmed this vulnerability for as far back as WeBWorK 2.17, but it most likely has been there since 2010 when the `insertGraph` method was added.

To utilize the vulnerability one needs to create a custom package in a problem that has the `imageName`, `ext`, and `draw` methods, and make the path that the `insertGraph` method uses be what one wants. There are actually several ways to do that.

One way is to use multiple `../`s in the file path to get out of the temporary directory specified in the PGcore object and then the `surePathToTmpFile` method will create any needed directories and write to whatever path it resolves to.

Another is to change the temporary directory specified in the `$PG` PGcore object. That can either be done by directly changing the `$PG->{tempDirectory}` value or by changing the `$envir{tempDirectory}` value before the PGcore object is constructed for the problem.  Note that the PGcore object is constructed in the `DOCUMENT` call at the beginning of the problem.  This means that the environment can be modified in the problem before that.

Yet another is to replace the `$PG` object entirely with one defined in the problem.  If the object derives from the `PGcore` package it can still use its methods, but change what the `surePathToTmpFile` method returns to be whatever path is desired. Then the original `insertGraph` method can be used to write wherever one desires. As long as the directories leading to the file exist the file  will be written (or overwritten).

And still another is to "monkey patch" the `surePathToTmpFile` method to return whatever path is desired. Again as long as the directories leading to the file exist the file will be written (or overwritten).

To fix this there are a couple of changes needed.  First the `$pg_envir` object in the `WeBWorK::PG::IO` package needs to be more than it currently is.  Currently it is only a generic environment, and not specific to the current problem being rendered.  I.e., it doesn't have the html temporary directory for the current user and course. This is fixed by using a package variable in the `WeBWorK::PG::IO` package that is set when by the `WeBWorK::PG` object is constructed.  Since neither the `WeBWorK::PG::IO` package nor the `$pg_envir` object (now really the `$WeBWorK::PG::IO::pg_envir` object) are shared to the safe compartment, there is no way for a problem to access these things and change anything in them.

Note that since a package variable is set, this will only work if problems are compartmentalized into separate processes.  If problems in two courses were to run in the same process, then the second one to run would set the package variable to a WeBWorK::PG::Environment object with the wrong information for the first one, and that would mean that files could be potentially saved into the wrong location (the temporary directory for the other course).  Fortunately, all of the ways that PG problems are rendered now use a Mojolicious subprocess, and so problems are compartmentalized as needed.  That is with the exception of the unit tests or someone calling PG directly, but that is fine for those uses. The security isn't needed for those.

One upshot of this is that the `$pg_envir` object in the `WeBWorK::PG::IO` module is now a fully valid environment for the course instead of the generic environment for the server that it was before. This means that more information from it can be used in that module. For instance the `permitted_read_dir` that restricts which directory a problem can read from (usually for reading another problem file) can now be set to the course root directory instead of the entire webwork2 courses directory.

The next thing that is needed is that the `insertGraph` and `surePathToTmpFile` methods need to use more care when creating directories or writing files.  To this end, a `WeBWorK::PG::IO::writeDataToFile` method is added that handles writing files from data to the disk, and all of the methods in the `WeBWorK::PG::IO` package that write to disk now check that the requsted location is inside the html temporary directory for the problem (and user and course) currently being rendered. Note that the `createFile` method has this protection added, but that method is not actually used at this point.

This fixes the first of three ways that I know of that a problem can write to anywhere that the server has write permission.


drgrice1:
Attached are some problems you can use to test this.  With the develop branch (or any version of webwork since 2010), these problems will write files into the webwork2 courses directory.  With minor tweaks to these problems one could use this to change the course.conf file for their course even without admin privileges.  They could also change the course.conf file for other courses.  Or overwrite any other file the server has write permission for.
[VulnerabilityExploitProblems.zip](https://github.com/user-attachments/files/19457776/VulnerabilityExploitProblems.zip)

pstaabp (approved):
This looks good and fixes all of the posted vulnerabilities.

somiaj:
Is it worth making hotfixes for these vulns you are fixing?

drgrice1:
Since these vulnerabilities have existed for more than 10 years, I don't think that is particularly necessary.  These are not front facing vulnerabilities.  That is, a hacker can not directly exploit them.  Only instructors (or those that can author problems) can do so.

PR #1223, opened 2025-03-29
Improvements to the debugging `show` method of PGML.

drgrice1:
This was something that was useful for me while I was working on writing the CodeMirror 6 PGML parser.  It just formats the things in the PGML stack that are hashes in a nicer way, actually showing the contents instead of something like `HASH(0x5d3919403d78)` as it previously did.

The script in the attached zip can be used to test this: 
[pgml-parse.zip](https://github.com/user-attachments/files/19514375/pgml-parse.zip)

The script requires that the `PG_ROOT` environment variable be set to the PG root directory.  Execute the script by running `./pgml-parse.pl -f path/to/problem.pg`.  The script works with any version of PG that has PGML, but you can see the difference in the output.  The `combine` entries (which almost any problem will have) will be output as `HASH(0x...)` with the develop branch, and will show the hash in a "prettier" way with this pull request.

This does not affect normal PGML usage in problems at all.

Just cleaning up local branches.

### DIFF
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -966,6 +966,21 @@ sub new {
 	bless { type => $type, %$fields }, $class;
 }
 
+sub stringifyHash {
+	my ($self, $hash) = @_;
+	return '{ ' . join(
+		', ',
+		map {
+			"$_ => "
+				. (
+					ref($hash->{$_}) eq 'HASH'
+					? $self->stringifyHash($hash->{$_})
+					: ("'" . $self->quote($hash->{$_}) . "'"))
+		} PGML::Sort(keys %$hash)
+	) . ' }';
+
+}
+
 sub show {
 	my $self    = shift;
 	my $indent  = shift || "";
@@ -974,7 +989,9 @@ sub show {
 		next if $id eq "stack";
 		if (ref($self->{$id}) eq 'ARRAY') {
 			push(@strings,
-				$indent . $id . ": [" . join(',', map { "'" . $self->quote($_) . "'" } @{ $self->{$id} }) . "]");
+				$indent . $id . ": [" . join(', ', map { "'" . $self->quote($_) . "'" } @{ $self->{$id} }) . "]");
+		} elsif (ref($self->{$id}) eq 'HASH') {
+			push(@strings, $indent . $id . ": " . $self->stringifyHash($self->{$id}));
 		} else {
 			push(@strings, $indent . $id . ": '" . $self->quote($self->{$id}) . "'");
 		}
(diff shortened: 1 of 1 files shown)


pstaabp:
A couple of things.  First, I'm getting some uninitialized string errors when running this on problems.  Doesn't see to affect the output. 

Also, would it be useful to put the `pgml-parse.pl` script in the `bin` or make a `bin/dev` directory for this?

drgrice1:
I am not seeing any warnings when using the script.  Do you have the `PG_ROOT` environment variable set?  If so, the warnings are probably actually coming from the problem you are testing.  You are testing with a problem that has `BEGIN_PGML/END_PGML` in it right?  You probably will get uninitialized string warnings from the script if not.

The script is not really that versatile and was rather quickly thrown together.  I don't know if it is worth adding to the repository.

pstaabp:
I do have `PG_ROOT` defined. 

I'm seeing `Use of uninitialized value in concatenation (.) or string at /opt/webwork/pg/lib/PGalias.pm line 34.` for every problem.  Maybe the pg_environment is not loading.  

Other errors that I'm seeing are some of my problems like: `/Contrib/Fitchburg/Algebra/intervals/interval_notation.pg` and I'm getting the following:
```
Use of uninitialized value $string in substitution (s///) at (eval 233) line 1005.
Use of uninitialized value $string in substitution (s///) at (eval 233) line 1006.
Use of uninitialized value in concatenation (.) or string at (eval 233) line 996.
Use of uninitialized value $string in substitution (s///) at (eval 233) line 1005.
Use of uninitialized value $string in substitution (s///) at (eval 233) line 1006.
Use of uninitialized value in concatenation (.) or string at (eval 233) line 996.
Use of uninitialized value $string in substitution (s///) at (eval 233) line 1005.
Use of uninitialized value $string in substitution (s///) at (eval 233) line 1006.
Use of uninitialized value in concatenation (.) or string at (eval 233) line 996.
```

drgrice1:
The PGalias.pm warnings are because of a change in the code with the security vulnerability fixes in #1219 that the script doesn't account for.

The other warnings are actually expected.  What is happening is that the `PG_restricted_eval` is encountering variables that are not defined because the script doesn't do a full render of the problem.  It only parses the PGML section.  For my tests that I worked with for development of the PG CodeMirror 6 editor I actually implemented special handling for this to define the needed variables.

Here is a hacked up version of the script that will remove the warnings, although will be less accurate for the actual result.
[pgml-parse.zip](https://github.com/user-attachments/files/19930554/pgml-parse.zip)

Note that this script really is not the point of this pull request.  The PGML show method has been there pretty much since PGML was created and this is just a minor tweak on its output.  A script that utilizes it is something different entirely.  @dpvc probably has scripts that do so as well.

somiaj (approved):
Tested both the provided scripts and get JSON output, not really sure how to best parse the output, but it is there. The first script gave uninitialized warnings, and the second one didn't as stated.

PR #1224, opened 2025-04-04
Remove aria labels from radio buttons.

drgrice1:
This was requested by @glarose and the accessibility experts at the University of Michigan.  Having aria labels on these radio buttons that also have labels causes issues with screen readers.  It results in the aria label being read and not the actual label.

It seems to me that this is correct in any case.  An input should not have both a label and an aria label. Those clearly conflict.

Also improve the aria labels for the answers inside the radio buttons for a `RadioMultiAnswer`.  For this the `parserRadioMultiAnswer.pl` macro has its own generate_aria_label method for now as the `generate_aria_label` method in `PGbasicmacros.pl` does not work well for the `parserRadioMultiAnswer.pl` macro.

In addition add some missing `maketext` calls on aria labels, and take advantage that `maketext` substitutions now work in PG.

We need to find a better way to create aria labels in general.  The `generate_aria_label` method in `PGbasicmacros.pl` is neither versatile nor extensible. Attempting to construct meaningful labels from PG's answer name was never a good idea to begin with. By the way, since the aria label is currently constructed from the answer name, this is yet another reason to strongly discourage (actually flat out forbid) problem authors from using hard coded made up answer names in problems instead of calling `NEW_ANS_NAME` to obtain one.

### DIFF
--- a/lib/Value/AnswerChecker.pm
+++ b/lib/Value/AnswerChecker.pm
@@ -459,7 +459,7 @@ sub ans_matrix {
 		foreach my $j (0 .. $cols - 1) {
 			my $label;
 			if ($options{aria_label}) {
-				$label = $options{aria_label} . 'row ' . ($i + 1) . ' col ' . ($j + 1);
+				$label = $options{aria_label} . main::maketext('row [_1] col [_2] ', $i + 1, $j + 1);
 			} else {
 				$label = pgCall('generate_aria_label', ANS_NAME($ename, $i, $j));
 			}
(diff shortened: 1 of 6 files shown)


drgrice1:
All answers that come from `NEW_ANS_NAME` will have the `PGcore` `ANSWER_PREFIX` which is defined to be `AnSwEr`.  So at the very least, any answer that doesn't have that certainly didn't come from `NEW_ANS_NAME`.  The converse would not necessarily be true though.  Although all of the cases where answer names should be used but are and are using a hard-coded answer name that I have seen don't use something containing `AnSwEr`.

pstaabp:
I didn't test it.  I think having @glarose test may be the way to go. 

I see for radio buttons, that the aria-label is removed and seems that it is added for checkboxes though.

somiaj (approved):

somiaj (approved):

pstaabp (approved):

PR #1226, opened 2025-04-04
Switch from the `JSON` module to `Mojo::JSON`.

drgrice1:
This matches a change with WeBWorK and allows us to entirely drop the `JSON` dependency.  The `Mojo::JSON` package is part of the Mojolicious stack which is now required in any case.

This depends on https://github.com/openwebwork/webwork2/pull/2692.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -26,8 +26,6 @@ jobs:
             libencode-perl \
             libgd-perl \
             libhtml-parser-perl \
-            libjson-perl \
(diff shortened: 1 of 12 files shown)


Alex-Jordan (approved):

pstaabp:
I notice that in the `unit-tests.yml` (GitHub actions) file there is ` libjson-perl` and `libjson-xs-perl`.  There are similar files in the docker files here (and on the webwork2 side) that can be scrubbed as well.

pstaabp (approved):

PR #1227, opened 2025-04-11
upgrade degree-1 matrices to degree 2 when doing a comparison

Alex-Jordan:
This is to address an issue I raised in the thread for https://github.com/openwebwork/pg/pull/1215.

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -810,6 +810,8 @@ sub power {
 #  Do lexicographic comparison (row by row)
 sub compare {
 	my ($self, $l, $r) = Value::checkOpOrderWithPromote(@_);
+	$l = $l->transpose if ($l->degree == 1 && $r->degree == 2);
+	$r = $r->transpose if ($l->degree == 2 && $r->degree == 1);
 	Value::Error("Can't compare Matrices with different dimensions")
 		unless join(',', $l->dimensions) eq join(',', $r->dimensions);
 	my @l = @{ $l->data };


pstaabp:
I think it would be a good idea to add these cases to the test suite.

drgrice1:
Could you provide a PG example of this?  You gave an example mathematically in https://github.com/openwebwork/pg/pull/1215#issuecomment-2795586463, but I am having a hard time producing this in an actual PG problem.  For example, I tried the following

```pg
DOCUMENT();

$vector = Vector(1, 0, 0);

$matrix = Matrix([ [ 1, 0, 0 ], [ 0, 1, 0 ], [ 0, 0, 1 ] ]);

$result = $matrix * $vector;

$result2 = Matrix([ [ 1, 0, 0 ] ]);

BEGIN_PGML
[`[$vector]`] is a [@ ref($vector) @]

[`[$matrix]`] is a [@ ref($matrix) @]

[`[$result]`] is a [@ ref($result) @]

[`[$result2]`] is a [@ ref($result2) @]

[`[$matrix] \cdot [$vector] = [$result]`] is
[$matrix * $vector == $result ?  'true' : 'false']:

[`[$matrix] \cdot [$vector] = [$result2]`] is
[$matrix * $vector == $result2 ?  'true' : 'false']:
END_PGML

ENDDOCUMENT();
```

That problem renders the same for both the develop branch and this pull request, and the last comparison is an error in both cases.  So I must be missing something as to what this is doing.

Alex-Jordan:
Here is a MWE that works in 2.19 and works on this PR's branch. But on develop, gives an error:

```
DOCUMENT();

Context("Matrix");
$M = Matrix([1,2], [3,4]);
$s = Vector("<1,2>");
$Ms = $M*$s;
$b = Matrix(0,0)->transpose;

if ($Ms == $b || $Ms != $b) {
BEGIN_PGML
No error!

END_PGML
};

ENDDOCUMENT();
```

pstaabp (approved):
The MWE that @Alex-Jordan posted shows the error in develop and this fixes it.

somiaj (approved):
Tested the issue is fixed.

somiaj:
I also noticed that for the example problem given, that  #1223 also fixes the error (unsure if it just fixes the error vs allows the matrices to be accurately compared).

PR #1228, opened 2025-04-11
remove unnecessary line

Alex-Jordan:
This removes a line that clearly should not be here, given the line preceding it.

I do not recall what issue this causes, but it must have been something. This edit is in our production server, uncommitted, and I guess I made it there at some point while tracking down and fixing some issue I can no longer recall.


drgrice1 (approved):
Clearly both lines 83 and 84 aren't needed. I will assume that you know best which line that is.  I see that you added lines 79-83 to restore state for contextForm answers, and you probably meant to not leave line 84 in there at that time.

somiaj (approved):
I will also defer to @Alex-Jordan that the correct line was removed.

PR #1229, opened 2025-04-14
Fix a vulnerability in the Rserve code.

drgrice1:
This requires a change to the `Statistics::R::IO::Rserve` package. The vulnerability is directly in that package as its `get_file` method allows the caller to pass an arbitrary local file name, and it will retrieve the remove file and save it to that file as long as the serve has write permission to do so.
    
I have submitted a pull request to the GitHub repository for the `Statistics::R::IO` package that would work to prevent that, and also emailed the author directly but have received no response.  Furthermore, the package has not had any changes since 2017.  So at this point I think that we are going to need to consider the `Statistics::R::IO` package unmaintained and move on from that package.
    
As such this pull request adds PG's own implementation. It is pretty much the `Statistics::R::IO::Rserve` package, but all of the other things that aren't needed by WeBWorK and PG are removed.
    
Since this is part of PG the `get_file` method can save to the requested file name using the `WeBWorK::PG::IO::saveDataFile` method which refuses to save anything outside of the html temporary directory.
    
Note that this implementation does not use `Class::Tiny` or `Class::Tiny::Antler`, so those modules should be removed from the modules that are share to the safe compartment.  That is done in the `conf/pg_config.dist.yml` file, but a separate pull request will be needed to do this for webwork2.  It is not critical, the package is not dangerous to have shared.  Just not needed.

There are unit tests for both the `Rserve` package and the `RserveClient.pl` macro. Some of the tests require an R connection.  Those tests are skipped if that is not available.  The `r-base-core` and `r-cran-rserve` Ubuntu packages have been added to the docker build and the GitHub unit test action. Rserve is started by a docker entrypoint script, and in the GitHub action so that the tests are run in the docker container and the GitHub action.

You can use the attached file to test the vulnerability with the develop branch.
[r-serve-vulnerability.pg.txt](https://github.com/user-attachments/files/19743696/r-serve-vulnerability.pg.txt)


drgrice1:
Note this is built on top of #1213.

pstaabp:
I tried running the PG problem posted above on this branch and getting a `Statistics::R::IO::Rserve` not found error.  Seems that this module is just `Rserve` now.   Changing that resulted in other errors.

I was trying to test that the file writing is disallowed now.   Do you have another version of this problem that will show that this fails to write a file.

drgrice1:
Yes, the problem is designed to work with the develop branch (or any other branch of webwork).  It obviously won't work with this pull request because this removes the `Statistics::R::IO::Rserve` package from the safe compartment.  To use it with this branch just change the `Statistics::R::IO::Rserve` package usage in the problem to just `Rserve` which is what the PG package that replaces the `Statistics::R::IO::Rserve` is called.  You will also need to change the `to_pl` calls to `to_perl` because I changed that in the package (`to_pl` was a lame shortening).  However, that will still just throw an exception and the problem won't render.  You will get the error message `ERRORS from evaluating PG file: Write path /your/given/filename is unsafe from within WeBWorK::PG::IO::saveDataToFile`

pstaabp:
I do have `PG_ROOT` defined. 

I'm seeing `Use of uninitialized value in concatenation (.) or string at /opt/webwork/pg/lib/PGalias.pm line 34.` for every problem.  Maybe the pg_environment is not loading.  

Other errors that I'm seeing are some of my problems like: `/Contrib/Fitchburg/Algebra/intervals/interval_notation.pg` and I'm getting the following:
```
Use of uninitialized value $string in substitution (s///) at (eval 233) line 1005.
Use of uninitialized value $string in substitution (s///) at (eval 233) line 1006.
Use of uninitialized value in concatenation (.) or string at (eval 233) line 996.
Use of uninitialized value $string in substitution (s///) at (eval 233) line 1005.
Use of uninitialized value $string in substitution (s///) at (eval 233) line 1006.
Use of uninitialized value in concatenation (.) or string at (eval 233) line 996.
Use of uninitialized value $string in substitution (s///) at (eval 233) line 1005.
Use of uninitialized value $string in substitution (s///) at (eval 233) line 1006.
Use of uninitialized value in concatenation (.) or string at (eval 233) line 996.
```

pstaabp (approved):
This fixes the writing outside of the course vulnerability.

PR #1233, opened 2025-05-05
Remove 0_60_100 grader

dlglin:
This removes `custom_problem_grader_0_60_100`.  It doesn't work with quizzes and in other contexts, and only works with problems that have exactly five answer blanks.

The only problem in the OPL that used this grader has been rewritten to use `custom_problem_grader_fluid`.

I also clarified the POD for `custom_problem_grader_fluid` and fixed a typo.

### DIFF
--- a/macros/core/PGgraders.pl
+++ b/macros/core/PGgraders.pl
@@ -59,90 +59,18 @@ sub full_partial_grader {
 	return ($rh_problem_result, $rh_problem_state);
 }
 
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Deprecate? Yes.
Remove? Yes.

pstaabp (approved):
🥳

somiaj (approved):
Bye bye grader!

Issue #1234, opened 2025-05-10
parserRaidoMultiAnswer no longer works with '%s*' for ans_array.


somiaj:
This following problem renders in 2.19 but doesn't render in PG-2.20 with errors.

```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl parserRadioMultiAnswer.pl));

Context('Matrix');
$M   = Matrix([[1,0], [0,1]]);
$rma = RadioMultiAnswer([['The identity matrix is %s*.', $M]], 0);

BEGIN_PGML
Enter Identity Matrix.

        [_]{$rma}

END_PGML
ENDDOCUMENT();
```

Replacing the `%s*` with `%s` fixes the issue, but then I don't get the matrix answer array.

somiaj:
Fixed in #1235

PR #1235, opened 2025-05-10
Fix the maketext call added to `lib/Value/AnswerChecker.pm` in #1224.

drgrice1:
Since that is outside of the safe compartment the `maketext` method can not be called directly.  Instead it needs to be called via `pgCall`.

This fixes issue #1234.


somiaj (approved):
Fixes the issue.

PR #1237, opened 2025-05-24
Protect \columnbreak in DragNDrop TeX output.

somiaj:
The TeX output in DragNDrop puts the multicols environment behind an `\ifdefined` conditional, but the `\columnbreak` line is always included, causing the hardcopy build to fail when the multicols environment is not used. This puts the same conditional on the `\columnbreak` line so it is only included if the multicols environment is actually being used.

The following problem's hardcopy won't correctly build from the problem editor with the empty theme without this change. The error in the hardcopy.log is "! Package multicol Error: \columnbreak outside multicols."

```
DOCUMENT();
loadMacros(qw{PGstandard.pl PGML.pl draggableProof.pl});
$Proof = DraggableProof([ 'A', 'B', 'C', 'D' ], [ 'E', 'F', 'G' ]);
BEGIN_PGML
[_]{$Proof}
END_PGML
ENDDOCUMENT();
```

### DIFF
--- a/lib/DragNDrop.pm
+++ b/lib/DragNDrop.pm
@@ -187,8 +187,11 @@ sub TeX {
 		}
 		$out .= "}";
 
-		if ($i % 2 == 0) { $out .= "\\columnbreak\n\n" if $i != $#{ $self->{defaultBuckets} } }
-		else             { $out .= "\\ifdefined\\nocolumns\\end{multicols}\\else\\fi\n\\end{minipage}\n" }
+		if ($i % 2 == 0) {
+			$out .= "\\ifdefined\\nocolumns\\columnbreak\\else\\fi\n\n" if $i != $#{ $self->{defaultBuckets} };
+		} else {
+			$out .= "\\ifdefined\\nocolumns\\end{multicols}\\else\\fi\n\\end{minipage}\n";
+		}
 		$out .= "\\vspace{0.75\\baselineskip}\n";
 	}
 	$out .= "\n\\hrule\n\\vspace{0.25\\baselineskip}\n";


somiaj:
Thinking about this a little more and testing this with other themes, seems with the theme refactoring `\nocolumns` is never defined, so those `\ifdefined` statements are doing nothing. Is there a new way to detect if two columns is being used or not? Or is this an artifact of the past when the hardcopy themes where refactored? It appears that only `lib/DragNDrop.pm` and `macros/core/PGbasicmarcos.pl` still check for the definition of this tex variable which is no longer used by webwork themes.

So I guess the actual fix is to either remove all lines still checking if `\nocolumns` is defined, or update things to test if two columns is being used or not correctly. I assume the first is the better fix, just remove any remaining use of `\nocolumns`?

drgrice1:
This is actually a mistake in the hardcopy themes.  All of the single column themes should defined `\nocolumns` as the previous one column theme did.

somiaj:
I also think `addToTeXPreamble` is fairly fragile, and just assuming that what is needed is to break multicols in problem 0 may not work well. I'm thinking about the new box themes, with either columns or rows. Breaking the environment with those themes in problem 0 I think would have strange effects.

I think this original pr probably still needs to go in if the plan is to keep `\nocolumns`, because even in multicols cases where `\nocolumns` isn't defined, you don't want DragNDrop causing column breaks in the main columns.

Alex-Jordan:
Here is an idea that would work for themes that are one-column or use `multicol` for two columns. Perhaps the idea could be expanded to work for the other ways in which themes achieve multiple columns, like tcolorbox.

When potentially starting a `multicols` environment for the proof, more directly check if you are already in a `multicols` instead of using the `\nocolumns` hack, Like this:
```
\makeatletter\ifnum\numexpr\col@number>1\else\begin{multicols}{2}\newcommand\dragndropmulticols{}\fi\makeatother
```
This is checking if there is a `\col@number` greater than 1. It will be 2 for example if there is an ambient `multicols` environment with 2 columns.

But then the same test can't be used for the `\columnbreak` or for closing the `multicols` environment, because now `\col@number` will be 2. So create `\dragndropmulticols{}` as the marker that now we are in a `multicols` that is specifically for drag n drop.

Then for the columnbreak:

```
\ifdefined\dragndropmulticols{}\columnbreak\else\fi
```

And then to close the `multicols`:

```
\ifdefined\dragndropmulticols{}\end{multicols}\else\fi
```

drgrice1:
@somiaj: Needing to run `perltidy` more than once is common.  Each time it runs it can change.  They do have precautions to prevent an infinite loop of changes though, so it will always stabilize to a fixed point after running it enough times.

somiaj:
My thoughts were more on the line if we could implement something similar to `ADD_CSS_FILE()`, as `ADD_TEX_HEADER()` which just gets inserted in the header for hardcopy output and in the mathjax js header, and limit things to just defining custom macros, which seems to be the intent of `addToTeXPreamble`.

My thought on DragNDrop is in TeX output, could we detect the current width, and use that, though @Alex-Jordan checking if we are inside multicols could work too.

drgrice1:
That actually is not what `addToTeXPreamble` was created for.  It was implemented for the image display mode, and is so that you can add custom latex commands that could be used in equations displayed in HTML.  It is not actually used for hardcopy generation at all.

somiaj:
I saw it was for image output initially, and MathJax support was added later (though the POD states it doesn't work for MathJax). But it seems that the current macro is adding the tex to the first problem of hardcopy too, not just images.

If it was only for HTML output, then wouldn't any custom macros created fail in hardcopy output? It seems if used it should be included in both.

drgrice1:
It is included when hardcopy is used, but that is not the point of the method.  It does not work with MathJax display mode.  The POD is correct.  Hardcopy does not use MathJax display mode.  The point was that you add add a TeX command like `\myVec` and then you can use that in formulas.  The plan was to eventually make it so that with the MathJax display mode you could also use `\myVec`.  For example, in a problem you could do `[`\myVec{a}`]`.  That does not work.

somiaj:
Then in the code the `} else {    # for MathJax mode` doesn't actually work for MathJax mode despite it seems to insert the commands directly into the problem?

drgrice1:
It seems that it does work with MathJax now.  At one point MathJax didn't support defining commands like that.

 However, note that it does not actually add to the TeX preamble, and what you are talking about is not what the method was intended for.

drgrice1:
Also, as mentioned in the comments of the method it is quite fragile.  One problem could potentially mess up another problem.  It really shouldn't be used.  The whole concept of a problem or macro adding to the preamble is really not a good idea.

drgrice1:
Quite frankly all two column layouts should be killed.  That is not a good layout for WeBWorK problems in hardcopy.

Alex-Jordan:
I like @somiaj's idea to check the `\textwidth` and if it's below some threshold, don't do the `multicols`. Is that something you want to try?

About two-column themes in general. Probably it would be better if the theme you use were a property of the assignment instead of a coursewide setting. There are certain assignments that are nothing but short expressions and a two-column theme looks better (especially imo one that progresses horizontally). If it were an assignment attribute then you could only use a two-column theme where it works well.

somiaj:
@Alex-Jordan here is a branch which uses `\linewidth` to determine when to use multicols in DragNDrop.

https://github.com/somiaj/pg/tree/drag-n-drop-use-linewidth

I am kinda new to conditionals in TeX, but this is working in my tests.

somiaj:
@Alex-Jordan thanks, I opened a PR with the suggested updates, we can move discussion that that approach there. I went with 300pt, as in testing 240pt which was actually small enough to use multicols on the "Two Columns" theme felt too small, so I just increased it to 300pt. Since it is configurable, that seemed to be a nice reasonable round number that didn't use multicols on any of the default hardcopy themes, but used it on the non column themes.

Alex-Jordan:
I merged the other PR, so should this one be closed?

somiaj:
Sure, something still probably needs to be done about `addToTeXPreamble`.

somiaj:
I think having a way to add custom tex macros would be nice, though I have personally moved away from them due to writing so much portable code with mathjax.

I think a better way of dealing with them is as you mention, putting them in the actual document header would be preferable. But it might not be worth it as you pointed out only one OPL problem use it, and it has been broken since 2.19, and so far no one has noticed (though maybe just due to hardcopy not being used that much).

But the problem you mentioned, hardcopy is currently broken, even if we deprecate the macro, I don't think we should keep it in a broken state. At a minimum maybe we should at least make it not break hardcopy even if we depreciate it.

Alex-Jordan:
As for that problem, it can be edited to just use `\vec` instead of a special vector macro.

PR #1238, opened 2025-05-26
Change DragNDrop TeX logic on when to use multicols.

somiaj:
When creating the TeX output for DragNDrop hardcopy, update the logic to test if `\linewidth` is smaller than a configurable length, default of `300pt`, to determine if multicols should be used or not. This removes the dependency on `\nocolumns`, and any layout which has a current `\linewidth` that is greater than or equal to the configured size will use multicols.

This is done by defining a custom if statement `\ifdndcolumns`, then setting that to false/true depending on if `\linewidth` is less than `$self->{multicolsWidth}`, which defaults to 300pt.

This is an alternative approach to #1237 that came from the discussion there.

### DIFF
--- a/lib/DragNDrop.pm
+++ b/lib/DragNDrop.pm
@@ -77,10 +77,16 @@ This is the text label for the reset button.
 This is the text label for the button shown that adds new buckets.  The button
 is only shown if AllowNewBuckets is 1.
 
-=item removeButtonText (Default: C<< 'Remove' >>
+=item removeButtonText (Default: C<< 'Remove' >>)
 
 This is the text label for any remove buttons that are added to removable
 buckets.
+
+=item multicolsWidth (Default: C<< '300pt' >>)
+
+This sets the size for which the TeX output for hardcopy uses two columns or not.
+If the current C<\linewidth> is greater than or equal to this size then two columns
+will be used, otherwise only single column is used.
 
 =back
 
@@ -131,6 +137,7 @@ sub new {
 		resetButtonText   => 'Reset',
 		addButtonText     => 'Add Bucket',
 		removeButtonText  => 'Remove',
+		multicolsWidth    => '300pt',
 		%options,
 		},
 		ref($self) || $self;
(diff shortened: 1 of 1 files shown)


somiaj:
Here is a test problem that uses both 4 and 5 buckets to show what the various output could look like. It also shows that having two (or more) DragNDrop in the same document won't cause any errors.

[bucket-test.pg.txt](https://github.com/user-attachments/files/20433341/bucket-test.pg.txt)

Alex-Jordan:
It seems to work as expected, after I tried various themes. Of course, it is a long problem so using any of the boxed themes, things don't look good. But still work as designed.

This does not work as one would hope when I put the two parts of the exercises into columns of a table. But without getting into details, I think that case can't be addressed in any easy way. And it's probably not a use case worth pursuing anyway.

It also does not work as hoped for when there is indentation, say from putting it all in a list item. But again that is probably not a use case worth pursuing.

somiaj:
Though the PGML indenting and tables output nothing new, this was mostly to address that hardcopy wasn't building for DragNDrop and to deal with `\nocolumns` not being defined anymore. I tested this with the other PR, and get the same behavior, so the indention and tables are beyond the scope of this PR.

To me if someone wanted to make this work better with PGML indention or tables, that should be a separate PR.

lib/DragNDrop.pm
```diff
@@ -159,18 +166,20 @@ sub HTML {
...
-			if   ($i % 2 == 0) { $out .= "\n\\hrule\n\\vspace{0.5\\baselineskip}\n" }
-			else               { $out .= "\n\\ifdefined\\nocolumns\\else\\hrule\n\\vspace{0.5\\baselineskip}\n\\fi" }
+			if   ($i % 2 == 0) { $out .= "\n\\hrule\n\\vspace{0.5\\baselineskip}\n"; }
+			else               { $out .= "\n\\ifdndcolumns\\else\\hrule\n\\vspace{0.5\\baselineskip}\\fi\n"; }
 		}
 
 		$out .=
 			"\n\\begin{minipage}{\\linewidth}\n\\setlength{\\columnseprule}{0.2pt}\n"
-			. "\\ifdefined\\nocolumns\\begin{multicols}{2}\\else\\fi\n"
+			. "\\ifdndcolumns\\begin{multicols}{2}\\else\\fi\n"
```
Alex-Jordan, line 182: The `\else` here is not needed. Same for two places below.

```suggestion
			. "\\ifdndcolumns\\begin{multicols}{2}\\fi\n"
```

somiaj: I'll remove them, they were just there before so I left them.

somiaj:
@Alex-Jordan `\ifdndcolumns` should have the same truth value every time it is used. That is why I define it outside of the multicols, so its truth will not change when inside the multicols (it failed the first time I tried this for a reason similar to like you describe). The `\ifdim` statement just sets that to be either true or false, and then once it is set, there are no more tests, so I think your understanding of the logic isn't correct. My understanding is this line:

```
\ifdim\linewidth<300px\relax\dndcolumnsfalse\else\dndcolumnstrue\fi\n
```

I believe this line is setting the truth value of the `\ifdndcolumns` conditional, and this is the only test for size that gets used, after that it is just a simple testing if that if statement is true or false.

lib/DragNDrop.pm
somiaj: Okay, removed the unneeded else statements.

PR #1240, opened 2025-05-28
Remove unload event listener from JavaScript.

somiaj:
The 'unload' event listener is deprecated, in addition since all javascript is stopped when the window unloads, it is unnecessary to also disconnect the mutation observer when the window unloads. This removes all the unload event listeners.


drgrice1 (approved):
Thanks for taking care of this.

PR #1241, opened 2025-06-01
Implement makeZero() for matrices (and other types).

### LINKED ISSUE #1239: Derivative formula not inheriting context variables
Below I have an exercise that I just tested with WW 2.18, and it works there. One correct answer you can use is `x` in the first answer blank and `x^2` in the second.

In 2.20 (as well as the 2.19+ that I have in production), this does not work. There's an error message in the feedback that a derivative operator needs to declare what variable to use. (Note the instances of the `D` method that do not specify a variable to differentiate with respect to.) But this shouldn't happen, because the context only has one variable, `x`.  There's not supposed to be a need to pass an argument to `D`.

I've found that it's the instances of using `D` a second time that matter. For example, `$s[0]->D` is not causing an error, but `$s[0]->D->D` is. So something about the result from  `$s[0]->D` is off, and doesn't know that its only context variable is `x`.

```
DOCUMENT();

loadMacros(qw(
      PGstandard.pl PGML.pl parserMultiAnswer.pl
));

$f = Formula("x");
$g = Formula("x^2");

$mp = MultiAnswer($f, $g)->with(
    singleResult => 1,
    checker => sub {
        my ($c,$s,$self) = @_;
        my @s = @{$s};
        Value::Error("Your first polynomial is not in \(\mathbb{P}_2\)") unless ($s[0]->D->D->D == 0);
        Value::Error("Your second polynomial is not in \(\mathbb{P}_2\)") unless ($s[1]->D->D->D == 0);
        Value::Error("Your first polynomial is not in \(\operatorname{ker}(T)\)") unless ($s[0]->eval(x => 0) == 0);
        Value::Error("Your second polynomial is not in \(\operatorname{ker}(T)\)") unless ($s[1]->eval(x => 0) == 0);
        Value::Error("Your polynomials do not span \(\mathbb{R}^2\)") unless ($s[0] != 0 && $s[1] != 0);
        my $ratio = $s[0]/$s[1];
        Value::Error("Your polynomials do not span \(\mathbb{R}^2\)") if ($ratio->D() == 0);
        return 1;
    },
);

BEGIN_PGML
Let [`T:\mathbb{P}_2\to\mathbb{R}^2`] be defined by [`T(f)=\begin{bmatrix}f(0)\\f(0)\end{bmatrix}`]. Find two polynomials that span the kernel of [`T`].

[`f(x)={}`][_]{$mp}{16} and [`g(x)={}`][_]{$mp}{16}
END_PGML
ENDDOCUMENT();
```

somiaj:
A new approach in the makeZero method of Parser::BOP to turn each operand into an object of the correct type with nothing but zeros as entries, and then returns a product of the results. If it doesn't know how to make an all-zero version, it leaves it as multiplication by zero.

This is used in the reduction rules `0*x` and `x*0` to return a zero object of the product. Then differentiation of variables uses this process to take derivatives of variables treated as constants to allow taking derivatives of variables that represent Points, Vectors, and Matrices. Error messages are added if attempting to take a derivative of variables that are not numbers.

This also corrects an issue to allow taking the derivative of a formula with a single variable without explicitly stating the variable via `$MO->D`. Fixes #1239.

Attached is a problem I used to test various cases.

[makeZero-tests.pg.txt](https://github.com/user-attachments/files/20541789/makeZero-tests.pg.txt)

### DIFF
--- a/lib/Parser/BOP.pm
+++ b/lib/Parser/BOP.pm
@@ -265,18 +265,30 @@ sub matchError {
 #  Return a zero, or a list of zeros of the proper length.
 #
 sub makeZero {
-	my $self = shift;
-	my $op   = shift;
-	my $zero = shift;
-	return $zero if ($op->isNumber);
-	if ($zero->isNumber && $op->type =~ m/Point|Vector/) {
-		my $context = $op->{equation}{context};
-		my $value   = $context->Package($op->type)->new($context, ($zero->{value}) x $op->length);
-		$value = $self->Item("Value")->new($op->{equation}, $value);
-		$value->{value}{ijk} = 1 if $op->class eq "Constant" && $op->{def}{value}{ijk};
-		return $value;
+	my ($self, $lop, $rop) = @_;
+	$lop = $self->zeroValue($lop);
+	$rop = $self->zeroValue($rop);
+	return $self unless defined $lop && defined $rop;
+	return $self->Item('Value')->new($self->{equation}, $lop * $rop);
+}
+
+sub zeroValue {
+	my ($self, $op) = @_;
+	my $type = $op->type;
+	return $self->Package('Real')->new(0) if $type eq 'Number';
+	my $typeDef = $op->{type};
+	my @coords  = (0) x $typeDef->{length};
+	if ($typeDef->{entryType}{list}) {
+		return if $type ne 'Matrix';
+		my @len;
+		do {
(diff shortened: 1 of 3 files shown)


dpvc (approved):
Thanks for making the PR

pstaabp (approved):
Looking at the sample problem, looks like this now works.

PR #1251, opened 2025-06-18
Add plots sample probs

pstaabp:
This adds some samples problems related to the plots.pl macros. 

In addition, there are some small updates to other problems.


Alex-Jordan (approved):

drgrice1 (approved):
Some of the added problems do not have `categories`.  Please make sure that they all do.  This makes sure they are represented in the sample problems categories list.

Other than the things mentioned, this looks good.

pstaabp:
Another commit that removes the Updated by ... comment, added categories to problems missing that tag and changed others for consistency.

PR #1254, opened 2025-06-20
Creation of a PG critic script.

pstaabp:
The `pg-critic` script analyzes a pgproblem (or list of them) for both good and bad features of a problem.  Currently this include:

Positive features:
- Uses PGML
- Provides a solution
- Provides a hint
- Uses Scaffolds
- Uses a custom checker
- Uses a multianswer
- Uses answer hints
- Uses nicetables

Old and deprecated features
- checking for deprecated macros
- Use of `BEGIN_TEXT/END_TEXT`
- Include the `TEXT(beginproblem)`
- Include old tables (for example from `unionTables.pl`)
- The use of `num_cmp`, `str_cmp` and `fun_cmp` in lieu of using MathObjects
- Including `Context()->TeXStrings`
- Calling `loadMacros` more than once.
- Using the line `$showPartialCorrectAnswers = 1` which is the default behavior and thus unnecessary.
- Using methods from `PGchoicemacros.pl`
- Including code or other text below the `ENDDOCUMENT();` line indicating the end of the problem.

Currently the script can also score a problem (there is a rubric that is built-in to the script).  This can be made more flexible. 

Also, there are plans to include more features on both the positive and negative side.


drgrice1 (commented):
Please set the executable bit on the `pg-critic.pl` script.

pstaabp:
This improves the pg-critic script.  Additional positive features are now scanned for including features in modern context, parsers and macros.  

The script also now updates the help page for the script and provides a JSON output format which writes to a file--this might be helpful for scanning large numbers problems.

duffee:
Nice work!

A suggestion for an optional critic policy is the amount of randomness available in the question. An anti-pattern I've seen by authors in a hurry (Contrib/BrockPhysics/College_Physics_Urone) is
```
$A1 = 63.5;
ANS(num_cmp("$A1"));
```
usually the result of values hardcoded in an image or just copied straight from an Answer Key. Other problems aside, this pgproblem has no randomness at all. It would be nice to know:
* if there is randomness in the answers
* estimate how many permutations available
* range of answers (because I can call `random`, but not use it in a calculation)

It's more a pedagogic issue than authoring style, but the first two should be a Simple Matter of Programming ;)

drgrice1 (changes requested):
I think that at this point, this and https://github.com/openwebwork/webwork2/pull/2748 will need to be deferred until the next release.  There are a lot of problems with the current code.  The webwork2 pull request has many issues including things like invalid HTML.  Both pull requests need code clean up.

The most serious issue though is with this pull request.  We need to make sure that issues (both positive and negative) are correctly detected.  It would be really bad if a positive aspect were detected as a negative aspect or vice versa, and that is happening at least to some extent.

Note that your code skips full line comments, but not code does not exclude comments at the end of a line.  That means there could be incorrect results from these end of line comments.

Any problem that uses macros like `parserPopUp.pl`, `parserRadioButtons.pl`, or `parserCheckboxList.pl` with randomization of the choices but does not have any other `random` call in the problem are currently flagged as not having "randomness".  That will definitely need to be fixed before this can be rolled out, since there are a lot of problems that use those macros.  I think in general the randomness check will need to be done quite differently in order to work.  I think that looking for the strings "random(" or "random_...(" or "random_list(" is not going to work.  Those strings could be contained in an end of line comment and the current check will count it.  Even if the end of line comment issue is fixed, these words could appear in an actual string in the code and the check will count that also.

Also, something is now wrong with the `bin/pg-critic.pl` script.  All problems are returning a score of 0.  This is because you must have changed from using "good" and "bad" in the `PGProblemCritic.pm` file to using "positive" and "negative", but didn't change the `pg-critic-.pm` script accordingly.  Even with that fixed there are issues because of other changes in the module not in the script (particularly the addition of "randomness").  In general that brings into question the techniques used.  The general data structure is not good.

In general, I don't think that this approach of parsing the code as text is going to work.  A more versatile and reliable approach is needed.

PR #1256, opened 2025-06-20
Deprecate some more macros and set a warning if using.

pstaabp:
Include the rest of the Union macros in the deprecate and move the PGnumericevaluators out of deprecate because every problem loads it.

In addition show a warning to instructors if a problem is using a deprecated macro.

Note: this will affect a lot of macros in the OPL.  Seems like the deprecated macro with most references in the OPL is the `answerFormatHelp` macro.  

I will plan to put a PR to remove that macro from all of the OPL problems that call it.

### DIFF
--- a/lib/PGloadfiles.pm
+++ b/lib/PGloadfiles.pm
@@ -53,7 +53,6 @@ package PGloadfiles;
 use strict;
 #use Encode(qw(encode decode));
 use Exporter;
-
 use PGcore;
 use WeBWorK::PG::Translator;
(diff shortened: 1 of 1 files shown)


pstaabp:
Remove the detection of deprecated macros on the PG side and now uses a different mechanism on the WeBWorK side.

drgrice1 (approved):
Lets get this in so that the macros that shouldn't have been deprecated are back where they belong.

PR #1258, opened 2025-06-29
AllowReusingItems

sfiedle1:
Adds functionality to drag-n-drop problems to permit the repository bucket to be continually replenished. This is enabled by setting the flag, AllowReusingItems => 1 in the DraggableSubsets call in the pg problem code.

### DIFF
--- a/htdocs/js/DragNDrop/dragndrop.js
+++ b/htdocs/js/DragNDrop/dragndrop.js
@@ -101,7 +101,7 @@
 			this.allowReusingItems = this.bucketPool.allowReusingItems;
 
 			if (this.allowReusingItems) {
-				if (id == 0) {
+				if (id === 0) {
 					this.sortable = Sortable.create(this.ddList, {
 						animation: 150,
 						sort: false,
@@ -119,23 +119,8 @@
 						removeOnSpill: true,
 						group: {
 							name: bucketPool.answerName,
-							put: (
-								to,
-								from,
-								dragEl,
-								event //Prevents buckets from storing multiple clones
-							) => {
-								this.flag = 0;
-								Array.from(to.el.children).some((child) => {
-									if (child.dataset.id === dragEl.dataset.id) {
-										this.flag = 1;
-									}
-								});
-								if (this.flag == 1) {
-									return false;
-								}
-								return true;
-							}
+							put: (to, _from, dragEl) =>
+								!Array.from(to.el.children).some((child) => child.dataset.id === dragEl.dataset.id)
 						}
 					});
 				}
--- a/htdocs/js/DragNDrop/dragndrop.js
+++ b/htdocs/js/DragNDrop/dragndrop.js
@@ -9,6 +9,7 @@
 			this.answerName = el.dataset.answerName ?? '';
 			this.buckets = [];
 			this.removeButtonText = el.dataset.removeButtonText ?? 'Remove';
+			this.allowReusingItems=el.dataset.allowReusingItems;
 
 			this.answerInput = el.parentElement.querySelector(`input[name="${this.answerName}"]`);
 			if (!this.answerInput) {
@@ -89,21 +90,50 @@
 		constructor(bucketPool, id, bucketData) {
 			this.id = id;
 			this.bucketPool = bucketPool;
-
 			this.el = this.htmlBucket(bucketData.label, bucketData.removable, bucketData.indices);
 			bucketPool.bucketContainer.append(this.el);
-
-			// Typeset any math content that may be in the added html.
 			if (window.MathJax) {
 				MathJax.startup.promise = MathJax.startup.promise.then(() => MathJax.typesetPromise([this.el]));
 			}
+			this.allowReusingItems=this.bucketPool.allowReusingItems;
 
-			this.sortable = Sortable.create(this.ddList, {
-				group: bucketPool.answerName,
-				animation: 150,
-				onEnd: () => this.bucketPool.updateAnswerInput()
-			});
-		}
+			if(this.allowReusingItems==false){
+				this.sortable = Sortable.create(this.ddList, {
+					group: bucketPool.answerName,
+					animation: 150,
+					onEnd: () => this.bucketPool.updateAnswerInput()
+				});
+			}
+			else if(id==0){  
+   			this.sortable=Sortable.create(this.ddList, {
+					animation:150, 
+					sort: false,
+ 					onEnd:()=>this.bucketPool.updateAnswerInput(),
+					group:{
+						name:bucketPool.answerName,
+						pull:'clone',
+            put: false,
+					}
+				});
+			}
+			else{
+				this.sortable=Sortable.create(this.ddList, {
+					animation:150, 
+					onEnd:()=>this.bucketPool.updateAnswerInput(),
+        	removeOnSpill: true,
+        	group:{
+						name:bucketPool.answerName,
+            put: (to, from, dragEl, event) =>  //Prevents buckets from storing multiple clones
+            { 
+							this.flag=0;
+							Array.from(to.el.children).some(child=> 
+								{if(child.dataset.id === dragEl.dataset.id){this.flag=1;}}
+							);
+							if(this.flag==1){return false} return true
+						} 
+				} });
+			}
+	}
 
 		htmlBucket(label, removable, indices = []) {
 			const bucketElement = document.createElement('div');
@@ -182,4 +212,7 @@
 		}
 	});
 	observer.observe(document.body, { childList: true, subtree: true });
+
+	// Stop the mutation observer when the window is closed.
+	window.addEventListener('unload', () => observer.disconnect());
 })();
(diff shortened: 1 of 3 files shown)


drgrice1:
To fix the JavaScript formatting issues in this pull request run `npm run prettier-format` from the htdocs directory.  To fix the Perl formatting issues run `./bin/run-perltidy.pl` in the PG root directory.

lib/DragNDrop.pm
```diff
@@ -132,7 +126,7 @@ package DragNDrop;
 use strict;
 use warnings;
 
-use Mojo::JSON qw(encode_json);
+use JSON;
```
pstaabp, line 129: We've switched to Mojo::JSON for this now.  Please switch this back.

drgrice1: In general, there are lots of places in this pull request that revert changes that have been made to the upstream code.  @pstaabp pointed out one here, but the new `multicolsWidth` option is removed in this pull request as well.  The `multicols` TeX handling fix is also reverted.  Please check over all of the changes, and make sure only the things that are for this new option are changed.

htdocs/js/DragNDrop/dragndrop.js
```diff
@@ -89,20 +90,57 @@
...
 				MathJax.startup.promise = MathJax.startup.promise.then(() => MathJax.typesetPromise([this.el]));
 			}
+			this.allowReusingItems = this.bucketPool.allowReusingItems;
 
-			this.sortable = Sortable.create(this.ddList, {
-				group: bucketPool.answerName,
-				animation: 150,
-				onEnd: () => this.bucketPool.updateAnswerInput()
-			});
+			if (this.allowReusingItems == false) {
```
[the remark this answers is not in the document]
drgrice1: In fact, since this is set from a `dataset` value, it will have type `string` or `undefined`.  So `!this.allowReusingItems` is definitely not the correct comparison here.  If this is set to `0` by the Perl code, then that will be the string "0", and `!this.allowReusingItems` will be false, and that is certainly not correct.

macros/math/draggableSubsets.pl
```diff
@@ -148,6 +149,14 @@ =head1 EXAMPLE
         # The default value if not given is 1.
         AllowNewBuckets => 1,
 
+				# 0 is the conventional approach, by which the repository bucket is 
+				# depleted if an item an item is dragged from that bucket. 
+        # 1 means that items are replenished in the repository bucket once 
+				# dragged. This option is designed to be used in conjunction with 
+				# $showPartialCorrectAnswers = 0 in the problem code.
+				# The default value if not given is 0.
```
somiaj, line 157: Same here, comments don't appear to line up properly.

htdocs/js/DragNDrop/dragndrop.js
```diff
@@ -97,12 +98,54 @@
...
+								Array.from(to.el.children).some((child) => {
+									if (child.dataset.id === dragEl.dataset.id) {
+										this.flag = 1;
+									}
+								});
+								if (this.flag == 1) {
+									return false;
+								}
+								return true;
+							}
```
drgrice1, line 138: I recommend changing this to
```suggestion
							put: (to, _from, dragEl) =>
								!Array.from(to.el.children).some((child) => child.dataset.id === dragEl.dataset.id)
```
That is much more succinct and does the same thing.

Note that the fourth `event` argument that I removed is optional, and since it is unused in this method it should not be added to the parameter list of the function.  Also the `_from` argument is unused, but obviously it is needed so that the third parameter that is needed can be accessed.  So the underscore should prefix it to denote that.  That is a standard way of denoting unused but necessary variables, and is recognized by default by many linters (like the `tsserver`).

Now the array `some` method checks that some element in the array returns true for the given test.  It will exit immediately and return true if some element does so, and returns false otherwise.  So the whole point of the method is that it already returns what you want this method to return if done right as it is in my suggestion.

By the way setting an attribute of `this` (as you did with the `this.flag`) was certainly not something that should have been done as it was not used anywhere outside of this method.  A local variable would have been the proper thing to use.

somiaj:
Thanks for the sample problem, in my initial testing everything worked for me. The first subset/bucket as part of the student's answer felt a bit odd to me. I see that you designed it to work with `$showPartialCorrectAnswers=0` to disable partial credit, but it could be nice to allow using partial credit and not include the first subset in the student's answer.

A different approach could be to create a new bucket that contains all items, not include this bucket as part of the student's answer, and also not require having to include it in the problem author's code. Though seems another option
to set any custom label of the new bucket would be needed, giving something like:

```
$draggable = DraggableSubsets(
    [ 'orange', 'blue', 'apple' ],
    [ [ 0,1 ], [ 0,2 ] ],
    DefaultSubsets => [
        { label => 'Color', indices => [] },
        { label => 'Fruit', indices => [] },
    ],
    AllowNewBuckets   => 0,
    AllowReusingItems => 1,
    ReusableItemLabel => 'Drag each of the below words to classify them as a color or a fruit.'
);
```

I haven't looked into the code to see how to achieve this approach, just presenting as an alternative approach to think about.

htdocs/js/DragNDrop/dragndrop.js
drgrice1: By the way the `===` is not a mistake.  I used that in my earlier suggestion for the `id === 0` for what is now line 104 of this file.  That was also not a mistake.  `===` is better to use than `==` in javascript in most cases.  `==` should only be used when a type check is not desirable.

macros/math/draggableSubsets.pl
sfiedle1: I suspect that the comment alignment issues are due to the use of spaces and tabs prior to running run-perltidy.pl. As I'm new to this process, I'm a bit hesitant to make edits to the resultant "tidied" code as it could conceivable trip the github formatting test again. My thought is to complete this clean-up as one of the last steps in the process but I am happy to abide by the norms of the community.

somiaj: This is part of the POD, and the indents should be four spaces to create a code block, so here you want to remove the tabs, replace with spaces and match the indenting of the rest of the example. You can view the POD from the problem editor in webwork to see the result. The issue with perltidy was using tabs in the first place, perltidy will leave this code alone when using spaces.

drgrice1: Perltidy never touches anything inside the POD.

sfiedle1:
@somiaj I too have been troubled by the improper calculation for partial credit. As you noted, I do not need it for my use case, but I could envision that it would be helpful for such bucket problems in a Test. However, I failed to make much headway after looking into this for a couple days. I appreciated your idea for a "ghost bucket" but my suspicion is such an approach may run into structural limitations of the sortablejs library. I also agree that your pg syntax would be a nice way for problem authors to specify the correct solution when using the AllowReusingItems option. I would be happy to revisit the partial credit issue if any folks have an idea on how this could be accomplished.

sfiedle1:
@somiaj I am not targeting any particular release and suspect that 2.20 may be a bit too ambitious.

drgrice1:
I think in general the problem with this pull request is that it is not in the spirit of the draggable proof and draggable subset design which is that objects can be dragged to and from one bucket to another.  As such, it introduces situations that are not going work right, such as the example I just mentioned.  So in order for this to work, situations like that are going to need to be thought out, and figured out how they should be dealt with.

Another issue is that this claims that this allows reusing items.  However, it limits how items can be reused.  It only allows one copy of an item for each bucket.  What if an author wants items to be reusable with multiple copies of an item in a single bucket?

I think that more thought in general is needed as to how this should really work.

sfiedle1:
I have appreciated the attention to this pull request. One small note if anyone comes back on these comments- unlimited spawning can be accomplished by commenting out the "put" statement in the dragndrop.js file   
               
// put: (to, _from, dragEl) =>  !Array.from(to.el.children).some((child) => child.dataset.id === dragEl.dataset.id)

macros/math/draggableSubsets.pl
```diff
@@ -333,7 +347,18 @@ sub cmp_defaults {
...
+		$tmp = $self->SUPER::cmp(%{ $self->{cmpOptions} }, %options)->withPostFilter(sub {
+			$ansHash          = shift;
+			$ncor             = $ansHash->{score} * $nbuc;    #number of buckets correct
+			$ansHash->{score} = ($ncor - 1) / ($nbuc - 1);
+			return $ansHash;
+		});
+		return $tmp;
+	} else {
+		return $self->SUPER::cmp(%{ $self->{cmpOptions} }, %options);
+	}
```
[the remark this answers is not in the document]
drgrice1: There is another issue that is related to this.  The answers that are shown for both the correct answer and the student answer include this first bucket which is really not part of the answer for this setup.  So that needs to be fixed so that the first bucket is not shown.  I think that if this is done right it will remove the need for the special handling with the `cmp_equal` method (or a post filter).

sfiedle1:
This pull request has been superseded by #1263.

PR #1259, opened 2025-07-03
also close image description details if you touch outside of the details

Alex-Jordan:
This makes it so if you are using a touchscreen device you can close the image long description by touching outside of the details, just extending what @drgrice1 already did for clicking outside of the details.


drgrice1:
Something is off with this.  The `pointerdown` event works for both the mouse and touch which is why I chose that event rather than the `mousedown` event that is specifically for the mouse.  So a touch should outside of the details already close the details without this pull request, and in my testing it does.

drgrice1:
By the way, see https://developer.mozilla.org/en-US/docs/Web/API/Element/pointerdown_event.  The documentation is in agreement with my assessment.

Issue #350, opened 2018-04-23
NumberWithUnits and arithmetic operations


heiderich:
When forming products and quotients of two MathObjects that are created using the subroutine `NumberWithUnits` from `parserNumberWithUnits.pl`, the "number part" is computed correctly, but the "unit part" is not correct. It seems that the unit of a product and a quotient is the unit of the left operand. The example below illustrates this problem. A workaround is to work with Perl numbers first and only add the units after all computations are done. I think however that it would be nice if an example like this would work too.

```
DOCUMENT();
loadMacros("PGbasicmacros.pl",
           "PGML.pl",  
           "parserNumberWithUnits.pl",
);
$F = NumberWithUnits("1 N");
$s = NumberWithUnits("2 m");
$W = $F / $s;
$t = NumberWithUnits("2 s");
$P = $W/$t;
TEXT(beginproblem());
BEGIN_PGML
[` [$W]=[$F] \cdot [$s] `]

[` [$P]=[$W] / [$t] `]
END_PGML
ENDDOCUMENT();
```

duffee:
My preference has been to create a new NumberWithUnits using the desired unit

```perl
$F = NumberWithUnits( 1, 'N');
$s = NumberWithUnits( 2, 'm');
$W = NumberWithUnits( $F * $s, 'J');    # or 'N*m' depending on which I'd like to display in the solution

$t = NumberWithUnits( 2, 's');
$P = NumberWithUnits( $W / $t, 'W');
```
It loses some of the 'do what I mean'-ness, but the clarity has helped every so often.

Alex-Jordan:
I think the original example meant to have `$W = $F * $s;` instead of `$W = $F / $s;`.

Confirming in 2025, this is still an issue.

pstaabp:
Fixed in #1260

PR #1260, opened 2025-07-03
Update NumberWithUnits to work more like other MathObjects (#350)

dpvc:
This PR makes the `NumberWithUnits` and `FormulaWithUnits` objects work more like regular MathObjects.  It adds addition, subtraction, multiplication, division, and powers (by non-zero integers) to both so that you can use the objects in perl formulas as with other MathObjects.  It also improves the comparison functions to allow comparisons among numbers and formulas with units (the original implementation only works in the context of an answer evaluator).

The `NumberWithUnits()` function will now return a formula-with-units if passed a string that evaluates to a formula.  For example

``` perl
$f = NumberWithUnits("x m/s");
```

will produce a `FormulaWithUnits` object rather than throwing an error.

The `compare` methods are significantly simplified to avoid the complications involved with handling the student units.  This is done by creating an actual object-with-units version of the student answer rather than re-parsing the student answer and having the `compare` method handle the student units.  This also allows `compare` to be used between object with units directly, which didn't work before.  The `$other` object is promoted as needed to make sure that the comparison can be made, and to generate error messages if the other operand isn't an object-with-unit.

The handling of the `cmp` method is refactored to use a pre-filter to strip off the units and let the usual `cmp` method do the rest of the work.  New `cmp_preprocess` and `cmp_postprocess` functions handle putting the units back and making an object-with-units from the student answer and providing the the message about incorrect units.  The `cmp_equal` is overridden to skip the check if there is an error.

There are other technical details that I won't go into, but if you have questions, I will try to answer them.

Resolves issue #350.

### DIFF
--- a/lib/Parser/Legacy/NumberWithUnits.pm
+++ b/lib/Parser/Legacy/NumberWithUnits.pm
@@ -20,7 +20,7 @@ sub makeValue {
 	my $self    = shift;
 	my $value   = shift;
 	my %options = (context => $self->context, @_);
-	Value::makeValue($value, %options);
+	bless Value::makeValue($value, %options), $options{class};
 }
 
 sub initializeUnits {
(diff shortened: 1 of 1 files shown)


pstaabp (approved):
Fixes the issue and did some other testing with it.

drgrice1:
Note that this was merged into the `develop` branch.  We want this in the 2.20 release right?  If so, I will make that happen behind the scenes.

drgrice1:
I went ahead and just added this to the `PG-2.20` branch.

Issue #1261, opened 2025-07-03
maketext error if PGanswermacros.pl is not loaded


dlglin:
Consider the following MWE:
```
DOCUMENT();

loadMacros("PG.pl",
           "PGbasicmacros.pl",
           "MathObjects.pl",
           #"PGanswermacros.pl",
           "PGcourse.pl",
);

$ans1 = Real(1);

BEGIN_TEXT
\{ans_rule(30)\}\{ans_rule(30)\}
END_TEXT

ANS($ans1->cmp());
ANS($ans1->cmp());

ENDDOCUMENT() ;
```
The problem renders in the problem editor, but if I click any of the buttons (preview, check, show answers) I get "Undefined subroutine &WeBWorK::PG::Translator::maketext called at /opt/webwork/pg/lib/WeBWorK/PG/Translator.pm line 1080."

If the problem is viewed within a set then the error appears immediately.

If you load `PGanswermacros.pl`, then there are no errors.  This makes no sense to me since `maketext` is not defined in that macro file or anything that it loads.

drgrice1:
This was fixed by #1262.

PR #1262, opened 2025-07-03
Fix direct `maketext` calls in the Translator.

drgrice1:
`maketext` is not defined in the Translator package, and so it cannot be called directly in the default graders defined in the translator. Instead an eval is needed to access the `maketext` in the safe compartment.

This fixes issue #1261.  See that issue for MWE to test this with.


drgrice1:
By the way, this only affects problems that are not correctly written to load `PGstandard.pl` (and hence `PGanswermacros.pl`) as all problems should.

pstaabp (approved):

PR #1263, opened 2025-07-05
Add an option to show the "universal set" in `draggableSubsets.pl` problems.

drgrice1:
This extends @sfiedle1's work in #1258, and implements the structural rework that I suggested in the conversation for that pull request (albeit slightly modified).

There are two new options for the `draggableSubsets.pl` macro.  They are `ShowUniversalSet` and `UniversalSetLabel`.

If `ShowUniversalSet` is 1, then the set of all elements that are passed as the first argument for the `DraggableSubsets` method will be shown as a separate drag and drop bucket.  This bucket is always above the other usual drag and drop buckets (both in HTML and TeX).  The elements of the universal set can be dragged multiple times to the other buckets.  Really this is a copy of the element, so all elements always remain in the universal set.

The `UniversalSetLabel` option is a string that will be shown as the label of the universal set bucket.

Note that this approach doesn't require any special handling for grading the answer regardless of if `$showPartialCorrectAnswers` is true or not, because the universal set bucket is not part of the answer in any case.  It is not listed in any of the answer previews, and is not part of the grading in any way. It is purely a source of elements.

I also added some validation of the answers and default subsets that are passed to the DraggableSubsets method.  Previously invalid inputs would have caused rather unexpected things.  Now they will be errors (specifically `Value::Error`s).

In my original comment to #1258 I suggested a `ShowUniversalSet` option and a related `AllowReusableElements` option.  The latter option was dropped because I realized it doesn't really make sense in the context of a universal set. I initially implemented it, but basically it made the "universal set" bucket really just a regular bucket just like all of the others, and it really wasn't a "universal set" anymore since it didn't always have all elements. I thought of a way that it could make sense in which it always has all elements, but only one copy was allowed to be dragged to the other buckets.  So only one duplicate, no reuse after that.  That would take some effort to implement (mostly in the javascript code), but could be done. I don't know that this would really be that useful though, so I didn't implement it for now.

A MWE for testing this is as follows:

```perl
DOCUMENT();

loadMacros(qw{PGstandard.pl PGML.pl draggableSubsets.pl PGcourse.pl});

$draggable = DraggableSubsets(
    [ 'orange', 'blue', 'apple' ],
    [ [ 0, 1 ], [ 0, 2 ] ],
    DefaultSubsets => [
        { label => 'Color', indices => [] },
        { label => 'Fruit', indices => [] },
    ],
    ShowUniversalSet  => 1,
    UniversalSetLabel =>
        'Classify each item below as a color, fruit, or both.',
    AllowNewBuckets => 0
);

BEGIN_PGML
[_]{$draggable}
END_PGML

ENDDOCUMENT();
```

To test this you should also test modifications of the above problem with other options.  For example, change to `AllowNewBuckets => 1`, and test addition and removal of buckets particularly removal after elements have been added to the bucket.  Add indices to the `DefaultSubsets` so elements are initially shown in those and behavior when elements are dragged in an out.

### DIFF
--- a/htdocs/js/DragNDrop/dragndrop.js
+++ b/htdocs/js/DragNDrop/dragndrop.js
@@ -123,6 +123,7 @@
 
 			if (bucketPool.showUniversalSet) {
 				if (bucketData.isUniversalSet) {
+					options.sort = false;
 					options.group.pull = 'clone';
 					options.group.put = false;
 				} else {
--- a/htdocs/js/DragNDrop/dragndrop.js
+++ b/htdocs/js/DragNDrop/dragndrop.js
@@ -9,6 +9,7 @@
 			this.answerName = el.dataset.answerName ?? '';
 			this.buckets = [];
 			this.removeButtonText = el.dataset.removeButtonText ?? 'Remove';
+			this.allowReusingItems = el.dataset.allowReusingItems;
 
 			this.answerInput = el.parentElement.querySelector(`input[name="${this.answerName}"]`);
 			if (!this.answerInput) {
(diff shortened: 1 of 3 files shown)


pstaabp:
Playing with the MWE, I added math into the  with \( \) and it seems like as soon as math is it, the label text gets put below the buckets.  

The `label` option on the `DefaultSubsets` option produces mathematics correctly.

sfiedle1:
Glenn, I appreciate this pull request. This functionality will be very helpful. A few minor observations... 
- The default label for UniversalSetLabel is unspecified. This appears to be at odds with the POD for DragNDrop.pm and draggableSubset.pl files.
- In the POD for the DragNDrop.pm, the description of the showUniversalSet variable states "Elements can be dragged from this set and into other buckets, but not into it". It's nitpicky, but technically elements can be dragged into the Universal Set, they are just spilled as you noted in the previous conversation.
- If $showPartialCorrectAnswers = 0; is set in the problem code, I could envision some of the weaker students getting hung up on rearranging the order of elements/items in the Universal Set in an attempt to complete the problem. This could be rectified by specifying "options.sort = false;" at line 128 in the dragndrop.js code.

PR #1244, opened 2025-06-09
Restructure the POD of macros

pstaabp:
For consistency the POD of macros have been updated so that there is a `=head1 NAME` with the following structure:

```
macroName.pl - short description.
```

Also, any method or function names will have the form:
```
=head2 function_name
```

This is helpful in conjunction with openwebwork/webwork2#2733 to extract information from the macro files.

PR #1269, opened 2025-07-12
Restructire the POD of macros (graph)

Alex-Jordan:


pstaabp (approved):

drgrice1 (commented):
There are a few places where the added name section has an sentence after the dash added that is not capitalized, and the `PGgraphmacros.pl` has inconsistent "Usage" handling.

I will add a pull request to this branch that fixes these things, and then this pull request looks like it is in order.

drgrice1 (approved):

PR #1272, opened 2025-07-12
Restructire the POD of macros (answers)

Alex-Jordan:

### DIFF
--- a/macros/answers/Generic.pl
+++ b/macros/answers/Generic.pl
@@ -1,30 +1,28 @@
-#R. Byerly, Texas Tech University, 2005
 
-sub _Generic { }    #Don't reload this file.
+=head1 NAME
 
-#Documentation:
+Generic.pl - provide some generic answer checkers.
 
-=head1 Generic Answer Checker
+=head1 DESCRIPTION
 
-(Based on David Cervone's vector_cmp.  See
-doc/parser/extensions/8-answer.pg under the webwork2 directory.)
-Place in macros directory and load this file (Generic.pl) using the
-loadMacros command.  (To just copy it into a pg file, replace
-occurences of "\&" by "~~&".)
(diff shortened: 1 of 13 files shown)


macros/answers/Generic.pl
```diff
@@ -1,30 +1,28 @@
...
 
-(Based on David Cervone's vector_cmp.  See
-doc/parser/extensions/8-answer.pg under the webwork2 directory.)
-Place in macros directory and load this file (Generic.pl) using the
-loadMacros command.  (To just copy it into a pg file, replace
-occurences of "\&" by "~~&".)
+This macro provides a set of functions that provide a generic answer checker.
 
- Usage:
+Usage:
```
Alex-Jordan, line 10: In one of the related PRs, I noted that "Usage" was a head2. Should that happen here?

drgrice1: I think it should actually be a `head1` in this case.  It seems that the "Usage" shouldn't be part of the description, and should not be a sub header of the description section.  I am not sure where the other instance of this is, but that may need to be reviewed as well.

pstaabp: I would say most of the cases where Usage is in the POD occurs inside of function/method names and there is a short usage statement.  It's been set up as 

```
Usage:

      CODE_SNIPPET
```

If we want to give this the next level header down (often `=head3`), that would be fine, except it would show up in the left sidebar for links for every case--I believe all headers to head3 are shown as link on the left.

drgrice1:
This pull request is generally in order other than some capitalization and punctuation issues.  I put in a pull request that fixes those.  I also added a comment to the `weightedGrader.pl` macro stating that it should not be used anymore, and that its functionality is now built in to the `avg_problem_grader` with a link to its documentation.

PR #1273, opened 2025-07-12
Restructire the POD of macros (miscallaneous)

Alex-Jordan:


macros/deprecated/CofIdaho_macros.pl
```diff
@@ -1,17 +1,14 @@
 
-=pod
-
 =head1 NAME
 
-    extra macros for Intermediate Algebra problems at The College of Idaho
+CofIdaho.pl - extra macros for Intermediate Algebra problems at The College of Idaho
```
Alex-Jordan, line 4: Is it a mistake that the file is `CofIdaho_macros.pl` and this here says "CofIdaho.pl"?

macros/deprecated/PGsequentialmacros.pl
```diff
@@ -1,13 +1,10 @@
 
 =head1 NAME
 
-PGsequentialmacros.pl
-
-Provides support for writing sequential problems, where certain parts
+PGsequentialmacros.pl -  Provides support for writing sequential problems, where certain parts
```
Alex-Jordan, line 4:
```suggestion
PGsequentialmacros.pl - Provides support for writing sequential problems, where certain parts
```

macros/deprecated/CofIdaho_macros.pl
drgrice1: Yes.  The name should be the same as the file name.

One thing to note though, is that @pstaabp originally made this change for how his parsing of the POD for the sample problem search was set up.  I was using this name to show in the search.  I changed that to actually use the file name.  That means if this is wrong (as I noticed in several places when I was working on that), then the correct file name is still displayed in the search.

pstaabp (approved):

PR #1281, opened 2025-07-18
Fix a warning in the `TriangleGraphTool.pg` sample problem.

drgrice1:
The image alt tag inserted via the PGML image syntax cannot wrap multiple lines.  So that will need to be all on one line.  This is a limitation of the PGML image syntax that we should eliminate.  Image alt text can be lengthy (and still not long enough to neccessitate the new long description), so it would be nice if multiple lines were allowed.


drgrice1:
I have gone through all of the sample problems now, and they all render without warning once this and #1279 are merged.

Issue #1296, opened 2025-07-22
getPrimesInRange and randomPrime doubling truncating range


djordan99:
https://github.com/openwebwork/pg/blob/0c7f42dcaae26f074096fa3c843daa92a1e47d89/macros/contexts/contextInteger.pl#L151
https://github.com/openwebwork/pg/blob/0c7f42dcaae26f074096fa3c843daa92a1e47d89/macros/contexts/contextInteger.pl#L159

I suggest changing the less than into less than or equal to in the while condition in line 159. The strict less than prevents the list of primes generated from including the $end. This means randomPrime(3,11), for example, cannot return 11. This seems counterintuitive given the behavior of random(3,11).

Alternatively, changing the condition in line 151 to include the equality would make the error message more useful. (Although also note the typo "or" in line 151).

https://github.com/openwebwork/pg/blob/0c7f42dcaae26f074096fa3c843daa92a1e47d89/macros/contexts/contextInteger.pl#L238-L239

In line 238, I believe the 0 was meant to be -1. If there were no primes in the range, then _getPrimesInRange would return an empty array with the final index negative one.

In line 239, `($#primes - 1)` should be `$#primes`. As is, the code prevents the largest prime in the range from ever being selected.

Perhaps the intent was for randomPrime(a,b) to return a prime in [a,b). Although, again, that seems counterintuitive. But, in any case, as things stand, neither b nor the greatest prime less than b can result from randomPrime. For example, randomPrime(2,5) is always 2. 

Thanks,
Dan

drgrice1:
There are a lot of problems with this macro.  For instance, in the `_getPrimesInRange` function the variable `$self` is used, but there is no such variable defined.  So if those conditions happen, the intended error is not thrown, and instead you get "can't call Error on an undefined value".  The macro is most likely is going to be replaced soon, perhaps with a proper integer context.  The current macro will probably be renamed, or the methods moved somewhere else and fixed.

Line 238 should be `$self->Error("Could not find any prime numbers in range.") if @primes == 0;`.  The incorrect count was used.  It should be checking the number of elements of the array, instead of the index of the last element in the array.

somiaj:
Fixed in #1297.

PR #1297, opened 2025-07-22
Fix several issues with the `contextInteger.pl` macro.

drgrice1:
Several times in the file the variable `$self` is used where it is not defined.  So if the `$self->Error` method is actually called it results in the error "can't call Error on an undefined value".

Also the wrong counts are used in the `randomPrime` method resulting in unexpected behavior.

Finally, if the `$end` parameter to the `_getPrimesInRange` is prime, then include it in the returned primes in the range.  Thus, the `randomPrime` method will potentially be the `$end` parameter.  I.e., `randomPrime(1, 11)` could return 11.

This is to address the things discussed in #1296.

### DIFF
--- a/macros/contexts/contextInteger.pl
+++ b/macros/contexts/contextInteger.pl
@@ -74,7 +74,7 @@ Determine if the given integer is prime.
 
 =cut
 
-loadMacros('MathObjects.pl');
+loadMacros('PGbasicmacros.pl', 'MathObjects.pl');
 
 sub _contextInteger_init { context::Integer::Init() }
 
@@ -124,7 +124,7 @@ sub Init {
 sub _divisor {
 	my $power = abs(shift);
 	my $a     = abs(shift);
-	$self->Error("Cannot perform divisor function on Zero") if $a == 0;
+	Value::Error('Cannot perform divisor function on Zero') if $a == 0;
 	$result = 1;
 	$sqrt_a = int(sqrt($a));
 	for (my $i = 2; $i < $sqrt_a; $i++) {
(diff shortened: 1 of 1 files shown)


macros/contexts/contextInteger.pl
```diff
@@ -235,8 +235,8 @@ sub isPrime {
 sub randomPrime {
 	my ($start, $end) = @_;
 	my @primes = context::Integer::_getPrimesInRange($start, $end);
-	$self->Error("Could not find any prime numbers in range.") if $#primes == 0;
-	my $primeIndex = $main::PG_random_generator->random(0, ($#primes - 1), 1);
+	Value::Error('Could not find any prime numbers in range.') if @primes == 0;
+	my $primeIndex = main::random(0, $#primes);
 	return $primes[$primeIndex];
```
Alex-Jordan, line 240: Lines 239 and 240 could be replaced with:
```
return main::list_random(@primes);

File: macros/contexts/contextInteger.pl [hunk 2]
```diff
@@ -235,8 +235,8 @@ sub isPrime {
 sub randomPrime {
 	my ($start, $end) = @_;
 	my @primes = context::Integer::_getPrimesInRange($start, $end);
-	$self->Error("Could not find any prime numbers in range.") if $#primes == 0;
-	my $primeIndex = $main::PG_random_generator->random(0, ($#primes - 1), 1);
+	Value::Error('Could not find any prime numbers in range.') if @primes == 0;
```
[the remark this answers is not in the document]
drgrice1: I certainly prefer `... unless @primes;`.  I switched it to that.  I don't know why I didn't do that to begin with.

macros/contexts/contextInteger.pl
drgrice1: That is also better.  Thanks.

Alex-Jordan (approved):

Issue #1275, opened 2025-07-14
PG uniq does not preserve order


dlglin:
The `uniq` function in PGbasicmacros.pl returns an array of unique items from the input array, but it doesn't maintain a consistent order of the items in the returned array.

Consider the following MWE
```
DOCUMENT();
loadMacros("PGstandard.pl",
           "parserRadioButtons.pl",
          "PGML.pl");

## Do NOT show partial correct answers
$showPartialCorrectAnswers = 0;

@answers = ('a','b','c','b','a');
@options = uniq(@answers);

$mc = RadioButtons([@options],'b');

BEGIN_PGML
[_]{$mc}
END_PGML

ENDDOCUMENT();
```
Each time you load the problem the options can be presented in a different order, which means that even if you select the correct answer it may be marked incorrect since the array could be reordered on submit.

All of the examples that I found in the OPL (Contrib actually) take this into account and sort the resulting array, so it may be that this behaviour is understood, but there is a comment in the code that says "sort is causing trouble with Safe.??", which makes me think that the intention was to have the function sort the array.

drgrice1:
Note that it is documented in the POD for the `uniq` method that the order is unspecified.  So this behavior is technically not an issue, as it is the expected behavior of the method.  Of course the reason it is documented that way is probably also related to the comment you mentioned regarding sort in the safe compartment.  But regardless of the reason it is implemented this way, at this point it means that sorting the result is something that the caller is expected to do if it is desired to have the result sorted.

One problem with sorting the `uniq` return value is choosing which sort to use.  `lex_sort` and `num_sort` are different sorts and would give different results depending on the types of the elements in the list that is passed to the `uniq` method.  In fact , you really couldn't use `num_sort` because then you can't even pass non-numeric values to the method (as in your example) without getting warnings.  So you really have to use `lex_sort`.  But then it may be unexpected when you get back `1, 11, 2, 3` when you call `uniq(1, 2, 3, 11)`.

So in the end I think that it is actually best to leave this as it is, and tell the caller to sort the result if desired.  The POD documentation should be made more concise regarding this though.

Another possibility is to reimplement the `uniq` method as

```perl
sub uniq {
	my @in = @_;
	my %seen;
	return grep { !$seen{$_}++ } @in;
}
```

Then the order will always be the same, but will be dependent on the order of the input.  For example, if you call `uniq('a', 'c', 'b', 'a', 'c')` it will always return `'a', 'c', 'b'`.

PR #1298, opened 2025-07-24
Rewrite the PG uniq function to return the array in a consistent order

dlglin:
This is to address #1275.  Find a MWE to test there.

With this update (supplied by @drgrice1 ) `uniq` now always returns the array in the same order.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -2634,13 +2634,9 @@ Usage:
 
 #  uniq gives unique elements of a list:
 sub uniq {
-	my @in   = @_;
-	my %temp = ();
-	while (@in) {
-		$temp{ shift(@in) }++;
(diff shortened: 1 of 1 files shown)


duffee:
Your other option is to use the List::Util's [uniq](https://metacpan.org/pod/List::Util#uniq) function. The module has been in core since [5.7-5.10](https://stackoverflow.com/questions/8694638/isnt-listutil-part-of-standard-perl-distribution#8694656) which is old enough to expect it in all(?) distributions. That decision hinges on your feeling about adding dependencies. Like the `grep` solution above, it also returns results in the same order and _may_ have been optimized for speed. Just a thought.

drgrice1:
> Does this work because `grep` keeps the same order as the input? Whereas using a hash doesn't preserve order?

Yes.  That is basically it.

Issue #1300, opened 2025-08-06
checkboxList failure in hardcopy if within a table


Alex-Jordan:
`parserCheckboxList` (and perhaps other multiple more) uses `$BR` as the default separator between items. Something like that is necessary for the HTML output, but not necessary for TeX or PTX output. In TeX, the use of `\item` for each item naturally moves to a new line.

But worse, the use of `$BR` in hardcopy can cause a problem. For hardcopy, `$BR` makes `\leavevmode\\\relax `. And if the checkbox list is inside a table cell, this causes the hardcopy to fail to build.

I'm not sure exactly what the solution should be. Several options occur to me. One is that since a separator doesn't really make sense for TeX or PTX, to make it so the separator is just ignored in those output formats.

PR #1304, opened 2025-08-10
Ignore the "separator" for parserCheckboxList.pl and parserRadioButtons.pl in TeX display mode.

drgrice1:
The default "separator" is a $BR which is defined to be `\\leavemode\\\\\\relax` if the display mode is TeX.  This doesn't work if the CheckboxList or RadioButtons happen to be in a table. Furthermore, the separator doesn't make sense in TeX display mode in any case.  So this makes it so that the separator is just not used in the TeX display mode.

This is to address issue #1300.

### DIFF
--- a/macros/parsers/parserCheckboxList.pl
+++ b/macros/parsers/parserCheckboxList.pl
@@ -614,7 +614,10 @@ sub CHECKS {
 		$checks[-1] .= "</div>";
 	}
 
-	return wantarray ? @checks : join($main::displayMode eq 'PTX' ? '' : $self->{separator}, @checks);
+	return
+		wantarray
+		? @checks
+		: join($main::displayMode eq 'PTX' || $main::displayMode eq 'TeX' ? '' : $self->{separator}, @checks);
 }
 
 sub protect {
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
Should the POD get into this? There's a place where it explains `separator` but not in a way that clarifies it will only be used in HTML.

pstaabp:
I think there's a general latex issue with itemize inside of tabular.   There's a suggestion: https://tex.stackexchange.com/questions/150492/how-to-use-itemize-in-table-environment to using the `p{}` option in the column specifier.  I switched the code to that and this did work.

drgrice1:
@pstaabp:  Yes, to use a `CheckboxList` in a table you need to use `p{}`.  I don't think there is any way to make `niceTables` automatically detect that you are putting a `CheckboxList` into a table cell, so the author will need to add that.  You can do this with
```
DOCUMENT();

loadMacros('PGstandard.pl', 'PGML.pl', 'parserCheckboxList.pl');
$checks = CheckboxList(['First', 'Second', 'Third'], [0, 1]);

BEGIN_PGML
Select First and Third

[# [. [_]{$checks} .] #]*{ align => 'p{2in}' }
END_PGML

ENDDOCUMENT();
```

The problem also looks quite bad without that in HTML.  I changed to "First", "Second", and "Third" to emphasize the point here.  This is a screenshot of what you get without the `p` alignment:

checkboxs-in-table

PR #1307, opened 2025-08-24
Fix parameter check in BeginList function

pbhenson:
This fixes two issues with a check in BeginList; first, it accessed a parameter without checking it if existed first, resulting in undefined value warnings; and second it tested the validity of the uppercase version of the parameter but then used the literal parameter in the assignment.


Alex-Jordan:
Can you redo this based on the develop branch? Anything that is not so urgent that it should be a hotfix to the current version of webwork/pg should be submitted as a pull request to the develop branch.

PR #1309, opened 2025-08-25
Fix parameter check in BeginList function

pbhenson:
This fixes two issues with a check in BeginList; first, it accessed a parameter without checking it if existed first, resulting in undefined value warnings; and second it tested the validity of the uppercase version of the parameter but then used the literal parameter in the assignment.


drgrice1 (approved):
This looks good.   Thanks for this.

PR #1310, opened 2025-08-29
Charge notation in contextReaction.pl 

sfiedle1:
This pull request adds a context flag "showUnity", which governs the TeX display of charges in singly charged ions. Presently, ions are displayed in TeX using 1+/1- notation. This pull request proposes to: A) Change the default notation of singly charged ions to a more conventional +/- notation and B) Permit the original 1+/1- notation with use of the showUnity=>1 flag. Such functionality is desirable for writing chemistry problems since since the display of ions in the problems is passed in TeX form via the [` `] syntax.

Below is a MWE that demonstrates its use.

     DOCUMENT();
     loadMacros('PGstandard.pl','PGML.pl','contextReaction.pl',"contextArbitraryString.pl");

     $ion_name="sodium ion";

     Context("Reaction");
     $ion=Formula("Na^+1");

     Context("ArbitraryString");
     BEGIN_PGML
      Give the correct ion name for [`[$ion]`] 

       [`\hspace{3em}`][_]{$ion_name} (The answer is: [$ion_name]).

     END_PGML

     Context("Reaction")->flags->set(showUnity=>1);
     $ion=Formula("Na^+1");
     Context("ArbitraryString");
     BEGIN_PGML
      Give the correct ion name for [`[$ion]`] 

       [`\hspace{3em}`][_]{$ion_name} (The answer is: [$ion_name]).

     END_PGML

    ENDDOCUMENT();

A small additional edit was also proposed in the POD to use C<Ca^2+> in lieu of C<Na^2+> since alkaline earth metals are more commonly used to exemplify divalent ions. The second ionization energies of alkali elements such as sodium are typically prohibitively high for them to be observed in nature.


drgrice1:
I changed the target to the develop branch.  It seems to have been created on the develop branch, and that is the correct branch for pull requests.  If desired a hot fix pull request can be created to go to main.

sfiedle1:
Thank you Glenn. I am targeting the develop branch.

macros/contexts/contextReaction.pl
```diff
@@ -747,7 +752,11 @@ sub TeX {
 	my $self = shift;
 	my $uop  = $self->{def};
 	my $op   = (defined($uop->{TeX}) ? $uop->{TeX} : $uop->{string});
-	return $self->{op}->TeX($uop->{precedence}) . $op;
+	#	my $unity     = $self->context->flags->get('showUnity');
+	my $magnitude = $self->{op}->TeX($uop->{precedence});
+	#if ($unity == 0 && $magnitude == 1) { return $op }
```
somiaj, line 757: I don't think this previous code you commented out is needed in the final commit.

sfiedle1: Thank you. I removed the comments.

macros/contexts/contextReaction.pl
```diff
@@ -744,10 +749,12 @@ sub string {
...
 sub TeX {
-	my $self = shift;
-	my $uop  = $self->{def};
-	my $op   = (defined($uop->{TeX}) ? $uop->{TeX} : $uop->{string});
-	return $self->{op}->TeX($uop->{precedence}) . $op;
+	my $self      = shift;
+	my $uop       = $self->{def};
+	my $op        = (defined($uop->{TeX}) ? $uop->{TeX} : $uop->{string});
+	my $magnitude = $self->{op}->TeX($uop->{precedence});
+	return $op if $self->context->flags->get('showUnity') == 0 && $magnitude == 1;
```
dpvc, line 756: Since `$magnitude` is a string, I would recommend using `$magnitute eq '1'` instead of the numeric comparison.  Although that works, you can avoid making the numeric conversion from the string that way.

Also, note that the `string()` method above (lines 742 to 746 in your file) already does the removal of the `1`, so you should probably work your `showUnity` check into that as well.

Finally, I would recommend using `!$self->context->flags->get('showUnity')` rather than a test specifically against 0, since that will cover any false-like value (and your test will produce true if `showUnity` were set to some non-numeric string, since that will convert to 0 for the numeric test).

sfiedle1:
Thank you Davide. I agree that one return statement makes it easier to read. To avoid a display issue with di/tri-valent ions, I was a bit verbose when consolidating the return statements to one line.

drgrice1 (approved):
This looks good to me.  I tested with the provided examples at least.

pstaabp (approved):

PR #1317, opened 2025-09-28
fix logic for when to check answer hints

Alex-Jordan:
I was using answerHints.pl for an answer that should be like ln|x|, and wanted a special message if the student entered ln(x). I noticed the answer hint was given even when the student enters ln|x|. And even though the default setting for `checkCorrect` is 0.

It turned out that `checkCorrect` was only respected if the specified wrong answer was given as code, rather than given as a MathObject. So I adjusted the logic to wrap a check around the entire loop, checking if `checkCorrect` is 1 or if the score is less than 1. While doing this, it seemed a few other things should be checked outside of the loop as well. Namely, the check for `checkTypes`, and the check for if a prior message should be replaced.

### DIFF
--- a/macros/answers/answerHints.pl
+++ b/macros/answers/answerHints.pl
@@ -32,8 +32,10 @@ set options for the message.  These can include:
 
 =item C<S<< checkCorrect => 0 or 1 >>>
 
-1 means check for messages even
-if the answer is correct.
+In the list of incorrect answers to compare the student's answer with, 1 means
+for each CODE reference incorrect answer, check the student's answer against
+that CODE reference incorrect answer for messages even if the student answer is
+correct.
 Default: 0
 
 =item C<S<< replaceMessage => 0 or 1 >>>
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I took back the proposed changed, and I think I made the changes that @dpvc has suggested. Assuming I followed the thread. Not sure my commit message is accurate, and I could force push a better one if someone proposes something.

Alex-Jordan:
Wait, I thought I took away the proposed changes. Stand by.

Alex-Jordan:
OK, now it's just @dpvc's proposed changes. The title of the PR is no longer correct and the objective has changed.

dpvc (approved):
Looks good.

somiaj:
@Alex-Jordan Assuming you want to retarget this to the release branch?

Alex-Jordan:
So moved!

somiaj (approved):
Here is my approval. I see what it is doing, by temporary changing the answer hash to compare answers. I don't have anything to test. Is there some test we should do?

Alex-Jordan:
I'm not sure what to test. I tried something that @dpvc suggests would not work before this change, but it seems to work. Both with and without this change.

PR #1318, opened 2025-10-06
Fix invalid html for ans_array answers.

drgrice1:
The `format_matrix_HTML` method uses `span` tags to format the html output for matrices. This is a problem when that is used to format the output for an `ans_array` answer because the answer inputs inside now are wrapped in `div` tags (so that the feedback button works in a valid html way).  So this just switches to using `div`s instead.  This doesn't change the result at all since the containg array layout div already has `display:inline-block` set.

There is one minor tweak to the style.  I added `text-align:center;` to the cells.  This just looks better in about all of the cases that I have observed. Note that this is also consistent with when these objects are displayed in math mode via MathJax or as an image with the image display mode.

Note that method is used ans_array matrix, vector, and point answer rules, as well as the "Entered" feedback preview of the student answer for those answers. It also affects the "textual" correct answer, but that actually isn't used for anything anymore.

### DIFF
--- a/lib/Value/AnswerChecker.pm
+++ b/lib/Value/AnswerChecker.pm
@@ -555,57 +555,57 @@ sub format_matrix_HTML {
 	my ($rows, $cols) = (scalar(@{$array}), scalar(@{ $array->[0] }));
 	my $HTML  = "";
 	my $class = 'class="ans_array_cell"';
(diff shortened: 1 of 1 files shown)


dpvc:
Everything gets explicit `display` CSS settings now (that wasn't the case when this was originally written), so I don't see any problem with this.

dpvc (approved):

PR #1322, opened 2025-10-08
Fix some issues with plots.pl and the graph tool.

drgrice1:
First, fix the alt attribute for TikZ and GD image types for plots.pl.

The `image` method of `PGbasicmacros.pl` currently sets the `aria_description` with the value of the `alt` option if it is passed to the method, and it does so regardless of the plots image type.  However, that is only used for `html` output with JSXGraph.  Thus for TikZ or GD image types the alt attribute is lost.  This makes it so that the `aria_description` is only set for `html` output, and for all other outputs the given alt tag is left in the `@alt_list` so that the later code that inserts the `<image>` tag can get it and add it as an attribute to the `<image>` tag.

You can test this with the following MWE:

```perl
DOCUMENT();

$plot = Plot();
$plot->add_function('x^2', 'x', -10, 10, color => 'blue');
$plot->image_type('tikz');

BEGIN_PGML
[!graph of x^2!]{$plot}{300}
END_PGML

ENDDOCUMENT();
```

With that example and the current code, the image will not have an alt attribute, but will with this pull request.  If you remove the line that sets the image type to 'tikz', then the JSXGraph image will get the aria description (with the second fix below).

Second, fix the aria description for both JSXGraph output of plots.pl and the graph tool.

This is caused by the removal of the `description` option for the `JXG.Board` object in the JSXGraph library. I must have missed this when this happened three years ago. Although, it seems to have been done rather quietly, as this is not listed in the change log for JSXGraph.

To fix this, I just do the same thing that the `description` option used to do, and add a visually hidden span that the graph is `aria-describedby`. Note that there is a new `aria-description` attribute that could be used in the future for this, but it is in a future aria specification, and I don't know how well supported it is at this point.

Finally, fix some issues with GD output of the plots.pl macro.  This is caused when an Plots::Plot object does not have the height explicitly set.  For TikZ and JSXGraph output, the `size` method is called which determines the height if it is not set explicitly.  So GD output should do the same.

### DIFF
--- a/htdocs/js/GraphTool/graphtool.js
+++ b/htdocs/js/GraphTool/graphtool.js
@@ -69,7 +69,6 @@ window.graphTool = (containerId, options) => {
 	if ('htmlInputId' in options) gt.html_input = document.getElementById(options.htmlInputId);
 	const cfgOptions = {
 		title: 'WeBWorK Graph Tool',
-		description: options.ariaDescription ?? 'Interactively graph objects',
 		showCopyright: false,
 		pan: { enabled: false },
 		zoom: { enabled: false },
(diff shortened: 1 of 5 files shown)


somiaj (approved):
These changes look good. We should probably get this in for when testing your other upcoming changes.

PR #1329, opened 2025-10-15
Restructure the warning handling for the LaTeXImage.pm module.

drgrice1:
This moves the error warnings up to immediately after latex or pdflatex/xelatex (whichever is the latex2pdf program) runs.  This way these errors are always warned.

When the `TikZ.pm` module was converted the the `LaTeXImage.pm` module and restructured for PTX output, it resulted in the case that a `pdf` file is the end result (such as for hardcopy) not having these warnings issued.

This is to address issue https://github.com/openwebwork/webwork2/issues/2834.


dlglin (approved):
This now gives the full latex output when using "download hardcopy" in the problem editor, which I guess is the desired result (though it's a lot to read through).

somiaj:
Two approvals so merging.

PR #1332, opened 2025-10-28
Change all webwork.maa.org links to openwebwork.org

dlglin:
Update links to point to openwebwork.org instead of webwork.maa.org, and change those links to https.

Remove code coverage badges from README.md

### DIFF
--- a/macros/contexts/contextReaction.pl
+++ b/macros/contexts/contextReaction.pl
@@ -25,14 +25,15 @@ reactions.  For example:
 
     $R = Formula("4P + 5O_2 --> 2P_2O_5");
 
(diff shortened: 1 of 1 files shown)


somiaj:
@dlglin seems you now have a conflict since @pstaabp removed the badges as well and that has been merged.

somiaj (approved):
Outside of fixing the minor conflict, looks good. Here is an approval based on fixing the conflict. Though I do find `wiki.openwebwork.org/wiki/` redundant, it is what it is.

drgrice1 (approved):
This looks good.

I am sure you are probably aware of this, but webwork2 will need the same thing.

Issue #1340, opened 2025-11-06
contextFraction interaction with changing Context()->{format}{number}


somiaj:
Consider the following example problem (I have a problem that mixes fractions and decimals, and I wanted the decimals to all be rounded to three decimal digits that this came from).

```
DOCUMENT();
loadMacros(qw(PGstandard.pl contextFraction.pl PGML.pl));

Context('Fraction');
Context()->{format}{number} = "%.3g";
$x = Fraction('27375/17');
BEGIN_PGML
[`[$x]`]
END_PGML
ENDDOCUMENT();
```

This problem gives the error:

```
ERRORS from evaluating PG file:
Fraction numerators must be integers at line 1 of (eval 1729)
   from within Value::Error called at line 855 of [PG]/macros/contexts/contextFraction.pl
   from within context::Fraction::Value::Fraction::new called at line 1 of (eval 1729)
   from within main::Fraction called at line 6 of undefined
```

The issue I ran into was a little different, where fractions where being printed using SI notation, as in this example.

```
DOCUMENT();
loadMacros(qw(PGstandard.pl contextFraction.pl PGML.pl));

Context('Fraction');
Context()->{format}{number} = "%.3g";
$x = Fraction('11/70') / Fraction('3/55') * 40;
BEGIN_PGML
[`[$x]`]
END_PGML
ENDDOCUMENT();
```

The faction should have been `2420/21` and if I create the fraction with `Fraction('2420') / Fraction('21')` I don't see an issue, but creating it using fraction operations did cause the fraction to be displayed using SI notation, `2.42E+3/21`.

drgrice1:
This seems to have to do with the change to line 927 of contextFraction.pl made in #1208.  If I change that line back to `return $a / $b`, then both of the examples you gave work as expected.  However, that causes issue #1204 that was fixed by that pull request to recur.

In general it seems that this is a rounding issue.  The `context::Fraction::Value::Fraction::new` method calls `toFraction` to convert the number into a fraction.  However, at that point the value it is passed is rounded to two decimal places (even without `Context()->{format}{number} = "%.3g";`).  As a result the returned fraction is incorrect.

dpvc:
[...]

PR #1345, opened 2025-11-07
Possible fix for the eval issue. (hotfix of #1344)

drgrice1:
This is a possible fix for #1340.  This is now @dpvc's suggested code from his comment in #1340.


somiaj (approved):

dlglin (approved):
#1344 has two approvals

PR #1346, opened 2025-11-08
Add a workflow to generate and publish PG POD and sample problem documentation.

drgrice1:
This is intended to replace the openwebwork/pg-docs repository.  The approach in that repository has some issues.

* The workflows must be manually triggered in sequence.
* It isn't tied to the PG repository and so can become and in fact is now out of date (actually I updated it so it is not out of date anymore, but could easily become so again).
* It unnecessarily commits the generated documentation files to a repository as well as publishing them on GitHub pages.
* It uses the `bin/dev_script/generate-ww-pg-pod.pl` script from the webwork2 repository to generate the PG and webwork2 POD, and the webwork2 POD should not be on a site that is intended for PG documentation. Furthermore, the generated html POD site does not fit into the PG documentation site well.  For, example there is no link to get back to the main PG documentation index page, and it is labeled as WeBorK documentation, not PG documentation.

The single workflow added in this pull request runs anytime anything is merged into the main branch.  It can also be manually triggered but that generally won't be needed.  This means that anytime a new version of PG is released the documentation will automatically be updated for that release.

The workflow generates the PG POD and sample problem documentation and publishes the documentation to the PG GitHub pages site at https://openwebwork.github.io/pg.  The generated documentation is not committed to a repository.  It is published to the pages site, and that is enough.

A new `bin/generate-pg-pod.pl` script is added that generates only the PG POD.  The generated html is tailored for the PG documentation on GitHub pages.  There is some minor redundancy as there are some files that are now in both the webwork2 and PG repositories (needed for both the `generate-ww-pg-pod.pl` and `generate-pg-pod.pl` scripts).  However, it certainly would not be a good idea to have the PG repository depend on the webwork2 repository for the `generate-ww-pg-pod.pl` script, and as mentioned that script doesn't really generated the right HTML to begin with.

Move the `SampleProblemParser` module into the `WeBWorK::PG` namespace. All of the other modules in the root namespace in the PG repository are directly for problem rendering except this one.  It is out of place.  So it is now the `WeBWorK::PG::SampleProblemParser` module.

Change the options for the `parser-problem-doc.pl` script and the `WeBWorK::PG::SampleProblemParser` module.  Using underscores for options is annoying.  Underscores require an extra keyboard button press (the shift key), and underscore are not standard for command line options.  Also the `pod_root` and `pg_doc_home` options are badly named. The `pod_root` option name seems to indicate it should be the directory location for the POD, but it is a URL.  So make that clear by renaming it to `pod-base-url`.  The `pg_doc_home` is not the URL for the PG documentation home, but for the sample problem base URL.  So rename that to `sample-problem-base-url` (and its shortcut from "h" to "s").  Also change the corresponding variable names in the `SampleProblemParser.pm` module.

Copy the PODParser.pm and PODtoHTML.pm modules from the webwork2 repository to lib/WeBWorK/Utils where webwork2 can also use them.  Those modules will be deleted from the webwork2 repository.

Copy podviewer.css and podviewer.js from the webwork2 repository into htdocs/js/PODViewer, also where webwork2 can use them. The files will be deleted from the webwork2 repository.

Copy the bin/dev_scripts/pod-templates category-index.mt and pod.mt files into assets/pod-templates here, and make changes to the files so that they will work for both webwork2 and pg. The files will also be deleted from the webwork2 repository.

Add the sample problem and macro POD search to PG docs.  The sample problem and macro POD search data is generated by the `WeBWorK::PG::SampleProblemParser::getSearchData` method.  This is used by webwork2 and the `bin/generate-search-data.pl` script.  The script is very simple.  It just calls the method passing in the file name to save the data to.  The workflow runs the script and copies the resulting file to the PG github pages site.

The `assets/stop-words-en.txt` file is moved here from the webwork2 repository since the `WeBWorK::PG::SampleProblemParser::getSearchData` needs it.

The `htdocs/js/SampleProblemViewer/documentation-search.js` is moved from the webwork2 repository (webwork2 uses it from here now), and this is also copied to the PG github pages site in the workflow.  It is slightly modified to work for both webwork2 and on the PG github pages site.

The necessary changes to the webwork2 repository are in https://github.com/openwebwork/webwork2/pull/2842.


somiaj (approved):
POD and sample problem generation works. Didn't test the workflow. I trust it will work and can be fixed needed.

PR #1356, opened 2025-12-21
Revert JSXGraph back to version 1.10.1.

drgrice1:
Something has changed in version 1.11.1 that is messing up the tab order when using the tab key to change focus between points in the graph and such.  I will need to update the keyboard handling code in the graphtool to fix this.  So for now this reverts back to version 1.10.1 for which the current graphtool code still works.

I will try to get this fixed an update back to 1.11.1 before the next release.


somiaj:
Closing now that #1357 is merged.

PR #1357, opened 2025-12-22
Upgrade JSXGraph to the latest version and fix the graphtool for that.

drgrice1:
The latest version of JXSGraph is 1.12.2.  However, that version (and all versions after 1.11.1) have an issue with tab order when keyboard events are enabled.  Basically, it is impossible to use shift-tab to progress in reverse in the tab order.  See https://github.com/jsxgraph/jsxgraph/issues/773. So to work around that I had to override the board's `keyDownListener` method with one that does not call `preventDefault` on a keydown event that comes from a tab key being used.

Another thing that is a bit annoying with versions 1.11.1 and later is that you now have to set the tabindex on elements that are not fixed yourself.  By default they set the tabindex to -1, which means they are not in the tab order.  So `gt.definingPointAttributes` is now a function, and if it is called with `gt.isStatic` true, a tabindex of -1 is used, but if `gt.isStatic` is false, then a tabindex of 0 is used (and so those points will be keyboard focusable).

### DIFF
--- a/htdocs/js/GraphTool/circletool.js
+++ b/htdocs/js/GraphTool/circletool.js
@@ -188,7 +188,7 @@
 					const center = this.center;
 					delete this.center;
 
-					center.setAttribute(gt.definingPointAttributes);
+					center.setAttribute(gt.definingPointAttributes());
 					center.on('down', () => gt.onPointDown(center));
 					center.on('up', () => gt.onPointUp(center));
 
--- a/htdocs/js/GraphTool/cubictool.js
+++ b/htdocs/js/GraphTool/cubictool.js
@@ -12,7 +12,7 @@
 
 				constructor(point1, point2, point3, point4, solid) {
 					for (const point of [point1, point2, point3, point4]) {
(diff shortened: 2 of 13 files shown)


pstaabp:
I notice that is both the develop version and this branch that in a simple problem with a graph tool, that the tab order starts with tex outside the problem, then any points on any tools, then all of the labels.  

I was expecting to go through the buttons as well to be able to focus.  This doesn't have to do with this PR, but just noticed it.

pstaabp:
> I don't know what you mean by "I was expecting to go through the buttons as well to be able to focus". It does go through the tool buttons below, both with the develop branch and this pull request.

This looks like another Safari quirk/bug.  Skips the buttons in Safari.  I notice it works as expected in Chrome.

drgrice1:
That is certainly odd.  The buttons are just standard HTML buttons which should be in the tab order, and should work in all browsers in the expected way.

pstaabp:
I guess it's a setting in Safari.  I put that on and all of the buttons now are focusable.

Issue #1355, opened 2025-12-15
PGML Par can create bad HTML by inserting empty divs for spacing


somiaj:
PGML Implements the Par block in HTML by inserting an empty `div` with a margin on it for spacing. When used inside of a list this puts these div's as children of `ul` or `ol` elements, which is considered invalid HTML.

I don't know the PGML parser that well, but can block elements look forward/backwards, so instead of Par placing the margin as a separate `div`, the margin is added to the previous block. For example:

```
BEGIN_PGML
1. First

2. Second

3. Third
END_PGML
```

Produces the following HTML:

```html
<div class="PGML">
    <ol style="margin:0; padding-left:2.25em; list-style-type: decimal;">
        <li>First</li>
        <div style="margin-top:1em"></div>
        <li>Second</li>
        <div style="margin-top:1em"></div>
        <li>Third</li>
    </ol>
</div>
```

Instead could the Par blocks some how tag the previous block element so the style is applied there instead like (note placing `margin-top` on the following block element might also be a possibility):

```html
<div class="PGML">
    <ol style="margin:0; padding-left:2.25em; list-style-type: decimal;">
        <li style="margin-bottom:1em;">First</li>
        <li style="margin-bottom:1em;">Second</li>
        <li>Third</li>
    </ol>
</div>
```

Tagging @dpvc as I suspect they understand the PGML parser best and if this is possible and how much work it would take to implement.

PR #1358, opened 2025-12-27
PGML::Format Refactoring

somiaj:
Refactor the PGML::Format package, which includes changing how it is used, and cleaning up HTML output (including fixing some invalid HTML issues).

+ Refactors how the methods to format the different PGML::Format item methods are called. Instead of using
an odd syntax for switch/cases, `/foo/ && do { ....; last }`, just convert the item `type` into a method name and test if that method exists. Also all item methods are sent the same input, a `state` hash which includes the data they need to determine what output to produce.

+ Fix some issues with invalid HTML, this means that the margin created by `par` blocks are now applied
to the previous list, list item, or indent div, instead of creating an empty div. Also fixed an issue where an `hr` tag was inside a `span` tag.

+ Remove newlines and extra spaces from HTML output. These are extra characters that don't need to be sent to the client.

+ Use `main::tag` to create the HTML tags used in the HTML output.

Based on the code from @dpvc in #1355 and in the review of this PR.

### DIFF
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -1292,7 +1292,7 @@ sub nl {
 	return $nl;
 }
 
-sub Escape { shift; shift }
+sub escape { shift; shift }
 
 sub Indent   { return "" }
 sub Align    { return "" }
@@ -1439,7 +1439,7 @@ sub Command {
 	my $item = $state->{item};
 	my $text = $self->{parser}->replaceCommand($item);
 	$text = PGML::LaTeX($text) if ($item->{hasStar} || 0) == 3;
-	$text = $self->Escape($text) unless $item->{hasStar};
+	$text = $self->escape($text) unless $item->{hasStar};
 	return $text;
 }
 
@@ -1448,7 +1448,7 @@ sub Variable {
 	my $item = $state->{item};
 	my $text = $self->{parser}->replaceVariable($item, $state->{block});
 	$text = PGML::LaTeX($text) if ($item->{hasStar} || 0) == 3;
-	$text = $self->Escape($text) unless $item->{hasStar};
+	$text = $self->escape($text) unless $item->{hasStar};
 	return $text;
 }
 
@@ -1456,7 +1456,7 @@ sub Text {
 	my ($self, $state) = @_;
 	my $text = $self->{parser}->replaceText($state->{item});
 	$text =~ s/^\n+// if substr($text, 0, 1) eq "\n" && $self->nl eq "";
-	return $self->Escape($text);
+	return $self->escape($text);
 }
 
 sub Image {
@@ -1486,7 +1486,7 @@ sub Image {
 package PGML::Format::html;
 our @ISA = ('PGML::Format');
 
-sub Escape {
+sub escape {
 	my ($self, $string) = @_;
 	return '' unless defined $string;
 	$string =~ s/&/\&amp;/g;
@@ -1647,7 +1647,7 @@ sub Rule {
 sub Verbatim {
 	my ($self, $state) = @_;
 	my $item = $state->{item};
-	my $text = $self->Escape($item->{text});
+	my $text = $self->escape($item->{text});
 	$text = "<code>$text</code>" if $item->{hasStar};
 	return $text;
 }
@@ -1702,7 +1702,7 @@ my %escape = (
 	'~'  => '{\ttfamily\char126}',
 );
 
-sub Escape {
+sub escape {
 	my ($self, $string) = @_;
 	return '' unless defined($string);
 	$string =~ s/(["\#\$%&<>\\^_\{|\}~])/$escape{$1}/eg;
@@ -1805,7 +1805,7 @@ sub Rule {
 sub Verbatim {
 	my ($self, $state) = @_;
 	my $item = $state->{item};
-	my $text = $self->Escape($item->{text});
+	my $text = $self->escape($item->{text});
 	$text = "{\\tt{}$text}" if $item->{hasStar};
 	return $text;
 }
@@ -1833,7 +1833,7 @@ sub Tag {
 package PGML::Format::ptx;
 our @ISA = ('PGML::Format');
 
-sub Escape {
+sub escape {
 	my ($self, $string) = @_;
 	return '' unless defined $string;
 	$string =~ s/&/&amp;/g;
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -1270,7 +1270,7 @@ sub string {
 	my $stack = $block->{stack};
 	my @strings;
 	my $state = { block => $block, strings => \@strings, i => 0 };
-	while ($state->{i} <= $#$stack) {
+	while ($state->{i} < @$stack) {
 		$state->{item} = $stack->[ $state->{i}++ ];
 		$self->{nl}    = (!defined($strings[-1]) || $strings[-1] =~ m/\n$/ ? '' : "\n");
 		my $method = ucfirst($state->{item}{type});
@@ -1323,7 +1323,7 @@ sub Table {
 	my $table = [];
 	my $row   = [];
 	for my $cell (@{ $item->{stack} }) {
-		push(@$row, $self->Table($cell));
+		push(@$row, $self->Table({ item => $cell }));
 		if ($cell->{hasStar}) {
 			push(@$table, $row);
 			$row = [];
--- a/macros/core/PGML.pl
+++ b/macros/core/PGML.pl
@@ -1510,32 +1510,25 @@ sub Indent {
 		$bottom = '1em';
 		$state->{i}++;
 	}
-	return $self->nl . "<div style=\"margin:0 0 $bottom ${em}em\">\n" . $self->string($item) . $self->nl . "</div>\n";
+	return main::tag('div', style => "margin:0 0 $bottom ${em}em", $self->string($item));
 }
 
 sub Align {
 	my ($self, $state) = @_;
 	my $item = $state->{item};
-	return
-		$self->nl
-		. '<div style="text-align:'
-		. $item->{align}
-		. '; margin:0">' . "\n"
-		. $self->string($item)
-		. $self->nl
-		. "</div>\n";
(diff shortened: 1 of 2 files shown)


macros/core/PGML.pl
```diff
@@ -1266,19 +1266,22 @@ sub format {
...
-	my $block   = shift;
+	my ($self, $block) = @_;
+	my $stack   = $block->{stack};
+	my $n       = scalar(@$stack);
 	my @strings = ();
 	my $string;
-	foreach my $item (@{ $block->{stack} }) {
+	for my $i (0 .. $n - 1) {
+		my $item     = $stack->[$i];
+		my $next_par = $i + 1 < $n && $stack->[ $i + 1 ]{type} eq 'par' ? $stack->[ $i + 1 ] : undef;
```
drgrice1: Change this to
```suggestion
	for my $i (0 .. $#$stack) {
		my $item     = $stack->[$i];
		my $next_par = $i < $#$stack && $stack->[ $i + 1 ]{type} eq 'par' ? $stack->[ $i + 1 ] : undef;
```
and delete `my $n = scalar(@$stack);` above (line 1271).  Perl is optimized for this and stores this number as a property of the array.  So there is no reason to save the array length to yet another location.  Also, perl offers the `$#` syntax for getting the index of the last entry in an array, so use that rather than adding or subtracting one from the array length.

drgrice1 (dismissed):
This looks good.

I think that the essential thing is that the `div`s that were added between list items be fixed, since that is invalid HTML.  Adding the style to other block items (such as the lists themself) instead of injecting an empty `div` after it is not particularly important since that is at least valid HTML.  I am not entirely certain what benefit that provides either.

somiaj:
I am unsure on the benefit as well, but I do feel the empty divs for spacing is not ideal. I was also wondering if there were any accessibility issues with not using `<p>` tags or grouping text. But to me the PGML output is doesn't feel correct.

```html
Paragraph one.
<div style="margin-top:1em"></div>
Paragraph two.
<div style="margin-top:1em"></div>
Paragraph three.
```

And this just feels better to me. But I couldn't figure out a way to do this, and assume that if the PGML parser was setup for this, `<p>` tags would have just been used in the first place.

```html
<div style="margin-bottom:1em">
Paragraph one.
</div>
<div style="margin-bottom:1em">
Paragraph two.
</div>
<div style="margin:0">
Paragraph three.
</div>
```

somiaj:
@dpvc Thanks. I am looking at implementing what you stated, but unsure how to modify the `Quote` methods. That method is sent the previous string, not the full block. It appears it is doing a test to determine if it needs to open or close the quote.

dpvc:
Sorry, I missed the string being sent to the `Quote` method.  You could add that to the `$self->$method(...)` call arguments.  Alternatively, you could create a `$state` variable that holds any data needed to be passed to the item implementations.  Something like

``` perl
my $state = {
  block => $block,
  strings => [],
  i => 0,
};
while ($state->{i} < $#$stack) {
  my $item = $self->{item} = $stack->[$state->{i}++];
  $self->{nl} = (!defined($state->{strings}[-1]) || $state->{strings}[-1] =~ m/\n$/ ? "" : "\n");
  my $method = ucfirst($item->{type});
  if (Value::can($self, $method)) {
    my $string = $self->$method($item, $state);
    push(@{$state->{strings}}, $string) unless !defined $string || $string eq '';
  } else {
    PGML::Warning "Warning: unknown block type '$item->{type}' in " . ref($self) . "::format\n";
  }
}
```

Then `Quote()` can look in the `$state->{strings}` array, and `Bullet()` can use `$state->{block}` and increment `$state->{i}` if it needs to.  Note that here `$state->{i}` will be pointing to the *next* item in the stack, as it is post-incremented when getting `$item`, so `Bullet()` can use `$state->{i}` rather than `$state->{i} + 1`.

Of course, this requires changes to the implementation functions that previously used `$block` and `$i` as arguments, but there were only a couple of those.

somiaj:
Implemented the suggested changes to the main `PGML::format::string` loop as suggested by @dvpc. One difference is I am treating the `$block` and `$state` as two different variables passed to all methods, this way no update is needed to the existing methods that just use the `$block`.

somiaj:
Seems my change missed adding a new line character that makes one of the unit tests fails. Trying to track that down unless someone else sees where a newline wasn't added correctly.

drgrice1 (commented):
While we are fixing HTML validation issues here, there is one more that I have observed.  On line 1638 (in this pull request at least), there is a `span` tag that contains an `hr` tag.  It is invalid for a `span` to contain an `hr`.  That can be easily fixed by changing that `span` to a `div`.  That will not affect the resulting display, and is then valid HTML.

You could add that to this pull request if you like.

macros/core/PGML.pl
```diff
@@ -1605,7 +1588,7 @@ sub Heading {
 sub Par {
 	my $self = shift;
 	my $item = shift;
-	return $self->nl . '<div style="margin-top:1em"></div>' . "\n";
+	return '<div style="margin-top:1em"></div>' . "\n";
```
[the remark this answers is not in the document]
drgrice1: Yeah, lets leave that for another pull request.

macros/core/PGML.pl
```diff
@@ -1523,17 +1501,15 @@ sub Escape {
...
-	return
-		$self->nl
-		. '<div style="margin:0 0 0 '
-		. $em . 'em">' . "\n"
-		. $self->string($item)
-		. $self->nl
-		. "</div>\n";
+	my $em     = 2.25 * $item->{indent};
+	my $bottom = '0';
+	if (defined $block->{stack}[ $state->{i} ] && $block->{stack}[ $state->{i} ]{type} eq 'par') {
```
dpvc: It might be useful to do

``` perl
my $next = $block->{stack}[ $state->{i} ];
if (defined $next && $next->type eq 'par') {
```

to reduce the number of dereferences that need to be performed.  Similarly for lines 1544 and 1558.

macros/core/PGML.pl
```diff
@@ -1523,17 +1501,15 @@ sub Escape {
...
+		$bottom = '1em';
+		$state->{i}++;
+	}
+	return $self->nl . "<div style=\"margin: 0 0 $bottom ${em}em;\">\n" . $self->string($item) . $self->nl . "</div>\n";
```
dpvc: If you really are trying to save individual characters being part of the HTML output, then remove the semi-colon after '${em}em' and the `\n` at the end, here.

macros/core/PGML.pl
```diff
@@ -1562,20 +1538,27 @@ sub Align {
...
 
 sub List {
-	my $self = shift;
-	my $item = shift;
-	my $list = $bullet{ $item->{bullet} };
+	my ($self, $item, $block, $state) = @_;
+	my $list   = $bullet{ $item->{bullet} };
+	my $margin = '0';
+	if (defined $block->{stack}[ $state->{i} ] && $block->{stack}[ $state->{i} ]{type} eq 'par') {
+		$margin = '0 0 1em 0';
```
dpvc: Again, if you want to reduce characters, leave of the final ` 0`, which is implied by the second `0`.

dpvc (dismissed):
I'm OK with this as is, but regarding passing `$block`, I think it makes more sense to put that into the state and modify the few functions that use that.  I almost put `$item` into that as well, but didn't want you to have to edit *every* function.  But there are only a couple that use `$block`.

dpvc:
> Do either of you happen to have a problem file which uses all of PGML syntax just to double check I didn't put a typo somewhere?

Unfortunately, no.  I had some PG-based slides from when I helped teach a summer course in WeBWorK authoring, and I did some sessions on PGML and other things like MathObjects.  They had live examples, as I recall, that went through most of the features.  But the PG problems had a lot of version-specific hacks to slim down the user interface and adjust the CSS styling, and I suspect they would no longer work in today's version of WW.  But I could package them up for you if you want them.

Alex-Jordan:
Would it help to test by going to the PGML samples from the editor?

somiaj:
@Alex-Jordan that is all I am doing. For somethings I looked at the inspector, but mostly just want to make sure the output looks like it should and no warnings appear. I have created a test problem with most common PGML syntax, but I am afraid I might miss some less common setups. Since I didn't change the parser at all (only output), only have to worry that the output looks correct.

dpvc (approved):
Good work on this.  I approve.

drgrice1 (changes requested):
There is a problem with the `Table` method that is breaking `niceTables.pl` usage.  The `Table` method calls itself with the cell items.  So a "stack" object needs to be passed to its inner call.  If you change line 1326 to `push(@$row, $self->Table({ item => $cell });` that fixes the issue.  The `Table` method doesn't need anything else from the `$state`, so this should be enough.  @dpvc: Can you check this?

Here is a MWE for testing:
```
DOCUMENT();

BEGIN_PGML
[#
    [.[`x`].]{ headerrow => 1 } [.[`x^2`].]*
    [.[`1`].] [.[`2`].]*
    [.[`2`].] [.[`4`].]*
    [.[`3`].] [.[`9`].]
#]
END_PGML

ENDDOCUMENT();
```

macros/core/PGML.pl
```diff
@@ -1266,43 +1266,22 @@ sub format {
...
-			/forced/   && do { $string = $self->Forced($item);                    last };
-			/comment/  && do { $string = $self->Comment($item);                   last };
-			/table/    && do { $string = $self->Table($item);                     last };
-			/tag/      && do { $string = $self->Tag($item);                       last };
-			PGML::Warning "Warning: unknown block type '$item->{type}' in " . ref($self) . "::format\n";
+	my ($self, $block) = @_;
+	my $stack = $block->{stack};
+	my @strings;
+	my $state = { block => $block, strings => \@strings, i => 0 };
+	while ($state->{i} <= $#$stack) {
```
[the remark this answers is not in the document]
dpvc: About the `$self->Table()` call, I mean.  The `<=` is also a good change.

drgrice1 (approved):
I think this is good now.

PR #1359, opened 2025-12-27
PTX stuff for parserRadioMultiAnswer

Alex-Jordan:
I think I never got around to helping make parserRadioMultiAnswer work well for PreTeXt. These changes do that with my testing.

1. For the static PTX output, the wrapper around everything should be a `ul`. And the attributes being used here make PTX understand the `ul` is for answering something.
2. No need to have anything in between one button and the next.
3. The static PTX output looks bad to have a button, and then immediately the label like "A", "B", etc. I made the call to remove the "tag" there. The correct answer will still say things like "B", even though now the radio buttons are not labeled A, B, etc. But I think this is still understandable. And this is all for static PTX output where we don't really need the labels.

### DIFF
--- a/macros/parsers/parserRadioMultiAnswer.pl
+++ b/macros/parsers/parserRadioMultiAnswer.pl
@@ -16,7 +16,7 @@ the problem.  For example:
     $rma = RadioMultiAnswer([
                ['The unique solution is \(x=\) %s and \(y=\) %s.', 5, 6],
                ['There are an infinite number of solutions parameterized by '
-                    . '\(x=\) %s and \(y=\) %s.', '23-3t', 't']
+                    . '\(x=\) %s and \(y=\) %s.', '23-3x', 'x'],
                ['There are no solutions.']
            ], 0);
 
@@ -233,6 +233,14 @@ is checked.  If this is set to "shift", unchecking requires the shift key to be 
 
 =back
 
+=item showInStatic (Default: showInStatic => 1)
+
+In static output, such as PDF or PTX, this controls whether or not the list of answer options is
+displayed.  (The text preceding the list of answer options might make printing the answer option
+list unnecessary in a static output format.)
+
+=back
+
 =cut
 
 BEGIN { strict->import }
@@ -288,6 +296,7 @@ sub new {
 		size              => undef,
 		checked           => undef,
 		uncheckable       => 0,
+		showInStatic      => 1,
 		@inputs
 	);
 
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Thanks for doing this.  Hopefully this means that usage of this macro is starting to take hold!

macros/parsers/parserRadioMultiAnswer.pl
```diff
@@ -288,8 +288,10 @@ sub new {
 		size              => undef,
 		checked           => undef,
 		uncheckable       => 0,
+		showInStatic      => 1,
 		@inputs
 	);
+	$options{originalLabels} = $options{labels};
```
drgrice1: This line isn't needed.  Line 600 below takes care of setting the `originalLabels` for the `$self` object.  This value might not be the correct value anyway, since the label option could be set using the `with` method at a later time.

drgrice1 (approved):
I think this looks good still other than a few suggestions.

macros/parsers/parserRadioMultiAnswer.pl
Alex-Jordan: I think I hit an error though, because this might never get set by line 600, if line 598 executes. And then this is referenced at line 719.

But what's the best way to do this? I want to capture if the problem author intends for 123, ABC, etc labeling and I was hoping to avoid actually inspecting the array of labels.

drgrice1: I see.  So in the case that the author passes an array of labels to begin with (and there are enough labels), then line 600 won't execute.  So just move line 600 to before line 598, and make it `$self->{originalLabels} //= $self->{labels};`.  Then the `//=` will only set `$self->{originalLabels}` the first time since it will be defined for all later calls.

Alex-Jordan:
> Hopefully this means that usage of this macro is starting to take hold!

Well, I made use of it a lot over the last year in a linear algebra course, and then in a differential equations course. Any situation where maybe something can't be done, or maybe it can, but if it can then go ahead and do it. That was all direct PG problem writing though, not with PTX.

So I set out to add an example of using this in the PTX sample docs, and that led to this PR.

macros/parsers/parserRadioMultiAnswer.pl
```diff
@@ -638,6 +648,9 @@ sub generate_aria_label {
 # Produce the answer rule.
 sub ans_rule {
 	my ($self, $size, @options) = @_;
+
+	return '' if !$self->{showInStatic} && ($main::displayMode eq 'TeX' || $main::displayMode eq 'PTX');
```
drgrice1: To ensure that the answer is recorded, this line will need to be moved to after the for loop that begins at line 661 below.  Then another copy of this line will need to be added at line 773 in the `begin_radio` method.  Also add `next if !$self->{showInStatic} && ($main::displayMode eq 'TeX' || $main::displayMode eq 'PTX');` after line 663 in the for loop below (the `named_ans_array_extension` and `named_ans_rule_extension` methods do not need to be called contrary to what I said before).

Alex-Jordan: Thanks for working that out. I made those changes and force pushed.

drgrice1:
I think this can be merged.  I don't know if @somiaj ever checked the changes made after his initial review, but it should still be okay.

PR #1363, opened 2026-01-08
PG::Critic add policy for weightedGrader.pl functions.

somiaj:
Add a `PG::Critic` policy to catch use of the deprecated `weightedGrader.pl` function calls.

### DIFF
--- a/lib/Perl/Critic/Policy/PG/ProhibitDeprecatedWeightedGrader.pm
+++ b/lib/Perl/Critic/Policy/PG/ProhibitDeprecatedWeightedGrader.pm
@@ -1,0 +1,55 @@
+package Perl::Critic::Policy::PG::ProhibitDeprecatedWeightedGrader;
+use Mojo::Base 'Perl::Critic::Policy', -signatures;
+
+use Perl::Critic::Utils qw(:severities :classification :ppi);
+
+use constant DESCRIPTION     => 'The deprecated %s function is called';
+use constant EXPLANATION     => 'The deprecated %s function should be replaced with a modern alternative.';
+use constant SCORE           => 5;
+use constant SAMPLE_PROBLEMS => [ [ 'Weighted Grader' => 'ProblemTechniques/WeightedGrader' ] ];
+use constant WEIGHTED_GRADER_METHODS => {
+	install_weighted_grader => 1,
+	WEIGHTED_ANS            => 1,
+	NAMED_WEIGHTED_ANS      => 1,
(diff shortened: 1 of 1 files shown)


lib/Perl/Critic/Policy/PG/RequireMetadata.pm
```diff
@@ -48,4 +48,15 @@ metadata attributes should be set at the beginning of the problem file before
...
+The C<DBsubject>, C<DBchapter>, C<DBsection> tags can only be specific values.
+The prefered way to edit these tags is to use the tag editor (click "Edit Tags"
+when viewing the problem). The tag editor provides drop down menus to choose
+between the valid possibilities.
+
+Since OPL metadata tags are only needed for problems in the OPL, for local
+problems it is possible to override this check by adding the following as
+the first line of the PG file.
+
+    ## no critic (PG::RequireMetadata)
```
drgrice1: This is not true.  If this comment is going to be added then it needs to be more specific.  You seem to be assuming that this is only used through the webwork2 interface.  This can also be used with the `bin/pg-critic.pl` script (or in theory other implementations).  The `no critic` annotations are shown by default, but can be disabled.  If the `-s` flag is passed to the `pg-critic.pl` script. then the `no critic` annotation is ignored and all policies are forced to apply.

drgrice1 (changes requested):
Technically the `weightedGrader.pl` macro is not yet officially deprecated.  It has not been moved to the `macros/deprecated` directory, which is the official mark of deprecation at this point.

I really don't think that it is a good idea to add a specific macro deprecation rule for a macro that is not deprecated.  Particularly when there is already a general rule for actually deprecated macros.

somiaj:
I just dropped my second commit with modifying the `PG::Metadata` POD. I wasn't exactly sure what others would think, but overall I think the descriptions in the POD can be expanded to help authors who may not be that familiar with PG know how to fix the issue.

somiaj:
I mostly removed it due to I would rather just disable it per your previous comment, than tell people to ignore it. So I think it should also be removed/disabled until the tag situation is better worked out. I find I'm just ignoring/disabling it in my use and thought others would want to do the same.

But to me, if many are just going to disable that policy, might as well just not have it, or at the least I don't want to couple it with the `weightedGrader.pl` policy for this PR at this time, and it should probably be discussed further.

drgrice1 (approved):

pstaabp (approved):

PR #1365, opened 2026-01-20
give feedback about variable potentially using wrong case

Alex-Jordan:
Suppose a context has variable `x`, but not `X` (or vice versa). And a student enters an answer with `X` in it. The usual feedback message is `variable 'X' is not defined in this context`. The change here will recognize that a case change on `X` would make a valid variable, and adds a suggestion in the feedback message that maybe this is the student's issue.

An alternative I was considering is to replace all instances of  `variable '%s' is not defined in this context` with `variable '%s' is not defined in this context; only variables '%s' are defined in this context` where the second string is the actual list of the context's variables. Would that be better?

For that matter, is what I'm proposing here a bad idea for some reason I am overlooking? Do problems ever try to hide what constitutes a valid variable?

### DIFF
--- a/lib/Parser/Variable.pm
+++ b/lib/Parser/Variable.pm
@@ -26,6 +26,14 @@ sub new {
 			$ref->[3] = $ref->[2] + length($1);
 			$equation->Error([ "'%s' is not defined in this context", $1 ], $ref);
 		}
+		my ($iname) = grep {/^$name$/i} $equation->{context}->variables->variables;
+		$equation->Error(
+			[
+				"Variable '%s' is not defined in this context, but perhaps you mean '%s', which is defined",
+				$name, $iname
+			],
+			$ref
+		) if $iname;
 		$equation->Error([ "Variable '%s' is not defined in this context", $name ], $ref);
 	}
 	$equation->Error([ "Variable '%s' is not defined in this context", $name ], $ref)


dlglin:
I also think that making these messages more meaningful to students is the right direction to go.  If the wording is clear enough, then problem authors should be able to understand what the messages are saying about the code, even if the message is less technical.

I can think of many cases where I would like to have the message tell students which variables are permitted, but as @somiaj mentioned there are other scenarios where this could be more confusing than helpful.  Another case to be careful about would be adaptive parameters, which are added as variables to the context but students shouldn't be using in their answers.

> I am for changing the feedback, I wonder if not using the word 'context' would help out students. Wonder if something like `The variable '%s' is not allowed in this answer.` could be more useful.

I agree (and so does @drgrice1) that we should get rid of the word "context" in these messages, but I don't think we can use the word "answer".  The same error message is used if a problem author creates a formula object containing a variable not defined in the context (e.g. `$f = Formula('X');`).  How about something like `You can't use the variable '%s' here.`?

dpvc:
> I wonder if not using the word 'context' would help out students

Remember that students are unaware that "context" has a technical meaning in the PG problem code, so they should understand "context" as its normal English meaning, which would be "the setting of the problem as a whole", or of that particular answer.  (The authors will understand the technical meaning, but that will have a different meaning for them than for the students.)

That being said, you can certainly change that message if the community feels there is a better one.

> I don't like this idea. I have problems that have way too many variables defined for this to be useful.

I also think this is probably a bad idea.  There definitely are cases where there are "hidden" variables (there are some in the FormulaUpToConstant objects, as I recall).  Also, the student might not have been trying to type a variable, but rather a function name, or a constant, but have gotten it wrong.  MathObjects defaults to thinking an undefined thing is a variable, but telling them the variables might not be the help they need.

> the messages try to be close to literally describing what happened in the Parser code, more than they try to assume what is happening for the user reading the message.

That is because trying to tell what the user *really* meant is very hard.  I prefer not to give misleading error messages (by making assumption about what the user was trying to do and getting that wrong), so often the only way to be sure to give a true message is to give a message about how the parsing has failed.  But I'm sure some messages could be improved.

> A much bigger project would be to find a way to centralize these messages, and make a way to translate them

Well, the `Value->Error()` method does include a means of substituting messages, and one of the goals behind that was to make translations possible.  All messages are looked up through the `Context()->{error}{msg}` hash which can be used to replace a message with another one (the original message before substituting values is used as a key in this hash, and if present, is replaced by the key's value, and then again after substitutions are made).  It would be possible to seed the `Context()->{error}{msg}` hash with translations for a given language, at least for the basic messages.  One could look for `->Error(...)` calls to find the strings and do something like a PO file from them.  Of course, problems can have their own error strings, so that complicates things, but the basic infrastructure is already there.

Alternatively, the `Context()->setError()` method could be modified to use the PG `maketext()` function (or whatever it is called) that ties into the translation mechanism.

Alex-Jordan:
This could be looked at again. Please look past the extended discussion about feedback messages in general. What this is about is captured in this screenshot:

Screenshot 2026-05-19 at 9 45 26 PM

dpvc:
> Maybe it is the word "defined" that is the issue.

How about something like "The variable 'X' has no meaning here. Perhaps..." or "This problem doesn't include the variable 'X'.  Perhaps..."?  The latter may not be true, however, if the problem uses multiple contexts, one of which has `X` and one that doesn't. Maybe "The variable 'X' should not be used here.  Perhaps...".  Note that page authors will also get this message when writing problems if a formula uses an undefined variable, so the message should also make sense for them.

That being said, I'm fine with the message in Alex's image.

drgrice1:
I think I am okay with the message in @Alex-Jordan's image. It seems to be hard to find a message that is good for students (particularly for lower level courses) and meaningful for problem authors.

drgrice1 (approved):

pstaabp (approved):

PR #1366, opened 2026-01-22
flash feedback message badges

Alex-Jordan:
This is an animation and may be controversial. Consider it an experiment.

I'm on a mission to see what can be done to get more students to realize the yellow badge means they should click to see feedback. There's a lot of not remembering this from the orientation. And a lot of just not visually registering that there's a yellow badge there. And/or not thinking that means something special.

So this change flares up the brightness of the badge when it first appears, and fades 1 second later to the normal brightness. I'm hoping to get students' attention, and have them realize "oh, maybe I should click this".

I kind of expect mixed reactions to an animation, for several good reasons. This is one place where I wouldn't mind having it.

### DIFF
--- a/htdocs/js/Problem/problem.scss
+++ b/htdocs/js/Problem/problem.scss
@@ -164,6 +164,19 @@
 			border-radius: 50%;
 			background-color: var(--bs-warning);
 			padding: 0.25rem;
+			animation: flash-in 0.75s;
+
+			@keyframes flash-in {
+				0% {
+					filter: brightness(2);
+					padding: 0.25rem;
+				}
+
+				50% {
+					filter: brightness(2);
+					padding: 0.5rem;
+				}
+			}
 		}
 	}
 


drgrice1 (approved):
I have no problem with this.  Although, the first time I tested this, I didn't even see it.  When I did it again and was looking more closely at where I knew the feedback button would appear on the page the next time, I did see it.  I also noticed that MathJax rendering could move things around and make it hard to see as well.  So in general, I wonder if this will help much.  Maybe we could make the animation continue with a few cycles or something?

Alex-Jordan:
I just pushed an update to this. The changes from before are that the opacity only drops to 50% instead of 0, it repeats the animation three times, and it alternates forward, backward, forward.

I'm sure it could be improved, but this meets the bar for me for good enough.

somiaj (approved):
I'm okay with this as is, and think it could help.

Here are some thoughts, but none need to be addressed.

I find the changing the opacity a bit odd and seeing the corner of the button through the dot.

I still wonder if flashing the whole button would make it stand out more, I could still see the this being missed as it is a small effect due to dot size.

Alex-Jordan:
OK, this is ready to look at again.

Animating the padding is something I had rejected earlier. Because changing the size of something like that introduces choppiness to the animation. But at the time I was doing things on the order of a 2 second animation. I reduced the duration to 0.75 seconds and it all happens quickly enough that I don't really notice choppiness.

Alex-Jordan:
This has two approvals. Is it OK to merge? I should not merge my own PRs.

drgrice1:
Sorry, I should have merged it when I added the second approval.  I think I missed that there already was an approval.

PR #1374, opened 2026-01-30
StatisticalPlots macro

pstaabp:
This creates a `StatisticalPlots` macro, which contains the methods add_barplot, add_histogram, add_boxplot and add_scatterplot.

There are many options for each and there is documentation as well.

This also includes an `add_rectangle` method to the `plots.pl` macro which is a wrapper for the add_dataset for creating rectangles.

This is all in a single perl file, but can split out into modules like the plots macro.  

I'm making this a draft and there are some additional options I'd like to add.  I'm also open to renaming items and adding other plots before this goes in. 

Here's a few test problems to show how to use it:

[stats.zip](https://github.com/user-attachments/files/25132560/stats.zip)

Note: updated zip file.


macros/graph/plots.pl
```diff
@@ -291,13 +291,30 @@ =head2 PLOT ARCS
 the direction of the third point. Arcs always go in the counter clockwise
 direction.
 
-    $plot->add_arc([$start_x, $start_y], [$center_x, $center_y], [$end_x, $end_y], %options);
+    $plot->add_arc([$center_x, $center_y], [$start_x, $start_y], [$end_x, $end_y], %options);
```
drgrice1: Good catch on this.  I changed the two below in #1336, but seem to have missed this one.  These all originally listed the start point first, then the center, and then the end, and the instructions above also said that was how it should be.  Contrary to the actual implementation.

lib/Plots/Plot.pm
```diff
@@ -392,6 +392,18 @@ sub add_arc {
...
+sub add_rectangle {
+	my ($self, $pt0, $pt2, %options) = @_;
+
+	Value::Error('The first point must be an array ref of length 2')
+		unless ref($pt0) eq 'ARRAY' && scalar(@$pt0) == 2;
+	Value::Error('The second point must be an array ref of length 2')
+		unless ref($pt2) eq 'ARRAY' && scalar(@$pt2) == 2;
+	my $pt1 = [ $pt2->[0], $pt0->[1] ];
+	my $pt3 = [ $pt0->[0], $pt2->[1] ];
+	return $self->add_dataset($pt0, $pt1, $pt2, $pt3, $pt0, %options);
```
drgrice1: This could succinctly be written as:

```suggestion
	return $self->add_dataset($pt0, [ $pt2->[0], $pt0->[1] ], $pt2, [ $pt0->[0], $pt2->[1] ], $pt0, %options);
```

There is no need for the intermediate variables `$pt1` and `$pt3`.

macros/graph/StatisticalPlots.pl
```diff
@@ -0,0 +1,507 @@
...
+
+StatisticalPlots.pl - A macro to create dynamic statistics plots to include in PG problems. 
+
+=head1 DESCRIPTION
+
+This macro includes a number of methods to include statistical plots in PG problems. 
+This is based on L<plots.pl> which will draw in either C<TikZ> or C<JSXGraph> format with the 
+default for the former to be used for hardcopy and the latter for HTML output.  
+
+The statistical plot available are
```
drgrice1: This should be

```suggestion
The statistical plots available are
```

somiaj:
I am a little confused, is the `lib/Plots/StatPlot.pm` an initial attempt, before making this fully a macro? Seems the same object and methods (which are more robust) are all included in `macros/graph/StatisticalPlot.pl`, which is probably a better place, and the file in `lib/Plots` should be removed?

drgrice1 (commented):
I haven't delved to deeply into this yet, but here are some initial observations.

Lines 100 and 105 are too long.  Shorten them.  Ideally the POD would be kept even shorter than the usual 120 character limit.  I think a max of 100 characters is a better limit for that, maybe even the classical 80 character limit would be better.  The POD is all textual and if you are trying to read it directly in the code it makes it easier to read if it is not so long.  Generally, you have kept the lines to 100 characters or less, but there are a few others that are inconsistently longer.

I think you should eliminate the `StatPlot.pm` module entirely.  There is not much code there, and so just put that directly into the `StatisticalPlots.pl` macro file.  The macro file is not that large anyway (at least at this point compared to many others).  The `plots.pl` macro is bigger, and it only has 5 lines of actual code!

macros/core/PGbasicmacros.pl
```diff
@@ -2923,7 +2923,7 @@ sub image {
 			);
 			next;
 		}
-		if (ref $image_item eq 'Plots::Plot') {
+		if (ref $image_item eq 'Plots::Plot' || ref $image_item eq 'Plots::StatPlot') {
```
somiaj: Would it be worthwhile to make this more general, for future extension macros. Maybe call your package `Plots::Plot::StatPlot`, and then have this only check if `ref $image_item` starts with `Plots::Plot`, that way this won't have to be updated each time someone wants to extend Plots?

lib/Plots/StatPlot.pm
```diff
@@ -0,0 +1,38 @@
...
+
+See L<StatisticalPlots.pl> for more details.
+=cut
+
+package Plots::StatPlot;
+
+use strict;
+use warnings;
+
+use WeBWorK::Utils qw(min max);
```
drgrice1: I just noticed this.  This definitely can't be.  A PG package can not use a WeBWorK package or the methods from it.

macros/core/PGbasicmacros.pl
drgrice1: Also, line 2946 below needs to check the `ref` also.

drgrice1 (commented):
I see now that this pull request doesn't even use the `lib/Plots/StatPlot.pm` file.  So just delete it, and ignore my comments in that file.

drgrice1:
Here are some issues seen with your example problems:

The `histogram.pg` example in your zip file is not working.  It gives the warning `Use of uninitialized value $options{"orientation"} in string eq at line 358 of [PG]/macros/graph/StatisticalPlots.pl.` and the resulting graph looks quite bad.

The `boxplot_vertical.pl` problem does not show a vertical box plot.  In fact the problem is identical to the `boxplot.pl` file.  Even adding the `orientation => 'vertical'` option to the `add_boxplot` call does not make it so.

The `scatter01.pg` file is the new problem template, and not a scatter plot at all.

lib/Plots/StatPlot.pm
pstaabp: Nevermind, this is in `PGauxillaryFunctions.pl`, and I use these.

pstaabp:
- Added pie charts and donut/ring charts. 
- Added aliases `mark_color`, `stroke_color` and `stroke_width` to the stat plot to help distinguish other colors/widths.
- Added the beginnings of a color_palette function to return a default color palette.  Note there is only one there right now, but will add more .
 
Here is an update of the problems I've been testing with: 
 
[stats.zip](https://github.com/user-attachments/files/25132655/stats.zip)

drgrice1 (commented):
Please remove the changes for custom labels in the `plots.pl` macro and associated modules from this pull request in favor of #1389.  The custom labels are not actually used by anything here, so it shouldn't be a big deal to do so.

The only place you do use it is in an example problem that you posted. Please test pull request #1389.  That should more than meet the needs of what you do in that problem. Furthermore, the histogram code in the `StatisticalPlots.pl` macro could use it to ensure that the class ends are labeled.  That is what I do in my problems that directly use the plots macro to construct histograms.

lib/Plots/Plot.pm
```diff
@@ -392,6 +392,18 @@ sub add_arc {
...
+
+	Value::Error('The first point must be an array ref of length 2')
+		unless ref($pt0) eq 'ARRAY' && scalar(@$pt0) == 2;
+	Value::Error('The second point must be an array ref of length 2')
+		unless ref($pt2) eq 'ARRAY' && scalar(@$pt2) == 2;
+	# If the fill_color option is set, set the fill to 'self'.
+	$options{fill} = 'self' if $options{fill_color} && !defined($options{fill});
+	return $self->add_dataset($pt0, [ $pt2->[0], $pt0->[1] ], $pt2, [ $pt0->[0], $pt2->[1] ], $pt0, %options);
+}
+
```
drgrice1: This is rather inconsistent with the rest of the `Plot.pm` code.

Most of the plots methods allow adding one or multiple of each object at a time.  The `add_rectangle` method should do this as well.
    
No where in the plots code is the `Value::Error` method used to give messages about incorrect usage, and for good reason.  Calling `Value::Error` dies and prevents the entire problem from rendering.  That should really only be done in answer checkers where those exceptions are caught, and not for something like this where you really just want to notify the problem author of incorrect usage.  A `warn` is certainly sufficient for that, and then the problem can still render.  There is also no need for a separate message for the two points. Just give one.

In short change this to

```suggestion
sub _add_rectangle {
	my ($self, $pt0, $pt2, %options) = @_;
	unless (ref($pt0) eq 'ARRAY' && @$pt0 == 2 && ref($pt2) eq 'ARRAY' && @$pt2 == 2) {
		warn 'A rectangle requires two points defined by length two array references.';
		return;
	}
	$options{fill} = 'self' if $options{fill_color} && !defined $options{fill};
	return $self->_add_dataset($pt0, [ $pt2->[0], $pt0->[1] ], $pt2, [ $pt0->[0], $pt2->[1] ], $pt0, %options);
}

sub add_rectangle {
	my ($self, @data) = @_;
	if (ref($data[0]) eq 'ARRAY' && ref($data[0][0]) eq 'ARRAY') {
		return [ map { $self->_add_rectangle(@$_) } @data ];
	}
	return $self->_add_rectangle(@data);
}
```

macros/graph/plots.pl
```diff
@@ -291,13 +291,30 @@ =head2 PLOT ARCS
...
+The following makes a filled rectangle with a thicker blue border.
+
+    $plot->add_rectangle([2,1], [6,3],
+        color        => 'blue',
+        width        => 1.5,
+        fill         => 'self',
+        fill_color   => 'yellow',
+        fill_opacity => 0.1,
+    );
+
```
drgrice1, line 327: This is inconsistent with the other POD in this file.  This should be much like the POD for the `add_circle` and `add_arc` methods. Those methods are really just dataset methods also, and so there is no need to remark on this being a convenience method that just calls `add_dataset`.

So change this to something like

```suggestion
=head2 PLOT RECTANGLES

A rectangle can be plotted with the C<< $plot->add_rectangle >> method. This
method takes two points which are opposite corners of the rectangle. Multiple
rectangles can be plotted at once by passing references to arrays of data for
each rectangle.

	$plot->add_rectangle([$lower_left_x, $lower_left_y], [$upper_right_x, $upper_right_y], %options);
	$plot->add_rectangle(
		[[$lower_left_x1, $lower_left_y1], [$upper_right_x1, $upper_right_y1], %options1],
		[[$lower_left_x2, $lower_left_y2], [$upper_right_x2, $upper_right_y2], %options2],
		...
	);

```

That also incorporates in adding multiple rectangles for the suggested code in my other comment.

Edit: Oops.  I cut and pasted the wrong section of POD.

drgrice1 (changes requested):
This is just a review of the `add_rectangle` method.  This completes what I have seen for the changes of the `plots.pl` code.  The typo fixes are really the only things that should stay as they are.

I will try to look at the `StatisticalPlots.pl` macro deeper soon.

macros/graph/plots.pl
pstaabp: updated.

lib/Plots/Plot.pm
pstaabp: updated.

lib/Plots/StatPlot.pm
pstaabp: This has been fixed.

macros/core/PGbasicmacros.pl
pstaabp: Wondering what to do with this.  When you say that the `Scalar::Util::blessed` method isn't available, is this a safe issue?  

Suggestions for updating this?

pstaabp:
I have updated this PR to 
1. get rid of the custom tick code in favor of #1389 
2. Other comments above. 
3. switch to the PG2.21 branch.

drgrice1 (commented):
You either need to rebase this pull request on to the `PG-2.21` release candidate branch, or merge that branch into this one.  This is showing things that aren't right.

Revert all changes to the `htdocs/js/Plots/plots.js`, `lib/Plots/Axes.pm`, `lib/Plots/JSXGraph.pm` files in this pull request.  Those files should not be modified or be listed as changed at all by this pull request.  The changes shown now are incorrect.

Alex-Jordan:
@pstaabp If you would like to get this in the release, I am happy to test it after you follow up with @drgrice1's review.

pstaabp:
@drgrice1 Rebased onto PG-2.21

htdocs/js/Plots/plots.js
```diff
@@ -387,9 +387,17 @@ const PGplots = {
...
 				xAxis.defaultTicks.formatLabelText = plot.formatLabelText;
 
+				if (options.xAxis.ticks?.customLabels) {
+					xAxis.defaultTicks.generateLabelText = function (tick) {
+						return options.xAxis.ticks.customLabels[tick.usrCoords[1] / options.xAxis.ticks.distance - 1];
+					};
+				} else {
+					xAxis.defaultTicks.generateLabelText = plot.generateLabelText;
+				}
+
```
drgrice1: Remove all changes to this file.  This pull request no longer should have any changes to this file.

lib/Plots/Axes.pm
```diff
@@ -350,6 +350,7 @@ sub axis_defaults {
 		tick_labels       => 1,
 		tick_label_format => 'decimal',
 		tick_label_digits => 2,
+		tick_label_custom => undef,       # NEW: Array of custom labels
```
drgrice1: Remove this.  It should not be here anymore.

lib/Plots/JSXGraph.pm
```diff
@@ -128,21 +128,29 @@ sub HTML {
...
-		$options->{xAxis}{ticks}{labelDigits} = $axes->xaxis('tick_label_digits');
+		$options->{xAxis}{name}          = $axes->xaxis('label');
+		$options->{xAxis}{ticks}{show}   = $axes->xaxis('show_ticks');
+		$options->{xAxis}{ticks}{labels} = $axes->xaxis('tick_labels');
+		if ($axes->xaxis('tick_label_custom')) {
+			$options->{xAxis}{ticks}{customLabels} = $axes->xaxis('tick_label_custom');
+		} else {
+			$options->{xAxis}{ticks}{labelFormat} = $axes->xaxis('tick_label_format');
+			$options->{xAxis}{ticks}{labelDigits} = $axes->xaxis('tick_label_digits');
+		}
```
drgrice1: Revert all changes to this file.  This pull request should not have any changes to this file anymore.

drgrice1 (commented):
There are still some remnants of things from before that need to be removed.

pstaabp:
@drgrice1 Reverted the changes to those files.

macros/graph/StatisticalPlots.pl
```diff
@@ -0,0 +1,876 @@
...
+
+=item Pie Charts
+
+=back
+
+=head2 USAGE
+
+First, start with a C<StatPlot> object as in 
+
+    loadMacros('StatisticsPlots.pl');
```
Alex-Jordan: Typo here. This should be `StatisticalPlots.pl`.

macros/graph/StatisticalPlots.pl
```diff
@@ -0,0 +1,876 @@
...
+			# take $alpha mod 2pi
+			$alpha = $alpha - (2 * $pi * int($alpha / (2 * $pi)));
+
+			$self->add_label(
+				1.1 * $options{radius} * cos($alpha),
+				1.1 * $options{radius} * sin($alpha),
+				$options{labels}->[$_],
+				(0 <= $alpha && $alpha < $pi / 4)
+					|| (7 * $pi / 4 < $alpha && $alpha < 2 * $pi) ? (h_align => 'left')
+				: $pi / 4 <= $alpha < 3 * $pi / 4     ? (v_align => 'bottom')
```
Alex-Jordan: I'm getting a syntax error from this line. Should `$pi / 4 <= $alpha < 3 * $pi / 4` be `$pi / 4 <= $alpha && $alpha < 3 * $pi / 4`? And similarly for the next line.

macros/core/PGbasicmacros.pl
Alex-Jordan: Crude regex hack?

```
if (ref $image_item =~ /^Plots::Plot(?!\w)/) {
```

drgrice1 (approved):
I think that this is good now, and ready for review for others.  I changed quite a bit in the pull request that @pstaabp just merged.

Alex-Jordan (approved):

PR #1377, opened 2026-02-04
Remove partial credit note in avg_problem_grader.

somiaj:
This note isn't that useful and can be misleading as it only is about number of answer blanks.


drgrice1 (approved):
Looks good, and I have been lobbying for this for a while.

dlglin (approved):

PR #1379, opened 2026-03-04
Some improvements and additions to the `SimpleGraph.pl` macro.

drgrice1:
First, change the weight color used from red to FireBrick.  This color has a contrast ratio of 6.68 against the white background.  The red color has a contrast ration of 4.00 which is not sufficient for accessibility purposes.

Second, improve label positioning. Instead of trying to place the labels using the coordinates, use the `anchor` and padding options to get better positioning.  The primary advantage is now the labels don't float away when the image is enlarged.  This was not done originally because at the time this macro was implemented the `anchor` and `padding` options didn't exist.

The weights for the default layout and the wheel layout are still positioned along the perpendicular vector for now.  The problem is that those labels are rotated, and that does not work well with the anchor.  This is a TikZ issue (I implemented the anchor for JSXGraph to work the same as the TikZ anchor option).  The problem is that the rotation is around the position of the anchor, and not around the center of the text. The usual solution for this in TikZ is to use a `\rotatebox` on the node contents. Perhaps another rotation option could be addded to the plots macro that would rotate the text instead of rotating around the anchor position.

Third, add a `components` method that returns the components of the graph.  The method returns an array containing references to arrays that form a partition the vertex indices into the connected components of the graph.  For example, for the graph with vertices E, F, G, H, I, J, K, and L, and edge set {{{E, L}, {F, G}, {F, L}, {G, J}, {H, I}, {J, K}, {J, L}}}, the method will return ([E, F, G, J, K, L], [H, I]).

### DIFF
--- a/macros/math/SimpleGraph.pl
+++ b/macros/math/SimpleGraph.pl
@@ -819,7 +819,7 @@ sub image {
 						$u * $iVertex->[0] + $v * $jVertex->[0] + $perp[0] * 0.06,
 						$u * $iVertex->[1] + $v * $jVertex->[1] + $perp[1] * 0.06,
 						label  => "\\\\($self->{adjacencyMatrix}->[$i][$j]\\\\)",
-						color  => 'red',
+						color  => 'FireBrick',
 						rotate => ($perp[0] < 0 ? 1 : -1) *
 							atan2(sqrt(1 - $perp[1] * $perp[1]), $perp[1]) * 180 /
 							$main::PI - ($perp[1] < 0 ? 180 : 0)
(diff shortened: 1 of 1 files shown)


drgrice1:
One more addition.  I added a `description` method.  This returns a translated string description of the graph.  The string that is returned is suitable for use as the "alt" text for the image returned by one of the image methods.  For now, this is just a graph theoretical description.  In other words, it just lists vertices and edges (with weights if `includeWeights => 1` is passed in the options).  For example, the description returned will be something like "A graph with vertices L, M, N, O, P. There are edges between L and N, L and O, L and P, M and N, M and P, N and P, and O and P." or if `includeWeights => 1` is set then it will be "A graph with vertices T, U, V, W. There are edges between T and U with weight 6, T and V with weight 7, T and W with weight 5, U and V with weight 8, U and W with weight 10 , and V and W with weight 9.".

Translation of general lists is rather challenging, but I think the way it is done should work for most languages.

PR #1383, opened 2026-03-10
License clarification.

somiaj:
Update the license to match what is in the webwork2 repo. This includes clarifying we use artistic license 1.0, remove the fsf physical address and provide links to both the artistic license 1.0 and gpl 2.0.

The old license and new license mention copies of the license should be provided in the "package", but they were not in the pg repo. This adds them to the doc directory to match what is done in webwork2.

Remove the old README which was just a copyright statement.

Note, I'm unsure if the `doc` directory is the best place for the licenses, since there is POD in there, but the licensees didn't seem to get picked up by the POD generation, unsure if there is a better place to put them, so I put them there to match what is done in webwork2.


drgrice1 (approved):
Looks good.  I don't think there is a better place to put the licenses than the `doc` directory.

somiaj:
One update, I change the license to say "Program Generation Language" to match what was in the old README from the "Online Homework Delivery System" which was there. Though I always thought it was "Problem Generation", so TIL.

drgrice1:
> One update, I change the license to say "Program Generation Language" to match what was in the old README from the "Online Homework Delivery System" which was there. Though I always thought it was "Problem Generation", so TIL.

I am not sure about this.  PG does not stand for "Problem Generation" (I also used to think that), but it also doesn't stand for "Program Generation".  According to https://wiki.openwebwork.org/wiki/History_of_WeBWorK (and Mike and Arnold agree with that) it stands for "Pretty Good".

somiaj:
I can revert that change, I was just using what was in the old README, so that must have been a mistake.

somiaj:
What about removing that line altogether, I just called it "WeBWorK PG" and that is all.

drgrice1:
Yeah, maybe that would be best.  Either that or @mgage  could offer a suggestion?

drgrice1:
I suppose I should tag @apizer also.  He could give a suggestion as well.

dlglin (approved):

PR #1384, opened 2026-03-11
Update Value/AnswerChecker diagnostics output.

somiaj:
Use plots.pl to create the graphs instead of the old graphmacros.

Remove use of tables for formatting. Using bordered divs instead. Only use of a table is for the table of test points, which now uses row headers properly and padding and no longer uses blank cells for spacing. The test point table puts each variable in its own row for multivariable functions.

Here are some test problems, one using an adaptive parameter, and one with multivariables to show each variable test points are put in their own row.

```
DOCUMENT();
$ans = Compute('(3 x^4 + 4)/x');
BEGIN_PGML
Enter [``[$ans]``]: [_]{$ans->cmp(diagnostics => 1)}
END_PGML
ENDDOCUMENT();
```

```
DOCUMENT();
Context()->variables->are(t => 'Real', C0 => 'Parameter');
$aSoln = Compute('C0 * e^t - 1')->cmp(diagnostics => 1);
BEGIN_PGML
Find one solution to the differential equation.

    [`\frac{dy}{dt} = y + 1`]

[`y =`] [_]{$aSoln}
END_PGML
ENDDOCUMENT();
```

```
DOCUMENT();
Context()->variables->are(x => 'Real', y => 'Real', z => 'Real');
$ans = Compute('(4x - 3y)/z');
BEGIN_PGML
Enter [``[$ans]``]: [_]{$ans->cmp(diagnostics => 1)}
END_PGML
ENDDOCUMENT();
```
 
I don't think stating the domain, range, and aspect ratio of the graphs is useful. I left that there, but suggest removing it.

@drgrice1 for some reason the mathjax for the graph labels only renders when clicking on the graph, but not the original graph output. I didn't debug as to why.

### DIFF
--- a/lib/Value/AnswerChecker.pm
+++ b/lib/Value/AnswerChecker.pm
@@ -2050,48 +2050,44 @@ sub cmp_postprocess {
 	$self->cmp_Error($ans, "The dimension of your result is incorrect");
 }
 
-#
-#  Diagnostics for Formulas
-#
+# Diagnostics for Formulas.
 sub cmp_diagnostics {
-	my $self        = shift;
-	my $ans         = shift;
-	my $isEvaluator = (ref($ans) =~ /Evaluator/) ? 1            : 0;
-	my $hash        = $isEvaluator               ? $ans->rh_ans : $ans;
-	my $diagnostics = $self->{context}->diagnostics->merge("formulas", $self, $hash);
+	my ($self, $ans) = @_;
+	my $isEvaluator = ref($ans) =~ /Evaluator/ ? 1            : 0;
+	my $hash        = $isEvaluator             ? $ans->rh_ans : $ans;
+	my $diagnostics = $self->{context}->diagnostics->merge('formulas', $self, $hash);
 	my $formulas    = $diagnostics->{formulas};
 	return unless $formulas->{show};
 
-	my $output = "";
+	my $output = '';
 	if ($isEvaluator) {
-		#
-		#  The tests to be performed when the answer checker is created
-		#
-		$self->getPG('loadMacros("PGgraphmacros.pl")');
-		my ($inputs) = $self->getPG('$inputs_ref');
-		my $process = $inputs->{checkAnswers} || $inputs->{previewAnswers} || $inputs->{submitAnswers};
-		if ($formulas->{checkNumericStability} && !$process) {
-			### still needs to be written
-		}
+		# The tests to be performed when the answer checker is created.
+		# Still needs to be written. Commenting it out for now.
+		#my ($inputs) = $self->getPG('$inputs_ref');
+		#my $process = $inputs->{checkAnswers} || $inputs->{previewAnswers} || $inputs->{submitAnswers};
+		#if ($formulas->{checkNumericStability} && !$process) {
+		### Still needs to be written.
+		#}
 	} else {
-		#
-		#  The checks to be performed when an answer is submitted
-		#
+		# The checks to be performed when an answer is submitted.
 		my $student = $ans->{student_formula};
-		#
-		#  Get the test points
-		#
-		my @names = $self->{context}->variables->names;
-		my $vx    = (keys(%{ $self->{variables} }))[0];
-		my $vi    = 0;
-		while ($names[$vi] ne $vx) { $vi++ }
-		my $points = [ map { $_->[$vi] } @{ $self->{test_points} } ];
-		my @params = $self->{context}->variables->parameters;
-		@names = $self->{context}->variables->variables;
 
-		#
-		#  The graphs of the functions and errors
-		#
+		# Get the test points. This requires first determining what variables are in use and ignore parameters.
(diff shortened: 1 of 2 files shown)


somiaj:
In testing this I ran across a bug where test points were not being graphed correctly when using adaptive parameters, and that was because the adaptive parameters are included in the list of variable names, but don't have test points assigned to them, so the test points being used were either nonexistent or assigned to a different variable depending on the random order of the keys from the `variables` hash.

So I updated this to remove adaptive parameters from the list of variable names.

I also updated it so that the list of test points only shows the values of variable test points used in either the student or correct answer. This does make the table more useful to me when there are additional variables in the context not used.

somiaj:
There is another possible issue where alias variables aren't correctly dealt with in determining what the test points are for the graph. I think this is similar to the parameter issue where these variables aren't given test points so the wrong list of test points is chosen to be graphed.

drgrice1:
> @drgrice1 for some reason the mathjax for the graph labels only renders when clicking on the graph, but not the original graph output. I didn't debug as to why.

This is not happening for me.  The MathJax renders on initial page load for me.

lib/Value/AnswerChecker.pm
```diff
@@ -2131,157 +2123,152 @@ sub cmp_diagnostics {
...
-						. ($error->value < $tolerance ? '00AA00' : 'AA0000') . '">'
-						. $error
-						. '</SPAN>'
-						if $tolType eq 'absolute';
-				} else {
-					$error = "---";
+		if (@P) {
+			my @i      = sort { $P[$a] <=> $P[$b] } 0 .. $#P;
+			my @points = ([] x scalar(@names));
+			for $p (@P) {
```
drgrice1: This should be

```suggestion
			for my $p (@P) {
```

somiaj:
@drgrice1 wonder if it is a race condition where mathjax isn't fully loaded by the time the graph loads, I think I've had this happen before.

lib/Value/AnswerChecker.pm
somiaj: Thanks for catching that.

somiaj:
Okay, I think I got the variables vs parameters vs aliases figured out. Aliases didn't need to be dealt with as the formula already does the substitution, parameters just need to be filtered otherwise they make the count of which list of test points to use incorrect. @dpvc Care to let me know if I'm missing something, the main bit of code I have is here:

```perl
		# Get the test points. This requires first determining what variables are in use and ignore parameters.
		my $points    = [];
		my @params    = $self->{context}->variables->parameters;
		my %pnames    = map { $_ => 1 } @params;
		my @names     = $self->{context}->variables->variables;
		my @variables = grep { !$pnames{$_} } keys %{ $self->{variables} };

		# If a single variable is found, use its test points for the graph.
		# Functions with more than one variable are not graphed so test points are not needed.
		if (scalar(@variables) == 1) {
			my $vi = 0;
			while ($names[$vi] ne $variables[0]) { $vi++ }
			$points = [ map { $_->[$vi] } @{ $self->{test_points} } ];
		}
```

dpvc:
@somiaj, I think you are OK with what you have done.  I haven't done any testing, but I don' see anything wrong with the approach.

drgrice1:
The reason that the MathJax is not rendering when this is viewed in the problem editor (or any case that the `render_rpc` route is used is because this is wrapped in a `code` tag.  When viewing the problem in a set via `Problem.pm`, this is inserted via the `warnings` method defined at line 638 of `lib/WeBWorK/ContentGenerator/Problem.pm` and this is NOT placed inside a `code` tag.  So the MathJax works.

MathJax by default includes the `code` tag in the `skipTags` setting.  We probably don't want to change that.  So I think to fix this the debug messages should not be placed inside a `code` tag.  This will take a change to WeBWorK.  This is done in the `templates/RPCRenderFormats/default.html.ep` file on line 176.  The `code` tag there should probably just be a `div`.

Although, in general this goes along with what I was talking about in the developer meeting this week.  The error/warning/debug output needs to be reworked.  Another issue is that when viewing this in a set all of this is wrapped in a `p` tag.  Although it works because browsers close the `p` tag when they encounter this content that can't be in a `p` tag, it is invalid HTML.  That `p` tag should also be changed to a div (and the `br` tags removed.

somiaj:
That sounds like stuff that should be fixed in another PR, and we should probably rework what is shown. I do think that it is better to not show debugging stuff as a warning as well.

dpvc:
> I do think that it is better to not show debugging stuff as a warning as well.

Just for some context, when the debugging stuff was written (long ago), because the answer checkers run *after* the problem output has been built, there was really no other way to get the debugging output into the page.  The interface has undergone a *lot* of changes since then, and there may now be a good way to do it (or one could be added).

drgrice1 (approved):
I think this looks good.

dlglin (approved):
Looks good

PR #1386, opened 2026-03-24
Remove the internal debug messages.

drgrice1:
The only place this is used is for error messages in `lib/PGresponsegroup.pm`. That is now reworked to use the `PGcore::warning_message` method instead. The `internal_debug_messages` is just not needed anymore in addition to the `warning_message` and `debug_message` methods. Furthermore, it is inappropriately named. It has "debug" in the name, but is used for error messages, not debugging messages.

Note that this depends on https://github.com/openwebwork/webwork2/pull/2935 for webwork2 to work with this.

### DIFF
--- a/lib/PGUtil.pm
+++ b/lib/PGUtil.pm
@@ -89,7 +89,7 @@ sub pretty_print_html {    # provides html output -- NOT a method
 	if (!$ref) {
 		return $r_input =~ s/</&lt;/gr;
 	} elsif (eval { %$r_input || 1 }) {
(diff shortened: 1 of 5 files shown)


somiaj:
Anything particular to test with this?

drgrice1:
The problem that I posted in the corresponding webwork2 pull request is what I used to test this.  The `INSERT_RESPONSE` and `APPEND_RESPONSE` calls in that problem are specifically for this.

somiaj (approved):

dlglin (commented):
Seems to work.

PR #1389, opened 2026-03-25
Make it possible to set custom tick positions and labels with the `plots.pl` macro.

drgrice1:
To set custom tick positions use the `tick_positions` axis option.  Set that to a reference to an array containing the positions on the axis that ticks are desired.  For example, `tick_positions => [ 2, 5, 9 ]` will place ticks at positions 2, 5, and 9 on the axis.  Note that when this option is used the `tick_delta`, `tick_scale`, and `tick_distance` options are not used. So only the given tick positions will appear in the graph.

To set custom tick labels use the `tick_labels` option.  Note that this is not a new option, but now it accepts a new type of value.  Previously this was purely boolean (0 or 1), and it only determined if tick labels would be shown or not.  Now it can take a value that is a reference to a hash.  The keys of the hash are tick positions, and the values are the labels to be placed at those positons. Note that formatting of the label must be done by the auther, and the `tick_label_format` option is ignored for any label provided in this hash. If a major tick is not listed in the hash, then the position will be used for the label and it will be formatted according to the `tick_label_format` option.

This is intended to replace what is done in #1374 and is a more flexible approach than what is done there.  In that pull request the capability for custom tick labels only is added, and it is extremely restrictive in what it can do.  Only positive tick labels can be customized, and it requires that the problem author label all major ticks (there is no fallback and a tick is labeled "undefined" if one is missing).

Here is a rather rudimentary example of using this new feature:
```
DOCUMENT();

$plot = Plot(
    xmin            => -2,
    xmax            => 12,
    xtick_distance  => 2,
    xminor          => 1,
    xtick_positions => [ 2, 4, 6 ],
    xtick_labels    => { 2 => '\(a\)', 6 => '\(b\)' },
    ymin            => -2,
    ymax            => 12,
    ytick_distance  => 2,
    yminor          => 1,
);

BEGIN_PGML
[@ $plot->image_type('Tikz'); @]*[!TikZ Image!]{$plot}

[@ $plot->image_type('JSXGraph'); @]*[!JSXGraph Image!]{$plot}
END_PGML

ENDDOCUMENT();
```

### DIFF
--- a/htdocs/js/Plots/plots.js
+++ b/htdocs/js/Plots/plots.js
@@ -86,9 +86,18 @@ const PGplots = {
 					}
 				},
 
-				// Override the default axis generateLabelText method so that 0 is displayed
-				// using MathJax if the axis is configured to show tick labels using MathJax.
+				// Override the default axis generateLabelText method to show custom tick labels if they are set, and so
+				// that 0 is displayed using MathJax if the axis is configured to show tick labels using MathJax.
 				generateLabelText(tick, zero, value) {
+					for (const axis of ['xAxis', 'yAxis']) {
+						if (
+							this === plot[axis]?.defaultTicks &&
+							typeof options[axis]?.ticks?.labels === 'object' &&
+							tick.usrCoords[1] in options[axis].ticks.labels
+						) {
+							return options[axis].ticks.labels[tick.usrCoords[1]];
+						}
+					}
 					if (JXG.exists(value)) return this.formatLabelText(value);
 					const distance = this.getDistanceFromZero(zero, tick);
 					return this.formatLabelText(Math.abs(distance) < JXG.Math.eps ? 0 : distance / this.visProp.scale);
@@ -216,7 +225,7 @@ const PGplots = {
 								strokeColor: options.grid.color ?? '#808080',
 								strokeOpacity: options.grid.opacity ?? 0.2,
 								insertTicks: false,
-								ticksDistance: options.xAxis?.ticks?.distance ?? 2,
+								ticksDistance: options.xAxis.ticks?.positions ?? options.xAxis.ticks?.distance ?? 2,
 								scale: options.xAxis?.ticks?.scale ?? 1,
 								minorTicks: options.grid.x.minorGrids ? (options.xAxis?.ticks?.minorTicks ?? 3) : 0,
 								ignoreInfiniteTickEndings: false,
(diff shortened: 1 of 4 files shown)


somiaj (approved):

somiaj:
Not related to this pull request, but TikZ doesn't allow `\dfrac{}{}`? I decided to see what labels would do if they were really tall and the `ymin` was zero, so I replaced your problem with `ymin => 0` and `xtick_labels    => { 2 => '\(\dfrac{\dfrac{a}{b}}{\dfrac{c}{d}}\)', 6 => '\(b\)' }`. The TiKz version only showed `abcd` for the label. But if I replaced `dfrac` with `frac` it worked. Does some additional TiKz package need loaded?

TiKz:

image

JSXGraph

image

drgrice1:
I will look into the `\dfrac` thing.  I am not sure what is happening there.  latex is giving an undefined control sequence for a `\dfrac`.

drgrice1:
Ahh, it is because `\dfrac` is part of the amsmath package.  It is not default latex.  So you will need to add the `amsmath` package to use it.

drgrice1:
This is a limitation of the `plots.pl` macro.  It does not give you access to the `LaTeXImage` object before it is drawn, and so there is not way to add TeX packages.

somiaj:
Might be worth adding the `amsmath` package by default? I didn't consider that someone would need access to it, but seems like there is a reason. Ether way, with this people may want to use `\dfrac` for fraction labels, so we could address this here or another PR if we want to make a more extensible method to allow access to the `LaTeXImage` object.

pstaabp:
Could we add another option for `tick_labels` like:
```
xtick_labels => ['\(a\)', '\(b\)', '\(c\)'],
```
that would replace the labels in tick position?  I would argue that this would be easier for many authors that don't know/use hashes.

drgrice1:
If you are writing PG problems, you really need to understand at least basic Perl, and knowing how to work with hashes is a basic part of that.  The problem is that doing so introduces a lot of ugliness in the code because that approach is entirely unreliable and you have to protect against all of the issues it creates.  The fact is that is the wrong data structure for this usage.

So I am going to say no.

pstaabp:
Seems like the JSXgraph custom labels aren't working.  Here's a version of your code above:
```
DOCUMENT();

$plot = Plot(
    xmin            => -2,
    xmax            => 12,
    xtick_distance  => 2,
    xminor          => 1,
    ytick_positions => [ 2, 4, 6 ],
    ytick_labels    => { 2 => '\(a\)', 6 => '\(b\)' },
    ymin            => -2,
    ymax            => 12,
    ytick_distance  => 2,
    yminor          => 1,
);

BEGIN_PGML
[@ $plot->image_type('Tikz'); @]*[!TikZ Image!]{$plot}

[@ $plot->image_type('JSXGraph'); @]*[!JSXGraph Image!]{$plot}
END_PGML

ENDDOCUMENT();
```
and I see the tikz change the y labels, but not the JSXGraph.

drgrice1:
@pstaabp: That is fixed.  I don't know why I missed that.  The y-axis needs to use the correct coordinate!

pstaabp:
Fixed!

pstaabp (approved):
👍

somiaj:
Two approvals so merging.

PR #1391, opened 2026-03-26
Inject macros via %envir

drdrew42:
Consider this as a mechanism to bypass (or roll-your-own) hierarchy of search locations for macro files...

Existing pathways all remain behaviorally consistent, we just skip `findMacroFile` if content for a named macro is already provided in the environment hash.

### DIFF
--- a/lib/PGloadfiles.pm
+++ b/lib/PGloadfiles.pm
@@ -150,19 +150,33 @@ sub loadMacros {
 			if $debugON;
 		unless ($macro_file_loaded) {
 			warn "loadMacros: loading macro file $fileName" if $debugON;
-			my $filePath = $self->findMacroFile($fileName);
-			#### (check for renamed files here?) ####
-			warn "loadMacros:  look for $fileName at |$filePath|" if $debugON;
-			if ($filePath) {
-				$self->compile_file($filePath);
-				warn "loadMacros is compiling $filePath" if $debugON;
+
+			# Check for injected macro source (from OPL content-addressed macros)
+			# before searching the filesystem. This allows custom macros to be
+			# delivered via %envir without requiring them on disk.
+			my $injected = $self->{envir}{injectedMacros}{$fileName};
+			if ($injected) {
+				warn "loadMacros: using injected source for $fileName" if $debugON;
+				my ($result, $error, $fullerror) =
+					$self->PG_macro_file_eval($injected, "injected:$fileName");
+				if ($error) {
+					die "Error detected while loading injected macro $fileName:\n$fullerror";
+				}
 			} else {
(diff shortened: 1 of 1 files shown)


lib/PGloadfiles.pm
```diff
@@ -150,19 +150,33 @@ sub loadMacros {
...
-				warn "loadMacros is compiling $filePath" if $debugON;
+
+			# Check for injected macro source (from OPL content-addressed macros)
+			# before searching the filesystem. This allows custom macros to be
+			# delivered via %envir without requiring them on disk.
+			my $injected = $self->{envir}{injectedMacros}{$fileName};
+			if ($injected) {
+				warn "loadMacros: using injected source for $fileName" if $debugON;
+				my ($result, $error, $fullerror) =
+					$self->PG_macro_file_eval($injected, "injected:$fileName");
```
drgrice1, line 161: This should be

```suggestion
				my ($result, $error, $fullerror) = $self->PG_macro_file_eval($injected, "injected:$fileName");
```

That is the reason this is failing the `perltidy` check.

lib/PGloadfiles.pm
```diff
@@ -150,19 +150,33 @@ sub loadMacros {
...
+				warn "loadMacros:  look for $fileName at |$filePath|" if $debugON;
+				if ($filePath) {
+					$self->compile_file($filePath);
+					warn "loadMacros is compiling $filePath" if $debugON;
+				} else {
+					my $pgDirectory       = $self->{envir}{pgMacrosDir};
+					my $templateDirectory = $self->{envir}{templateDirectory};
+					my @shortenedPaths    = @{$macrosPath};
+					@shortenedPaths = map { $_ =~ s|^$templateDirectory|[TMPL]/|; $_ } @shortenedPaths;
+					@shortenedPaths = map { $_ =~ s|^$pgDirectory|[PG]/macros/|;  $_ } @shortenedPaths;
```
drgrice1, line 177: I know that this is the original code, but it should be fixed.  It should not modify `$_`.  See https://metacpan.org/release/THALJEF/Perl-Critic-1.06/view/lib/Perl/Critic/Policy/ControlStructures/ProhibitMutatingListFunctions.pm.  Yeah, the reason for the policy is moot since the intent is to modify the array element, but then again, it doesn't make sense to make the assignment with that modification either and it is also good to prevent the Perl critic warning.

So could you fix this and change this to

```suggestion
					@shortenedPaths = map { $_ =~ s|^$templateDirectory|[TMPL]/|r } @shortenedPaths;
					@shortenedPaths = map { $_ =~ s|^$pgDirectory|[PG]/macros/|r } @shortenedPaths;
```

somiaj:
Closing in favor of #1397

PR #1392, opened 2026-03-26
Harden system calls

drdrew42:
Please consider this commit for hotfix.

### DIFF
--- a/lib/LaTeXImage.pm
+++ b/lib/LaTeXImage.pm
@@ -271,12 +271,18 @@ sub draw {
 
 sub use_svgMethod {
 	my ($self, $working_dir) = @_;
-	if ($self->svgMethod eq 'dvisvgm') {
-		system WeBWorK::PG::IO::externalCommand('dvisvgm')
-			. " $working_dir/image.dvi --no-fonts --output=$working_dir/image.svg > /dev/null 2>&1";
+
+	# Validate svgMethod against known SVG converters to prevent command injection.
+	my $method = $self->svgMethod;
+	if ($method eq 'dvisvgm') {
+		my $cmd = WeBWorK::PG::IO::externalCommand('dvisvgm');
+		system {$cmd} $cmd, "$working_dir/image.dvi", '--no-fonts', "--output=$working_dir/image.svg";
+	} elsif ($method eq 'pdf2svg') {
+		my $cmd = WeBWorK::PG::IO::externalCommand('pdf2svg');
+		system {$cmd} $cmd, "$working_dir/image.pdf", "$working_dir/image.svg";
 	} else {
-		system WeBWorK::PG::IO::externalCommand($self->svgMethod)
-			. " $working_dir/image.pdf $working_dir/image.svg > /dev/null 2>&1";
+		warn "Unknown svgMethod '$method'. Must be 'dvisvgm' or 'pdf2svg'.";
+		return;
 	}
 	warn "Failed to generate svg file." unless -r "$working_dir/image.svg";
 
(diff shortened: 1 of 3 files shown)


lib/LaTeXImage.pm
```diff
@@ -271,12 +271,18 @@ sub draw {
 
 sub use_svgMethod {
 	my ($self, $working_dir) = @_;
-	if ($self->svgMethod eq 'dvisvgm') {
-		system WeBWorK::PG::IO::externalCommand('dvisvgm')
-			. " $working_dir/image.dvi --no-fonts --output=$working_dir/image.svg > /dev/null 2>&1";
+
+	# Validate svgMethod against known SVG converters to prevent command injection.
```
drgrice1, line 275: The places that you are doing the validation in this file are not the right places for the validation.  It should be done in the `$self` method when the `svgMethod` or `convertOptions` is set.  That is defined on lines 32 - 44 of this file.  Validation to protect system calls against injection is exactly the point of using a blessed `$self` method instead of a blessed `$self` hash, and it is already validating the `ext` field in this way.

I can't believe that I was this sloppy when the `svgMethod` and `convertOptions` were added by not taking care of validation.  The `ext` validation was in the initial code, and those were added later.

lib/LaTeXImage.pm
```diff
@@ -271,12 +271,18 @@ sub draw {
...
+	my $method = $self->svgMethod;
+	if ($method eq 'dvisvgm') {
+		my $cmd = WeBWorK::PG::IO::externalCommand('dvisvgm');
+		system {$cmd} $cmd, "$working_dir/image.dvi", '--no-fonts', "--output=$working_dir/image.svg";
+	} elsif ($method eq 'pdf2svg') {
+		my $cmd = WeBWorK::PG::IO::externalCommand('pdf2svg');
+		system {$cmd} $cmd, "$working_dir/image.pdf", "$working_dir/image.svg";
```
drgrice1, line 282: Actually, looking closer at this I see that the `svgMethod` is not directly used in the `system` command.  It is just the name for the lookup via the `WeBWorK::PG::IO::externalCommand` method.  So if it is set to something that is not a valid `externalProgram`, then the `WeBWorK::PG::IO::externalCommand` method will just return undefined.  So I don't think extra validation is needed.  The `elsif` doesn't hurt to prevent an exception I suppose.

lib/PGalias.pm
```diff
@@ -270,7 +270,7 @@ sub convert_file_to_png_for_tex {
 	$self->debug_message('convert filePath ', $sourceFilePath, "\n");
 
 	my $conversion_command = WeBWorK::PG::IO::externalCommand('convert');
-	my $returnCode         = system "$conversion_command '${sourceFilePath}[0]' $targetFilePath";
+	my $returnCode         = system {$conversion_command} $conversion_command, "${sourceFilePath}[0]", $targetFilePath;
```
drgrice1, line 273: I don't think there is really any benefit or added security by using `system {$conversion_command} $conversion_command, "${sourceFilePath}[0]", $targetFilePath` versus just `system $conversion_command, "${sourceFilePath}[0]", $targetFilePath`.

This is true for all of the places where you have done this.  Since the program name is the same it won't actually do anything different.  It is just making the code line longer.

There is benefit to switching from the pure string argument to the list argument since it avoids invoking the shell and thus avoids potential shell expansion, but adding the program name does not add anything since they are the same.

lib/LaTeXImage.pm
```diff
@@ -285,11 +291,24 @@ sub use_svgMethod {
...
+
+	# Validate convertOptions keys and values to prevent shell injection.
+	# Only alphanumeric option names and simple values are permitted.
+	my @args;
+	for my $phase (qw(input output)) {
+		my $opts = $self->convertOptions->{$phase} // {};
+		for my $key (keys %$opts) {
+			warn("Invalid convert option name: $key"), next unless $key =~ /^[a-zA-Z][a-zA-Z0-9\-]*$/;
+			my $val = $opts->{$key};
+			warn("Invalid convert option value for $key"), next unless defined $val && $val =~ /^[a-zA-Z0-9.\-+:x%]+$/;
```
drgrice1, line 303: As I mentioned this validation should be done in the `$self` method.  Also,
this should be

```suggestion
			unless ($key =~ /^[a-zA-Z][a-zA-Z0-9\-]*$/) {
				warn("Invalid convert option name: $key");
				next;
			}
			my $val = $opts->{$key};
			unless (defined $val && $val =~ /^[a-zA-Z0-9.\-+:x%]+$/) {
				warn("Invalid convert option value for $key");
				next;
			}
```

Don't use commas to separate statements.  This is bad practice.  See https://metacpan.org/pod/Perl::Critic::Policy::ValuesAndExpressions::ProhibitCommaSeparatedStatements.

somiaj:
Closing in favor of #1397

PR #1394, opened 2026-04-08
Add keyboard controls for drag and drop problems.

drgrice1:
This is for problmes that use the `dragndrop.js` JavaScript and the `DragNDrop.pm` module via the `draggableProof.pl` and `draggableSubsets.pl` macros.

The elements in a drag and drop list can now be focused using tab and shift-tab.  Once focused the arrow keys move the elements around.  If an element is moved with the keyboard controls the changes are announced in a visually hidden span.  Note that drag and drop actions via the mouse cursor are not aria announced for now.  I am assuming that a screen reader user would not be using the mouse for drag and drop.

There is now also a "Drag and Drop Help" button that is shown below the drag and drop lists.  If pressed, help is shown describing the keyboard controls.  This help can be customized by the problem.

There are new options for the macros that allow for customizing the help and associated texts.  See the updated POD in the module and macros for details.  Also, I don't like that there are some options that are for the `DragNDrop.pm` package that are passed in from the macros, but are different in the macros than in the module.  Basically the first letter is upper case in the macros, but lower case in the module.  So the upper case first letter variants is deprecated (but will still work with a compatibility layer).  I don't like that the macros use Pascal case to begin with for options.  Options should be camel case.

Note that I also made all of the texts for the drag and drop object translatable via `maketext`.  The texts can be customized per problem as mentioned above, but if only a translation is needed there is no need for such customization.

### DIFF
--- a/lib/DragNDrop.pm
+++ b/lib/DragNDrop.pm
@@ -65,8 +65,8 @@ drag and drop bucket when clicked on.
 If the C<bucketLabelFormat> option is defined, then buckets for which an
 explicit label is not provided will be will be created with the label with the
 C<%s> in the string replaced with the bucket number in the pool.  This also
-applies to new buckets that are added by JavaScript.  An example value for this
-option is C<< 'Subset %s' >>.
+applies to new buckets that are added by the user via JavaScript if
+C<allowNewBuckets> is 1.  An example value for this option is C<'Subset %s'>.
 
 =item resetButtonText (Default: C<< 'Reset' >>)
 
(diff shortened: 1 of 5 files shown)


pstaabp (approved):
Was a little tricky to get started, but the help was helpful and I got a handle on this.  Great improvement.

lib/DragNDrop.pm
```diff
@@ -98,6 +98,77 @@ buckets, but not into it.
 
 Label shown for the universal set bucket if C<showUniversalSet> is 1.
 
+=item addFromUnversalText
```
pstaabp: This looks like a typo.

drgrice1: I am not sure where you see a typo here?  `addFromUniversalText` is correct.

Alex-Jordan: It's missing an "i" in "Universal".

somiaj (approved):
Tested my problems work just fine with the keyboard. Nice improvement.

PR #1399, opened 2026-04-17
Update the `AskSage` method.

drgrice1:
The code that is submitted via the AskSage method no longer works with the newer versions of Sage and needs to be updated.

Also the public Sage cell server at https://sagecell.sagemath.org/service no longer accepts public code via its `service` endpoint.  So make the URL for a sagecell server configurable.

Files for building a sagecell server and deploying it via docker have been added in the `sagecell-docker` directory. Instructions for doing so are in the `sagecell-docker/README.md` file.

Remove the `accepted_tos` parameter from the `AskSage` call.  Just add the parameter internally, and stop requiring that the author do so.  It is a meaningless gesture to force the author to do so.  Furthermore, the docker build is set to not require that parameter.

Also make it so that the `AskSage` method can be called without passing any arguments, since there are now none that are required.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ htdocs/static-assets.json
 htdocs/**/*.min.js
 htdocs/**/*.min.css
 htdocs/tmp
+
+sagecell-docker/sagecell-docker.service
--- a/conf/pg_config.dist.yml
+++ b/conf/pg_config.dist.yml
@@ -74,6 +74,17 @@ URLs:
     - .
     - $pg_root_url/images
 
+  # By default the AskSage method uses the public sagecell service located at
+  # https://sagecell.sagemath.org/service. However, the server at that address
+  # is no longer accepting public requests. So in order for problems that use
+  # the AskSage method to work a different sagecell server must be used.
+  # Uncomment the line below and set the value to a custom sagecell server
+  # address to use with the AskSage method.  If using the docker container built
+  # with the instructions and files in the sagecell-docker directory in this
+  # repository, the use the line below as is.
+
+  #sagecellService: http://localhost:3500/service
+
 # Flat-file database used to protect against MD5 hash collisions. TeX equations
 # are hashed to determine the name of the image file. There is a tiny chance of
 # a collision between two TeX strings. This file allows for that. However, this
--- a/lib/PGcore.pm
+++ b/lib/PGcore.pm
@@ -703,9 +703,8 @@ sub directoryFromPath {
 }
 
 sub AskSage {
-	my $self    = shift;
-	my $python  = shift;
-	my $options = shift;
+	my ($self, $python, $options) = @_;
+	$options = {} unless ref $options eq 'HASH';
 	$options->{curlCommand} = WeBWorK::PG::IO::externalCommand('curl');
 	WeBWorK::PG::IO::AskSage($python, $options);
 }
--- a/lib/WeBWorK/PG/IO.pm
+++ b/lib/WeBWorK/PG/IO.pm
@@ -315,7 +315,7 @@ sub externalCommand {
 
 # Isolate the call to the sage server in case we have to jazz it up.
 sub query_sage_server {
-	my ($python, $url, $accepted_tos, $setSeed, $webworkfunc, $debug, $curlCommand) = @_;
+	my ($python, $url, $setSeed, $webworkfunc, $debug, $curlCommand) = @_;
 
 	# Validate url to prevent shell injection — must look like a URL.
 	if ($url && $url !~ m{^https?://[^\s;|&`\$]+$}) {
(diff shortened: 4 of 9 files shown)


Alex-Jordan:
I'm following the docker installation instructions. When I get to run the docker container for the first time, I get:

```
Building Maxima command completion list (this takes
a few seconds only the first time you do it).
To force rebuild later, delete /home/sage/.sage//maxima_commandlist_cache.sobj.
```

This has been going on for a long time, maybe 10 minutes. Certainly more than a few seconds. Does that sound normal?

drgrice1:
Generally, that sounds right in terms of what you are seeing.  The command completion list build will take a few seconds, but then it will just sit there doing nothing.  It is now running.  Now to test it set up a server to connect to it using the setting added in https://github.com/openwebwork/webwork2/pull/2951.  Usually, you will set up the systemctl service to run this docker container.

Alex-Jordan:
When I hit ctrl-C to end, I then see:
```
^C
Done!

Building R command completion list (this takes
a few seconds only the first time you do it).
To force rebuild later, delete /home/sage/.sage//r_commandlist.sobj.
```

It seems odd to me that the R command completion list is built after I quit.

drgrice1:
That is what happens for me too.  So everything you are seeing is normal.   I thing that the output that is shown is actually the result of a job that ran earlier, but the output isn't flushed until you hit `Ctrl-C`.

drgrice1:
In any case, the `AskSage` problems work with this.

Alex-Jordan (approved):

Alex-Jordan:
This works for me with a minimal problem that uses AskSage. Nice update! Having a local sagecell server running removes a barrier that always kept me from investing in using this.

Alex-Jordan:
My server did not already have docker installed. I'm not sure if that requirement should be mentioned in the release notes or not.

drgrice1:
Using this is optional, but yes, if you want use it you would need to install docker.  It would be worth mentioning in the release notes, and adding optional installation instructions to the installation manual.

pstaabp:
I got the docker image up and running.   Tested it with a `docker exec -it` command and entered some sage on the command line and got reasonable response.   Pretty sure that is running. 

I added the line to `localoverrides` from the WeBWorK side as is.  It looks like it's running on port 3500.  

I tried the problem on the wiki: https://wiki.openwebwork.org/wiki/Sage_in_WeBWorK

and just seeing this:

Safari 2026-06-09 13 03 48

I must be missing something...

drgrice1:
That isn't an `AskSage` problem.  That problem uses the `sage.pl` macro for an interactive sage cell.  That does not use this at all.

Here is an `AskSage` problem you can test with:
```
DOCUMENT();

Context()->{format}{number} = '%.4f#';

$a = random( 2, 6);
$b = random( 2, 17);
$n = random( 3, 6);
$c = random(-3, 0);
$d = random( 1, 4);

$integrand =
    Formula('sqrt(1 + (' . $a * $n . ' x^' . ($n - 1) . " + $b)^2)")->reduce;

# Use sage to compute the arc length
$sageCode = <<~ "END_CODE";
    x = SR.var('x')
    WEBWORK['arc_length'] = numerical_integral($integrand, $c, $d, eps_abs = 1e-10, eps_rel = 1e-10)[0]
    END_CODE

$sageReply = AskSage($sageCode, { accepted_tos => 'true' });

if (sageReturnedFail($sageReply)) {
    # Fallback value if the sage call fails.
    $arc_length = Compute(0);
} else {
    $arc_length = Compute($sageReply->{webwork}{arc_length});
}

# Main Text
BEGIN_PGML
The length of the curve defined by

[```y = [$a]x^{[$n]} + [$b]x```]

from the point [`([$c], [$a * ($c**$n) + $b * $c])`] to the point
[`([$d], [$a * ($d**$n) + $b * $d])`] is [`[$arc_length->value]`].
END_PGML

ENDDOCUMENT();
```

pstaabp:
I went back through and wasn't sure I had the port, but now pointing to localhost:3500 (either using `curl` or a browser) shows it's working. 

However running the above problem just times out.  I turned on debug, but am seeing anything sage related or why it may not be connecting.

drgrice1:
Use the command `docker run --rm -p 127.0.0.1:3500:3500 sagecell:latest` to run the sage cell server.  Also, sometimes you have to wait a bit until the sage cell server is ready.  Basically, at least until you see the odd message that @Alex-Jordan saw.

pstaabp:
I'm pretty sure the service is running.  Putting `localhost:3500` brings up a webpage indicating this is true.

somiaj (approved):
Here is my approval to get this in. I haven't setup docker to test it.

drgrice1:
I just tested this with `podman` and it works.

Install `podman` (on Ubuntu use `sudo apt install podman`).

Then make sure that `docker.io` is in the `podman` registry.  For this create the directory `~/.config/containers`, and edit the file `~/.config/containers/registries.conf` and add

```
[registries.search]
registries = ['docker.io']
```

Then follow the directions in the `sagecell-docker/README.md` file in this pull request, except change `docker` to `podman` in all of the commands.

The `podman` syntax is identical to the `docker` syntax for almost everything.

somiaj:
@drgrice1 thanks that worked for me. When I tested podman, I tested the actual webwork docker images and not this one, and in that case I was mixing commands I found in the podman docs and didn't find the right docs to run the webwork docker image (the wiki had multiple pages of docs I didn't fully understand). But was straight forward and worked.

Testing the sample problem added in #1444 and the problem you gave above both work fine with the sage cell.

Merging now that my approval was tested and there are two approvals.

drgrice1:
@somiaj: It seems you merged the webwork2 pull request with the configuration option, but not this pull request that uses that option.

somiaj:
@drgrice1 thanks I meant to get both.

Issue #1349, opened 2025-11-17
List inside a RadioMultiAnswer not working if the list has only one thing.


Alex-Jordan:
I think this is an issue with parserRadioMultiAnswer.pl. But maybe with Lists.

Consider the exercise below. Imagine a problem where you ask if the MVT applies to a function on a certain closed interval or not. The first option is something like "The MVT does not apply.". The second option is for when it does apply, and there is a list of numbers c where f'(c) = (f(b) - f(a))/(b - a).

That answer blank should expect a List. And for the problem in question, it turns out the list only has one element. Then the problem is not accepting the correct answer as correct.

```
DOCUMENT();

loadMacros(qw(PGstandard.pl PGML.pl parserRadioMultiAnswer.pl));

Context("Numeric");

$list = List(Real(1));
$list2 = List(Real(1),Real(2));

$rma = RadioMultiAnswer([
    ['No numbers.'],
    ["The list \($list\):$BR %s", $list]
], 1);

BEGIN_PGML
[_]{$rma}
END_PGML
ENDDOCUMENT();
```

If there are two numbers in the list, then it's OK.

drgrice1:
This is an issue with both the `parserRadioMultiAnswer.pl` and `parserMultiAnswer.pl` macros.  The default graders don't work for lists.  So you have to provide your own grader in this situation at this point.  Eventually, it would be nice to fix this of course.

drgrice1:
You claim that for a list with two numbers this works, but that isn't actually true.  With your example, if you switch to `$list2`, and then enter `2, 1`, that answer is not accepted as correct and it should be with the usual list defaults (`ordered` is false).  Part of the issue in that case is that `ordered` is an answer checker option, and not a property of the list itself.  So if `$l1 = List(1, 2)` and `$l2 = List(2, 1)`, then `$l1 == $l2` is false.

The issue for a single element list is different though.  The issue in that case is that the correct answer is a MathObject list since you pass it as such.  However, the student answer is not recognized as a list since it does not have the list operator (a comma).

somiaj:
This is more general than just the multianswer macros, fixing it so unordered lists worked with the overloaded  `==` operator would be nice, as this also causes other problems.

dpvc:
[...]

PR #1405, opened 2026-04-30
Make list answers work in MultiAnswer and RadioMultiAnswer part answers.

drgrice1:
This fixes issue #1349 and is the result of the discussion in that issue.  See the issue for details and example problems.

I have had this sitting around for a while and so I am going to go ahead and put it in.  Note there were some comments at the end of the discussion about partial credit.  I implemented what @dpvc and I discussed, but @dpvc said he wanted to think more about the partial credit case.   I think that what is in this should work though.

### DIFF
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -14,7 +14,6 @@ our $answerPrefix = "MuLtIaNsWeR_";    # answer rule prefix
 our $separator    = ';';               # separator for singleResult previews
 
 my @ans_defaults = (
-	checker             => sub {0},
 	showCoordinateHints => 0,
 	showEndpointHints   => 0,
 	showEndTypeHints    => 0,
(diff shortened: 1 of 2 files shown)


drgrice1:
I should also point out that the `partialCredit` option is not  implemented for the `parserRadioMultiAnswer.pl` macro at this point.  To be honest, I am not entirely certain how to score partial credit for that.

somiaj (approved):
Fixes the issue.

Would adding how to use the `$self->{scores}` in custom checkers be useful so people using custom checkers know about this and take advantage of the changes.

pstaabp (approved):
Looks like this fixes the original problem.

somiaj:
Note, it still might be nice to update the POD to let users know how to take advantage of this in their own custom graders, not just the default grader.

drgrice1:
I forgot about that.  I will create a new pull request that does that.  Thanks for reminding me.

PR #1408, opened 2026-05-12
Add method to load texPackages and tikzLibraries to Plots 

somiaj:
This adds two options to plots, `texPackages` and `tikzLibraries` to Plots, that can be used to add additional packages and libraries fro generating TikZ output. This also adds "amsmath" in the list of default packages that are loaded.

### DIFF
--- a/lib/Plots/Plot.pm
+++ b/lib/Plots/Plot.pm
@@ -29,10 +29,12 @@ sub new {
 		axes            => Plots::Axes->new,
 		colors          => {},
 		data            => [],
+		texPackages     => [],
+		tikzLibraries   => []
 	}, $class;
 
 	# Besides for these core options, pass everything else to the Axes object.
-	for ('width', 'height', 'tex_size', 'rounded_corners') {
+	for ('width', 'height', 'tex_size', 'rounded_corners', 'texPackages', 'tikzLibraries') {
 		$self->{$_} = delete $options{$_} if $options{$_};
 	}
 	$self->axes->set(%options) if %options;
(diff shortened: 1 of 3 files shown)


drgrice1 (approved):

drgrice1 (approved):

Alex-Jordan:
Maybe it would be good to also add POD (to `plot.pl`?) to document which latex packages are available.

I wondered how long until someone hits an issue with some other tex package being unavailable? Not to add mission creep, but have you considered making `texPackages` configurable by the problem author using `plots.pl`?

somiaj:
@Alex-Jordan this did come up in the meeting that plots doesn't have a way to configure the LaTeXImage object directly. But this kinda leads down a rabbit hole of how configurable should plots give the various outputs it supports. What about compatibility with jsxgraph, should this allow creating images that cannot also be created in jsxgraph, or whatever other plot output that could be added in the future?

There is more that can be done here, and maybe we should expose more of the LaTeXImage object to plots. This was just a quick fix to allow common things that many would expect so the tikz/pgfplots label output supports most of the stuff in jsxgraph via MathJax. If someone is really wanting to create a plot object with lots of custom libraries, maybe they should use LaTeXImage directly if they want that much control of the latex/tikiz output, instead of plots which is meant to be more independent of the back end of the plot being created. I think the real issue is someone creates something that shows up nicely in jsxgraph/html output only to find at a later date the hardcopy output is missing something.

It might be a loosing battle to try to deal with each use case, so a future PR could expose more of the LaTeXImage configuration to the plots object. But then authors will have to know what additional libraries to load since MathJax auto loads them.

Alex-Jordan:
Ok. But still the limitations could be documented in POD, so it's not a
mystery why some amsmath macro works now, but some amssymb (or whatever)
doesn't.

Alex Jordan
Mathematics Instructor
Portland Community College

pstaabp:
How about staying with only the addition of the amsmath package, but document and amsmath and pgfplots are included?  That might tip others that might want other packages to request them.

somiaj:
I updated this to give an option so a user can load additional `texPackages` and `tikzLibraries`, and documented the options. "amsmath" is still loaded by default.

drgrice1 (approved):
This looks fine.

pstaabp (approved):

PR #1411, opened 2026-05-19
Pg critic fixes for older perl

drgrice1:
Using `//=` in signatures is a feature that is not available until version 5.28 of perl.  In this case that isn't needed since the `walkPGMLTree` is never called without that argument.

I am not entirely certain of why the added `to_string` is needed or why it works with newer versions of perl, but it causes an issue with older perl versions as well.

This fixes issue https://github.com/openwebwork/webwork2/issues/2978.

### DIFF
--- a/lib/WeBWorK/PG/Critic/Utils.pm
+++ b/lib/WeBWorK/PG/Critic/Utils.pm
@@ -62,7 +62,7 @@ use Scalar::Util        qw(blessed);
 use Mojo::Util          qw(md5_sum encode);
 use Env                 qw(PG_ROOT);
 
-use lib curfile->dirname->dirname->dirname->dirname->dirname->child('lib');
+use lib curfile->dirname->dirname->dirname->dirname->dirname->child('lib')->to_string;
 
 require Value;
 
@@ -84,7 +84,7 @@ sub main::Context()                  { return; }
 
 do "$PG_ROOT/macros/core/PGML.pl";
 
-sub walkPGMLTree ($block, $results //= {}) {
+sub walkPGMLTree ($block, $results) {
 	for my $item (@{ $block->{stack} }) {
 		next unless blessed $item && $item->isa('PGML::Block');
 		if ($item->{type} eq 'command') {


Alex-Jordan (approved):

lib/WeBWorK/PG/Critic/Utils.pm
```diff
@@ -62,7 +62,7 @@ use Scalar::Util        qw(blessed);
 use Mojo::Util          qw(md5_sum encode);
 use Env                 qw(PG_ROOT);
 
-use lib curfile->dirname->dirname->dirname->dirname->dirname->child('lib');
+use lib curfile->dirname->dirname->dirname->dirname->dirname->child('lib')->to_string;
```
somiaj, line 65: This seems to be a very specific number of `->dirname`, would this cause issue if `PG_ROOT` was equal to something like `/very/deep/nested/dir/in/narnia/pg`, and not just inside of `/opt/pg`?

somiaj: Or is 'lib` always relative to `PG_ROOT/lib`? Anyways I was just double checking.

drgrice1: `curfile` will always be `/path/to/lib/WeBWorK/PG/Critic/Utils.pm`, where of course `/path/to` depends on where you have the PG repository.  The specific `dirname` calls is exactly what is needed to traverse down to `/path/to`.  I suppose that one of those could actually be dropped and then the `->child('lib')` could be dropped now that I look at this, since this is inside the `lib` directory already.  In any case, this will work regardless of where you call the script from.  You can even make a symbolic link to the script, and it will work.

somiaj: Ahh I see, thanks for the clarification, so this is relative to the `Utils.pm` script.

Alex-Jordan:
@somiaj If you are good with this, would you go ahead and merge it? I am not easily able to test a lot until this is in, or else I have to separately track having this patch in place.

somiaj:
I've only looked at the code and haven't actually had time to test it. I assume it works, but want to test it before merging. Note you could always just `git rebase pg-critic-fixes-for-older-perl` as a quick way to get this patch when testing. I'll test it in a bit after I'm done doing some school work.

pstaabp:
Let's merge this.  It's only the perl critic code and shouldn't affect the general use.

pstaabp (approved):

pstaabp:
@Alex-Jordan let us know if this fixes it.

drgrice1:
This does affect general use as issue https://github.com/openwebwork/webwork2/issues/2978 shows.

Alex-Jordan:
Yes, the changes here resolved the issues I was hitting.

PR #1415, opened 2026-05-23
Enable disable toolbar width

Alex-Jordan:
A bootstrap class that is used with the "Disable Toolbar" popover for the MQ editor has `min-width` set to something, and it makes excess whitespace following the "Disable Toolbar".

Before:

Screenshot 2026-05-23 at 11 17 05 AM

After:

Screenshot 2026-05-23 at 11 16 57 AM

### DIFF
--- a/htdocs/js/MathQuill/mqeditor.js
+++ b/htdocs/js/MathQuill/mqeditor.js
@@ -527,6 +527,7 @@
 
 			const menuEl = document.createElement('ul');
 			menuEl.classList.add('dropdown-menu');
+			menuEl.style.setProperty('min-width', 'unset');
 			const li = document.createElement('li');
 			menuEl.append(li);
 			const action = document.createElement('a');


htdocs/js/MathQuill/mqeditor.js
```diff
@@ -527,6 +527,7 @@
 
 			const menuEl = document.createElement('ul');
 			menuEl.classList.add('dropdown-menu');
+			menuEl.style.setProperty('min-width', 'unset');
```
drgrice1, line 530: Rather than setting this with JavaScript, you should add a custom class on line 529 above (so make that line something like `menuEl.classList.add('dropdown-menu', 'mqeditor-menu');`), and then add
```css
.mqeditor-menu {
	min-width: unset;
}
```
to the `mqeditor.scss` file.

Also, a nicer way to do this in JavaScript is with `menuEl.style.minWidth = 'unset';`.  Although, that is just my preference.

drgrice1: Actually, I king of like just setting this in JavaScript.  Just less code.  So leave it as is, or change to the suggested.  Either is fine.

drgrice1 (approved):

somiaj:
Should this be delayed or not needed after the toolbar is redone, in #1417?

drgrice1:
It can be merged, but I also added this to the code in https://github.com/openwebwork/mathquill/pull/46.  So thanks for finding this in any case.

Alex-Jordan:
If the issue is covered by #1417, I'll close this. Reopen if I misunderstood.

drgrice1:
It is covered by #1417, but only because you opened this pull request and I added the changes here to the changes to the toolbar code that is now part of the MathQuill code.  So thanks for noticing this.

PR #1420, opened 2026-05-25
empty node following booktabs rules

Alex-Jordan:
This addresses https://github.com/openwebwork/webwork2/issues/2697.

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -1616,7 +1616,7 @@ sub getWidthPercent {
 sub hrule {
 	my ($booktabs, $type, $thickness) = @_;
 	if ($booktabs) {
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Seems to fix the issue.

somiaj:
When testing this, it seems to add an extra cell to the table. It does fix the hang, but the generated table in hardcopy has an extra cell.

image

To test this I removed the `PopUp` causing issues, and here is the difference between `PG-2.21` and this branch.

PG-2.21 branch:

image


This branch:

image

Alex-Jordan:
Nice catch @somiaj. Try it now. It only adds the empty node if it is not the last row.

somiaj (approved):

PR #1421, opened 2026-05-25
method to set a particular entry in a matrix to a value

Alex-Jordan:
This is one of the methods that was planned in #1076, but not yet implemented. If `$A` is a matrix, then for example
```
$A->set(7, [1, 2]);
```
replaces `$A` with the same matrix, except the (1,2) position is replaced with `7`. The `7` could be Real, Complex, or Fraction if we are in the appropriate ambient context.

1.  I'm not sure the way that I am mutating `$self` for this is the best. @dpvc, do you have a better suggestion?
2. This doesn't work for a matrix of formulas (or is it a formula that returns a matrix?). For example, `$A->set(Formula('x'), [1, 2]);` returns an error that `set` is not a method for formulas. Does that mean this approach needs to be scrapped for something more robust?
3. I need to add unit testing. I'm waiting for feedback on 1 and 2 first.

### DIFF
--- a/t/math_objects/matrix.t
+++ b/t/math_objects/matrix.t
@@ -215,30 +215,30 @@ subtest 'Extract a column' => sub {
 	}, qr/Column must be a positive integer/, 'Test that an error is thrown for passing a non-positive integer';
 };
 
-subtest 'Set a value' => sub {
+subtest 'Replace a value' => sub {
 	my $A = Matrix([ 1, 2, 3 ]);
 	my $B = Matrix([ [ 1, 2, 3 ], [ 4, 5, 6 ] ]);
 	my $C = Matrix([ [ 1, 2, 3 ], [ 4, 5, 6 ] ]);
 	my $D = Matrix([ [ [ 1, 2 ], [ 3, 4 ] ], [ [ 5, 6 ], [ 7, 8 ] ] ]);
 
-	is $A->set(9, [2]),         2, 'Replace an element from a degree 1 matrix, returning replaced element';
-	is $A->TeX,                 Matrix([ 1, 9, 3 ])->TeX, 'Replace an element from a degree 1 matrix';
-	is $B->set(9, [ 2, 1 ]),    4, 'Replace an element from a degree 2 matrix, returning replaced element';
-	is $B->TeX,                 Matrix([ [ 1, 2, 3 ], [ 9, 5, 6 ] ])->TeX, 'Replace an element from a degree 2 matrix';
-	is $C->set(9, 2, 1),        4, 'Replace an element from a degree 2 matrix, returning replaced element';
-	is $C->TeX,                 Matrix([ [ 1, 2, 3 ], [ 9, 5, 6 ] ])->TeX, 'Replace an element from a degree 2 matrix';
-	is $D->set(9, [ 2, 1, 2 ]), 6, 'Replace an element from a degree 3 matrix, returning replaced element';
+	is $A->replace(9, [2]),      2, 'Replace an element from a degree 1 matrix, returning replaced element';
+	is $A->TeX,                  Matrix([ 1, 9, 3 ])->TeX, 'Replace an element from a degree 1 matrix';
+	is $B->replace(9, [ 2, 1 ]), 4, 'Replace an element from a degree 2 matrix, returning replaced element';
+	is $B->TeX,                  Matrix([ [ 1, 2, 3 ], [ 9, 5, 6 ] ])->TeX, 'Replace an element from a degree 2 matrix';
+	is $C->replace(9, 2, 1),     4, 'Replace an element from a degree 2 matrix, returning replaced element';
+	is $C->TeX,                  Matrix([ [ 1, 2, 3 ], [ 9, 5, 6 ] ])->TeX, 'Replace an element from a degree 2 matrix';
+	is $D->replace(9, [ 2, 1, 2 ]), 6, 'Replace an element from a degree 3 matrix, returning replaced element';
 	is $D->TeX, Matrix([ [ [ 1, 2 ], [ 3, 4 ] ], [ [ 5, 9 ], [ 7, 8 ] ] ])->TeX,
 		'Replace an element from a degree 3 matrix';
 
 	like dies {
-		$B->set(10, [ 1, 2, 3 ]);
+		$B->replace(10, [ 1, 2, 3 ]);
 	}, qr/There should be 2 indices/, 'Check correct number of indices passed';
 	like dies {
-		$B->set(10, [ 3, 1 ]);
+		$B->replace(10, [ 3, 1 ]);
 	}, qr/The first index is outside of the array bounds/, 'Check index is within bounds';
 	like dies {
-		$B->set(Point(1, 2), [ 2, 1 ]);
+		$B->replace(Point(1, 2), [ 2, 1 ]);
 	}, qr/The new entry value should be a Number not a Point/, 'Check replaced value has the same type';
 
 };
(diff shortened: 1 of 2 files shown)


dpvc:
This last version is a significant improvement over your original one.  But I do think it is a bit aggressive to tear down the entire matrix data structure and build a completely new one when all you want to do is replace one entry.  It seems more efficient just to walk into the existing structure and replace the one entry.

Another issue is that it doesn't deal with some error conditions very well.  For example, if you give a column index that is too large, you will get a message about rows needing to be the same size (perhaps not the most clear message for that), but if you give a row that is too large, you get no error, and the matrix is unchanged.  Also, if you give an index of 0, you get the *last* entry, and a -1 gets the second to last, and so on.  This is non-intuitive, at best.

Here is a version that doesn't replace the matrix data but instead just replaces the one entry, and that issues errors about incorrect indices, and so on.  It doesn't allow negative indices, but you could add support for that if you want (but I would make -1 be the last row or column, and disallow a 0 index).

``` perl
sub set {
  my ($self, $value, $indices) = @_;
  $value = Value::makeValue($value) unless Value::isValue($value);
  my @indices = @$indices;
  my $dim = scalar($self->dimensions);
  $self->Error("There should be $dim indices") unless $dim == @indices;
  my $data = $self->{data};
  my $i = pop(@indices) - 1;
  for (my $n = 0; @indices; $n++) {
    my $j = shift(@indices) - 1;
    $self->Error("The " . $self->NameForNumber($n) . " index is outside of the array bounds")
      if $j < 0 || $j >= scalar(@$data);
    $data = $data->[$j]->{data};
  }
  $self->Error("The last index is outside of the array bounds")
    if $i < 0 || $i >= scalar(@$data);
  $self->Error("The new entry value should be " . $data->[$i]->showType . " not " . $value->showType)
    unless $value->type eq $data->[$i]->type;
  my $old = $data->[$i];
  $data->[$i] = $value;
  return $old;
}
```

This also has the added effect of returning the entry that was originally in the matrix.  I thought that might be useful, but you can remove the two lines concerning `$old` if you don't think so.

Alex-Jordan:
Thanks @dpv, although I can't keep committing your work under my name :) I will change to this code, and attribute the commit to you.

I like the extra feature of returning the replaced value.

dpvc:
> I can't keep committing your work under my name

Sure you can.  I'm happy not to have to go through the PR process myself (I do enough of that for MathJax).

I don't care about attribution for these things.  I'm sorry if I'm being too pushy with my code.  In th future, I could just suggest ideas rather than providing code; but I usually test out my ideas before presenting them, and that often means working out the details, and it seems a shame to then make someone else go through the same work all over again.

I'm happy to handle things however you all want, though.

drgrice1:
Any way that good code that works well gets in is good. I agree that if you have worked out the details, then there is certainly no point in making someone else do it again.  And @Alex-Jordan is always good with attribution too (better than most of the rest of us I think).

Alex-Jordan:
I added a unit test for this.

t/math_objects/matrix.t
```diff
@@ -215,6 +215,23 @@ subtest 'Extract a column' => sub {
...
+	my $A = Matrix([ 1, 2, 3 ]);
+	my $B = Matrix([ [ 1, 2, 3 ], [ 4, 5, 6 ] ]);
+	my $C = Matrix([ [ 1, 2, 3 ], [ 4, 5, 6 ] ]);
+	my $D = Matrix([ [ [ 1, 2 ], [ 3, 4 ] ], [ [ 5, 6 ], [ 7, 8 ] ] ]);
+
+	is $A->set(9, [2]),         2, 'Replace an element from a degree 1 matrix, returning replaced element';
+	is $A->TeX,                 Matrix([ 1, 9, 3 ])->TeX, 'Replace an element from a degree 1 matrix';
+	is $B->set(9, [ 2, 1 ]),    4, 'Replace an element from a degree 2 matrix, returning replaced element';
+	is $B->TeX,                 Matrix([ [ 1, 2, 3 ], [ 9, 5, 6 ] ])->TeX, 'Replace an element from a degree 2 matrix';
+	is $C->set(9, 2, 1),        4, 'Replace an element from a degree 2 matrix, returning replaced element';
```
dpvc: Should this be `set(9, [2,1])`?

dpvc (commented):
Should there be tests for the error conditions as well?

t/math_objects/matrix.t
Alex-Jordan: OK, in there now too. I'm not sure how exhaustive to be. Like testing for an index that is too high, but also for one too low. Or testing the first index, a middle index, and the last index. I just went with a representative subset.

Alex-Jordan: No, I made it so you can pass the indices together as an array reference, or just let them be additional trailing inputs. So this test at line 227 should be identical to the test at line 225, just using different syntax. And checking the alternate syntax works.

Alex-Jordan: Sorry, I think I referenced the wrong line numbers. Should have wrote 226 and 228.

dpvc: > No, I made it so you can pass the indices together as an array reference, or just let them be additional trailing inputs.

OK, I didn't notice that (I hadn't looked closely at the code, I'm afraid).  Good to know.

Alex-Jordan: Yeah, I'm not sure it's a good idea. But it's what motivated putting the new entry as the first argument. I think it is more natural in English to say "replace the entry at (1,2) with 3" than to say "overwrite 3 into the entry at (1,2)". But for this purpose of allowing the indices to not be inside an array ref, I made it that way.

I suppose I could have used `pop` to keep the new entry at the end of the input array.

Or just not offer this flexibility.

dpvc:
I forgot to comment on this question:

> This doesn't work for a matrix of formulas (or is it a formula that returns a matrix?)

MathObjects doesn't have an object for a Matrix of Formulas.  Any matrix that has a formula entry is turned into a Formula returning a Matrix.  Some methods or properties on a Matrix require the entries to be constants (e.g., `isZero`, and the various RealMatrix-based functions).  Trying to do something like compute the determinant of a matrix of formulas is potentially possible, but could result in entries that are excessively complex, and the MathObject reduction rules are not sufficient to simplify these adequately.

When a Matrix-valued Formula is just a matrix of formulas, you could probably implement `set()` and `extract()` methods that would walk the Matrix structure to locate the needed entry.  But it would be a bit complicated, as constant rows are stored as Parser::Value objects, while other rows are `Parser::List::Matrix` objects, so you would have to deal with those differently.  And if you set an entry in a constant row to have a Formula entry, you would have to convert it from `Parser::Value` to `Parser::List::Matrix`, and would probably have to deal with its `typeRef` in case the needs adjusting (I haven't looked closely at what would have to be changed).  There are some internal conversion functions (`toFormula()`) that might be helpful, however.

(It would be possible to define more reduction rules for Matrix operations in Matrix-Valued formula, so that, for example, matrix addition could produce the matrix of sums rather than a sum of matrices as the formula.  That would allow the use of `set()` in more circumstances.  But I'm not sure it is worth it in most cases.)

This discussion brings up another potential issue of the current version.  If you do

``` perl
$A = Matrix([[1, 2], [3, 4]]);
$A->set(Formula("x"), [1,1]);
```

you will not get an error (because the Formula type will be `Number`), but the resulting matrix will not operate properly.  It would be good to produce an error for this case as well (using `Value::isFormula()` to detect it).  Because the `set()` method alters the initial matrix (rather than returning a new one), it would be hard to coerce the resulting matrix to a Matrix-valued Formula, which is what it would need to be in order to work properly.

Alex-Jordan:
OK, I added a check that the new entry cannot be a Formula.

Also, I changed the name of the method from `set` to `replace`. Note it was `setElement` in #1076. I can't seem to make up my mind what is a good name for this method.

pstaabp (approved):
Looks good. 

About the name, I think either set or replace is fine.  I guess replace is more descriptive of what this does.  I think I used `setElement` based on another language, but I'm glad you changed it.

dpvc (approved):

PR #1424, opened 2026-05-28
subMatrix method for Matrix Math Objects

Alex-Jordan:
Here is another one from #1076. This one is ripped directly from #1076, and I committed it as @pstaabp's commit. Only change I made was to move the auxiliary subroutine `extractElements` inside this method.  I don't see a need to make that accessible as a method, but you can let me know if it should be.

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -1056,7 +1056,7 @@ perform row operations.
 
 =item * Row Swap
 
-The method C<< Value::Matrix->E(n,[i, j]) >> returns the n by n elementary matrix that 
+The method C<< Value::Matrix->E(n,[i, j]) >> returns the n by n elementary matrix that
 upon right multiplication performs the row swap between rows C<i> and C<j>.
 
 Usage:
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I've discovered a few issues with this. Pausing the PR until I make a few corrections.

Alex-Jordan:
Now three commits total. The first is basically @pstaabp's code from #1076 for `subMatrix`. There were some issues to fix, and also I extended the method's functionality, and this is all in the second commit. The third commit is adding `removeRow` and `removeColumn`, which just call `subMatrix` in the appropriate way. This can be reviewed now.

pstaabp (approved):
Looks good.  Thanks for looking back to that PR to include these.

drgrice1 (approved):
I think this looks good.  Just fix the merge conflicts from merging #1421.  They are rather straightforward to fix, and are just because this code is next to the code from that pull request.

Alex-Jordan:
Merge conflicts are fixed now.

Alex-Jordan:
This has two approvals and no conflicts. Would anyone care to merge?

PR #1425, opened 2026-05-28
Convert to PGML: add the ability to parse loadMacros if the arguments are in a qw

pstaabp:
Add the ability to parse loadMacros if the arguments are in a qw block.  Also, removes empty macros and duplicate macros.

In addition, if it appears that the problem is already in PGML mode, then return the file without changes.

This is in response to openwebwork/webwork2#2908

### DIFF
--- a/bin/convert-to-pgml.pl
+++ b/bin/convert-to-pgml.pl
@@ -78,7 +78,7 @@ sub convertFile ($filename) {
 	my $pg_source = $path->slurp;
 	my $result    = convertToPGML($pg_source);
 	if (ref($result) eq 'HASH' && $result->{errors}) {
-		warn "Error parsing $filename. " . $result->{errors} . "\n";
+		warn "Error parsing $filename. " . $result->{errors};
 		return;
 	}
 
--- a/bin/convert-to-pgml.pl
+++ b/bin/convert-to-pgml.pl
@@ -83,7 +83,7 @@ sub convertFile ($filename) {
 	}
 
 	# Copy the original file to a backup and then write the file.
-	my $new_path    = $backup ? $path : Mojo::File->new($filename =~ s/\.pg/.$suffix/r);
+	my $new_path    = $backup ? $path : Mojo::File->new($filename =~ s/\.pg/.$suffix\.pg/r);
 	my $backup_file = $filename =~ s/\.pg$/.pg.bak/r;
 	$path->copy_to($backup_file) if $backup;
 	$new_path->spurt($result->{pgmlCode});
--- a/bin/convert-to-pgml.pl
+++ b/bin/convert-to-pgml.pl
@@ -70,14 +70,18 @@ sub convertFile ($filename) {
 	my $path = Mojo::File->new($filename);
 	die "The file: $filename does not exist or is not readable" unless -r $path;
 
-	my $pg_source        = $path->slurp;
-	my $converted_source = convertToPGML($pg_source);
+	my $pg_source = $path->slurp;
+	my $result    = convertToPGML($pg_source);
+	if (ref($result) eq 'HASH' && $result->{errors}) {
+		warn "Error parsing $filename. " . $result->{errors};
+		return;
+	}
 
 	# copy the original file to a backup and then write the file
 	my $new_path    = $backup ? $path : Mojo::File->new($filename =~ s/\.pg/.$suffix/r);
 	my $backup_file = $filename =~ s/\.pg$/.pg.bak/r;
 	$path->copy_to($backup_file) if $backup;
-	$new_path->spurt($converted_source);
+	$new_path->spurt($result->{pgmlCode});
 	print "Writing converted file to $new_path\n"      if $verbose;
 	print "Backing up original file to $backup_file\n" if $verbose && $backup;
 }
--- a/bin/convert-to-pgml.pl
+++ b/bin/convert-to-pgml.pl
@@ -78,7 +78,7 @@ sub convertFile ($filename) {
 	my $pg_source = $path->slurp;
 	my $result    = convertToPGML($pg_source);
 	if (ref($result) eq 'HASH' && $result->{errors}) {
-		warn "Error parsing $filename. " . $result->{errors};
+		warn "Error parsing $filename. " . $result->{errors} . "\n";
 		return;
 	}
 
--- a/bin/convert-to-pgml.pl
+++ b/bin/convert-to-pgml.pl
@@ -8,6 +8,14 @@ convert-to-pgml.pl -- Convert pg problem with non-pgml structure to PGML structu
 
     convert-to-pgml -b -s pgml file1.pg file2.pg ...
 
+Options:
+
+    -b|--backup         Create a backup of the original file before converting.
+    -s|--suffix=s       Suffix for the converted files. Default is 'pgml'.
+    -v|--verbose        Print verbose output.
+    -h|--help           Show the help message.
+
+
 =head1 DESCRIPTION
 
 This converts each pg file to PGML formatting.  In particular, text blocks are
@@ -20,14 +28,14 @@ any variables of the form C<$var> to C<[$var]>, scripts from \{ \} to [@ @], and
 to the form C<[_]{}>
 
 Many code features that are no longer needed are removed including
-C<TEXT(beginproblem())>, C<<Context()->texStrings;>> and C<<Context()->normalStrings;>>.
+C<TEXT(beginproblem())>, C<< Context()->texStrings; >> and C<< Context()->normalStrings; >>.
 Any C<ANS> commands are commented out.
 
 The C<loadMacros> command is parsed, the C<PGML.pl> is included and C<MathObjects.pl>
 is removed (because it is loaded by C<PGML.pl>) and C<PGcourse.pl> is added to the
 end of the list.
 
-Note: many of the features are converted correctly, but often there will be errors
+Note: many of the features are converted correctly, but there may be errors
 after the conversion.  Generally after using this script, the PGML style answers
 will need to have their corresponding variable added.
 
@@ -42,28 +50,25 @@ C<pgml> is used. If the C<-b> flag is used, this option will be ignored.
 
 =cut
 
-use strict;
-use warnings;
-use experimental 'signatures';
+use Mojo::Base -signatures;
 
 use Mojo::File qw(curfile);
 use Getopt::Long;
+use Pod::Usage;
 
 use lib curfile->dirname->dirname . '/lib';
 
 use WeBWorK::PG::ConvertToPGML qw(convertToPGML);
 
-my $backup  = 0;
-my $verbose = 0;
-my $suffix  = 'pgml';
-
 GetOptions(
-	"b|backup"   => \$backup,
-	"s|suffix=s" => \$suffix,
-	"v|verbose"  => \$verbose,
+	'b|backup'   => \my $backup,
+	's|suffix=s' => \my $suffix,
+	'v|verbose'  => \my $verbose,
+	'h|help'     => \my $show_help
 );
+pod2usage(2) if $show_help || @ARGV == 0;
 
-die 'arguments must have a list of pg files' unless @ARGV > 0;
+$suffix //= 'pgml';
 convertFile($_) for (grep { $_ =~ /\.pg$/ } @ARGV);
 
 sub convertFile ($filename) {
@@ -73,11 +78,11 @@ sub convertFile ($filename) {
 	my $pg_source = $path->slurp;
 	my $result    = convertToPGML($pg_source);
 	if (ref($result) eq 'HASH' && $result->{errors}) {
-		warn "Error parsing $filename. " . $result->{errors};
+		warn "Error parsing $filename. " . $result->{errors} . "\n";
 		return;
 	}
 
-	# copy the original file to a backup and then write the file
+	# Copy the original file to a backup and then write the file.
 	my $new_path    = $backup ? $path : Mojo::File->new($filename =~ s/\.pg/.$suffix/r);
 	my $backup_file = $filename =~ s/\.pg$/.pg.bak/r;
 	$path->copy_to($backup_file) if $backup;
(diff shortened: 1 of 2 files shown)


lib/WeBWorK/PG/ConvertToPGML.pm
```diff
@@ -70,12 +70,21 @@ This returns a string that is the converted input string.
...
 
+use Mojo::Util qw(dumper);
+
 # This stores the answers inside of ANS and related functions.
 my @ans_list;
 
 sub convertToPGML {
 	my ($pg_source) = @_;
 
+	print dumper ($pg_source);
```
somiaj: Did you forget to remove this?

pstaabp: Yep.  Just fixed.

drgrice1:
There are some outlier cases that are doing some weird things with this.  These cases probably are not worth putting much effort into, but they could happen, and are valid Perl, and this turns them into invalid Perl.

Either `loadMacros(qw{PGstandard.pl MathObjects.pl}, 'draggableProof.pl');` or `loadMacros(qw{PGstandard.pl MathObjects.pl draggableProof.pl},);` becomes
```perl
loadMacros(qw{
	PGstandard.pl
	PGML.pl
	draggableProof.pl
	PGcourse.pl
');
```

The second case is probably more of an issue, because that is probably not that unlikely.

Closely related and more serious though is the following.  `loadMacros(qw{PGstandard.pl MathObjects.pl draggableProof.pl}  );` becomes
```perl
loadMacros(qw{
	PGstandard.pl
	PGML.pl
	draggableProof.pl}
	PGcourse.pl
 );
```
This needs to handle the possibility of white space between the `qw` ending and the ending parenthesis.

pstaabp:
This addresses the cases of @drgrice1.  

I also tried to format the output on a single line if the input was on a single line and multiple lines if the input was that way.

lib/WeBWorK/PG/ConvertToPGML.pm
```diff
@@ -76,6 +76,11 @@ my @ans_list;
 sub convertToPGML {
 	my ($pg_source) = @_;
 
+	# Check that the file is not already in PGML format by looking for PGML.pl in the loadMacros statement.
+	# and there are no BEGIN_TEXT, BEGIN_SOLUTION, etc. blocks.
```
drgrice1: There is something wrong with this comment.  The first line ends with a period, and the second line starts with a lowercase "and".

lib/WeBWorK/PG/ConvertToPGML.pm
```diff
@@ -101,29 +106,67 @@ sub convertToPGML {
...
+				# Remove comments within loadMacros block (should we keep them?)
+				$row =~ s/#.*$//;
+				$macros .= $row;
+				last if ($row =~ /(.*)\);/);
+				++$num_macro_lines;
 				$row = shift @rows;
-				my @mrow = split(/#/, $row);
-				# This only adds the row if there is something relevent to the left of a #
-				$macros .= $mrow[0] if $mrow[0] !~ /^\s*$/;
 			}
```
drgrice1: This could potentially loop forever.  I tried to convert the following to PGML, and that happened:
```
DOCUMENT();

loadMacros('PGstandard.pl', 'PGML.pl', 'parserMultiAnswer.pl', 'PGcourse.pl')

BEGIN_TEXT
Bad stuff
END_TEXT
```

Yes, it is intentionally designed to make this loop forever, but someone using the PG problem editor should not be able to crash the server by trying to convert bad code.  Note that the above code does not cause an infinite loop with the code for this in the PG-2.21 branch.

lib/WeBWorK/PG/ConvertToPGML.pm
```diff
@@ -101,29 +106,67 @@ sub convertToPGML {
...
-			# Split by commas and pull out the quotes.
-			my @macros =
-				grep { $_ !~ /^#/ }
-				grep {
-					$_ !~
-					/(PGstandard|PGML|PGauxiliaryFunctions|PGbasicmacros|PGanswermacros|MathObjects|PGcourse|AnswerFormatHelp).pl/
+			my @macros = ();
```
drgrice1: This should be `my @macros;`.  There is no need for `= ()` as that is the default value for an array variable.

lib/WeBWorK/PG/ConvertToPGML.pm
```diff
@@ -101,29 +106,67 @@ sub convertToPGML {
...
 
-			push(@all_lines,
-				'loadMacros('
-					. join(', ', map {"'$_'"} ('PGstandard.pl', 'PGML.pl', @macros, 'PGcourse.pl'))
-					. ');');
+				@macros =
+					grep {
+						$_
+						&& $_ !~
+						/(PGstandard|PGML|PGauxiliaryFunctions|PGbasicmacros|PGanswermacros|MathObjects|PGcourse|AnswerFormatHelp).pl/
```
drgrice1: Use the `x` flag on this regular expression so that the cases can be separated onto multiple lines and so that this is not longer than 120 characters.

drgrice1 (commented):
This still needs some more work.

lib/WeBWorK/PG/ConvertToPGML.pm
pstaabp: Good point.  I'm wondering the best way to handle this situation.   We could throw an error in this situation, but I don't want this to become a syntax checker.

drgrice1: To an extent, any code formatter needs to be a syntax checker.  If the code it is trying to format is not valid, then it really cannot proceed. All of the code formatters such as `perltidy` and `prettier` work this way.  Basically, what should happen is that if something invalid is encountered, then the formatter should just stop and give some error message.

Regardless, the point is that the infinite loop cannot happen.  So this pull request is stopped until this is a definite no go until this is resolved.

pstaabp: I have a fix for this now.

pstaabp:
This now checks for some parsing of the loadMacros call.  

Additionally, this returns a hash for the output with any parsing errors.  I will post a PR for WeBWorK to handle the errors in the PGeditor.

The command line works with no additional code.

pstaabp:
Note: there is now updated PGEditor code in openwebwork/webwork2#3046 that handles errors in the PG problem editor.

bin/convert-to-pgml.pl
```diff
@@ -70,14 +70,18 @@ ($filename)
 	my $path = Mojo::File->new($filename);
 	die "The file: $filename does not exist or is not readable" unless -r $path;
 
-	my $pg_source        = $path->slurp;
-	my $converted_source = convertToPGML($pg_source);
+	my $pg_source = $path->slurp;
+	my $result    = convertToPGML($pg_source);
+	if (ref($result) eq 'HASH' && $result->{errors}) {
+		warn "Error parsing $filename. " . $result->{errors};
```
drgrice1: Add a newline to the end of the warn so that the line number in the `convert-to-pgml.pl` script is not reported.  That does not pertain to the error, and so should not be shown.  So change this to

```suggestion
		warn "Error parsing $filename. " . $result->{errors} . "\n";
```

lib/WeBWorK/PG/ConvertToPGML.pm
```diff
@@ -76,7 +76,17 @@ my @ans_list;
 sub convertToPGML {
 	my ($pg_source) = @_;
 
-	# First get a list of all of the ANS, LABELED_ANS, etc. in the problem.
+	# Check that the file is not already in PGML format by looking for PGML.pl in the loadMacros statement,
+	# and there are no BEGIN_TEXT, BEGIN_SOLUTION, etc. blocks.
+
+	return { pgmlCode => $pg_source }
+		if ($pg_source =~ /loadMacros\((.*)PGML\.pl(.*)\)/m && $pg_source !~ /BEGIN_(TEXT|HINT|SOLUTION)/);
```
drgrice1: This extremely naive regular expression check is just not going to work.  First, since you do not include the `s` flag on the regular expression, the `.` character does not match newlines.  So unless `PGML.pl` is literally on the same line as the `loadMacros` call, this check will not find it.  So this will catch something like `loadMacros( #PGML.pl)` which it shouldn't, but it will not catch something like
```
loadMacros(qw{
    PGstandard.pl
    PGML.pl
});
```

Note that just adding the `s` flag won't fix this either because then it would still catch the false positive noted above, but would also catch something like
```
loadMacros('PGstandard.pl');
# (This does not use PGML.pl)

In addition, the second regular expression check will also catch many incorrect things.  For example, if someone mentions `BEGIN_TEXT` in a comment in the file, but does not use it.

**Review by drgrice1 — CHANGES_REQUESTED** [review]
Something is definitely not working.  The example given by @somiaj in issue [#2908](https://github.com/openwebwork/webwork2/issues/2908) doesn't even work.  It gives the error `PGML conversion errors: The loadMacros command cannot be parsed.`  In fact, every single file I have tested, regardless of any content that it has, gives that error unless `PGML.pl` happens to be on the same line as the `loadMacros` call.

The `bin/convert-to-pgml.pl` script is really poorly written.  There should at the very least be a `help` option using `pod2usage` from `Pod::Usage`.  If the script is called with no arguments which are required, then instead of dying with the message that arguments must have a list of files the help should be shown.  Also, the `SYNOPSIS` section in the POD should be updated to show the usual help that is shown with the help option with the `pod2usage` method that shows the way the script should be called with all options described.

Also, the `ConvertToPGML.pm` module exports symbols by default.  That shouldn't happen.  `@EXPORT_OK` should be used instead of `@EXPORT`.

I must not have reviewed this well before when you first added this.  Those things should not have been allowed.

File: lib/WeBWorK/PG/ConvertToPGML.pm (hunk 7 above)
Comment (pstaabp) [reply lib/WeBWorK/PG/ConvertToPGML.pm]: I had noticed this a few days ago too.  I think changing to the /s on the RegEx and then 
```
$pg_source =~ /loadMacros\((.*)PGML\.pl(.*)\)\s*;/s
```
parses your first example correctly, but not the 2nd one.  

I do see that putting BEGIN_TEXT comment  will catch it incorrectly.  Would 
```
$pg_source !~ /^\s*BEGIN_(TEXT|HINT|SOLUTION)/ms
```
be sufficient?  

My hope was to catch most of the cases that have already been converted.

Comment (drgrice1) [reply lib/WeBWorK/PG/ConvertToPGML.pm]: You will need `$pg_source !~ /^\s*BEGIN_(TEXT|HINT|SOLUTION)/m`.  Remove the `s` flag for this one.  Then this will be pretty much what the translator uses.  Note that the `s` flag makes `.` match a new line, so the `^` anchor would be nullified if the `s` flag is added.

The `loadMacros` check is more complicated.  As I said, adding the `s` flag is not going to be enough.  I am not sure what to do for this.

**Comment by pstaabp** [comment]
New version of this.  

- Added better documentation of the convert-to-pgml script.
- Separated the loadMacros to a separate subroutine and perform error handling in the subroutine to better capture possible syntax errors.
- Also, add a better way to parse the answer blanks in PGML form.

**Comment by drgrice1** [comment]
This is still not working.  Testing with the source
```
DOCUMENT();

loadMacros(    # PGML.pl)
    qw{
        PGstandard.pl
        MathObjects.pl
        parserPopUp.pl
        PGcourse.pl
    }
);
# (This file does not use PGML.pl.)

$mc = DropDown([ '1.5', '2 ', 3 ], 2);    # Correct answer is 3

BEGIN_TEXT
Problem text.
\{ $mc->ans_rule \}
END_TEXT

ANS($mc->cmp);

ENDDOCUMENT();
```
gives
```
DOCUMENT();

# (This file does not use PGML.pl.)

$mc = DropDown([ '1.5', '2 ', 3 ], 2);    # Correct answer is 3

BEGIN_PGML
Problem text.
[_]{$mc}
END_PGML

# ANS($mc->cmp);

ENDDOCUMENT();
```

It removes the `parserPopUp.pl` macro.

**Comment by pstaabp** [comment]
The newline between the `loadMacros(` and the `qw{` was a problem.  Just pushed another fix.

**Comment by drgrice1** [comment]
The first example I mentioned in https://github.com/openwebwork/pg/pull/1425#issuecomment-4970711382 is still not working.

**Comment by pstaabp** [comment]
I forgot to check that one again.  I was double checking for comments.  That issue is fixed now. 

I decided to strip all comments from any loadMacros call.  I was trying to figure out how to keep them, but also adding PGML.pl and removing any duplicates or unnecessary macros makes that quite difficult.

File: lib/WeBWorK/PG/ConvertToPGML.pm (hunk 5 above)
Comment (drgrice1) [reply lib/WeBWorK/PG/ConvertToPGML.pm]: This was not really resolved.  You added the `x` flag, but did not use it in any meaningful way.  The point was to change this code to
```
		@macros =
			grep {
				$_ && $_ !~ /(
				PGstandard |
				PGML |
				PGauxiliaryFunctions |
				PGbasicmacros |
				PGanswermacros |
				MathObjects |
				PGcourse |
				AnswerFormatHelp
				).pl/x
			}
			map {s/['"]//gr} @macros;
```
That fits the 120 character limit and is very nicely readable.

drgrice1:
Why does the `bin/convert-to-pgml.pl` have a `-b|--backup` option? It never writes to the original `.pg` file, so that option is entirely pointless.

Why does the `bin/convert-to-pgml.pl` write the converted problem to a file with a different extension than `.pg`?  It uses the `-s|--suffix` option for the extension defaulting the `.pgml`.  That is not a good way to go.  It should use an option like that but instead it should use an extension like `.pgml.pg` (using the suffix option instead of `pgml`).  That way you get an actual PG problem file.

drgrice1:
I see, when the `-b|--backup` option is used it does write to the original `.pg` file, but creates a backup first.  In that case it also does not use the `-s|--suffix` option.  Sorry, I missed that.

My second comment regarding the `-s|--suffix` option still stands.

pstaabp:
Another update to add .pg to problem, so .pgml.pg is the default converted file suffix.

pstaabp:
This is closed in favor of #1482

PR #1427, opened 2026-06-02
fuzzy test for row equivalency

Alex-Jordan:
This adds an `isREQ` method for Matrix objects. Used like `$A->isREQ($B)`.

This is a "fuzzy " check that avoids both machine rounding issues and student rounding issues (for example if they enter something like 0.3333 instead of 1/3). Instead of reducing each matrix to RREF (which could introduce extraneous pivot positions from machine rounding), this approach compares the two row spaces. If each row from one matrix is "close enough" to the other matrix's row space, then the algorithm effectively declares that row to be in the other matrix's row space. "Close enough" is based on normalizing nonzero rows to length 1, and then the usual fuzzy tolerances for comparing that normalized row and its projection onto the other matrix's row space.

For example, `Matrix([ 3, 1 ], [ 1, 1 / 3 ])` and `Matrix([ 1, 0.3333 ], [ 0, 0 ])` will be considered row equivalent.

But `Matrix([ 3, 1 ], [ 1, 1 / 3 ])` and `Matrix([ 1, 0.33 ], [ 0, 0 ])` will not be.

And of course `Matrix([ 3, 1 ], [ 1, 1 / 3 ])` and `Matrix([ 1, 0], [ 0, 1 ])` will not be.

For questions that ask a student to completely reduce `$A`, a checker could check `$student->isRREF() && $student->isREQ($A)`.

### DIFF
--- a/lib/Value/Matrix.pm
+++ b/lib/Value/Matrix.pm
@@ -669,36 +669,14 @@ sub isREQ {
 		}
 		return $lk * $r == $rk * $l;
 	} elsif (scalar @ldim == 2) {
-		# get Gram-Schmidt basis for each row space
-		my @rrows = map { $r->new($_) } $r->value;
-		@rrows = grep { !$_->isZero } @rrows;
-		@rrows = map  { $_ / ($_ * $_)**0.5 } @rrows;
-		my @rgs = ($rrows[0]);
-		for my $i (1 .. $#rrows) {
-			my $proj = $r->new((0) x $rdim[1]);
-			for my $v (@rgs) {
-				$proj += ($rrows[$i] * $v) * $v;
-			}
-			my $gs = $rrows[$i] - $proj;
-			push @rgs, $gs / ($gs * $gs)**0.5 unless $rrows[$i] == $proj;
-		}
-		my @lrows = map { $l->new($_) } $l->value;
-		@lrows = grep { !$_->isZero } @lrows;
-		@lrows = map  { $_ / ($_ * $_)**0.5 } @lrows;
-		my @lgs = ($lrows[0]);
-		for my $i (1 .. $#lrows) {
-			my $proj = $r->new((0) x $ldim[1]);
-			for my $v (@lgs) {
-				$proj += ($lrows[$i] * $v) * $v;
-			}
-			my $gs = $lrows[$i] - $proj;
-			push @lgs, $gs / ($gs * $gs)**0.5 unless $lrows[$i] == $proj;
-		}
+		my @rgs = $r->NGS;
+		my @lgs = $l->NGS;
 
 		return 0 if scalar @lgs != scalar @rgs;
 
 		# project each row from $rrows onto @lgs;
 		# if the complement is nonzero, the row spaces disagree
+		my @rrows = map { $r->new($_) } $r->value;
 		for my $v (@rrows) {
 			my $proj = $r->new((0) x $ldim[1]);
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I updated the top comment here, partly to fix typos, but also to explain a little more about what the comparison actually does.

In another thread, @somiaj and I were talking about actually implementing an RREF algorithm. We could do that too, I just couldn't think of a way to avoid extraneous pivot positions popping up from machine rounding. That situation would create "false negatives" where two matrices that should be considered row equivalent would not be. But here, the algorithm goes the other way. It may create "false positives" because of the fuzzy tolerances.

drgrice1:
I might be misunderstanding something, but I tested with the following problem:
```Perl
DOCUMENT();

loadMacros('PGstandard.pl', 'MathObjects.pl', 'PGML.pl');

$A = Matrix([ [ 1, 2, 3 ], [ 4, 5, 6 ], [ 7, 8, 9 ] ]);
$B = Matrix([ [ 1, 2, 3 ], [ 4, 5, 6 ], [ 7, 8, 9 ], [ 14, 16, 18 ] ]);

BEGIN_PGML
[``[$A]``] and [``[$B]``] are [$A->isREQ($B) ? 'row equivalent' : 'not row equivalent']
END_PGML

ENDDOCUMENT();
```
and I get the error `Error evaluating variable $A->isREQ($B) ? 'row equivalent' : 'not row equivalent': Cannot compare row equivalency because matrices differ in the first dimension`.  It seems that these matrices should be comparable, and should be row equivalent.  They have the same row space.

Alex-Jordan:
I could change it for that interpretation. But the implementation is based on A~B meaning that you could apply row operations to one to get the other. And applying row operations to a 3x3 matrix can't produce a 4x3 matrix. Even if they have the same row space.

drgrice1:
I see.  Yeah, that is the standard definition of row equivalent matrices.

pstaabp:
Tested with a few matrices.  Seems to work well.  For the code if you're in the matrix (2 dimension) then there are two blocks of repetitive code that seems like can be simplified.   For example, could make a gram-schmidt sub which is called on both matrices.

drgrice1 (approved):
I think this generally looks good.  I agree with @pstaabp that there could be some refactoring of the code to reduce redundancy.

Alex-Jordan:
I added a (fuzzy) normalized Gram Schmit method (`NGS`), including tests. And now the `isREQ` method calls that in the two places it needs to.

Alex-Jordan:
Stand by. I'm working on a variant that handles Gram Schmidt for the columns.

Alex-Jordan:
OK. Now you can use this like:

```
Value::Matrix->NGS([row1], [row2], ...);
```
or like
```
$A->NGS
```
or like
```
$A->NGS(cols => 1)
```

In array context, you get an array of Matrices for a GS basis. The Matrices are 1D in the first two cases above, giving a GS basis for the indicated rows. The Matrices are 2D, nx1, in the last case; and they give a GS basis for `$A`'s columns.

In scalar context, you get a Matrix that glues the previously described array of vectors together. Either as rows in the first two cases, or as columns in the third.

pstaabp (approved):
Looks good.

pstaabp:
Since @drgrice1 's approval was before refactoring, let's get another approval before merging.

drgrice1 (approved):
This still looks good to me.

PR #1428, opened 2026-06-03
Update the pgaux.t test for subtests and add some new

pstaabp:
Add some new rounding test that will be updated with #1422.   This also switches tests to subtest for organization.

### DIFF
--- a/t/macros/pgaux.t
+++ b/t/macros/pgaux.t
@@ -13,105 +13,115 @@ do "$ENV{PG_ROOT}/t/build_PG_envir.pl";
 
 loadMacros('PGauxiliaryFunctions.pl');
 
-# test step functions
+subtest 'Step function' => sub {
+	is(step(8),  1, "step: positive number");
+	is(step(-8), 0, "step: negative number");
+	is(step(0),  0, "step: step(0)=0");
+};
 
-is(step(8),  1, "step: positive number");
-is(step(-8), 0, "step: negative number");
-is(step(0),  0, "step: step(0)=0");
+subtest 'Floor function' => sub {
+	is(floor(0.5),   0, "floor: positive non-integer");
+	is(floor(-0.5), -1, "floor: negative non-integer");
+	is(floor(1),     1, "floor: positive integer");
+	is(floor(0),     0, "floor: floor(0)=0");
+	is(floor(-1),   -1, "floor: negative integer");
+};
 
-# test floor function
+subtest 'Ceiling function' => sub {
+	is(ceil(0.5),   1, "ceil: positive non-integer");
+	is(ceil(-0.5),  0, "ceil: negative non-integer");
+	is(ceil(1),     1, "ceil: positive integer");
+	is(ceil(0),     0, "ceil: floor(0)=0");
+	is(ceil(-1),   -1, "ceil: negative integer");
+};
 
-is(floor(0.5),   0, "floor: positive non-integer");
-is(floor(-0.5), -1, "floor: negative non-integer");
-is(floor(1),     1, "floor: positive integer");
-is(floor(0),     0, "floor: floor(0)=0");
-is(floor(-1),   -1, "floor: negative integer");
+subtest 'Min and Max functions' => sub {
+	is(max(1, 2, 3, 9, 4, 5, 6, 8),             9,       "max: set of integers");
+	is(max(0.1, -2.3, 1.345, 2.71712, -1000.1), 2.71712, "max: set of decimals");
+	is(min(1, 2, 3, 9, 4, 5, 6, 8),             1,       "min: set of integers");
+	is(min(0.1, -2.3, 1.345, 2.71712, -1000.1), -1000.1, "min: set of decimals");
+};
 
-# test ceiling function
+subtest 'round and Round functions' => sub {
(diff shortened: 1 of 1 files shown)


t/macros/pgaux.t
```diff
@@ -13,105 +13,115 @@ do "$ENV{PG_ROOT}/t/build_PG_envir.pl";
...
+	is(Round(1.793,  2),  1.79, "Round to 2 digits: test 1");
+	is(Round(1.797,  2),  1.80, "Round to 2 digits: test 2");
+	is(Round(1.795,  2),  1.80, "Round to 2 digits: test 3");
+	is(Round(-1.793, 2), -1.79, "Round to 2 digits: test 1");
+	is(Round(-1.797, 2), -1.80, "Round to 2 digits: test 2");
+	is(Round(-1.795, 2), -1.80, "Round to 2 digits: test 3");
+
+	is(Round(15.793, -1), 20, "Round to -1 digits (nearest 10)");
+
+	# tests that round handles some cases related to precision loss in calculations
```
Alex-Jordan: At this point, without the proposed new Round, this comment is perhaps incorrect. The cases below show that Round doe *not* handle that sort of thing.

pstaabp: Are you recommending I remove the comment?

Alex-Jordan: I'm not sure the new Round() will be merged. @drgrice1 found another issue with it, and it's the kind of thing where I will never feel sure that there isn't still some undiscovered issue. We just have to weigh that against the known issue with the current Round().

But since the new Round() might not be merged, this should probably stick to describing the current Round().

t/macros/pgaux.t
```diff
@@ -13,105 +13,113 @@ do "$ENV{PG_ROOT}/t/build_PG_envir.pl";
...
+	is(lcm(5, 6),                30,  "lcm: relatively prime numbers");
+	is(lcm(2, 3, 4),             12,  "lcm: 3 numbers");
+	is(lcm(2, 3, 4, 5, 6, 7, 8), 840, "lcm: 7 numbers");
+
+	# gcd
+	is(gcd(16, 8), 8, "gcd: 2 powers of 2");
+	is(gcd(10, 9), 1, "gcd: 2 relatively prime");
+
+	is(gcd(10, 20, 30, 40), 10, "gcd: 4 multiples of 10");
+
```
drgrice1: There is an unnecessary empty line here.

t/macros/pgaux.t
```diff
@@ -13,105 +13,113 @@ do "$ENV{PG_ROOT}/t/build_PG_envir.pl";
...
+		$sum1 += gcd(@coprimes);
+		$sum2 += gcd($coprimes[0], $coprimes[1]);
+		$sum3 += gcd($coprimes[0], $coprimes[2]);
+		$sum4 += gcd($coprimes[1], $coprimes[2]);
+	}
+	is($sum1 + $sum2 + $sum3 + $sum4, 400, "random_pairwise_coprime: 100 tests of [-9..-1,1..9],[1..9],[1..9]");
+};
+
+subtest 'reduce function' => sub {
+
```
drgrice1: There is another unnecessary empty line here.

t/macros/pgaux.t
```diff
@@ -13,105 +13,113 @@ do "$ENV{PG_ROOT}/t/build_PG_envir.pl";
...
+	# Note: this is testing reduction of fractions. Not sure why this is here.
```
drgrice1: This is a test for the `PGauxiliaryFunctions.pl` macro correct?  It does define a `reduce` function, and that is what this is testing.  Please delete this comment.

t/macros/pgaux.t
```diff
@@ -13,105 +13,113 @@ do "$ENV{PG_ROOT}/t/build_PG_envir.pl";
...
+	my @my_arr = (3, 4);
+	my @res    = reduce(15, 20);
+	is($my_arr[0], $res[0], "reduce: correct numerator");
+	is($my_arr[1], $res[1], "reduce: correct denominator");
```
drgrice1: The previous comment in the code that you added, but are now removing stated that "it would be nicer to directly compare the arrays".  Why is that not done?  If this were changed to


```suggestion
	is(\@res, \@my_arr, "reduce gives correct numerator and denominator");
```

it would do that.

Note that the order of the arguments for these tests is incorrect (corrected above).  The result being tested should be first, and the expected result should be second. 

Also, this test could be made quite concise by changing it to
```
subtest 'reduce function' => sub {
	is([ reduce(15, 20) ], [ 3, 4 ], "reduce gives correct numerator and denominator");
};
```

t/macros/pgaux.t
```diff
@@ -13,105 +13,113 @@ do "$ENV{PG_ROOT}/t/build_PG_envir.pl";
...
+};
 done_testing;
```
drgrice1, line 118: Please add an empty line before `done_testing;` to separate the tests from this line.

drgrice1 (approved):
There is some code clean up needed, but the separation into subtests is nice.

Alex-Jordan (approved):
Marking for approval, but I won't merge given the unresolved comments.

pstaabp:
Updated.

drgrice1 (approved):
This looks good now.

PR #1430, opened 2026-06-06
change ptx 'me' to 'md'

Alex-Jordan:
The PTX `me` tag (math equation) has been deprecated. Now, you just use `md` (display math) with bare content. Previously, `md` was only valid if it had `mrow` children. But now if it has bare content, it serves the function that `me` served.


drgrice1 (approved):
As usual, I will take your word for it on this!

somiaj (approved):
I too will take @Alex-Jordan word that this is correct.

PR #1431, opened 2026-06-07
Rework the `MathJax.typesetPromise` usage.

drgrice1:
The `MathJax.typesetPromise` calls no longer need to be passed through the `MathJax.startupPromise` according to @dpvc. See his comments on this in https://github.com/openwebwork/webwork2/issues/2955.

Also, make the `MathJax.typesetPromise` calls to typeset popover contents specific to the particular popover that needs to be typeset instead of typesetting all popovers in the page. This is not done in the way that @dpvc suggested because moving the typesetting from the `show.bs.popover` event to the `shown.bs.popover` event results in visual motion as the popover is typeset and possibly resized, and that is not desirable. Instead a timeout is used.  This works because Bootstrap creates the `popover.tip` element immediately after Bootstrap triggers the `show.bs.popover` event, and so when the timeout handler is executed it is available to work with. But that is still before the css transitions are complete which is when the `shown.bs.popover` event is triggered.

Note that in `htdocs/js/Knowls/knowl.js` and `htdocs/js/DragNDrop/dragndrop.js`, the `typesetPromise` method is often not yet defined on the `MathJax` object when those calls are made.  That is okay, because if it isn't yet defined, that means that MathJax is still typesetting the page, and gets those elements.  Generally the call is only needed for cases where those things are added to the page later.

### DIFF
--- a/htdocs/js/Feedback/feedback.js
+++ b/htdocs/js/Feedback/feedback.js
@@ -13,7 +13,7 @@
 
 		feedbackBtn.addEventListener('inserted.bs.popover', () => {
 			// Render MathJax previews.
-			if (window.MathJax) MathJax.typesetPromise([feedbackPopover.tip]).then(() => feedbackPopover.update());
+			if (window.MathJax) MathJax.typesetPromise?.([feedbackPopover.tip]).then(() => feedbackPopover.update());
 
 			// Execute javascript in the answer preview.
 			feedbackPopover.tip?.querySelectorAll('script').forEach((origScript) => {
--- a/htdocs/js/Feedback/feedback.js
+++ b/htdocs/js/Feedback/feedback.js
@@ -11,20 +11,10 @@
 		});
 		feedbackPopovers.push(feedbackPopover);
 
-		// Render MathJax previews.
-		if (window.MathJax) {
-			feedbackBtn.addEventListener('show.bs.popover', () => {
-				// The feedbackPopover.tip element is not yet available to work with, but will be immediately after this
-				// event fires, since Bootstrap sets it later in the same method in which this event is triggered. So
-				// delay until it is with a timeout. Note that this still occurs earlier than the shown.bs.popover
-				// event, and so prevents visual motion as the MathJax content is typeset and the popover resizes.
-				setTimeout(() => {
-					MathJax.typesetPromise([feedbackPopover.tip]).then(() => feedbackPopover.update());
-				});
-			});
-		}
+		feedbackBtn.addEventListener('inserted.bs.popover', () => {
+			// Render MathJax previews.
+			if (window.MathJax) MathJax.typesetPromise([feedbackPopover.tip]).then(() => feedbackPopover.update());
 
-		feedbackBtn.addEventListener('shown.bs.popover', () => {
 			// Execute javascript in the answer preview.
 			feedbackPopover.tip?.querySelectorAll('script').forEach((origScript) => {
 				const newScript = document.createElement('script');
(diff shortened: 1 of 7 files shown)


Alex-Jordan (approved):
I'm not seeing any issues from using this branch. I'm not sure what to look at for testing. But I'll mark it as approved.

dpvc:
I still think that using the `show.bs.popover` to initiate the MathJax typesetting is the wrong place for this, and that you are working harder than necessary to overcome the fact that this is the wrong event to use for this.  It relies on subtle timing issues: the fact that the popover isn't currently in the DOM, but MathJax's typesetting is in a promise, and so it not done until the next javascript idle time, with the expectation that the popover will be in place by the time MathJax runs.  That seems fragile to me.

Instead, I'd suggest using the [`inserted.bs.popover` ](https://www.geeksforgeeks.org/bootstrap/bootstrap-5-popovers-events/) event rather than `show.bs.popover`, since that seems to be exactly what you are looking for (the popover is in the DOM but not yet revealed).  Then `popover.tip` will be available and you don't need all the fragile `setTimeout()` commands.

Then in `feedback.js`, you can combine the current `show.bs.popover` and `shown.bs.popover` to a single `inserted.bs.popover` event handler.

If you really want to avoid jitter from MathJax's typesetting of expressions, not just in popovers, but in the page itself, you could add

``` js
document.querySelector('#problem_body').style.visibility="hidden";
```

to the bottom of the `mathjax-config.js` file, and insert

``` js
      pageReady() {
        return MathJax.startup.defaultPageReady()
          .then(() => document.querySelector('#problem_body').style.visibility='');
      },
```

into the `startup` block of the configuration.  This will hide the problem until after MathJax has typeset all the expressions.

drgrice1:
I switched to the the `inserted.bs.popover` event.  Thanks for pointing that out @dpvc.  I didn't even know that event existed, and have missed that in the documentation for years.  It certainly is the correct event, and is better than a timeout.

htdocs/js/Feedback/feedback.js
```diff
@@ -11,16 +11,10 @@
...
-		if (window.MathJax) {
-			feedbackBtn.addEventListener('show.bs.popover', () => {
-				MathJax.startup.promise = MathJax.startup.promise.then(() =>
-					MathJax.typesetPromise(['.popover-body']).then(() => feedbackPopover.update())
-				);
-			});
-		}
+		feedbackBtn.addEventListener('inserted.bs.popover', () => {
+			// Render MathJax previews.
+			if (window.MathJax) MathJax.typesetPromise([feedbackPopover.tip]).then(() => feedbackPopover.update());
```
dpvc: I haven't tested it, but I wonder if the popover update is still needed, as it hasn't been shown, yet, and showing it may get the position/size right automatically.  Just a thought.

dpvc (approved):
Looks good, with one question about an update that might not be needed.

dpvc:
> I didn't even know that event existed, and have missed that in the documentation for years.

I didn't know about it, either, but hoped there might be one for "created but not yet shown", and so looked for the list of events, and there it was!

drgrice1:
Yeah, I might have missed it partially due to the order they list the events in the documentation.  The `inserted` event is listed before both the `show` and `shown` events.  Yeah, I am just making excuses for not reading carefully, but it might have helped if they were listed in the order they are triggered.

drgrice1:
I will check on if the `feedbackPopover.update` is still needed or not.

dpvc:
> it might have helped if they were listed in the order they are triggered.

Agreed.  That would make things clearer.

drgrice1:
I added protection against calling the `typesetPromise` method when it is not defined in all cases.  There were several instances where I was still seeing that it was not.

Also, it seems that the `feedbackPopover.update()` call is still needed.  I still see cases where the popover does not get resized correctly if it is removed.

dpvc:
OK, sounds good.  Thanks for checking!

somiaj (approved):
A quick test of various rendering, popups, and everything worked. There are multiple approvals, but some changes since the last one, I'm going to merge soon so it can get more testing.

PR #1435, opened 2026-06-10
Perltidy update

drgrice1:
Update to `perltidy` version 20260204.

Developers will need to install and use this version to match the workflow.

Also update Github workflows to use node 24, and the actions in the workflows to their newest versions. For now these are still running on Ubuntu 24.04, since an official runner image for Ubuntu 26.04 has not been released yet.  Also because there is a runner image, but some of the rserve unit tests are failing with it.  Note that the rserve tests are passing on my locak machine running Ubuntu 26.

The docker test build is updated to use Ubuntu 26.04.  The rserve tests also are passing with this build.

The first commit in this pull request sets everything up and makes the other changes noted above.  The second commit perltidies the perl files in the code base.

### DIFF
--- a/.github/workflows/check-formats.yml
+++ b/.github/workflows/check-formats.yml
@@ -18,9 +18,9 @@ jobs:
       image: perl:5.38
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
       - name: Install dependencies
-        run: cpanm -n Perl::Tidy@20240903
+        run: cpanm -n Perl::Tidy@20260204
       - name: Run perltidy
         shell: bash
         run: |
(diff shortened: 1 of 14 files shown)


somiaj (approved):
Let's get this in to update the workflows.

Did the sample problems get hit by this? Should they?

drgrice1:
The sample problems are never perltidied by this.  That uses the PG perltidy, and not the general perltidy.  That is a different matter, and for now must be done manually.  The workflow doesn't check those either.  So at this point, no.  They were not and should not be affected by this.

Alex-Jordan (approved):

Alex-Jordan:
I merged the perltidy update PRs. So everyone needs to change their perltidy version to 20260204.

Alex-Jordan:
And rebase PRs, possibly making changes for perltidy.

PR #1436, opened 2026-06-11
don't try to remove a Matrix's only row (or only column)

Alex-Jordan:
Without this PR, you can try to remove the only row of `[[1,2,3]]`, or the only column of `[[1],[2],[3]]`. It won't work, but the error message will be a bit cryptic, coming from the `subMatrix` method. It will say "Cannot use empty array reference for indices to keep".

This commit changes it so that `removeRow` and `removeColumn` catch that kind of thing earlier and give a direct message about it.


somiaj (approved):
Gives a nicer error.

pstaabp (approved):

Issue #503, opened 2020-09-06
texString display for micrometres missing greek prefix


duffee:
Variables declared in micrometres are displayed as 'um' where they really should display as '&mu;m'
```
loadMacros(
  "PGstandard.pl",
  "parserNumberWithUnits.pl",
);

$thickness = NumberWithUnits( 70, 'um');

Context()->texStrings;
BEGIN_TEXT
The thickness of a human hair is about \($thickness\)
END_TEXT
```
The rendered text should display as `\( 70 \ \rm \mu m\)`

I assume that this is also the case for the microNewton, microFarad, microSievert and microMol

duffee:
In a similar vein, the symbol for ohm is &Omega; `\( \Omega \)` which also goes for the k&Omega; and M&Omega;.  If someone could get degrees to display as `^{\circ}`, I'd be a happy camper.

Just getting all my unit display gripes in one place.  I realize that if it were easy it would have been done by now  :)

similar to #504

Alex-Jordan:
A difficulty with this is that we store a simple string for each unit. And this string is expected to display in HTML and in PDF hardcopy output, where `pdflatex` is used. And it is expected to work outside of math mode, and also inside math mode wrapped in `\rm{}`.

As long as `pdflatex` is the PDF production tool, I don't think it will be possible to get these symbols without engineering a new way for the display strings to be generated. Even using `xelatex`, I can get Ω in PDf output, but not μ using the default font.

Or maybe someone else will have better ideas.

drgrice1:
pdflatex has broad support for Greek letters.  `\mu m` will give 'μm'.   So the issue is just in expanding webwork's handling of units to not just store a simple string for the unit name.  There should be a TeX string also.  Much like the way many other things (functions, constants, operators, etc) are handled in contexts.

Alex-Jordan:
If I understand what you are saying, yes I do know that in math mode pdflatex has support for Greek letters. But I just tested `\mu` outside of math mode in a .tex document and it does not work.

This matters because a `NumberWithUnits` object will often be displayed _outside_ of math mode. So now we are talking about a situaton where _if_ the `NumberWithUnits` is in math mode, print `\mu` and if it is not in math mode, print something like `\(\mu\)` and that is the difficulty I was trying to explain.

drgrice1:
I see.  I thought you probably knew about Greek letters in math mode.  I missed that you were talking about getting them outside of math mode.

Alex-Jordan:
[...]

PR #1437, opened 2026-06-12
pretty versions of some units in math mode

Alex-Jordan:
This sort of addresses #503, except by way of the newer `contextUnits.pl`. Each of the supported units from `Units.pm` that has a special unicode character will now print that way, if the unit is being used in math mode. Test with the problem below. It all looks fine in HTML. In hardcopy, we must be careful that the default font we use has glyphs for these characters. This is why two characters are used for degrees Fahrenheit, but a single character can be used for degrees Celsius. Also if someone switched away from xelatex over to pdflatex, these won't work. I'm unsure if we support not using xelatex.

Test with:
```
DOCUMENT();
loadMacros(qw(PGstandard.pl PGML.pl contextUnits.pl));

Context("Units")->withUnitsFor('length', 'temperature', 'angles', 'time', 'force', 'electricity', 'radiation');

$thickness = Compute('70 um');
$boils = Compute('100 degC');
$freezes = Compute('32 degF');
$zero = Compute('0 degK');
$pi = Compute('180 deg');
$us = Compute('us');
$angstrom = Compute('angstrom');
$bacterium = Compute('0.01 uN');
$charge = Compute('1 uC');
$farad = Compute('1 uF');
$ohm = Compute('ohm');
$kohm = Compute('kohm');
$Mohm = Compute('Mohm');
$uSv = Compute('uSv');

BEGIN_PGML
The thickness of a human hair is about [$thickness], or [`[$thickness]`].  
Water boils at [$boils], or [`[$boils]`].  
Water freezes at [$freezes], or [`[$freezes]`].  
Absolute zero is [$zero], or [`[$zero]`].  
[`\pi`] is [$pi], or [`[$pi]`].  
A very short unit of time is a [$us], or [`[$us]`].  
A very short distance is an [$angstrom], or [`[$angstrom]`].  
A bacterium has weight [$bacterium], or [`[$bacterium]`].  
A very small charge is a [$charge], or [`[$charge]`].  
A very small unit of capacitance is a [$farad], or [`[$farad]`].  
[$ohm], [$kohm], [$Mohm], or [`[$ohm]`], [`[$kohm]`], [`[$Mohm]`]  
[$uSv] or [`[$uSv]`]
END_PGML

ENDDOCUMENT();
```

### DIFF
--- a/macros/contexts/contextUnits.pl
+++ b/macros/contexts/contextUnits.pl
@@ -781,7 +781,7 @@ sub addUnit {
 		$constants->add(
 			$name => {
 				value      => context::Units::Unit->new($name => $unit),
-				TeX        => "\\text{$name}",
+				TeX        => '\\text{' . pretty($name) . '}',
 				isUnit     => 1,
 				isConstant => 1
 			}
@@ -793,6 +793,27 @@ sub addUnit {
 		$self->addUnitAliases($unit);
 	}
 	return $self;
+}
+
+sub pretty {
+	my $name       = shift;
+	my %prettyName = (
+		angstrom => "\x{00C5}",
+		deg      => "\x{00B0}",
+		degC     => "\x{2103}",
+		degF     => "\x{00B0}F",
+		degK     => "\x{212A}",
+		ohm      => "\x{2126}",
+		kohm     => "k\x{2126}",
+		Mohm     => "M\x{2126}",
+		uC       => "\x{00B5}C",
+		uF       => "\x{00B5}F",
+		um       => "\x{00B5}m",
+		uN       => "\x{00B5}N",
+		us       => "\x{00B5}s",
+		uSv      => "\x{00B5}Sv",
+	);
+	return defined $prettyName{$name} ? $prettyName{$name} : $name;
 }
 
 #
@@ -1315,7 +1336,7 @@ sub cmp_postprocess {
 #
 sub string {
 	my ($self, $equation, $open, $close, $precedence) = @_;
-	my $string = $self->stringFor('nunits', 'dunits', $self->{order});
+	my $string = context::Units::Context::pretty($self->stringFor('nunits', 'dunits', $self->{order}));
 	$string = '(' . $string . ')' if $string =~ m![ /]! && defined($precedence) && $precedence > 2.9;
 	return $string;
 }
@@ -1502,7 +1523,7 @@ sub uString { (shift)->unit->uString(shift) }
 sub string {
 	my ($self, $equation, $open, $close, $precedence) = @_;
 	my $unit   = $self->unit;
-	my $u      = $unit->stringFor('nunits', 'dunits', $unit->{order}, 0, 1);
+	my $u      = context::Units::Context::pretty($unit->stringFor('nunits', 'dunits', $unit->{order}, 0, 1));
 	my $string = $self->number->string;
 	$string .= substr($u, 0, 1) eq '/' ? $u : " $u";
 	$string = '(' . $string . ')' if defined($precedence) && $precedence > 1;
--- a/macros/contexts/contextUnits.pl
+++ b/macros/contexts/contextUnits.pl
@@ -1525,9 +1525,15 @@ sub string {
 	my $unit   = $self->unit;
 	my $u      = context::Units::Context::pretty($unit->stringFor('nunits', 'dunits', $unit->{order}, 0, 1));
 	my $string = $self->number->string;
-	$string .= substr($u, 0, 1) eq '/' ? $u : " $u";
+	$string .= substr($u, 0, 1) eq '/' || noSeparator($u) ? $u : " $u";
 	$string = '(' . $string . ')' if defined($precedence) && $precedence > 1;
 	return $string;
+}
+
+sub noSeparator {
+	my $u = shift;
+	return 1 if ($u =~ /^(\x{00B0}|\x{2103}|\x{00B0}F)$/);
+	return 0;
 }
 
 #
@@ -1541,7 +1547,7 @@ sub TeX {
 	if (substr($u, 0, 1) eq '/') {
 		$tex = "\\frac{$tex}{" . $unit->raiseUnit(-1, 1)->with(negativePowers => {})->TeX . '}';
 	} else {
-		$tex .= '\\,' . $unit->TeX;
+		$tex .= (noSeparator($unit) ? '' : '\\,') . $unit->TeX;
 	}
 	$tex = '(' . $tex . ')' if defined($precedence) && $precedence > 1;
 	return $tex;


Alex-Jordan:
Now this also prints pretty versions of the same units outside of math mode.

macros/contexts/contextUnits.pl
```diff
@@ -781,7 +781,7 @@ sub addUnit {
 		$constants->add(
 			$name => {
 				value      => context::Units::Unit->new($name => $unit),
-				TeX        => "\\text{$name}",
+				TeX        => "\\text{" . pretty($name) . "}",
```
somiaj: Should probably switch to single quotes here since there is no more variable interpolation.

Alex-Jordan: Is this something perltidy can be configured to catch?

somiaj:
Small thing, but the space before the degree symbol seems a little out of place to me.

Alex-Jordan:
> Small thing, but the space before the degree symbol seems a little out of place to me.

That is not a change, in the sense that currently you have `180 deg`. This changes the `deg` to `°`. But the space in between the quantity and the unit is a global separator. Something could be done wherever that separator is used, to omit that for special units where spaces are not appropriate.

somiaj:
Looking at the mathjax speech text, some seems good, `0.01 micronewtons`, and others not so much, `micro s` or `cap A ring`. Also I noticed that degrees F vs degrees C get different speech text. degree C just gets `℃` but degree F gets `degrees cap F`. Though this might be something MathJax should deal with better, just some things I noticed in my testing.

somiaj:
> > Small thing, but the space before the degree symbol seems a little out of place to me.
> 
> That is not a change, in the sense that currently you have `180 deg`. This changes the `deg` to `°`. But the space in between the quantity and the unit is a global separator. Something could be done wherever that separator is used, to omit that for special units where spaces are not appropriate.

I wasn't staying it was a change, but to me `180 deg` seems proper, but `180 °` doesn't, I would always write it as `180°`. I'm okay with it as is, just some things I've noticed and unsure if we want to address or not.

Alex-Jordan:
Surely, even these speech strings are improvements over what currently
would be there. But I think some of this is controlled by which speech rule
engine the user chooses (or defaults to) in the MJ menu.

Alex Jordan
Mathematics Instructor
Portland Community College

Alex-Jordan:
I have a question. If the unit is initially defined like `Compute('10 microns')` should that display as "10 microns" or "10 µm"?  There are a few other units with word versions that we may or may not want to respect.

somiaj:
Agreed, I wasn't sure if there was a solution or something we could do to improve it, I was just pointing it out.

somiaj:
> I have a question. If the unit is initially defined like `Compute('10 microns')` should that display as "10 microns" or "10 µm"? There are a few other units with word versions that we may or may not want to respect.

Would having some flag that this can be configured be useful to authors? I sometimes like to write units as `10 seconds` and other times as `10 s`.

Alex-Jordan:
I added a commit that essentially creates a list of unit symbols for which there will be no separator. Not sure I implemented it in the best way, but see what you think.

`10 seconds` will be `10 seconds`. That is not the kind of thing happening here. What is happening here is that certain units like `10 angstrom` will now be `10 Å`. But at the same time, this does *not* change `10 angstroms` or `10 Angstrom`. Those still comes out as `10 angstroms` or `10 Angstrom`. My thinking was to only do this kind of prettying on the one lowercase singular-case unit where there is conventionally a special character. Not to just always print the unit the pretty way. The case of `micron` is especially worth thinking about, where it's actually a new word for `micrometer`.

Alex-Jordan:
There are also weird things like `Compute('10 ㏀')`, which will be recognized. And will display using that one weird ㏀ character. But won't display at all in print because our default font has no glyph for that. If an author used ㏀, I would be inclined to let that stay as ㏀, not try to change it to the two-character `kΩ`. But this PR does change `kohm` to `kΩ`.

My point is, I am wary about just deciding at the distribution level, how to display all equivalent units (like how `s` is equivalent to `seconds`). I think it is safer to just do things like specifically display `um` as `µm`.

Alex-Jordan:
I have a feeling I'm not doing one or both of these things here (pretty printing certain units, eliminating certain space separators) the right way. If @dpvc has a chance, it would be good to see what you think, as the designer of this context.

somiaj:
@Alex-Jordan I'll leave it to you to decide what is best, I haven't used units yet, so don't have any experience, just pointing out things I saw. I didn't realize that this only affected a specific case of the units.

macros/contexts/contextUnits.pl
drgrice1: I wish perltidy would do that. Unfortunately it doesn't though. Prettier and eslint do this for JavaScript.

drgrice1:
> Small thing, but the space before the degree symbol seems a little out of place to me.

I have also been bothered by the space between the number and the degree symbol. It is something I have thought about fixing at some point.

drgrice1:
I am not sure that I like the approach used in implementing this.  I think that this approach suffers from maintainability issues.  It seems that rather than defining `pretty` and `noSeparator` methods that have special cases for a few units, the determination as to how to display a unit should be a part of the original unit definition to begin with.  In other words, there should be new keys defined in `lib/Units.pm` for each unit that are looked up and used if defined by the code.  The legacy `NumberWithUnits.pm` code could just ignore these new keys, but the `contextUnits.pl` macro would use this.  This keeps the unit definitions centralized so that if one adds a new unit, then one only needs to make changes in one place, and does not need to look around in the code for oddball functions like these that add properties to a unit such as how it is displayed.

dpvc:
I think @drgrice1's suggestion makes sense.  It would also make it possible for those who use the `addUnits()` method of the units context to add author-defined units to also indicate if special spacing is needed.

Alex-Jordan:
My first attempt at this was like that. I added a "display" key to the "um"
unit. And then something happened that I could not understand. Having
*only* added that key and having changed no other code, I loaded the Units
context, with units for distance. And "um" was no longer an available unit.
I verified this by printing the available units to screen. Is there any
reason the Units context would reject a unit with a new key?

If this doesn't make sense, I'll see if I can reproduce it.

Alex Jordan
Mathematics Instructor
Portland Community College

dpvc:
> I added a "display" key to the "um" unit. ... I loaded the Units context, with units for distance. And "um" was no longer an available unit.

Right.  I just looked back at the `pg/lib/Units.pm` file, and it turns out that you can't easily add new keys to those entries.  There are two special keys, `factor` and `aliases`.  The `aliases` are stripped out by the code at

https://github.com/openwebwork/pg/blob/1603ef5f023131b5a402692561c1cdd2881b5304/lib/Units.pm#L795-L803

That leaves only `factor`, with the other fields representing the fundamental units that are involved with that unit.  When you add `display` there, that makes `um` have a factor of `1E-6` for fundamental units `m` and `display` (even though the latter fundamental unit doesn't exist).  When you ask for distance units, the units context will grab the units that have only `m` fundamental units, so not ones with `m` and `display`.

So it looks like it will not be straight-forward to add keys to the unit definitions.

One possibility would be to pre-process them, like the lines above do with `aliases`, and make a separate hash with the display values, kind of like your original, separate list, and then have the units context access that list, like it does the aliases.  Not quite so nice, but it could be done.

An alternative would be to change the data in the `pg/lib/Units.pm` file to be something like

``` perl
our %UNITS = (
	m => {
		factor  => 1,
		fundamental => {
			m => 1,
		},
		aliases => [ 'meter', 'meters', 'metre', 'metres' ]
	},
	...
);
```

and then build the current `%known_units` and alias hashes from that initial data.  That would keep things backward compatible, while making it possible to add additional keys, like `display` (or `spacing`, or whatever).  Then the units context could use the new `%Units::UNITS` has (rather than the separated `%Units::known_units` and `%Units::unit_aliased_to` hashes) to do its work, where it could read the `display` key.

Issue #1182, opened 2025-01-21
`niceTables.pl` layout table not terminated properly with mismatched cell count


drgrice1:
Here is a MWE that shows the problem:

```perl
DOCUMENT();
loadMacros(qw{PGstandard.pl PGML.pl});

BEGIN_PGML
[#
    [.A.] [.B.] [.C.]*
    [.D.] [.E.]
#]*{ align => 'ccc' }

Content after
END_PGML

ENDDOCUMENT();
```

In that example, the "Content after" is not after the table.  Instead it appears to be in a third row of the table.  Really it is just inside the `div` that is supposed to contain the table.  The following is the generated html for the  problem:

```html
<div class="PGML">
    <div style="margin: auto; display: table; border-collapse: collapse">
        <div style="display: table-row">
            <div style="vertical-align: top; padding: 0.85rem 0.85rem; text-align: center; display: table-cell">A</div>
            <div style="vertical-align: top; padding: 0.85rem 0.85rem; text-align: center; display: table-cell">B</div>
            <div style="vertical-align: top; padding: 0.85rem 0.85rem; text-align: center; display: table-cell">C</div>
        </div>
        <div style="display: table-row">
            <div style="vertical-align: top; padding: 0.85rem 0.85rem; text-align: center; display: table-cell">D</div>
            <div style="vertical-align: top; padding: 0.85rem 0.85rem; text-align: center; display: table-cell">E</div>
            <div style="vertical-align: top; padding: 0.85rem 0.85rem; text-align: center; display: table-cell"></div>
        </div>
        <div style="margin-top: 1em"></div>
        Content after
    </div>
</div>
```

dpvc:
The problem is with this function in `niceTables.pl`:

https://github.com/openwebwork/pg/blob/afae348457c7a312e164d0fbc3947cf89242422f/macros/ui/niceTables.pl#L1512-L1528

In order to pad a short line (when `align` is set), niceTables tries to insert empty cells, but the empty content prevents `tag()` from adding the close tag.  That means the resulting table is improperly nested.  You are seeing that here.

One possible solution is to change 

https://github.com/openwebwork/pg/blob/afae348457c7a312e164d0fbc3947cf89242422f/macros/ui/niceTables.pl#L469

so that the data is `' '` with a space instead of empty.  That way, `tag()` will properly close the `<div>` used for the table.

somiaj:
Fixed in #1438.

PR #1438, opened 2026-06-13
fully close most HTML tags in niceTables

Alex-Jordan:
This addresses #1182. If the `tag` function is making an HTML tag and there is no content, close it like `<div></div>`. Only for a few exceptional tags, close it like `<br>`. The list of exceptional tags is not complete, but the ones left out are not reasonably going to arise in niceTables. In fact, I'm not sure the ones I listed will arise form niceTables either, but it's harmless to list them.

### DIFF
--- a/macros/ui/niceTables.pl
+++ b/macros/ui/niceTables.pl
@@ -1514,8 +1514,7 @@ sub tag {
 		$return .= $inner;
 		$return .= "$separator</$name>";
 	} else {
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Looks good.  Fixes my initial MWE in the issue at least.

somiaj (approved):
Fixes the issue.

PR #1440, opened 2026-06-14
Feedback button changes.

somiaj:
I am using this PR as a discussion (so it is a draft). I'm unsure what others will think of these changes, but this is to address two issues I have with the feedback buttons. I have attempted to implement suggested changes to address the issues.

The first is themes can change the feedback button colors since they are using the bootstrap color classes. This may not be desirable that the feedback buttons match the webwork button colors, so new classes are created for the feedback buttons that can be themed independently of the bootstrap classes.

The second is when `$showPartialCorrectAnswers=0` and the student didn't get the problem correct, the feedback looks like a "Preview Answers" and it confuses students as they think they hit "Preview Answers" and not "Submit Answer" button. This makes it so the feedback has its own look for this case. It uses the yellow color for partially correct, a title of "Unknown" (I think something better could be used here to indicate it is unknown if the answer is correct or incorrect) and a question mark in a circle as the icon (my question mark needs to be improved, I had trouble finding a good SVG question mark).


pstaabp:
About the circle, I think the font awesome one looks good.  Here it is:

```
<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 640 640'><!--!Font Awesome Free v7.2.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2026 Fonticons, Inc.--><path d='M528 320C528 205.1 434.9 112 320 112C205.1 112 112 205.1 112 320C112 434.9 205.1 528 320 528C434.9 528 528 434.9 528 320zM64 320C64 178.6 178.6 64 320 64C461.4 64 576 178.6 576 320C576 461.4 461.4 576 320 576C178.6 576 64 461.4 64 320zM320 240C302.3 240 288 254.3 288 272C288 285.3 277.3 296 264 296C250.7 296 240 285.3 240 272C240 227.8 275.8 192 320 192C364.2 192 400 227.8 400 272C400 319.2 364 339.2 344 346.5L344 350.3C344 363.6 333.3 374.3 320 374.3C306.7 374.3 296 363.6 296 350.3L296 342.2C296 321.7 310.8 307 326.1 302C332.5 299.9 339.3 296.5 344.3 291.7C348.6 287.5 352 281.7 352 272.1C352 254.4 337.7 240.1 320 240.1zM288 432C288 414.3 302.3 400 320 400C337.7 400 352 414.3 352 432C352 449.7 337.7 464 320 464C302.3 464 288 449.7 288 432z'/></svg>
```

somiaj:
Thanks I updated this to be the icon.

somiaj:
After the meeting it seemed that others were okay with these changes so I'll change it from a draft so it can be fully reviewed.

drgrice1 (approved):
I think this looks good now.  This is a much better way to go than the previous attempt from the webwork2 side, and leaves the bootstrap classes available for theming and usage for the general user interface.

PR #1442, opened 2026-06-18
Fix a typo in the AdaptiveParameters sample problem.

drgrice1:
`$self` should be `$ans`.  It seems that I changed the third parameter of the checker from `$self` to the more customary `$ans`, but forgot to change the usage in the method. Since that is used to skip the adaptive parameter check when the equivalence message post filter is called, if you submit an answer that is incorrect, but differs by a constant, and then submit the correct answer you get the message that "your answer is equivalent to the previous answer" even though the previous answer was incorrect, but the current answer is correct.

### DIFF
--- a/tutorial/sample-problems/ProblemTechniques/AdaptiveParameters.pg
+++ b/tutorial/sample-problems/ProblemTechniques/AdaptiveParameters.pg
@@ -43,7 +43,7 @@ loadMacros('PGstandard.pl', 'PGML.pl', 'PGcourse.pl');
 #: If the answers may be a pure multiple of the correct answer without an
 #: additive constant, then `0` should not be considered correct. For example,
 #: if the general solution were `ce^x`. So the commented out line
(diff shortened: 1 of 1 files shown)


tutorial/sample-problems/ProblemTechniques/AdaptiveParameters.pg
```diff
@@ -55,7 +55,7 @@ $aSoln = Compute('e^x - 1')->cmp(
     checker => sub {
         my ($correct, $student, $ans) = @_;
         # return 0 if $student == Formula(0); # See comments.
```
somiaj: Not part of this pull request, but isn't this better to be `Formula(0) == $student`, as I thought it was preferable for the correct answer to be on the left when comparing with student answers, and this will better use the zero checkers?

somiaj (approved):

drgrice1:
Yes, it should be.  I added that.

Alex-Jordan (approved):

PR #1444, opened 2026-06-19
Add two more problem technique sample problems.

drgrice1:
These are both problems from https://wiki.openwebwork.org/wiki/Problem_Techniques that demonstrate valid and current techniques in problems that were not added before for some reason.

Add a `NamedAnswerRules.pg` problem that demonstrates how to use named answer rules to add buttons to the MathQuill toolbar for the named rule. This is an updated version of https://wiki.openwebwork.org/wiki/NamedAnswerRules. With PG 2.21 the way to do so is changing.  So this is needed to demonstrate the new way to do it.

Add an `AskSage.pl` problem that demonstrates how to use the `AskSage` method to send a request to a Sage cell server. This is an updated version of https://wiki.openwebwork.org/wiki/AskSage. Note that this version of the problem depends on the changes in #1399. Also note that without #1399 there is no hope for the `AskSage` method in general (so lets get that merged).


Alex-Jordan (approved):

somiaj (approved):

drgrice1:
Note that this has been merged with a comment in the `AskSage` problem that refers to things from the unmerged #1399 pull request.

somiaj:
Oops, forgot about that. I agree we need to get those in, I just don't use docker and have an easy way to test them which is the only reason I haven't put an approval, is I wanted to test it.

PR #1448, opened 2026-06-25
typos identified by github.com/crate-ci/typos

Alex-Jordan:
This is a massive collection of typo corrections after I used `github.com/crate-ci/typos`. Note that it had some false positives, and I examined each one of these to avoid the false positives.

Most of these are in code comments and POD.

A small number of these actually correct a meaningful typo in code.

An even smaller number of these correct some typo in a variable or method name. In those cases, I checked that nothing else was relying on the bad variable/method name.

### DIFF
--- a/macros/contexts/contextPolynomialFactors.pl
+++ b/macros/contexts/contextPolynomialFactors.pl
@@ -223,7 +223,7 @@ sub cmp_postprocess { }
 ##############################################
 
 package PolynomialFactors;
-our @ISA = ('LimitedPolynomal');
+our @ISA = ('LimitedPolynomial');
 
 sub markFactor {
 	my $self = shift;
--- a/macros/contexts/contextPolynomialFactors.pl
+++ b/macros/contexts/contextPolynomialFactors.pl
@@ -130,7 +130,7 @@ sub checkFactors {
 
 sub checkStrict {
 	my $self = shift;
-	$self->Error("You can only use '%s' between coefficents and variables or between factors", $self->{bop});
+	$self->Error("You can only use '%s' between coefficients and variables or between factors", $self->{bop});
 }
 
 ##############################################
--- a/macros/math/SimpleGraph.pl
+++ b/macros/math/SimpleGraph.pl
@@ -2789,7 +2789,7 @@ flag before constructing the C<Edge> or C<EdgeSet>.  For example, by calling
         map { $_ => { isVertex => 1, caseSensitive => 1 } } 'A' .. 'F'
     );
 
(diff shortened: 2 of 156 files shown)


macros/contexts/contextPolynomialFactors.pl
```diff
@@ -223,7 +223,7 @@ package PolynomialFactors::Formula;
 ##############################################
 
 package PolynomialFactors;
-our @ISA = ('LimitedPolynomal');
+our @ISA = ('LimitedPolynomial');
```
drgrice1, line 226: This change is going to need to be checked.  Although the change is to what the original code intended, at this point it makes a change that could cause unexpected behavioral changes.  Due to the typo, the `PolynomialFactors` package effectively had no super class.  Fixing the typo makes it so that it does.  This means, for instance, that previously calling `$f->isa('LimitedPolynomial')` where `$f` is a `PolynomialFactors` object would have returned false.  But with this fixed, it will return true.  There are other changes that this will cause as well, such as method resolution for the `PolynomialFactors` package.  These may not matter in this case, but it is worth noting and thinking about.

macros/math/SimpleGraph.pl
```diff
@@ -1,4 +1,4 @@
-BEGIN { strict->import; }
+BeEGIN { strict->import; }
```
drgrice1: This is adding a typo!  This is a bad one too since it breaks the macro entirely.

drgrice1 (commented):
This is good, but the code changes need to be checked carefully.

Alex-Jordan:
I'm separating this into commits that hopefully make it easier to review. First one is meant to only be POD typos.

Alex-Jordan:
No promises that I correctly sorted these, but I tried. The POD ones were hardest to categorize with accuracy when just manually looking at a small diff.

pstaabp:
Everything looks good except for the two macros:

- lib/Parser/Legacy/LimitedComplex.pm
- macros/contexts/contextPolynomialFactors.pl

which has a typo in the name.  It may be easier to check them separately.   How about pull those out and put them in separately?

macros/contexts/contextPolynomialFactors.pl
drgrice1: I think this is fine, and you can leave it.

drgrice1 (approved):

pstaabp (approved):
I think the only use of the typos in 
-lib/Parser/Legacy/LimitedComplex.pm
-macros/contexts/contextPolynomialFactors.pl

appear no where else.

PR #1450, opened 2026-06-27
Use `isa` to determine if a given image is a `Plots::Plot`, `WWPlot`, or `LaTeXImage`.

drgrice1:
This simplifies the checks in the `PGbasicmacros.pl` `image` method that determine if an image is an instance of a `Plots::Plot`, `WWPlot`, or `LaTeXImage` package and makes it so that any future package that derives from one of those will work without further modification to the `image` method. This uses the `isa` method wrapped in an `eval`.  So if something is not blessed, the `isa` call will throw an exception, but the `eval` catches the exception and ignores it returning false.

This is technically not quite as good as using `blessed $image_item && $image_item->isa('Package::Name')`, but here should be fine.  The one case that this will now throw an exception for that the `Scalar::Util::blessed` method would fix is if `$image_item` is a literal package name that satisfies the `isa` call. For example if someone were to call `image(WWPlot)` or `image('WWPlot')`. That will now throw an uncaught exception, since `WWPlot->isa('WWPlot')` will not throw an exception here and will return true.  So the `insertGraph` method will be called with the package name `WWPlot`, and that code will throw the exception because it cannot work with the `WWPlot` package name. That should not be a real problem though, since no one should do that. Without this pull request that still wouldn't work, but it would only give the warning "The file name "WWPlot" does not have an extension. Every file name used as an argument to alias must have an extension. The permissible extensions are .gif, .jpg, .png, .svg, .pdf, .mp4, .mpg, .ogg, .webm, .css, .js, .nb, .csv, .tgz, and .html," but the problem would still render.  The same would happen for any package name that derives from one of `WWPlot`, `Plots::Plot`, or `LaTeXImage`.

### DIFF
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -2923,7 +2923,7 @@ sub image {
 			);
 			next;
 		}
-		if (ref $image_item eq 'Plots::Plot' || ref $image_item eq 'Plots::StatPlot') {
+		if (eval { $image_item->isa('Plots::Plot') }) {
 			# Update image attributes as needed.
 			$image_item->{width}    = $width    if $out_options{width};
 			$image_item->{height}   = $height   if $out_options{height};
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I know it's not part of this, but reading that list make me wonder about
".jpeg" and ".svg".

Alex Jordan
Mathematics Instructor
Portland Community College

drgrice1:
You are talking about the listed allowed extensions in `PGalias.pm`, correct?  It would be an easy thing to allow the `jpeg` extension.  `svg` is already allowed though.  I think you missed that.

pstaabp (approved):

Alex-Jordan (approved):

PR #1476, opened 2026-07-21
Update npm dependencies once more.

drgrice1:
The only dependencies that were updated that are not for the `generate-assets.js` script are `jsxgraph` and `plotly.js-dist-min`. The `plot.js-dist.min` update seems to not cause any issues or need any code changes.  The `jsxgraph` update allows doing what the FIXME in `htdocs/js/GraphTool/graphtool.js` says and remove the workaround that was added to fix tab order.

The `generate-assets.js` script still seems to be working fine, so the development updates should be fine also.

The security vulnerabilities that just showed up were fixed with the upgrade.


drgrice1:
Apparently prettier changed something and the `js/GraphTool/segments.js` file needed to be formatted.

Alex-Jordan (approved):

somiaj (approved):

PR #1477, opened 2026-07-22
changes to support PTX sidebyside elements as PG layout tables

Alex-Jordan:
These changes support a PTX element called a `sidebyside`. Bascially, PTX converts a `sidebyside` to PGML for a niceTables layout table. Then the changes in this PR make sure that the static PTX output form such a layout table are reasoanable as static PTX.

All the changes here only matter for PTX display mode.


drgrice1 (approved):
This only affects the `PTX` display mode, so as usual I am deferring to your judgement on this.

pstaabp (approved):
Ditto.

PR #1480, opened 2026-07-24
changed units ml 

sfiedle1:
This proposed edit will change the milliliter symbol from ml (lowercase l) to mL (uppercase L).  This will make it consistent with the deciliter abbreviation (dL), which is already present.

### DIFF
--- a/lib/Units.pm
+++ b/lib/Units.pm
@@ -221,7 +221,7 @@ our %known_units = (
 		factor => 0.001,
 		m      => 3
 	},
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Larger scale work that will accomplish this is being developed in #1447. Within the next few days I will review the recent suggested changes there and I am hoping to get that PR approved for 2.21.

sfiedle1:
#1447 would supersede #1448.  Thank you for bringing it to my attention. I'll close this request.

PR #1484, opened 2026-07-31
Explicitly set start/end arrow color in JSXGraph output.

### LINKED ISSUE #1483: plots.pl - start_mark and end_mark don't respect the color option
When graphing a formula using plots.pl, `start_mark` and `end_mark` are plotted in black in jsgraph even if `color` is set.  In hardcopy they match the color of the graph.  Here's a MWE:

```
DOCUMENT();

loadMacros(
    "PGstandard.pl","PGML.pl",'plots.pl',"PGcourse.pl");

$plot=Plot();
$plot->add_function('.5x','x',-5,5,color=>'red',start_mark=>'arrow', end_mark=>'arrow');

BEGIN_PGML
[! !]{$plot}
END_PGML

ENDDOCUMENT();
```

somiaj:
In `Plots::Plot`, explicitly set the start/end arrow color instead of relying on inheritance, since safari SVG doesn't correctly inherit the color of the curve for the marks. Fixes #1483.

### DIFF
--- a/lib/Plots/JSXGraph.pm
+++ b/lib/Plots/JSXGraph.pm
@@ -233,9 +233,21 @@ sub get_options {
 		$drawOptions{dash}        = $self->get_linestyle($data);
 		$drawOptions{strokeColor} = $self->get_color($data->style('color'));
 		$drawOptions{strokeWidth} = $data->style('width');
-		$drawOptions{firstArrow}  = { type => 2, size => $data->style('arrow_size') || 8 }
+
+		# Safari doesn't correctly inherit the stroke color, so explicitly set it for arrows.
+		$drawOptions{firstArrow} = {
+			type        => 2,
+			size        => $data->style('arrow_size') || 8,
+			strokeColor => $drawOptions{strokeColor},
+			fillColor   => $drawOptions{strokeColor}
+			}
 			if $data->style('start_mark') eq 'arrow';
-		$drawOptions{lastArrow} = { type => 2, size => $data->style('arrow_size') || 8 }
+		$drawOptions{lastArrow} = {
+			type        => 2,
+			size        => $data->style('arrow_size') || 8,
+			strokeColor => $drawOptions{strokeColor},
+			fillColor   => $drawOptions{strokeColor}
+			}
 			if $data->style('end_mark') eq 'arrow';
 	}
 


pstaabp:
Although the code looks like it should fix the problem, this is what I see:

image

with the code: 
```
$plot = Plot(
    xmin        => -2.2,
    xmax        =>  2.2,
    ymin        => -2.2,
    ymax        =>  2.2,
    xtick_delta =>  0.5,
    ytick_delta =>  0.5,
    xlabel      => '\(x\)',
    ylabel      => '\(y\)',
);
$plot->add_dataset([0,0],[1.5,1.5], end_mark => 'arrow', color => 'red');
```

and I did confirm that the `$drawOptions` hash has been updated correctly.

somiaj:
Thanks for testing. I got the possible fix from AI and seems it hallucinated to can set the color of the `firstArrow` or `lastArrow`. According to the docs setting the color of the arrow cannot be done explicitly like I attempted, as the only options are arrow size and type.

This might just be a JSXgraph bug that needs to be filed over there and not fixable by us. I don't see any other way to set the color of the `firstArrow` or `lastArrow`.

somiaj:
I found a bug that was suppose to fix this, https://github.com/jsxgraph/jsxgraph/issues/772

Here is a test, does this also show a black arrow head? https://jsfiddle.net/tc9oud4g/

Trying to figure out if this is something we can fix or if we need to report this to jsxgraph as maybe their previous arrow head fix didn't deal with the case we are using with `firstArrow` and `lastArrow`.

pstaabp:
> Here is a test, does this also show a black arrow head? https://jsfiddle.net/tc9oud4g/

That still has a black arrow head. Reading up on this, sounds like quite a pain.  I even tested with Safari Technology Preview, which features an upcoming codebase, but no luck.  

We could add line in the documentation related to colors of the original Model T:  "You can choose any color you want as long as that color is black.".  😄

Alex-Jordan:
I started browsing jsxgraph's repo, and what I see is that there are instances of `strokecolor` that I suspect should be `strokeColor`. (Unfortunately, changing your patch to use `strokecolor` did not fix anything.)

[Here](https://github.com/jsxgraph/jsxgraph/blob/4b97351634951e3e36277f35270dc37edb26bbae/src/renderer/svg.js#L377) seems to be where the Safari deficiency is handled, and `strokecolor` is used.

But most places in the code base, I see `strokeColor`. I even see some places where `strokecolor` is used alongside `strokeWidth` which suggests it's a typo, not intentional. 

So it's just a hypothesis, but maybe this is the issue. Do you have a way to take jsxgraph's repo, make edits, build and test? Or maybe it's quicker just to ask the developers to check if this is a typo.

drgrice1:
JSXGraph options are actually case insensitive.  So both `strokeColor` and `strokecolor` will work.

Alex-Jordan:
>  I even tested with Safari Technology Preview

It sounds like even if Safari moves to support `context-stroke`, you wouldn't see things start working. Because currently jsgraph detects that you are using Safari (with or without support for `context-stroke`) and then goes off to do something special.

Alex-Jordan:
> JSXGraph options are actually case insensitive. So both `strokeColor` and `strokecolor` will work.

OK, but what if jsxgraph's code (javascript, etc) is mixing cases? Would line 379 at that link I posted care about case?

drgrice1:
The line you are referencing calls `el.evalVisProp('strokecolor')`.  The `evalVisProp` method is exactly what makes JSXGraph options case insensitive.  It converts the passed key to lowercase.  All properties are stored internally in all lowercase.

somiaj:
I am closing this. `firstArrow` and and `lastArrow` do not honor color settings, they only honor a very limited number of settings (I didn't double check the docs and believed an hallucinating google AI): 

https://jsxgraph.uni-bayreuth.de/docs/symbols/Line.html#firstArrow

The issue is with JSXGraph and not how we are using it. We should report the issue there.

Closing, future conversation should be done on the actual issue, #1483.

PR #1486, opened 2026-08-02
Transfer aria labels of text inputs to MathQuill inputs.

drgrice1:
This is something that has been missed with MathQuill inputs for a while.  The original text input has an `aria-label`, but when it is hidden that becomes meaningless.  So transfer that label to the MathQuill input.  The input also needs the `role` to be set to `textbox` for this to work.

### DIFF
--- a/htdocs/js/MathQuill/mqeditor.js
+++ b/htdocs/js/MathQuill/mqeditor.js
@@ -25,6 +25,8 @@
 
 		const answerQuill = document.createElement('span');
 		answerQuill.id = `mq-answer-${answerLabel}`;
+		answerQuill.setAttribute('aria-label', input.getAttribute('aria-label'));
+		answerQuill.role = 'textbox';
 		answerQuill.input = input;
 		input.classList.add('mq-edit');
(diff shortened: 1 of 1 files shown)


Alex-Jordan (approved):

somiaj (approved):

somiaj:
Thanks. I'm just curious how this works with a screen reader (I wasn't able to get orca to work for my window manager). Anyways, the label is now applied to a `span` which contains the mathQuill `textarea`. So do screen readers read both the label in the containing `span` and the actual mathQuill `textarea` which contains "Math Input" then the text for reading the actual formula inputted into MathQuill.

drgrice1:
Basically, it works much as for a regular text input.  When the MathQuill input is focused Gnome Orca reads "answer 1 entry ..." where "answer 1" is the aria label and `...` is the content of the input if any.  That is the same as what is read for a text input, except the content reads differently.  For the MathQuill input it reads the hidden mathspeak span, and for a text input it directly reads the content as text.

PR #1487, opened 2026-08-03
2 small changes to improve behavior in RTL problems/courses

taniwallach:
Changes for problems in courses in RTL languages:

1. The DIV containing the standard answer boxes are set to use `dir='ltr'` so that then result/feedback icon appears to their right even when a answer box is used directly in the text in a problem which is being rendered in an RTL setting. This will match the behavior which occurs when answers are inside an LTR span in order to handle interleaved math and answers in RTL language problems. The change makes things consistent for problems in RTL languages. This change was suggested to me by @drgrice1 (and has been in use with WW 2.19 on my servers as a local change).
    - A unit test needed to be updated as a result of this change.
3. Set the CSS for the feedback message box to have a default setting of `dir='ltr'`. This will be changed automatically to `dir='rtl'` when the course language is either set to Hebrew or Arabic (as those languages are set by `get_lang_and_dir` of `webwork2/lib/WeBWorK/Utils/LanguageAndDirection.pm` to put `dir='rtl'` on the main `HTML` tag, so will use the RTL version of the CSS.) That will make feedback messages in such courses (which are expected to mostly be in the RTL language) render in the appropriate RTL manner.

Neither change should have **any** visible effect on problems in English or other LTR languages.

I think the most critical thing to confirm that no changes will occur in courses in English and other LTR languages.

A PG problem for testing is provided below. Testing requires suitable setting made in `simple.conf` or via the config menu.

---

Testing of the fix to the location of the feedback message requires temporarily setting either 
`$language = 'ar';` (Arabic) or `$language = 'he-IL';` (Hebrew).

Testing with the course language in Arabic is probably easier, as there is not much actual translation of the UI to Arabic, so essentially everything will be in english but places in a right-to-left manner.

The goal of the first change is that the "math" in the feedback message when the first answer is incorrect will appear on the left side, while without the change it appears on the right side. (Images below.)

---

Testing of the fix to the location of the result/feedback button require setting 
`$perProblemLangAndDirSettingMode` to a value which will allow RTL rendering of a problem. The suggested setting is `$perProblemLangAndDirSettingMode = 'auto:en:ltr';` as that will default to leaving problems in English and LTR, but accept overrides made by in the PG code of any given problem as is done in every Hebrew problem.

Before the change, the button should appear on the left of the answer box on the bottom line, and after the change, it should appear on the right side. (Images below.)

---

PG problem for testing:

```
DOCUMENT();

SET_PROBLEM_LANGUAGE('he-IL');
SET_PROBLEM_TEXTDIRECTION('rtl');

Context("Numeric");
$ans = Compute('pi/3')->cmp(
    checker => sub {
        my ($correct, $student, $ansHash) = @_;
	$ansHash->{ans_message} = 'נסה שנית. לא מתקיים \( \cos(x) = \frac{\sqrt{3}}{2} \).'
            if cos($student) != sqrt(3) / 2 && !$ansHash->{isPreview};
        return cos($correct) == cos($student);
    }
);


$ans2 = Real(0)->cmp(
    checker => sub {
        my ($correct, $student, $ansHash) = @_;
        $ansHash->{ans_message} = 'נסה שנית. יש להזין \( 0 \).'
            if $student != 0 && !$ansHash->{isPreview};
        return $correct == $student;
    }
);

BEGIN_PGML

הזן ערך של
[`x`]
עבורו
[`\cos(x) = [@ Compute('cos(pi/3)') @]`].

[@ openSpan( { 'dir' => 'ltr' } )  @]*
Enter a value of [`x`] for which [`\cos(x) = \frac{\sqrt{3}}{2} `].
[@ closeSpan() @]*

[@ openSpan( { 'dir' => 'ltr' } )  @]*
[`x =`] [___]{$ans}
[@ closeSpan() @]*

רשום 0 כאן:
[___]{$ans2}

END_PGML

ENDDOCUMENT();
```

---

Image of the confusing placement of the result/feedback button before these changes. It is the second button which is placed in a confusing manner. After the changes - it should be on the right side.

rtl-issue-1

---

Image of the flawed (LTR) layout of a Hebrew feedback message before the change. The `cos` term appears on the far right, but needs to be on the left.

rtl-issue-2

---

Image of the correct (RTL) layout of a Hebrew feedback message after the change. (Will only appear this way for a course with the main language set to Hebrew or Arabic, as otherwise the RTL version of the CSS files will not be used.)

rtl-issue-2-after-fix

### DIFF
--- a/htdocs/js/Problem/problem.scss
+++ b/htdocs/js/Problem/problem.scss
@@ -412,6 +412,7 @@
 				}
 
 				&.feedback-message {
+					direction: ltr;
 					background-color: #ede275;
 					&:not(:last-child) {
 						border-bottom: 1px solid black;
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -301,6 +301,7 @@ sub NAMED_ANS_RULE {
 		HTML => tag(
 			'div',
 			class => 'text-nowrap d-inline',
+			dir   => 'ltr',
 			tag(
 				'input',
 				type           => 'text',
--- a/t/pg_problems/problem_file.t
+++ b/t/pg_problems/problem_file.t
@@ -23,7 +23,7 @@ is(
 	qq{<div class="PGML">\n}
 		. qq{Enter a value for <script type="math/tex">\\pi</script>.}
 		. qq{<div style="margin-top:1em"></div>}
-		. qq{<div class="text-nowrap d-inline">}
+		. qq{<div class="text-nowrap d-inline" dir="ltr">}
 		. qq{<input aria-label="answer 1 " autocapitalize="off" autocomplete="off" class="codeshard" }
 		. qq{dir="auto" id="AnSwEr0001" name="AnSwEr0001" size="5" spellcheck="false" type="text" value="3.14159">}
 		. qq{<input id="MaThQuIlL_AnSwEr0001" name="MaThQuIlL_AnSwEr0001" type="hidden" value="">}


Alex-Jordan:
The button uses bootstrap's `ms-2` (margin "start") class. So there is a margin to its left (when in LTR) that separates it from the answer blank some. In some of your images, that margin is missing. Because now in RTL, the "start" is to the right. If the button is to always appear to the right of an answer blank, even in RTL languages, would you want that bootstrap class to change to `ml-2`, always to the left?

Then there is the position of the badge (the little yellow circle). I don't know what feels correct to an RTL language reader. Does it make sense for that to be in the upper left as in your screen shots, or the upper right, the opposite side from the answer blank?

These things I mention here are perhaps separate from your PR. But not unrelated, so I thought I'd ask what you think.

drgrice1:
Bootstrap does not have a `ml-2` class.  The only options are `s` (start), `e` (end), or `x` (both) for horizontal margins and padding.  So something else would need to be done to force a margin to always be on the left.

taniwallach:
@Alex-Jordan Thanks. I really did not think of such fine tuning.

> The button uses bootstrap's `ms-2` (margin "start") class. So there is a margin to its left (when in LTR) that separates it from the answer blank some. In some of your images, that margin is missing. Because now in RTL, the "start" is to the right. If the button is to always appear to the right of an answer blank, even in RTL languages, would you want that bootstrap class to change to `ml-2`, always to the left?

I really did not worry about the margin... and just tried to see if I could quickly figure out how to change this, and was not successful (`ml-2` did not work, and @drgrice1 just explained why). I doubt that forcing margins on both sides is that nice visually for English courses. 

Bottom line: The margin is not mission critical, so for now - I think this is good enough. 

Note: For other types of answers (radio button groups, pull downs, etc.) the placement of the button was not modified, so it will appear differently from what occurs in English. Over about 2 years of use of WW 2.19 where - no one has complained that the location of the buttons on other types of answers was confusing.

> Then there is the position of the badge (the little yellow circle). I don't know what feels correct to an RTL language reader. Does it make sense for that to be in the upper left as in your screen shots, or the upper right, the opposite side from the answer blank?

About the badge - I don't think it really matters where it appears. Students should notice it.

somiaj (approved):
Looks okay to me. I will leave it @taniwallach to have tested it works well.

As for the margin, in most cases I don't think using `mx-2` would cause problems. Though there are problems where the button is put inside a sentence/paragraph and that might have a little more spacing, but don't think it would cause to much problems to have a little extra margin there.

Alex-Jordan (approved):
Approving.

@taniwallach will understand best what to do for RTL support. But conceptually, I would guess that if you want the button to come after the input field in the DOM, and also always appear to the right of the input field, then you would wand an LTR span wrapping the input field and the button. And then do what is needed within those elements to go back to RTL (like display the feedback message RTL; don't let it inherit LTR).

PR #1489, opened 2026-08-04
Add ansLabels option to parserMultiAnswer.

somiaj:
`ansLabels` is an array ref which is a list of suffixes to append to the end of the `aria-label` of each answer blank. These can be used to improve the accessibility of the answer blanks when used in more complex answers, such as stating which blank is the left/right hand side of an equation, the different parts of an integral, etc.

I think being able to add to the `aria-label` of answer boxes could be helpful. For instance stating things like "answer 2 part 1 left hand side" or "answer 3 part 3 upper bound" could be useful with multianswer problems to describe what the different answer blanks are for in more complicated settings.

I thought about allowing custom `aria-label` but thought knowing what answer/part of the problem is useful as well, so I went with a suffix that is appended to the default `aria-label`.

I was also looking on how to do this to regular answer blanks (maybe mathObjects specifically), so I setup `generate_aria_label` to be able to do this for that too, but I couldn't figure out the best way to allow a user to send that information to this method. Was thinking using `cmp_options`, but didn't see a proper way to do this so leaving that for now. This can be extended to other answer blanks depending on what others thing of this feature.

### DIFF
--- a/macros/parsers/parserMultiAnswer.pl
+++ b/macros/parsers/parserMultiAnswer.pl
@@ -48,6 +48,7 @@ sub new {
 		singleResult        => 0,
 		namedRules          => 0,
 		cmpOpts             => undef,
+		ansLabels           => [],
 		checkTypes          => 1,
 		allowBlankAnswers   => 0,
 		tex_separator       => $separator . '\,',
@@ -416,33 +417,38 @@ sub NEW_NAME {
 	main::RECORD_FORM_LABEL(shift);
 }
 
+sub generate_aria_label {
+	my ($self, $name, $part) = @_;
+	return main::generate_aria_label($self->{singleResult} ? $answerPrefix . $name . '_' . $part : $name,
+		$self->{ansLabels}[$part]);
+}
+
 #
 #  Produce an answer rule for the next item in the list,
 #    taking care to use names or extensions as needed
 #    by the settings of the MultiAnswer.
 #
 sub ans_rule {
-	my $self = shift;
-	my $size = shift || 20;
-	my $data = $self->{data}[ $self->{part} ];
-	my $name = $self->ANS_NAME($self->{part}++);
-	if ($self->{singleResult} && $self->{part} == 1) {
-		my $label = main::generate_aria_label($answerPrefix . $name . "_0");
-		main::RECORD_IMPLICIT_ANS_NAME($name) unless $self->{namedRules};
-		return $data->named_ans_rule($name, $size, @_, aria_label => $label);
-	}
-	if ($self->{singleResult} && $self->{part} > 1) {
-		my $extension_ans_rule = $data->named_ans_rule_extension(
+	my $self  = shift;
+	my $size  = shift || 20;
+	my $part  = $self->{part};
+	my $data  = $self->{data}[$part];
+	my $name  = $self->ANS_NAME($self->{part}++);
+	my $label = $self->generate_aria_label($name, $part);
+	if ($self->{singleResult}) {
+		if ($part == 0) {
+			main::RECORD_IMPLICIT_ANS_NAME($name) unless $self->{namedRules};
+			return $data->named_ans_rule($name, $size, @_, aria_label => $label);
+		}
+		return $data->named_ans_rule_extension(
 			$name, $size,
 			answer_group_name => $self->{answerNames}{0},
-			@_
+			@_,
+			aria_label => $label
 		);
-		# warn "extension rule created: $extension_ans_rule for ", ref($data);
-		return $extension_ans_rule;
-	} else {
-		main::RECORD_IMPLICIT_ANS_NAME($name) unless $self->{namedRules};
-		return $data->named_ans_rule($name, $size, @_);
 	}
+	main::RECORD_IMPLICIT_ANS_NAME($name) unless $self->{namedRules};
+	return $data->named_ans_rule($name, $size, @_, aria_label => $label);
 }
 
 #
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
Could you provide a sample problem that uses this, for testing?

somiaj:
I had to use the inspector to see the labels actually being generated.

```
DOCUMENT();
loadMacros(qw(
    PGstandard.pl
    PGML.pl
    parserMultiAnswer.pl
    PGcourse.pl
));
$ma = MultiAnswer("x + 2", "2x - 3")->with(
    singleResult => 1,
    ansLabels    => [ 'left hand side', 'right hand side' ]
);

BEGIN_PGML
[_]{$ma} [`=`] [_]*{$ma}
END_PGML
ENDDOCUMENT();
```

somiaj:
Note, without #1486 the labels won't be transferred over and only be on the hidden inputs.

somiaj:
rebased and retargeted develop.

somiaj:
After a comment in the meeting about just a way to 'label' answer boxes, I decided to just call the option `ansLabels` and describe them as a label not a suffix.

Alex-Jordan:
I think what this makes me realize, is that we should have something like the following supported by PGML:

```
[_]{answer}{width}{name}{label}
```

Or at least like:
```
[_]{answer}{label => my_label}
```

Not just for `parserMultiAnswer` (where of course it probably helps most) but to give authors a way to do this any time. Not sure if the aria label should jsut be what would be specified here, or appended to the usual "answer 1" and so on.

somiaj:
@Alex-Jordan agreed, I want this to be for all answers, but wasn't sure the best way to add it and retrieve the data when creating the answer boxes and was hoping someone who understands the code better could help with that.

I just started here because I could see how to do it and this macro is used in cases it would be the most useful.

dpvc:
The proposal from @Alex-Jordan suggests that the original problem could be written as

``` perl
DOCUMENT();
loadMacros(qw(
    PGstandard.pl
    PGML.pl
    parserMultiAnswer.pl
    PGcourse.pl
));
$ma = MultiAnswer("x + 2", "2x - 3")->with(singleResult => 1);

BEGIN_PGML
[_]{$ma}{5}{'left-hand side'} [`=`] [_]*{$ma}{5}{'right-hand side'}
END_PGML
ENDDOCUMENT();
```

or 

``` perl
DOCUMENT();
loadMacros(qw(
    PGstandard.pl
    PGML.pl
    parserMultiAnswer.pl
    PGcourse.pl
));
$ma = MultiAnswer("x + 2", "2x - 3")->with(singleResult => 1);

BEGIN_PGML
[_]{$ma}{label => 'left-hand side'} [`=`] [_]*{$ma}{label => 'right-hand side'}
END_PGML
ENDDOCUMENT();
```

I like this because it puts the data at the location where it is used rather than having it at some earlier place where it is more detached.  Of course, the original form could also be allowed.  I'm just noting the possibilities.

macros/parsers/parserMultiAnswer.pl
```diff
@@ -416,33 +417,38 @@ sub NEW_NAME {
...
-		return $data->named_ans_rule($name, $size, @_, aria_label => $label);
-	}
-	if ($self->{singleResult} && $self->{part} > 1) {
-		my $extension_ans_rule = $data->named_ans_rule_extension(
+	my $self  = shift;
+	my $size  = shift || 20;
+	my $part  = $self->{part};
+	my $data  = $self->{data}[$part];
+	my $name  = $self->ANS_NAME($self->{part}++);
+	my $label = $self->generate_aria_label($name, $part);
```
drgrice1, line 437: Now that the initial value of  `$self->{part}` is being saved to a local variable, this should be

```suggestion
	my $part  = $self->{part}++;
	my $data  = $self->{data}[$part];
	my $name  = $self->ANS_NAME($part);
	my $label = $self->generate_aria_label($name, $part);
```

So the current value of `$self->{part}` is saved to `$part` and then `$self->{part}` is incremented and not used again in the method.

The same change should be made in the `ans_array` method.

macros/parsers/parserMultiAnswer.pl
```diff
@@ -566,6 +573,14 @@ =head2 allowBlankAnswers
 Indicates whether to remove the blank-check prefilter from the answer checkers used for type checking
 the student's answers. Default: 0.
 
+=head2 ansLabels
+
+An array reference of labels to be added to the assoicated answer box. By default answer boxes are
```
drgrice1, line 578: `assoicated` -> `associated`

macros/parsers/parserMultiAnswer.pl
```diff
@@ -566,6 +573,14 @@ =head2 allowBlankAnswers
...
+labeled (C<aria-label>) with "answer X" or "answer X part Y" if C<singleResult> is used. These labels
+are appeneded to the default label, e.g. "answer X part Y custom label", and can be used to improve
```
drgrice1, line 580: `appeneded` -> `appended`

macros/parsers/parserMultiAnswer.pl
```diff
@@ -566,6 +573,14 @@ =head2 allowBlankAnswers
...
+the accessiblity. For example stating the side of the equation or part of an integral the answer
```
drgrice1, line 581: `accessiblity` -> `accessibility`

drgrice1:
It seems a bit confusing that you changed `suffix` to `ansLabels`, but yet in the aria label it is still a suffix.  In your example problem I see the aria labels `answer 1 part 1 left hand side ` and `answer 1 part 2 right hand side `.

I also agree with @Alex-Jordan and @dpvc that a PGML way of setting this would be nice, and that this should be extended to all answers.  Perhaps that is for another pull request though.

drgrice1:
I also think that the option name should contain something indication which label it affects.  Something like `ariaLabels` instead of `ansLabels` (or `ariaLabelSuffixes`).  This could easily be confused with the PG answer rule name set by `NAMED_ANS` (which also has the alias `LABELED_ANS`).  There are also actual input labels which are actual labels in HTML (although general answer in PG lack these).

dlglin:
> I also think that the option name should contain something indication which label it affects. Something like `ariaLabels` instead of `ansLabels` (or `ariaLabelSuffixes`). This could easily be confused with the PG answer rule name set by `NAMED_ANS` (which also has the alias `LABELED_ANS`). There are also actual input labels which are actual labels in HTML (although general answer in PG lack these).

I agree that `ansLabels` could lead to confusion with named/labeled answers.  My concern with using `ariaLabels` is whether a problem author would know what an aria label is.  How about `accessibilityLabels`?

somiaj:
Looking closer I think I see how to do this in a more general way. But before I do that, I have a question.

Do we want to allow users to write the full `aria-label`? Do we want to go the method I used here of just having a `suffix` (i.e. do we want labels like "right hand side", "answer X right hand side", "answer X part Y right hand side"), or do we want to allow both options? I think having the answer and part as part of the label will be useful which is why I went with that approach originally, but before I make this more general figured I'd see what others think.

Alex-Jordan:
I imagine something where the author can write an aria label that replaces
the default one. And for authors, it's just a "label", not an aria-label.

Alex Jordan
Mathematics Instructor
Portland Community College

drgrice1:
> I imagine something where the author can write an aria label that replaces the default one. And for authors, it's just a "label", not an aria-label. Alex Jordan Mathematics Instructor Portland Community College

I am fine with it being called something that is indicative of it being for accessibility.  I am not fine with it being called just a "label" as I have already said with valid justification.  That is unless this is for an actual html label, and not an aria label.

dpvc:
> Do we want to allow users to write the full aria-label?

You could probably do something like allow `ariaLabel => "full label"` for the full label, and `ariaLabel => "+ postfix"` to allow a postfix label.  You could pick something else for `+`, but that might be one way to do it, or `ariaLabel => "%s postfix"` to allow the "answer 1" to be inserted at the `%s`.  Something like that might work without having too many extra named options, while still giving the flexibility to do what you want.

somiaj:
Closing in favor of #1493. Figured I'd start a new PR/discussion.

Alex-Jordan:
"accessible name" appears to be a recognized term for aria label, but
abstracted away from being html-centric. How do either accessible-name or
accessible-label sound? Maybe with underscore, not hyphen?

Alex Jordan
Mathematics Instructor
Portland Community College

drgrice1:
Those sound fine to me.

drgrice1:
I would go with accessible-label over accessible-name.

somiaj:
I went with `aria_label` just because that was already being used as the option being sent to the various answer rule methods, and this way the name is the same everywhere. I could use `accessible_label` in PGML, but then the option there would be different than someone wanted to send the option to the answer rule directly. I could update everything, but the code is already using `aria_label`. See the other PR for that and if you want me to change it, I can in that PR.

PR #1490, opened 2026-08-04
Move unionTables.pl to deprecated folder

pstaabp:
This puts unionTables.pl into the deprecated folder.  There are must better and more accessible options with `niceTables.pl` and PGML tables. 

This will flag these tables when running PGcritic.


drgrice1 (approved):
I think it is time for this.

somiaj (approved):

PR #1492, opened 2026-08-05
Fix a random issue with the `parse-problem-doc.pl` script.

drgrice1:
The script works by generating the metadata for the files in the `tutorial/sample-problems` directory.  That is a hash whose keys are the filenames and whose values are hash references containing the metadata for the files. Then the keys of that hash are iterated over, and the sample problem documentation rendered.  In that process a subdirectory of the output location is created for the file if it does not already exist. Since hashes are not ordered, the order the files are iterated over will be different each time. So if the `tutorial/sample-problems/VectorCalc/VectorFieldGraph3D/VectorFieldGraph3D1.pg` problem happens to be processed before any of the other problems in the `tutorial/sample-problems/VectorCalc` directory, then the `mkdir` call attempts to create the output subdirectory `VectorCalc/VectorFieldGraph3D` which fails because the `VectorCalc` subdirectory does not yet exist and `mkdir` is not recursive.

To fix this use the `Mojo::File` `make_path` method instead, which is recursive.

It is not essential to hotfix this.  If the job fails when a hotfix goes in, then we just need to rerun the job, and hope that file, which is the only one with directory depth 2, is not processed before the other files in the directory the next time it is run.

### DIFF
--- a/bin/parse-problem-doc.pl
+++ b/bin/parse-problem-doc.pl
@@ -31,8 +31,9 @@ use feature 'say';
 
 my $pg_root;
 
+use Mojo::File qw(curfile path);
+
 BEGIN {
-	use Mojo::File qw(curfile);
 	$pg_root = curfile->dirname->dirname;
 }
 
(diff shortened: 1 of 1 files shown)


drgrice1:
Note that I already reran the job from when the release candidate branch was merged into main, and it succeeded the second time.

somiaj (approved):

pstaabp (approved):

PR #1493, opened 2026-08-09
Add aria labels to PGML and unify their use.

somiaj:
Update the `generate_aria_label` method to be given an input label. If the input label is defined and doesn't contain `%s`, return the label. If the input label is undefined (or blank) generate a default label based on the problem number, answer number, and part number. If the input label contains `%s`, replace it with the default generated label and return the result.

Update answer rules to more uniformly accept the `aria_label` option and use the `generate_aria_label` method. This required updating some and adding the option to others.

Update macros that use the `generate_aria_label` method to take advantage of the new method. The parserRadioButtons and parserCheckboxList will append "option X" to the end of the generated label. Matrix arrays will append "col N row M" to the end of the generated label.

Update PGML to accept the `aria_label` option in answer blanks to customize the label, "[_]{$answer}{aria_label => '%s label'}".

The biggest change was in parserRadioMultiAnswer which allows a new option `ariaLabels` which is the labels to send to the associated answer rules. Update its internal `generate_aria_label` function to match the changes to the one in PGbasicmacros. In addition the "subpart X" is only added to the label if there are more than one subparts, and the "part Y" will use the label of the part if `displayLabels` is set, otherwise it uses the number of the part.

This is a more general approach and replacement of #1489. Due to the discussion I am calling the option `aria_label` to match what was already in the code, `label` is not a correct name, and `accessibility_label` is just longer and would require making the code better deal with that. `aria_label` was already in use and is now just being used more uniformly. `aria_label` can be used in PGML or sent directly to various `ans_rule` or `ans_array` calls.

I went with @dpvc suggestion of allowing a single `%s` sprintf option to include the default generated label as part of the custom label. To not get any errors from `sprintf`, I went with just regex find/replace of the first instance of `%s` in the string.

Here is a problem I was using to test various different checkers and macros. Edit it to remove the `aria_label` option to check out default generated labels or remove `%s` to see any effects.

```
DOCUMENT();
loadMacros(qw(
    PGstandard.pl
    PGML.pl
    parserMultiAnswer.pl
    parserPopUp.pl
    parserRadioMultiAnswer.pl
    parserCheckboxList.pl
    parserRadioButtons.pl
    PGcourse.pl
));
Context('Matrix');
$f     = Compute("7x - 5");
$ma    = MultiAnswer("x + 2", "2x - 3")->with(singleResult => 1);
$M     = Matrix([ [ 2, 0, -3 ], [ -4, -3, -2 ] ]);
$popup = DropDown([ A .. F ], 2, useHTMLSelect => 0);
$rma   = RadioMultiAnswer(
    [
        [ 'Matrix %s*',       Matrix([ [ 0, 1 ], [ 2, 3 ] ]) ],
        [ 'Formula %s',       Formula('x+1') ],
        [ 'Number %s and %s', Compute(5), Compute(-7) ],
        ['No Input']
    ],
    2,
    ariaLabels =>
        [ '%s Matrix', '%s Formula %s', '%s Number1', '%s Number2' ]
);
$checks  = CheckboxList([ A .. F ], [ 2, 5 ]);
$buttons = RadioButtons([ A .. F ], 4);

BEGIN_PGML
[_]{$f}{aria_label => '%s %s function %s'}

[_]{$ma}{aria_label => '%s left hand side'} [`=`] [_]*{$ma}{aria_label => 'right hand side'}

[_]*{$M}{aria_label => '%s augmented matrix'}

[_]{$popup}{aria_label => '%s letter of graph'}

[_]{$rma}{aria_label => '%s radio multi answer'}

[_]{$checks}{aria_label => '%s checkbox list'}

[_]{$buttons}{aria_label => '%s radio buttons'}
END_PGML
ENDDOCUMENT();
```


somiaj:
@drgrice1 One thing I was unsure about was parserPopUp when the `useHTMLSelect => 0` would create an empty visually hidden span that contained the `aria_label`. I assuming it was correct, but feels weird that the label doesn't wrap the full popup menu being generated.

macros/core/PGbasicmacros.pl
```diff
@@ -402,7 +402,7 @@ sub ANS_RULE {
 	my ($number, $col) = @_;
 	my $name = NEW_ANS_NAME();
 	RECORD_IMPLICIT_ANS_NAME($name);
-	return NAMED_ANS_RULE($name, $col);
+	return NAMED_ANS_RULE($name, $col, @_);
```
drgrice1: This certainly looks wrong.  `$number` and `$col` are the first two elements of `@_`, and `@_` is passed to `NAMED_ANS_RULE` still containing those two elements, and `$col` is also passed again.

somiaj: Thanks, I missed the values where not shifted off like usual. I also missed this really is only used by old problems, so this change isn't needed, since nothing else will be sending `aria_label` to this method.

somiaj:
I added @Alex-Jordan suggestion, PGML can now use `[_aria label_]` to set a custom label. Follow the same labeling rules as before, so `[_%s my custom answer_]` will replace `%s` with the generated label.

I made this its own commit to easily remove.

drgrice1:
> @drgrice1 One thing I was unsure about was parserPopUp when the `useHTMLSelect => 0` would create an empty visually hidden span that contained the `aria_label`. I assuming it was correct, but feels weird that the label doesn't wrap the full popup menu being generated.

That doesn't even make sense and wouldn't work.  If the visually hidden span wrapped the popup menu, then the popup menu would also be visually hidden.

somiaj:
Right, if the span did wrap the drop down, it shouldn't be visually hidden. I was just confused because the label is just by itself, `<span class="visually-hidden" aria-label="answer 2 "></span>`, and it isn't clear to me how this actually applies the label to the drop down button that follows it. Though I guess labels don't apply to children elements too, so why isn't this label just applied to the drop down button? Anyways, it was more just I didn't understand why it was sufficient to label this visually hidden element vs the button that is used for the drop down.

drgrice1:
There was some discussion on that in #1159.  It was about getting the screen reader to announce the right thing when the dropdown menu is open.

macros/parsers/parserRadioButtons.pl
```diff
@@ -608,25 +608,25 @@ sub Index {
...
-	my $size    = shift;
+	my ($self, $extend, $name, $size, %options) = @_;
 	my @choices = @{ $self->{orderedChoices} };
 	my @radio   = ();
 	main::RECORD_IMPLICIT_ANS_NAME($name = main::NEW_ANS_NAME()) unless $name;
 
-	foreach my $i (0 .. $#choices) {
-		my $value = $self->{values}[$i];
-		my $tag   = $choices[$i];
+	my $label = main::generate_aria_label($name, delete $options{aria_label});
```
drgrice1: This is adding aria labels back to `parserRadioButtons.pl` answers.  These were just removed a release or two ago. That is because these should not have an aria label when they already have an actual visible label.  See https://github.com/openwebwork/pg/pull/1224.

macros/core/PGbasicmacros.pl
```diff
@@ -454,10 +453,11 @@ sub NAMED_ANS_RADIO {
...
 				'input',
-				type  => 'radio',
-				name  => $name,
-				id    => $name,
-				value => $value,
+				type       => 'radio',
+				name       => $name,
+				id         => $name,
+				value      => $value,
+				aria_label => generate_aria_label($name, $options{aria_label}),
```
drgrice1: This also was removed in #1224, and should not be added back.  The same for line 492 below.

macros/parsers/parserRadioButtons.pl
somiaj: Ahh, I should have looked deeper as to why these were not present. It did feel odd they were not there, and I now see that in parserRaidoMultianswer, the aria labels are added to the additional inputs but not the actual radio select buttons.

Would it be worth adding a label to the radio-buttons-container or checkbox-container so there could be a label for the whole radio/checkbox select stating either what problem it is or allowing users to label?

drgrice1: The problem author sets the text which is the label.  I think that the author can also always add a statement before the radio or check container that is for everyone to see.  So I would say not to an aria label on  those.

somiaj: Okay removed that here and the other place you noted that was incorrect. Though there is still a small change with the code cleanups I did. I can remove those changes if you prefer.

I also noticed that `NAMED_ANS_RADIO_BUTTONS` was still using `aria_label`, but then passed it to a method that didn't use them, so I cleaned up that method to also no longer have that unneeded logic.

macros/core/PGbasicmacros.pl
```diff
@@ -501,58 +500,57 @@ sub NAMED_ANS_RADIO_EXTENSION {
...
-	my $name  = shift;
-	my $label = '';
+	my ($name, $inLabel) = @_;
+	my $outLabel = '';
 
-	# if we dont have an AnSwEr type name then we do the best we can
+	# Return the input label unless it contains '%s'.
+	if ($inLabel) {
+		if ($inLabel =~ /^\S$/) {
+			$inLabel = '';
```
drgrice1: This does not make sense to me.  Why is it here?  If the `$inLabel` is a single non-whitespace character, then it is set equal to the empty string?  Why?

somiaj: The return value depends on the truth of that variable, so I'm make it false to avoid doing a substitution.

```
return $inLabel ? $inLabel =~ s/%s/$outLabel/r : $outLabel;
```

Because if I didn't do that, a single space `$inLabel = ' '` would result in returning that single space as the label.

somiaj: This is to prevent labels that are effectively blank.

somiaj: Though I just realized a flaw in my logic, `$inLabel = 0` is also false, and so one could not make the label just "0". I don't think that is a good label, but probably something I shouldn't have unintentionally prevented.

somiaj: Okay, updated the logic so "0" is allowed as a label.

drgrice1: I think you are misunderstanding what `\S` means in a Perl regular expression. It means a NON whitespace character.  So `$inLabel =~ /^\S$/` will NOT match if `$inLabel` is equal to `' '`, and so it will NOT be set to the empty string.  `$inLable =~ /^\S$/` will only match a single NON whitespace character.  So it in fact will match if `$inLabel` is equal to `0`, and it will be set to the empty string in that case.

somiaj: Oh I see what I missed, yea I got myself turned around, but I was thinking `^` negated that, vs meant start with. Thanks.

somiaj: Wait, no the reason the code worked is it was mostly skipped over, since that would only match labels that were a single non white space character, which I never tested. I am fixing it now, thanks again.

somiaj: Okay, I updated it, thanks again.

somiaj:
Just realized another possible flaw with my approach I don't have a solution to yet. If the user provided label contains more than one "%s", each time `generate_aria_label` is called the next "%s" will get replaced with the generated label until there are no more "%s" remaining. In my limited testing `generate_aria_label` only gets called once, but unsure if it is possible some recursion can happen and what to do if a user provides more than one "%s" as part of their label.

PR #1494, opened 2026-08-10
Fill in missing PG translations with AI-assisted drafts

drgrice1:
Translations were performed by Claude Sonnet 5.

All empty msgstr entries in lib/WeBWorK/PG/Localize/*.po (de, cs-CZ, el, es, fr, fr-CA, he-IL, hu, ko, ru-RU, tr, zh-CN, zh-HK) were translated. All previously-translated entries were left untouched.

Confidence is solid for es, fr, fr-CA, cs-CZ, hu, ru-RU, and el; lower for he-IL (RTL), ko, tr, zh-CN, and zh-HK, so native-speaker review should prioritize those first. A few pre-existing (not newly added) translations look wrong and were left as-is since they weren't empty, e.g. fr.po's "Ungraded" -> "Mise à jour réussie" and cs-CZ.po's "The answer has been graded." -> a translation that reads as "the correctness of the answer was NOT evaluated". Terminology was kept consistent within each language (e.g. "bucket" -> "list", "universal set" and "circuit" mapped to the standard math term).

This is an approach to obtaining full translation. We could commit these and allow native speakers to review and suggest changes, or drop this if we don't think this is a good idea.

### DIFF
--- a/lib/WeBWorK/PG/Localize/cs-CZ.po
+++ b/lib/WeBWorK/PG/Localize/cs-CZ.po
@@ -25,47 +25,47 @@ msgstr ""
 #. ($edgeText[0], $edgeText[1])
 #: /opt/webwork/pg/macros/math/SimpleGraph.pl:813
 msgid " There are edges between %1 and %2."
-msgstr ""
+msgstr " Mezi %1 a %2 jsou hrany."
 
 #. (join($comma, @edgeText[ 0 .. $#edgeText - 1 ])
 #: /opt/webwork/pg/macros/math/SimpleGraph.pl:817
 msgid " There are edges between %1%2and %3."
-msgstr ""
+msgstr " Mezi %1%2a %3 jsou hrany."
 
 #. ($edgeText[0])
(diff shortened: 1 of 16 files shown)


Alex-Jordan (approved):
This is a neat application of AI. Maybe we could add support for new languages this way.

somiaj:
If we are going to use these AI translations, is it worth getting this into main?

drgrice1:
I would be fine with merging this into main if we think that these translations are good enough.  I suppose that would be one way to get feedback on how good the translations are.

@Alex-Jordan: Yeah, if there is a language that we have a strong desire to add with no translator available, then it could be added in this way.

somiaj:
Poor might be better than none. I cannot make any statement about the quality of these translations. But maybe better to have something rather than nothing. Anyways I'll approve this with the idea poor is better than none. I think if we do decided to use ai translations, putting them in main will help them be seen and might cause someone to come help fix them.

somiaj (approved):

Alex-Jordan:
I don't have an opinion about getting this into main right now. I will merge this one though.

PR #1495, opened 2026-08-11
Add unit tests for MathObjects (Value.pm, Parser.pm, and subclasses).

drgrice1:
Covers previously untested or thinly-tested classes: Complex, Vector, Point, Set, Union, String, AbsoluteValue, and Matrix linear algebra methods (det, inverse, trace, norms, solve, order, power), plus Parser.pm error paths for malformed expressions and differentiation of formulas with units.

Also fixes Value::Interval::D, Value::Set::D, and Value::Union::D in Differentiation.pm, which each had two extra leftover `shift` calls before `my $self = shift`, so $self was bound to the wrong argument (or undef) instead of the invocant, breaking the "can't differentiate" error methods. This is covered by the new differentiation.t tests.

I wrote the `absolute_value.t` and `differentiation.t` tests and fixed the issue in `lib/Parser/Differentiation.pm` a long time ago. I asked Claude to add some more unit tests for better MathObject test coverage.  This is one of the things that AI is quite good at.

### DIFF
--- a/lib/Parser/Differentiation.pm
+++ b/lib/Parser/Differentiation.pm
@@ -757,22 +757,16 @@ sub Value::List::D {
 }
 
 sub Value::Interval::D {
-	shift;
-	shift;
 	my $self = shift;
 	$self->Error("Can't differentiate intervals");
 }
 
 sub Value::Set::D {
-	shift;
-	shift;
 	my $self = shift;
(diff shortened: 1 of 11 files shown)


somiaj (approved):
All tests passed except for skipped rserv here (don't have that setup).

Alex-Jordan (approved):

PR #1497, opened 2026-08-12
Remove the image generator and the images display mode.

drgrice1:
The images display mode is not accessible, is server intensive, doesn't look very nice, and I don't think that anyone uses it anymore.  MathJax is the only real display mode that anyone uses for HTML output.  It is time for the images display mode to go.

The `WeBWorK::PG::ImageGenerator` package is only used for this, and so has been deleted.

The `makeTempDirectory` method which was part of the `WeBWorK::PG::ImageGenerator` package and is used by the `LaTeXImage` package is now a method in the `WeBWorK::PG::IO` package. Note that it is a little different though.  It does not give the caller a choice on the parent directory of the temporary directory.  It only allows creation of a temporary subdirectory inside the temporary directory set in the WeBWorK::PG::Environment.

Since the `WeBWorK::PG::EquationCache` and `WeBWorK::PG::RestrictedClosureClass` packages are only used by the `WeBWorK::PG::ImageGenerator`, they are also not needed and were deleted.

The only remaining constant defined in the `WeBWorK::PG::Constants` package was the `$WeBWorK::PG::TIMEOUT` variable.  That variable is now defined directly in the `WeBWorK::PG` package and works exactly the same as it did before. This is the correct way to defined package variables in any case.  They shouldn't be defined elsewhere in some other package. So the `WeBWorK::PG::Constants` package is not needed and is no more.


somiaj (approved):
Tested with the webwork pairing of this. All works as expected in my initial test of assignments, tests, problem editor, library browser.

Alex-Jordan (approved):

Issue #1498, opened 2026-08-12
Old display_matrix macro from PGmatrixmacros causes issues with BEGIN_TEXT/END_TEXT


somiaj:
There are probably some problems like this in the OPL, but something broke this macro when using old problem methods.

This fails to render correctly.

```
DOCUMENT();
loadMacros(qw(
    PGstandard.pl
    PGML.pl
    PGmatrixmacros.pl
));

BEGIN_TEXT
\{ display_matrix([
   [ans_rule(5),ans_rule(5),ans_rule(5),ans_rule(5)],
   [ans_rule(5),ans_rule(5),ans_rule(5),ans_rule(5)],
   [ans_rule(5),ans_rule(5),ans_rule(5),ans_rule(5)]
],
'align'=>"ccc|c") \}
END_TEXT
ENDDOCUMENT();
```

For some reason mathjax is creating a very large vertical bar and the matrix doesn't really render.

<img width="500" height="439" alt="Image" src="https://github.com/user-attachments/assets/afef121f-2177-478c-a84f-47b9e7432578" />

This renders fine:

```
DOCUMENT();
loadMacros(qw(
    PGstandard.pl
    PGML.pl
    PGmatrixmacros.pl
));

BEGIN_PGML
[@ display_matrix([
   [ans_rule(5),ans_rule(5),ans_rule(5),ans_rule(5)],
   [ans_rule(5),ans_rule(5),ans_rule(5),ans_rule(5)],
   [ans_rule(5),ans_rule(5),ans_rule(5),ans_rule(5)]
],
'align'=>"ccc|c") @]*
END_PGML
ENDDOCUMENT();
```

We should probably fix this method as it is used in the OPL.

somiaj:
I'm unsure why `BEGIN_TEXT` / `END_TEXT` will break this macro when it works fine in PGML. I am also unsure where to look, but I suspect this has broken things in the OPL as a result.

drgrice1:
This seems to be something that is browser specific.  I tested with Firefox first, and though you were full of **** as usual!  But then I tested in Google Chrome, and see something like your screen shot.  I also see console errors with several messages like `Error: <svg> attribute viewBox: Expected number, "0 -3295 NaN 6090".`  So it looks like this is a MathJax issue.

somiaj:
[...]

PR #1499, opened 2026-08-12
Stop wrapping math in `<script type="math/tex">...</script>`.

drgrice1:
This used to be handled directly by MathJax long ago, but it no longer does so.  Using `\(...\)` or `\[...\]` now works.  So just do that. Also make sure that `&`, `<`, and `>` are properly HTML escaped. This actually gets done when the `post_process_content` method is called, but just in case the `general_math_ev3` method also does this. Note the `Mojo::DOM` parser is smart enough not to doubly escape these things.

The script tag handling in the webwork2 `mathjax-config.js` file seems to be causing issues in some cases, and switching to using `\(...\)` or `\[...\]` fixes those issues.  See #1498 for details.

It is also time to start doing some code clean up. I just did some minimal clean up of the `general_math_ev3` method.

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -1204,8 +1204,7 @@ sub ENDDOCUMENT {
 					if ($main::displayMode eq 'HTML' || !$wrapPreviewInTex) {
 						return $preview;
 					} elsif ($main::displayMode eq 'HTML_MathJax') {
-						return Mojo::DOM->new_tag('script', type => 'math/tex; mode=display', sub {$preview})
-							->to_string;
+						return "\\[$preview\\]";
 					}
 				};
 
--- a/macros/core/PGbasicmacros.pl
+++ b/macros/core/PGbasicmacros.pl
@@ -1886,35 +1886,28 @@ sub general_math_ev3 {
 	my $in   = shift;
 	my $mode = shift || "inline";
 
-	$in = FEQ($in);                          # Format EQuations
-	$in =~ s/((^|[^\\])(\\\\)*)%/$1\\%/g;    # avoid % becoming TeX comments (unless already escaped)
+	$in = FEQ($in);                          # Format equations
+	$in =~ s/((^|[^\\])(\\\\)*)%/$1\\%/g;    # Avoid % becoming TeX comments (unless already escaped).
 
-	## remove leading and trailing spaces so that HTML mode will
-	## not include unwanted spaces as per Davide Cervone.
-	$in =~ s/^\s+//;
-	$in =~ s/\s+$//;
-	## If it ends with a backslash, there should be another space
-	## at the end
-	if ($in =~ /(^|[^\\])(\\\\)*\\$/) { $in .= ' ' }
+	# Remove leading and trailing spaces so that HTML mode will not include unwanted spaces.
+	$in =~ s/^\s+|\s+$//g;
+	# If it ends with a backslash, there should be another space at the end.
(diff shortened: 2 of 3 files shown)


somiaj (approved):
I can confirm it fixes the issue. Probably needs a hotfix.

dpvc:
If there are no more `<script type="math">` tags being generated, you should be able to remove the lines

https://github.com/openwebwork/webwork2/blob/c7060fe858cb27b17aad5cf77574ff7d1ae3e1fa/htdocs/js/MathJaxConfig/mathjax-config.js#L147-L171

from the MathJax configuration, as well as

https://github.com/openwebwork/webwork2/blob/c7060fe858cb27b17aad5cf77574ff7d1ae3e1fa/htdocs/js/MathJaxConfig/mathjax-config.js#L133

It must have something to do with the faked MathItem that is created there.  I'll look into it, but since these seems to resolve the problem, I'm n to sure ho much time I can put to it.

drgrice1:
I do plan to put in a pull request to webwork2 to remove that `script` tag handling in `mathjax-config.js`.  I have it already committed in a branch.  I was just tweaking this one with the HTML escaping.  I need to check on the escaping of the feedback case in PG.pl yet.

@somiaj: The `>` doesn't really need it as long as there is not matching unescaped `<` preceding it.

somiaj:
Should we worry about breaking third party macros and the likes if we remove lines that deal with the script tags? I am not worried if we break them, just something to consider.

drgrice1:
How would removing the `script` tag break a third party macro?

somiaj:
@drgrice1 Yea, I knew they didn't need to be escaped, I was just thinking in terms of consistency, someone might wonder why `<` changes to `&lt;` but `>` doesn't. Though might not affect the rendering in the end anyways.

somiaj:
I was just considering what was happening of some third party macro was using `<script type="math/tex">...</script>` for their math output in HTML_MathJax mode. But probably not being done or something to worry about.

drgrice1:
Okay, since the feedback case is handled in post processing anyway, and always goes through `Mojo::DOM`, that case is always escaped.  So this is good as is.

drgrice1:
@somiaj: You might note that in the usual way that problems are rendered via the translator, the `>` ends up getting escaped anyway when the `post_process_content` method of the translator is called, and the problem text is parsed by `Mojo::DOM`.

drgrice1:
I switched to using `HTML::Entities::decode_entities` and encode `>` as well.  Since the `HTML::Entities` package is available, might as well.  It is already used more generally for HTML display mode below.

lib/WeBWorK/PG.pm
```diff
@@ -160,7 +160,7 @@ sub new_helper ($invocant, %options) {
 		);
 	}
 
-	$translator->post_process_content if ref($translator->{rh_pgcore}) eq 'PGcore';
+	#$translator->post_process_content if ref($translator->{rh_pgcore}) eq 'PGcore';
```
somiaj: Is this a mistake from testing?

drgrice1: Yes. I will remove that later.

drgrice1: That is removed now.

somiaj (approved):
Double checked again that the issue is fixed. I still think this needs to be a hotfix as the broken problem was originally from the OPL.

Alex-Jordan (approved):

PR #1500, opened 2026-08-13
Mojo template updates

drgrice1:
Put proper extensions on the Mojolicious template files in this repository.
    
This is the same as was done with these files in the webwork2 repository.  The `.mt` extension is not a valid extension for a Mojolicious template file.  So rename them to `.html.epl` files.

Also format the template files with `prettier` and `@openwebwork/prettier-plugin-mojolicious-html-template`. Note that this is not enforced by the formatting workflow in this repository, and so this will need to be manually maintained.  There aren't that many of these, and they aren't modified often, so that should be fine for now.

Also update the versions of the javascript and css libraries used in the templates.


drgrice1:
Note the move of the files is done in the first commit, and formatting and version updating in the second.  This is so that git tracks the move properly.

somiaj (approved):

Alex-Jordan (approved):

PR #1501, opened 2026-08-13
Stop wrapping math in `<script type="math/tex">...</script>`. (hotfix of #1500)

drgrice1:
This used to be handled directly by MathJax long ago, but it no longer does so.  Using `\(...\)` or `\[...\]` now works.  So just do that. Also make sure that `&`, `<`, and `>` are properly HTML escaped. This actually gets done when the `post_process_content` method is called, but just in case the `general_math_ev3` method also does this. Note the `Mojo::DOM` parser is smart enough not to doubly escape these things.

The script tag handling in the webwork2 `mathjax-config.js` file seems to be causing issues in some cases, and switching to using `\(...\)` or `\[...\]` fixes those issues.  See #1498 for details.

It is also time to start doing some code clean up. I just did some minimal clean up of the `general_math_ev3` method.

### DIFF
--- a/macros/PG.pl
+++ b/macros/PG.pl
@@ -1206,8 +1206,7 @@ sub ENDDOCUMENT {
 					} elsif ($main::displayMode eq 'HTML_dpng') {
 						return $rh_envir->{imagegen}->add($preview);
 					} elsif ($main::displayMode eq 'HTML_MathJax') {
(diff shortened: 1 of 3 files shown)


somiaj (approved):

Alex-Jordan:
Before anyone merges this, please give me a little more time to complete a test with PreTeXt. Should be able to do that later today.

Alex-Jordan (approved):

PR #1516, opened 2026-08-17
Fix bug in problemRandomize hardcopy.

### LINKED ISSUE #1515: problemRandomize.pl doesn't generate hardcopy
When trying to generate hardcopy with any problem that calls `ProblemRandomize()` from problemRandomize.pl you get the error

Use of uninitialized value in string ne at line 176 of [PG]/macros/deprecated/problemRandomize.pl.

Here's a MWE to test with.  Save this to a problem set then generate a hardcopy.

```
DOCUMENT();

loadMacros("PGstandard.pl", "problemRandomize.pl", 'PGML.pl', "PGcourse.pl");

ProblemRandomize();

BEGIN_PGML
[_]{0}
END_PGML

ENDDOCUMENT();
```

somiaj:
When generating hardcopy with a problem using problemRandomize, `$input_ref` doesn't contain the `user` or  `effectiveUser` keys, which causes an error when trying to compare them. Only compare if `user`  and `efficitveUser` are equal in the case `effectiveUser` is a defined key in `$input_ref`.

Fixes #1515.

Note this is just a quick fix, as the macro is deprecated and there could probably be other issues with its approach.

### DIFF
--- a/macros/deprecated/problemRandomize.pl
+++ b/macros/deprecated/problemRandomize.pl
@@ -166,8 +166,10 @@ sub new {
 		when         => "correct",
 		onlyAfterDue => 1,
 		style        => "Button",
-		styleName    =>
-			($main::inputs_ref->{effectiveUser} ne $main::inputs_ref->{user} ? "checkAnswers" : "submitAnswers"),
+		styleName    => (
+			!$main::inputs_ref->{effectiveUser}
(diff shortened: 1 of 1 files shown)


drgrice1 (approved):
Looks good for this broken macro!

Alex-Jordan (approved):

PR #1520, opened 2026-08-19
Update `WWSafe.pm`.

drgrice1:
The updated `WWSafe.pm` is based on version 2.47 of the upstream `Safe.pm` module. It is essentially the upstream module except that some reentrancy issues with `reval` and `rdo`, and an issue with sharing an entire package to the safe compartment are fixed.

Claude helped quite a bit with finding and fixing those issues.

This should be tested with various types of problems to ensure that all problem rendering still works as expected without any new warnings or errors.  I have done quite a bit of testing, and have not found any issues at this point.

### DIFF
--- a/lib/WWSafe.pm
+++ b/lib/WWSafe.pm
@@ -2,9 +2,9 @@ package WWSafe;
 
 #use 5.003_11;
 use 5.12.0;
-use strict;
-use utf8;
-$Safe::VERSION = "2.16";
+use Scalar::Util qw(reftype refaddr blessed);
+
+$Safe::VERSION = "2.47";
 
 # *** Don't declare any lexicals above this point ***
 #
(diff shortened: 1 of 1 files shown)


somiaj (approved):

drgrice1:
I tried to get Claude to come up with a way to exploit the previous safe.  Something that this fixes. After quite a bit of effort the answer was that the update to WWSafe.pm (Safe.pm 2.47 base) does fix a bug.  A code reference captured from `reval`'s return value, if called later by trusted code outside an active `reval`, loses the compartment's runtime namespace-rooting protection for bless/symbolic method dispatch/symbolic sub references. This can be reproduced at the bare-WWSafe level. But Claude could not find a way to actually trigger it through PG's current code within a single render from a problem. Every call site that invokes compartment-produced code (process_answers, grade_problem, post_process_content) already wraps the call in a fresh `reval`. So it's a genuine fix at the library level, but not something exploitable.

Note that Claude did find some possibilities if multiple problems are rendered in the same process.  However, since that is not done in production anymore, that is not an issue.

I still think that this update is worthwhile (and Claude agrees), because future PG code could do something that would call a returned `reval` code reference, and with this PG might be prepared to do so securely.

pstaabp (approved):
I also tried a bunch of different types of problems and looks good.  

I suggest we merge and then can test over the next 10 months.

PR #1523, opened 2026-08-23
More vulnerability fixes

drgrice1:
Fix several more safe compartment vulnerabilities.

First, `HTML::Parser` was shared into the safe compartment. Its `parse_file` method opens and reads whatever path it is given, bypassing the permitted_read_dir restriction. Thus giving any PG problem arbitrary file read. It is not actually used by PG, and so that is removed from the modules that are shared. Note that the `HTML::Entities` package which is part of the `HTML::Parser` package on CPAN is still shared and is used.

Second, the `PGloadfiles::compile_file` method compiles whatever file it is given, and so a problem could call it directly (bypassing `findMacroFile`'s restriction of `$filePath` to being in a directory in the `$macrosPath` array) to compile and execute an arbitrary file.  So the `compile_file` method now validates the file it is asked to compile using the same restriction as the findMacroFile method.  Although this cannot be checked against `$self->{envir}` and `$self->{pwd}`, since problem code can modify those.  Instead `WeBWorK::PG` now saves the macrosPath and problem directory into `WeBWorK::PG::IO` which is not exposed to the safe compartment, and the `compile_file` method uses those.

Also fix WeBWorK::PG::IO::path_is_subdir to reject an empty or undefined directory argument. Previously that normalized (via canonpath) to '/', which every absolute path matches, silently turning a "restrict to this directory" check into "allow anything".

Third, restrict `PGalias::alias_for_tex` to reading files in allowed locations.  `alias_for_tex`, used when a problem is rendered in hardcopy, did not check that an absolute path passed to it was located in an allowed location before using it, unlike `alias_for_html` which routes such paths through `create_link_to_tmp_file`'s `permitted_read_dir` check.  Since `alias` reaches `alias_for_tex` directly for TeX mode, and methods like `image` embed its return value straight into \includegraphics, any problem could get an arbitrary file on the server embedded into its generated hardcopy PDF.

Fourth, restrict the GD::Image file-path-taking methods to permitted_read_dir.  GD is shared into the safe compartment for graphing macros.  Several of its methods (new, newFromPng, newFromJpeg, newFromGd, newFromXpm, and others) open a given file path directly with no restriction, bypassing WeBWorK::PG::IO's permitted_read_dir. Add WeBWorK::PG::SafeGD, which patches the GD::Image symbol table once after GD loads so that these methods reject a path outside permitted_read_dir before touching the filesystem. In their current form these methods can be used in a problem to implement a file existence check, and furthermore can reveal if a file exists but the server user does not have permission to access them.

Also, fix an incorrectly quoted string in `AnswerHash.pm`.  The backtick quoted string would have been evaluated as a shell command.  This only would occur if the `debug` key is set on an `AnswerHash` object. But it should not be backtick quoted.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

I will post problems that demonstrate these vulnerabilities in slack.

Note that this is built on top of #1522 since otherwise it would conflict with that pull request.

There will also be a paired pull request to webwork2 that removes the `HTML::Parser` package from the list of modules shared to PG.


drgrice1:
I believe that with this pull request and #1522, the safe compartment is now properly safe guarded against invalid file IO access.

drgrice1:
Clearly, we need to be more careful about sharing entire packages to the safe compartment.

pstaabp:
I checked out this PR and webwork2#3166 and for all of the problems you posted I am not able to access any of the files.  On the first couple, it assumes you're on linux with the paths and I tested on MacOS.  

I'm guessing this is what is expected--anything else to test with this?

drgrice1:
Yes, you will need to edit the problems and set them up to use paths and files that are appropriate for your system to properly test this.  You should test the problems with the develop branch where you will see the vulnerabilities in action.  Then test with this pull request, and of course the vulnerabilities will not work.

pstaabp:
I see that the fixed worked with everything except the two POSIX read/write problems.  I'm not sure that MacOS has the same file to load to use.  Basically those problems failed because there was no file.

drgrice1:
I suspect that MacOS does have the file.  It is just located at a different place and might have a different name.  The same will be true of other linux distributions also.  That is what I meant by may need to modify the problem files to properly test them.  An attacker could try different locations for the file, and find it using the right approach in the code.

From what I can tell based on a quick Google search, the file on MacOS will have the `.bundle` extension instead of `.so`.  Possibilities include `/System/Library/Perl/[version]/darwin-thread-multi-2level/auto/POSIX/POSIX.bundle` or `/opt/homebrew/lib/perl5/site_perl/[version]/darwin-[arch]/auto/POSIX/POSIX.bundle`.

pstaabp:
I found a file at `Users/XXXX/perl5/perlbrew/perls/perl-5.42.0/lib/5.42.0/darwin-2level/auto/POSIX/POSIX.bundle` and this PR does not allow those files to be loaded now.

pstaabp (approved):

somiaj (approved):

PR #1526, opened 2026-08-26
Cleanup of PGloadfiles.pl.

drgrice1:
This is just cleanup other than the removal of the `sourceAlias` method. That method is not used, and returns a completely dysfunctional link. It has nothing to do with the `source.pl` macro or how it works, and is not needed for the functionality of that macro (which does still work if proxying via apache2).

I have had this branch sitting here for a while.  Doing branch house cleaning!

### DIFF
--- a/lib/PGloadfiles.pm
+++ b/lib/PGloadfiles.pm
@@ -1,29 +1,162 @@
+package PGloadfiles;
+
+use strict;
+use warnings;
+use utf8;
+
+use Encode qw(decode);
+
+use WeBWorK::PG::Translator;
+
+our $debugON = 0;
+
+sub new {
+	my ($class, $envir) = @_;
+	die 'PGloadmacros must be called with an environment' unless ref($envir) eq 'HASH';
+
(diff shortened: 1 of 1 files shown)


lib/PGloadfiles.pm
```diff
@@ -1,213 +1,103 @@
...
-		macrosPath    => '',
-		pwd           => '',       # current directory -- defined in initialize
-	};
-	bless $self, $class;
-	$self->initialize;
-	#$self->check_parameters;
-	return $self;
-}
+	my ($class, $envir) = @_;
+	warn 'PGloadmacros must be called with an environment' unless ref($envir) eq 'HASH';
```
somiaj: I know this isn't new code, but this warn won't stop future code from being run, and below this `$envir` is used as a hash ref, so should this be a `die` instead of a `warn`?

drgrice1: I changed that to a `die` instead.

somiaj (approved):

pstaabp (approved):

PR #1527, opened 2026-08-26
Clean up PGrandom.pm

drgrice1:
This is just code clean up, and part of a push to get the core PG modules in order.

Another branch I have had for a while.


drgrice1:
Note that one thing to test here is that problems with a given seed are the same as before.

pstaabp (approved):
Tested many problems and getting the same problem from both this and the develop branch.

somiaj (approved):
Random numbers seem to stay the same with this in my limited testing.

PR #1528, opened 2026-08-28
Wait for the `MathJax.startup.promise` to update the feedback popoer.

drgrice1:
The `inserted.bs.popover` event still does not occur late enough for the MathJax `typesetPromise` to be defined on the page sometimes. I am not seeing this once the update is deferred a bit more with a `setTimeout` call.  So wait on the `MathJax.startup.promise` to be fulfilled before updating.
    
One way to see the issue is to set the "Assist with the student answer entry process" option to "MathView".  Then click the "Show Correct Answers" button in a problem with multiple answers (it can occur with one answer, but more reliably occurs with multiple answers).  The popovers are not positioned correctly.
    
Another way to see the issue is to add `console.log(MathJax.typesetPromise)` on the line before the change in this pull request (even with MathQuill enabled).  You will see that most of the time, that is undefined.

### DIFF
--- a/htdocs/js/Feedback/feedback.js
+++ b/htdocs/js/Feedback/feedback.js
@@ -13,7 +13,10 @@
 
 		feedbackBtn.addEventListener('inserted.bs.popover', () => {
 			// Render MathJax previews.
-			if (window.MathJax) MathJax.typesetPromise?.([feedbackPopover.tip]).then(() => feedbackPopover.update());
+			if (window.MathJax)
+				MathJax.startup.promise
+					.then(() => MathJax.typesetPromise([feedbackPopover.tip]))
+					.then(() => feedbackPopover.update());
 
 			// Execute javascript in the answer preview.
 			feedbackPopover.tip?.querySelectorAll('script').forEach((origScript) => {


dpvc:
Perhaps you should wait on `MathJax.startup.promise` before using `MathJax.typesetPromise()` rather than assuming the `setTimeout()` will have the correct delay?  Just a thought.

drgrice1:
That is what the code used to do, but I thought you told us to remove that for MathJax 4?  I will add that back.  Does the result need to be assigned to `MathJax.startup.promise` as it was before also?

drgrice1:
I changed that, but did not assign to `MathJax.startup.promise`.  That seems to be working, but if the assignment is needed, I will add that.

dpvc:
> I thought you told us to remove that for MathJax 4?

Sorry, I should have been clearer.  It is true that you no longer need to assign to the `MathJax.startup.promise`, but you can still use it to wait for when MathJax is ready to go, as you have done now.  I would probably do it as

``` js
if (window.MathJax) {
	MathJax.startup.promise
		.then(() => MathJax.typesetPromise([feedbackPopover.tip]))
		.then(() => feedbackPopover.update())
}
```

(chaining the `then` calls rather than nesting them), but that is just a style preference.

dpvc:
PS, you also should not need the `?.` for the `typesetPromise()` call, as that will be set up when the `MathJax.startup.promise` resolves.

drgrice1:
I switched to the promise chain.  I agree that is cleaner.

I also removed the `?.`.

dpvc (approved):
Looks good.

somiaj (approved):
Should this go into main too?

drgrice1:
Yeah, I was planning to make a hotfix of this, and did so now.

PR #1531, opened 2026-08-31
Fix for an empty macro when running pg-critic

pstaabp:
If there is an empty macro in a PG problem like

```
loadMacros('PGStandard.pl','PGML.pl',,'PGcourse.pl');
```
the `pg-critic.pl` script throws an error and stops.  This fix will replace an undefined macros with a ''.


drgrice1 (approved):
Looks good.

somiaj (approved):

PR #1533, opened 2026-09-04
Drop POST_HEADER_TEXT support.

drgrice1:
There is no valid reason for a problem to use `POST_HEADER_TEXT`. Anything that you want after the HTML `head` can be added to the problem text itself.  There was an invalid comment about iframes suggesting that this method be used if a page embedded in an iframe contains a form. However, it does not need to be outside of the problem form.  It is in an iframe which is a completely separate page, and will not conflict with the outer page's form.

Note that a problem author can still add to the `head` section of the HTML page with `HEADER_TEXT`.

This method is not used by any problem in the OPL or Contrib.

I have had this sitting here for 3 years.  I am pulling the trigger on it.

### DIFF
--- a/lib/PGcore.pm
+++ b/lib/PGcore.pm
@@ -52,10 +52,9 @@ sub new {
 	#warn "creating a new PGcore object";
 	my %options = @_;
 	my $self    = {
-		OUTPUT_ARRAY      => [],    # holds output body text
-		HEADER_ARRAY      => [],    # holds output for the header text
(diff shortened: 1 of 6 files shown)


somiaj (approved):

Alex-Jordan (approved):

somiaj:
This has a conflict that needs to be fixed now the spelling check workflow is in.

PR #1537, opened 2026-09-16
Always load `PGbasicmacros.pl` and `PGauxiliaryFunctions.pl`.

drgrice1:
This is done at the end of the `DOCUMENT` method in `PG.pl`. This means that problems no longer need to load `PGstandard.pl`.  The basic functionality that is needed is always there.

This does NOT load `PGanswermacros.pl`.  That is because that is entirely deprecated functionality. Problems that use it already load what is needed, and must continue to do so.  Newly written problems should not use it.

This could also load `customizeLaTeX.pl`, but I held off on that. That is an antiquated approach to customization of notation that the rest of PG never followed up on.

Edit: Also remove `PGstandard.pl` from the `loadMacros` calls in most sample problems.
    
There are two problems that `PGstandard.pl` was replaced with `PGanswermacros.pl` for now.  Those are the problems `Misc/Matching.pg` and `ProblemTechniques/Percent.pg`.

The problem `Misc/Matching.pg` was using `unionTables.pl` to layout the multiple choice questions and answers.  That was replaced with a `niceTables.pl` layout table via its PGML syntax.  However, it uses the `PGchoicemacros.pl` match list which is the one case that we still have not created a modern equivalent of. That needs the `str_cmp` method.

The problem `ProblemTechniques/Percent.pg` uses `contextPercent.pl` which uses the `prfmt` method which is defined in `PGanswermacros.pl`.  That is something that should be changed.  It should not depend on a macro that is essentially deprecated.


drgrice1:
As discussed, loading `PGML.pl` by default is troublesome because it loads `MathObjects.pl` which breaks `lib/Matrix.pm` and any macro that depends on it.  The `PGmorematrixmacros.pl` is one of these, and many problems in the OPL load this macro.  Note that many of those problems do not actually use the macro, they just load it.  However, that is enough to break the problem if `MathObjects.pl` is already loaded.  Generally that macro, and more generally `lib/Matrix.pm`, is entirely incompatible with `MathObjects.pl`.

Alex-Jordan:
Should the sample pg problem (and all distribution pg problems both here
and in PG) stop loading them then?

Alex Jordan
Mathematics Instructor
Portland Community College

drgrice1:
So this is just starting to feel out where we want to go with this.  This is the most basic approach that resists going further with loading `PGML.pl` by default as well.  I also have an approach that sort of handles loading `PGcourse.pl` by default, although, it is a hack.  @dlglin has suggested going so far as removing the need for even the `DOCUMENT();` and `ENDDOCUMENT();` calls at the beginning and end of the file.

But yes, if this is where we stop, then all of the sample problems should be updated accordingly.

somiaj (approved):
Good start.

drgrice1:
I went ahead and removed `PGstandard.pl` from most sample problems.  See the edited initial comment for this pull request.

somiaj:
Seems there could be more work done to remove use of PGstandard.pl if this is the route we are going (which I think is good).

+ `WeBWorK::PG::ConvertToPGML` should probably not include it anymore.
+ `PGbasicmacros.pl` POD for `helpLink` suggest loading `PGstandard.pl`.
+ Various macros load `PGstandard.pl` but may not need to.  I suspect some of those macros could have PGstandard.pl removed, though most are older/deprecated macros that may require it.

somiaj:
Also left out, many tests load PGstandard, maybe update those tests too, for which it is appropriate.

drgrice1:
Yes, there will be more work needed.  One of the reasons I proceeded with removing it in the sample problems was to see that it works well with those.  The result is that it does for all but those two problems, and it reveals the issues there.  One is a macro that is yet needed (a math object match list macro), and one is a current macro with a bad dependency.

If we decide that this is the way to go, and that it goes far enough (i.e., that, at least for now, we don't want to go to loading PGML by default, or include the PGcourse.pl loading hack, or go on to removing the need to call `DOCUMENT` and `ENDDOCUMENT`), then I will take it the rest of the way.

PR #1538, opened 2026-09-18
Fix some issues with the list handling for MultiAnswer and RadioMultiAnswer.

drgrice1:
There is a check in both macros that sets the `checkTypes` value to be 'compatible' when the default checker is used and `checkTypes` is not 'exact'. However, 'exact' is not a valid value ever.  That is supposed to be 'equal'.

Also in the `perform_check` method for a `RadioMultiAnswer` it should not return when the answer hash `ans_message` is set and `checkTypes` is `compatible`.  For a `MultiAnswer` object it is already this way.  This means that if an entry in a list answer is incorrect, the default or custom checker will not be called.  The default checker will return 0 in this case (with a minor tweak to ensure the score is 1 instead of just nonzero that was added), but a custom checker might want to score this case and give partial credit.


somiaj (approved):
Looks good.

somiaj:
With the problem in the forums,  I was wondering if we could make `compatible` the default for `checkTypes`, since we are basically doing that for the default checker. Though I assume this was not done in order to not break some older problems that expected the old behavior.

drgrice1:
Yes, the new `compatible` setting was created so that it would not break existing behavior.

pstaabp (approved):
Fixes the issue.

PR #1539, opened 2026-09-19
Fix an issue with GraphTool interval arrows at infinity disappearing.

drgrice1:
This is caused by a change in how JSXGraph sets opacity for arrow heads for version 1.13.  It sets the opacity using the opacity of the line the arrow head is attached to, and reasserts this setting after creation. Since the line for the points at infinity is always completely transparent, and so this makes the arrows transparent as well, since the interval tool setting of the `fill` attribute is overridden.  To fix this the fill is set via an inline style instead which overrides the JSXGraph attribute and so it doesn't matter what JSXGraph internally sets its attribute to.

To test this you can use the following problem:

```
DOCUMENT();

loadMacros('PGstandard.pl', 'PGML.pl', 'parserGraphTool.pl', 'PGcourse.pl');

$x = random(-5,     -1);
$y = random($x + 1,  5);

$gt = GraphTool("{interval,(-inf,$x]},{interval,($y,inf)}")->with(
    numberLine     => 1,
    bBox           => [ -7, 7 ],
    ticksDistanceX => 1,
    minorTicksX    => 0,
    useBracketEnds => 0,
);

BEGIN_PGML
Graph the solution set for the compound linear inequality [`x \leq [$x]`] and [`x > [$y]`].

[_]{$gt}
END_PGML

ENDDOCUMENT();
```

Graph a bounded interval, and then once it is graphed move an end point to infinity (the end of the board) using either the mouse or the keyboard.  Instead of it becoming an arrow, it disappears.  Hovering the mouse over it, or using the arrow key again makes it appear, but it shouldn't disappear at all.


somiaj (approved):
Fixes issue.

pstaabp (approved):
fixes the issue.

